PR #20, opened 2017-04-03
Minor fixes while building openjdk in native env (without docker)

neomatrix369:
Fixed a abs/relative path issue, also replaced hard-coded references to openjdk repo folder name with a variable

### DIFF
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -202,5 +202,5 @@ if [ "${USE_DOCKER}" ] ; then
 
 else
   echo "Calling sbin/build.sh $WORKING_DIR $TARGET_DIR"
-  $WORKING_DIR/sbin/build.sh $WORKING_DIR $TARGET_DIR
+  $WORKING_DIR/sbin/build.sh $WORKING_DIR $TARGET_DIR $OPENJDK_REPO_NAME
 fi
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -25,6 +25,7 @@
 # You can set the JDK boot directory with the JDK_BOOT_DIR environment variable
 
 REPOSITORY=AdoptOpenJDK/openjdk-jdk8u
+OPENJDK_REPO_NAME=openjdk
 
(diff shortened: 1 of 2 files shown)


sbin/build.sh
```diff
@@ -95,7 +96,7 @@ else
   fi
    
   # We get the files we need at $WORKING_DIR/installedfreetype
-  ./configure --prefix=$WORKING_DIR/installedfreetype $PARAMS && make all && make install
+  ./configure --prefix=$WORKING_DIR/openjdk/installedfreetype $PARAMS && make all && make install
```
karianna: Shouldn't this now be `$OPENJDK_REPO_NAME` as well?

neomatrix369: Yes, good catch!

neomatrix369:
@karianna this is more a bug-fix rather than an enhancement, fixing a folder reference issue in `sbin/build.sh` (just for posterity reasons good to label it closer to what it means)

makejdk.sh
```diff
@@ -119,25 +120,25 @@ else
 fi
 
 echo $git
-if [ -d "$WORKING_DIR"/openjdk/.git ] && [ $REPOSITORY == "AdoptOpenJDK/openjdk-jdk8u" ] ; then
+if [ -d "$WORKING_DIR"/$OPENJDK_REPO_NAME/.git ] && [ $REPOSITORY == "AdoptOpenJDK/openjdk-jdk8u" ] ; then
   # It does exist and it's a repo other than the AdoptOpenJDK one
```
karianna: I assume a later PR will take care of the hard coding of `openjdk-jdk8u`?

karianna:
@gdams - If you're happy with this, then lets merge it :-)

makejdk.sh
neomatrix369: Will create multiple PRs to move script to more generic and reusable state - focused PRs

PR #31, opened 2017-04-05
Extra arguments for jtreg execution

Bluevisitor:
The original plan for this branch was to add a series of options onto the jtreg command line.

However, the debug output from an earlier job shows us that these options are all used by default.

In the process of learning this, I discovered that there appears to be a variable defined entirely for the purpose of allowing the user to set jtreg command-line options without having to directly edit the makefile (which is what we used to do, using sed).

So now I'm moving us from sed to this environment variable.

### DIFF
--- a/sbin/jtreg.sh
+++ b/sbin/jtreg.sh
@@ -68,8 +68,7 @@ export JTREG_HOME=${JTREG_INSTALL}
 export JPRT_JTREG_HOME=${JT_HOME}
 export JPRT_JAVA_HOME=${PRODUCT_HOME}
 export JTREG_TIMEOUT_FACTOR=5
-export CONCURRENCY=2
-export EXTRA_JTREG_OPTIONS='-xml:verify -jcov\/classes:$(ABS_PLATFORM_BUILD_ROOT)\/jdk\/classes\/  -jcov\/source:$(ABS_PLATFORM_BUILD_ROOT)\/..\/..\/jdk\/src\/java\/share\/classes  -jcov\/include:*'
+export EXTRA_JTREG_OPTIONS='-concurrency:2 -xml:verify -jcov\/classes:$(ABS_PLATFORM_BUILD_ROOT)\/jdk\/classes\/  -jcov\/source:$(ABS_PLATFORM_BUILD_ROOT)\/..\/..\/jdk\/src\/java\/share\/classes  -jcov\/include:*'
 
 
 make test jobs=10 LOG=debug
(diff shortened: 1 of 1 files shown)


Bluevisitor:
Make documentation appears to imply that a variable passed into a makefile via an environment variable can be manipulated internally, unlike command-line arguments.

If that is the case, the test here should have the concurrency setting for jtreg set explicitly to 2. 

https://ci.adoptopenjdk.net/job/adam_test/6/

If this is not the case, the concurrency variable will not be explicitly set to 2, and we'll need to add -concurrency:2 to the EXTRA_JTREG_OPTIONS environment variable consumed by the makefile.

neomatrix369:
@Bluevisitor Changes look fine although https://ci.adoptopenjdk.net/job/adam_test/6/console show test run has broken

Bluevisitor:
I see 2 points for concern:

1) The test fails due to the 2 concurrency options on the jtreg command line. 
2) The concurrency option used to be set to "7", despite the fact that we were setting it to 8. Now it's still setting the concurrency option (both of them) to 7, despite the fact that I changed it to 2 in the environment variable.

Will investigate.

Bluevisitor:
The 2 concurrency options may be caused by bug: https://bugs.openjdk.java.net/browse/JDK-8157605

It's fixed in jdk9, but the request to backport is to 8 has gone unanswered. Also had difficulty finding why/how it takes the concurrency value and subtracts from it before passing the final value to the jtreg command.

Right now, my best plan is to replace the concurrency environment variable with a direct -concurrency option inside the EXTRA_JTREG_OPTIONS variable.

Bluevisitor:
Here's a jenkins run to test that the aforementioned change works: https://ci.adoptopenjdk.net/job/adam_test/7/

Bluevisitor:
Didn't work. Will investigate further.

gdams:
with regard to `3` we probably don't want the test to fail if no tests are found otherwise it will fail every time we build without the `--jtreg` flag

Bluevisitor:
It seems the only way around this is to get the fix for the bug (link below) backported to jdk8. Will raise a new bug to cover this.

https://bugs.openjdk.java.net/browse/JDK-8157605

As for this PR, the only change we can make now is the change to the jenkins job config to cause a job failure if there are no results files. Since that doesn't need a PR, I'm removing this PR and deleting the branch.

Bluevisitor:
Sorry George, I didn't spot your comment. Will leave the jenkins option alone.

Issue #7, opened 2017-03-20
Builds should run test/test-coverage, CLI/config option to switch this off


neomatrix369:
Our Cloudbees/Jenkins jobs currently runs the native test suite (with the help of `jtreg`), with code coverage enabled (using `jcov`).

It would be great to have the new jobs be able to do the same.

Examples on how this is done can be see from the bash scripts attached the Jenkin jobs.

tellison:
Yep, test is a significant area for further discussion and contribution.  We need to figure out what types of testing we need to do for different purposes (PR sanity check, build verification check, release level testing, etc).

Jtreg is fine as far as it goes, and should definitely feature in our overall test approach.

neomatrix369:
A number of pull requests have been merged to help resolve this issue:
#22 
#24 
#31

karianna:
@neomatrix369 What's left to go?  Can you summarise here?

neomatrix369:
~Our Cloudbees/Jenkins jobs currently runs the native test suite (with the help of `jtreg`), with code coverage enabled (using `jcov`).~ (done)

~It would be great to have the new jobs be able to do the same. If necessary, provide CLI/config option to switch this off and just build the artifact.~ (done)

**If the test fails, the artifact isn't created *** <=== needs doing, although is easy to do, once the script is refactored a bit

~when it passes we have three tarballs, one with just the JDK binaries and the other two with the test coverage and test output reports included in the tarball, see https://adopt-openjdk.ci.cloudbees.com/view/OpenJDK/job/openjdk-1.8-linux-x86_64/~

**Sanity check of JCov reports - is it getting generate correctly, is it legible **
**Publishing link to the JCov reports from the workspace **

Items in bold are remaining.

karianna:
@Bluevisitor reports that langtools jtreg fails:

```
/usr/bin/make -k -C  ../langtools/test  JT_JAVA=/home/jenkins/workspace/openjdk_build_x86-64_linux/openjdk/build/linux-x86_64-normal-server-release/images/j2sdk-image JTREG_HOME= TEST=jtreg jtreg
make[2]: Entering directory '/home/jenkins/workspace/openjdk_build_x86-64_linux/openjdk/langtools/test'
make[2]: *** No rule to make target '/bin/jtreg', needed by 'check-jtreg'.
```

neomatrix369:
[...]

PR #40, opened 2017-04-07
Allow user to specify tests we want jtreg to run

Bluevisitor:
Currently we allow the user to say "run the default jtreg tests after the build is complete", but we don't allow the user to specify a subset.

If we want to run less tests than the default, or more, or if we want to run focused testing (in the event that a specific module has been changed recently), we'll need this option or something like it.

### DIFF
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -76,8 +76,8 @@ while [[ $# -gt 0 ]] && [[ ."$1" = .-* ]] ; do
     "--jtreg" | "-j" )
     JTREG=true; shift;;
     
-    "--jdk_subsets" )
-    JTREG=true; JTREG_JDK_SUBSETS="$1" shift;;
+    "--jtreg_subsets" )
+    JTREG=true; JTREG_TEST_SUBSETS="$1" shift;;
 
     *) echo >&2 "${error}Invalid option: ${opt}${normal}"; man ./makejdk.1; exit 1;;
    esac
@@ -218,8 +218,8 @@ else
   $WORKING_DIR/sbin/build.sh $WORKING_DIR $TARGET_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JVM_VARIANT
 
   if [[ ! -z $JTREG ]]; then
-    if [[ ! -z $JTREG_JDK_SUBSETS ]]; then
-      $WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JTREG_JDK_SUBSETS
+    if [[ ! -z $JTREG_TEST_SUBSETS ]]; then
+      $WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JTREG_TEST_SUBSETS
     else
       $WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME
     fi
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -77,7 +77,7 @@ while [[ $# -gt 0 ]] && [[ ."$1" = .-* ]] ; do
     JTREG=true; shift;;
     
     "--jtreg_subsets" )
-    JTREG=true; JTREG_TEST_SUBSETS="$1" shift;;
+    JTREG=true; JTREG_TEST_SUBSETS="$1"; shift;;
 
     *) echo >&2 "${error}Invalid option: ${opt}${normal}"; man ./makejdk.1; exit 1;;
    esac
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -75,6 +75,9 @@ while [[ $# -gt 0 ]] && [[ ."$1" = .-* ]] ; do
 
     "--jtreg" | "-j" )
     JTREG=true; shift;;
+    
+    "--jdk_subsets" )
+    JTREG=true; JTREG_JDK_SUBSETS="$1" shift;;
 
     *) echo >&2 "${error}Invalid option: ${opt}${normal}"; man ./makejdk.1; exit 1;;
    esac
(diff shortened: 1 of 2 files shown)


Bluevisitor:
The test run didn't work: https://ci.adoptopenjdk.net/job/adam_test/8/consoleFull

My local run continues to work.

Will investigate.

makejdk.sh
```diff
@@ -215,6 +218,10 @@ else
   $WORKING_DIR/sbin/build.sh $WORKING_DIR $TARGET_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JVM_VARIANT
 
   if [[ ! -z $JTREG ]]; then
-    $WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME
+    if [[ ! -z $JTREG_TEST_SUBSETS ]]; then
+      $WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JTREG_TEST_SUBSETS
```
karianna: Could we default subsets to be "all" or empty string or whatever indicates all?

Bluevisitor:
We're missing a semicolon from the non-local version. Doh.

makejdk.sh
```diff
@@ -75,6 +75,9 @@ while [[ $# -gt 0 ]] && [[ ."$1" = .-* ]] ; do
 
     "--jtreg" | "-j" )
     JTREG=true; shift;;
+    
+    "--jtreg_subsets" )
+    JTREG=true; JTREG_TEST_SUBSETS="$1" shift;;
```
karianna: See comment below, if --jtreg runs with a default of all, wouldn't need this separate option

sbin/jtreg.sh
```diff
@@ -91,7 +92,11 @@ settingUpEnvironmentVariablesForJTREG()
 runJtregViaMakeCommand()
 {
   echo "Running jtreg via make command (debug logs enabled)"
-  make test jobs=10 LOG=debug  
+  if [ -z $JTREG_TEST_SUBSETS }; then
```
karianna: See comment above, is it possible to default to all

makejdk.sh
```diff
@@ -75,6 +75,9 @@ while [[ $# -gt 0 ]] && [[ ."$1" = .-* ]] ; do
 
     "--jtreg" | "-j" )
     JTREG=true; shift;;
+    
+    "--jtreg_subsets" )
+    JTREG=true; JTREG_TEST_SUBSETS="$1"; shift;;
```
karianna: Could we default subsets to be "all" or empty string or whatever indicates all?

makejdk.sh
Bluevisitor: Currently, when we run the jtreg without options, we default to jdk_core and langtools (the latter doesn't seem to work, but jdk_core has enough coverage that we don't mind atm).

We could default to jdk_core, which gives us a core set of tests and removes the non-functional langtools, or we could switch to jdk_all. I don't know how long jdk_all takes, but I'm guessing it's longer than jdk_core. 

The latter already takes 25 minutes on our ultra-fast linux server.

sbin/jtreg.sh
Bluevisitor: Currently, when we run the jtreg without options, we default to jdk_core and langtools (the latter doesn't seem to work, but jdk_core has enough coverage that we don't mind atm).

We could default to jdk_core, which gives us a core set of tests and removes the non-functional langtools, or we could switch to jdk_all. I don't know how long jdk_all takes, but I'm guessing it's longer than jdk_core.

The latter already takes 25 minutes on our ultra-fast linux server.

makejdk.sh
Bluevisitor: Currently, when we run the jtreg without options, we default to jdk_core and langtools (the latter doesn't seem to work, but jdk_core has enough coverage that we don't mind atm).

We could default to jdk_core, which gives us a core set of tests and removes the non-functional langtools, or we could switch to jdk_all. I don't know how long jdk_all takes, but I'm guessing it's longer than jdk_core.

The latter already takes 25 minutes on our ultra-fast linux server.

makejdk.sh
karianna: Do we have a separate issue tracking why langtools is failing?

Ideally we'd default to jdk_all for release builds and yeah probably jdk_core for nightly builds.  So perhaps this script should default to jdk_core (a decent speed/coverage) compromise, but we can override that in our various build scripts.

IIRC jdk_all will also perform UI tests (Swing/AWT) - so will expose other missing deps that our build boxes may need.

Bluevisitor: I'm not aware of any issue tracking why langtools is failing. The error is:

/usr/bin/make -k -C  ../langtools/test  JT_JAVA=/home/jenkins/workspace/openjdk_build_x86-64_linux/openjdk/build/linux-x86_64-normal-server-release/images/j2sdk-image JTREG_HOME= TEST=jtreg jtreg
make[2]: Entering directory '/home/jenkins/workspace/openjdk_build_x86-64_linux/openjdk/langtools/test'
make[2]: *** No rule to make target '/bin/jtreg', needed by 'check-jtreg'.

If we specify jdk_core specifically, langtools won't be run. If we don't specify any test, then jck_core and langtools will be run by default.

The question is: do we still want to be running langtools?

Perhaps we should have an issue for langtools, and also specify jdk_core in the jenkins setup by default?

karianna: Ok, I'm adding that failure to the overall test issue #7 - I think we probably do want to be running langtools, I'll let @neomatrix369 comment over on #7

Bluevisitor:
The test run passed: https://ci.adoptopenjdk.net/job/adam_test/10

I see two problems here.

1) The .jtr.xml files we find the results xml in are not in the jdk_core folder if we're running alternate tests (like jdk_lang). We can resolve this by replacing "jdk_core" with a wildcard character in the post-build "Junit xml interpretation" stage.

2) The results files are also not in jdk_core, so our test harness is having a tough time uploading it. I plan to tweak the relevant scripts so they can accommodate this.

Bluevisitor:
Ok, I think we're all good.

Note: The RUN_JTREG settings (including default value and comment) needs to be copied to the main linux jenkins job once this change has been merged. Also, the "jdk_core" in the "location of results xml files" post-build step needs to be replaced with a "*" (sans quotes).

Bluevisitor:
Here's a test run to verify the changes.

https://ci.adoptopenjdk.net/view/all/job/adam_test/11/

makejdk.sh
karianna: Should JTREG_TEST_SUBSETS default to 'the default *before* having $1 (if $1 exists / is non empty) assigned to it?  That way you may not need to have the separate --jtreg_subsets flag?

makejdk.sh
karianna: Hoping that ` if [[ ! -z $JTREG_TEST_SUBSETS ]]; then` won't be required...

makejdk.sh
Bluevisitor: After our call I understand what you mean. Will make those changes now.

Bluevisitor: After playing with the shell script, I've determined that shifting after assigning $1 to it has the nasty side-effect of swallowing the next variable passed in on the command line.

so "./makejdk.sh --jtreg --potato" assigns "--potato" to JTREG_TEST_SUBSETS, and removes --potato from the list of arguments to be processed.

If we don't want to add another "if", what we could do is require a value for jtreg.

'./makejdk.sh --jtreg "" --potato' works, for example.

Will make that change. Let me know what you think when you get back.

Bluevisitor: karianna suggests that we're making a mountain out of a molehill here, so he proposed (and I agreed) that we should add the functionality now and tidy it up later when we have spare time.

Bluevisitor:
Thanks. :)

PR #37, opened 2017-04-06
Refactor build script

neomatrix369:
Note: please do not merge this PR till previous PR (https://github.com/AdoptOpenJDK/openjdk-build/pull/36) has been merged. Once its are merge git will show the actual change (as commits).

- Refactored blocks of code in build.sh into respective functions. 
- Made function name consistent across scripts in one case.

PR #44, opened 2017-04-10
Build script dependencies refactor

neomatrix369:
- Externalised the dependencies for
   - Alsa
   - FreeType font
   (these can/should be overriden via environment variables, in the absence of them it will default to hard-coded values)
- Corrected spellings

- Uniform usages of {...}

PR created at the back of review comments on PR #37.

### DIFF
--- a/ansible/playbooks/centos_72_ppc64le.yml
+++ b/ansible/playbooks/centos_72_ppc64le.yml
@@ -1,5 +1,5 @@
 ---
-- hosts: rhel6_s390x
+- hosts: centos72_ppc64le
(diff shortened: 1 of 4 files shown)


neomatrix369:
I was wondering if the dependency version strings should contain ".tar.gz" as well? or is it a bit far fetched?

karianna:
@neomatrix369 Too far, I think we just want pure version numbers

neomatrix369:
If this is good, please merge so I can create some more to tidy up existing scripts - we are getting there soon

PR #48, opened 2017-04-11
Do not run tests if not specified

neomatrix369:
Fix to change made in the same area as changed by PR https://github.com/Bluevisitor/openjdk-build/commit/be63d41343d235dcb7a2c82472643157e82a85f8

Strangely this line was causing the issue:
```if [[ ! -z $JTREG ]]; then```

`If` block was executed irrespective of passing `--jtreg` or not

We couldn't have spotted this via pull requests, maybe we should discourage such constructs (as the above) reducing the chances of bugs slipping thru - @karianna @gdams any thoughts?

### DIFF
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -260,8 +260,10 @@ buildAndTestOpenJDKViaDocker()
 
 testOpenJDKInNativeEnvironmentIfExpected()
 {
-  if [[ ! -z $JTREG ]]; then
-    if [[ ! -z $JTREG_TEST_SUBSETS ]]; then
+  if [[ "$JTREG" == "true" ]];
+  then
+    if [[ ! -z $JTREG_TEST_SUBSETS ]];
+    then
       $WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JTREG_TEST_SUBSETS
     else
       $WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -260,7 +260,7 @@ buildAndTestOpenJDKViaDocker()
 
 testOpenJDKInNativeEnvironmentIfExpected()
 {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by neomatrix369** [comment]
Just ran a build at https://ci.adoptopenjdk.net/job/mani-test-x86-64_linux/1/console - was successful - did not run the jtreg tests

File: makejdk.sh [hunk 1]
```diff
@@ -260,12 +260,9 @@ buildAndTestOpenJDKViaDocker()
 
 testOpenJDKInNativeEnvironmentIfExpected()
 {
-  if [[ ! -z $JTREG ]]; then
-    if [[ ! -z $JTREG_TEST_SUBSETS ]]; then
+  if [[ "${JTREG}" == "true" ]];
+  then
```
karianna: Will this work if `$JTREG_TEST_SUBSETS` is not set?

gdams: we shouldn't need quotes as well as curly braces here...

neomatrix369: @karianna should work the build jobs above (see links) dont set `JTREG_TEST_SUBSETS`, the below line
`$WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JTREG_TEST_SUBSETS` becomes `$WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME`

neomatrix369: @gdams this is the line that failed strangely, so best its clear and consistent (we already have scripts that now use "${...}" ), also see conversation https://github.com/AdoptOpenJDK/openjdk-build/pull/43#discussion_r110635712. Not too fused just lets follow one standard.

karianna: @neomatrix369 I agree with @gdams here, I think we can go with:

`if [[ ${JTREG} == "true" ]];`

We can fix the other scripts later

neomatrix369: So we agree that all if block conditions will be represented as for all our scripts:  
`if [[ ${...} == "...." ]];`

neomatrix369:
Feel free to do the ceremonies if rest looks good, I'll amend the jenkins job to accept the new String param

PR #50, opened 2017-04-12
Pass make arguments as env variable in build-script

neomatrix369:
Extracted hard-coded params to the make command to environment variables, defaults set when these variables are not set

This should help when experimenting with cmd line args for make for different platforms from within jenkins or via scripts 

Should partially help issue https://github.com/AdoptOpenJDK/openjdk-build/issues/49


neomatrix369:
Build job kicked off - https://ci.adoptopenjdk.net/job/openjdk_build_x86-64_linux_mani_test/2/console to see if the defaults are being picked up, once merged we need to add two more params to jenkins job all across the board

neomatrix369:
And one for ARM platform - https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk_build_arm_linux_mani_test/1/console

sbin/build.sh
```diff
@@ -20,10 +20,16 @@ WORKING_DIR=$1
...
 BUILD_FULL_NAME=$4
+
 JVM_VARIANT=${5:-normal}
+
+OS_MACHINE=$(uname -m)
+
 ALSA_LIB_VERSION=${ALSA_LIB_VERSION:-1.0.27.2}
 FREETYPE_FONT_SHARED_OBJECT_FILENAME=libfreetype.so.6.5.0
 FREETYPE_FONT_VERSION=${FREETYPE_FONT_VERSION:-2.4.0}
+MAKE_ARGS_FOR_ALL_PLATFORMS=${MAKE_ARGS_FOR_ALL_PLATFORMS:-"images"}
```
neomatrix369: We can even do away with two different env variables - wish to discuss this?

sbin/build.sh
```diff
@@ -224,9 +230,9 @@ runTheOpenJDKConfigureCommandAndUseThePrebuildConfigParams()
 buildOpenJDK()
 {
   if [[ "$OS_MACHINE" == "s390x" ]] || [[ "$OS_MACHINE" == "armv7l" ]] ; then
```
neomatrix369: if block can vanish, we can store the logic in the jenkins job itself

tellison: The scripts need to run outside of Jenkins too.
To ensure our build process is transparent and reproducible, there should be as little as possible in the scripts that depends upon our specific build environment -- and certainly nothing from the build environment that results in modifications to the final binary.

neomatrix369: I would suggest a documentation in this case and passing the environment setting during build:
`export MAKE_ARGS_FOR_ALL_PLATFORMS="..." makejdk.sh {....}` could be a good start

So for ARM the command should look like 
`export MAKE_ARGS_FOR_ALL_PLATFORMS=" DEBUG_BINARIES=true all" makejdk.sh {....}` could be a good start.

Another option could be to maintain a script that we only use when running outside jenkins which has the checks for the different platform and passes the respective `MAKE_ARGS_FOR_ALL_PLATFORMS` settings.

`MAKE_ARGS_FOR_ALL_PLATFORMS` will also have a default setting of `images`

Which one do you think makes more sense?
I prefer the maintaining the top-level script which would mean it is the doc by itself and would be version controlled. Any other ideas?

karianna: I'm OK with the parameter / logic being passed in, but as Tim says it needs to be buildable from the CL for anyone.  Might be worth just leaving for now.

neomatrix369: We can still make it buildable from CLI, as I had suggested to provide a script i.e make-from-local-machine.sh that called makejdk with the different ENV variables.

karianna:
@neomatrix369 You'll need to resolve the conflicts here now :-)

sbin/build.sh
karianna: Ah I see, yeah I think that would be OK

neomatrix369:
Maybe the additional enhancement I spoke about can be done at a later stage, I'll create an issue on this

neomatrix369:
@karianna basically abstracting the dependencies and conditional checks outside this script and make it controllable via Env variables and/or a supporting script

Issue #53, opened 2017-04-12
Refactor jtreg_prep script (and dependent scripts)


neomatrix369:
The functionality added by https://github.com/AdoptOpenJDK/openjdk-build/pull/43 is useful and fulfills its purpose. On the back of the discussions that took place on the PR, we should consider the below:
- make script similar in format as the sbin/build.sh
- maintain consistency between the use of `${}` and `"${}"`
- any other tech debt that was left for resolution at this stage (refer to discussions on the PR)
- adding missing documentation

neomatrix369:
#58 helps this issue partially

neomatrix369:
Closing issue as related PRs are merged

PR #54, opened 2017-04-12
Refactoring jtreg prep script and dependent scripts

neomatrix369:
Fully resolves issue #53. Amends all .sh scripts in the repo that has to do with jenkins or local builds. 

Beware there are plenty of small changes, mainly due to the replacement of the existing use of $VAR, ${VAR} and "${VAR}" with the respective versions depending on where the VAR was placed:

- use "$VAR" in if conditions
- use ${VAR} if it was part of another string/variable i.e. "The source is ${VAR}"
- use $VAR in all other circumstances

See the [ref guide](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)

- Added documentation and updated both short and long param flags

### DIFF
--- a/makejdk.1
+++ b/makejdk.1
@@ -26,6 +26,9 @@ reuse docker container (prevents deleting)
 .BR \-j ", " \-\-jtreg
 run jtreg after building
 .TP
+.BR \-js ", " \-\-jtreg-subsets
+select one or more jtreg tests to run
+.TP
 .BR \-S ", " \-\-ssh
 use ssh when cloning git
 .TP
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -95,7 +95,7 @@ parseCommandLineArgs()
       "--jtreg" | "-j" )
       JTREG=true; shift;;
 
-    "--jtreg_subsets" )
+      "--jtreg-subsets" | "-js" )
     JTREG=true; JTREG_TEST_SUBSETS="$1"; shift;;
 
       *) echo >&2 "${error}Invalid option: ${opt}${normal}"; man ./makejdk.1; exit 1;;
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -38,7 +38,7 @@ fi
 
 BUILD_TYPE=normal
 
-BUILD_FULL_NAME=$OS_KERNAL_NAME-$OS_MACHINE-$BUILD_TYPE-$JVM_VARIANT-release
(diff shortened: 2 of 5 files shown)


neomatrix369:
Running a number of builds to do sanity check, see https://ci.adoptopenjdk.net/view/work%20in%20progress/

gibfahn:
>use "$VAR" in if conditions
>use ${VAR} if it was part of another string/variable i.e. "The source is ${VAR}"
>use $VAR in all other circumstances
>Note: ${VAR} is same as "$VAR", hence "${VAR}" is a bit overloaded

I don't think that's true, I'd say:
 - use `"${VAR}"` if you can't use `"$VAR"`
 - use `"$VAR"` otherwise

Double quotes are always good, you need them to handle spaces in your input (you have to be careful with spaces in bash).

The `{}` is useful when the character after the variable name is `a to z` or `_`, e.g.:

```bash
VAR=foo
[ "$VARbar" = foobar ] && echo true # Returns false as it thinks $VARbar is a variable name
[ "${VAR}bar" = foobar ] && echo true # Returns true, concatenates $VAR and "bar"
```

there are also some cool things you can do with `${}`:

```bash
echo ${VAR:-default} # Echo's $VAR, or "default" if $VAR is undefined
```

and some people just think they look neater (I find they usually don't help, but YMMV):

```bash
git clone "https://github.com/${ORG}/${REPO}"
git clone "https://github.com/$ORG/$REPO"    # Personal preference
```

karianna:
Thanks @gibfahn - that makes it much clearer.  I think we should add your comment to the Contributing.md file.

tellison:
@gibfahn has it right.  Also see [previous discussion](https://github.com/AdoptOpenJDK/openjdk-build/pull/58#discussion_r111420127).

There is a good, clear explanation on the [shellcheck wiki](https://github.com/koalaman/shellcheck/wiki/SC2086).  Take a read of their "minimal" and "canonical" style suggestion.  I don't have a strong feeling of which we adopt.

karianna:
@neomatrix369 - Conflicts :-).  @tellison and @gibfahn I learned something new today, thanks! I have a slight preference for canonical, but in all honesty I'll leave that decision to those actually doing the coding :-).

neomatrix369:
Temporarily closing PR till fixes the conflicts, too many of them caused due to other PRs

Issue #30, opened 2017-04-05
Compute the available threads/cores/whichever instead of hard-coding a number


neomatrix369:
When running tests we set CONCURRENCY to 8, and leave it like that for all machines - fast or slow.

Its best to find out (and its possible to get a accurate number) from the OS and use this number instead.

If this process fails we can agree to use 8 or another number.

Applicable to this [jtreg.sh script](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/jtreg.sh) and used by jenkins jobs i.e. https://ci.adoptopenjdk.net/job/openjdk_build_x86-64_linux/

tellison:
Have you tried using `getconf _NPROCESSORS_ONLN` from gcc?

neomatrix369:
@tellison thanks for the tip - will take a look at it when investigating this issue

gdams:
we could just add an env var to our build machines?

neomatrix369:
ENV is a good idea, although would be nice to auto generate it

tellison:
I'm going through and adding a (more) useful Jenkins description to the build machines now.  I'll add a `NUM_PROCESSORS` env var to each node while I'm there.

The scripts can use that var, and should have a sensible default value for when the env var is missing.

neomatrix369:
Potentially a good idea, but how do you compute this value (`NUM_PROCESSORS`) ? Is it manual or information from the platform (HW/OS)?

tellison:
I'm adding it manually.  As I wrote above, you can use `getconf _NPROCESSORS_ONLN` to get it from the OS.

neomatrix369:
Ok thanx for clarifying - new PR on the way i think ;)

neomatrix369:
ARM build does not make use of the cores provided, scales down to using 1 core only, see https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk_build_arm_linux_mani_test/6/console, check if this logic is specific to ARM builds

PR #55, opened 2017-04-12
Parameterise number of cores for test runs (jtreg)

neomatrix369:
Injected the environment variable NUM_PROCESSORS into the jtreg script to allow parameterising the number of cores to use when running jtreg tests.

Fixes issues #30


neomatrix369:
Fixes issues #30

Issue #51, opened 2017-04-12
Further refactoring on build script for eliminating platform specific settings 


neomatrix369:
On the back of the PR https://github.com/AdoptOpenJDK/openjdk-build/pull/50 we discussed that we want to be able to build OpenJDK outside Jenkins so the same script should be usable for both local machine and jenkins build.

PR #56, opened 2017-04-12
Parameterise JVM_VARIANT for jenkins and local env builds

neomatrix369:
- Parameterised `JVM_VARIANT` so environment variable(s) can be injected into the script when running via jenkins or via an external script
- Valid values for `JVM_VARIANT` are `server`, `client`, `minimal1`, `kernel`, `zero`, `zeroshark`, `core`
- A `String` parameter by the name `JVM_VARIANT` will need to be added to all the jenkins jobs with the respective values set as its default value in order for the correct images to be created from the builds  
- Introduced a new script (`makejdk-local-env.sh`) to enable local environment builds (outside jenkins)

Partially related to issue #51. Once this is in place, #51 should be easier to implement


makejdk-local-env.sh
```diff
@@ -0,0 +1,37 @@
...
+# and the target destination for the tar.gz e.g. -d or --destination $HOME/mytargetdir
+# Both must be absolute paths! You can use $PWD/mytargetdir
+
+# A simple way to install dependencies persistently is to use our Ansible playbooks
+
+# You can set the JDK boot directory with the JDK_BOOT_DIR environment variable
+
+set -eu
+
+OS_KERNAL_NAME=$(echo $(uname) | awk '{print tolower($0)}')
```
tellison: `OS_KERNAL_NAME` -> `OS_KERNEL_NAME`

Should `OS_KERNEL_NAME` be exported (it is not used)?
What is the purpose of the  `echo` ?

neomatrix369: echo => sets the final outcome of the whole command `$(uname) | awk '{print tolower($0)}'` to `OS_KERNEL_NAME` - otherwise the output might get executed

neomatrix369: Fixed `OS_KERNEL_NAME` via 3c7e1906c8c5e04c848cba33f02b13e2ef68eb28

PR #57, opened 2017-04-12
Travis file for lint checking script files

tellison:
This will run shellcheck on all files ending with '.sh' and report any static issues found.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: bash
+
+# Use container-based infrastructure for quicker build start-up
+sudo: false
+
+addons:
+  apt:
+    sources:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: bash
+
+# Use container-based infrastructure for quicker build start-up
+sudo: false
```
karianna: sudo: false means " Use container-based infrastructure for quicker build start-up"?

tellison: Yes, see [the Travis docs](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments) for details.  We don't need sudo access for this script execution.

PR #58, opened 2017-04-13
Fix shell script lint check errors

tellison:
This fixes a number of warnings and errors from running the shell lint checker.
There are some real bugs (e.g. use of undefined vars), but most are safety features; notably 
1. using quotes around var expansion to avoid word splitting and globbing by the shell,
1. fail safe when `cd`'ing to a directory, especially where we `cd` then `rf -f`

Special bonus addition is the image in our README.md to show the testing status.

### DIFF
--- a/git-hg/git-hg.sh
+++ b/git-hg/git-hg.sh
@@ -18,7 +18,7 @@ hg clone http://hg.openjdk.java.net/jdk8u/jdk8u/ openjdk_hg
 
 cd openjdk_hg || exit
 # get latest mercurial tag
-hgTag=`hg log -r "." --template "{latesttag}\n"`
+#hgTag=`hg log -r "." --template "{latesttag}\n"`
 
 bash ./get_source.sh
 for i in corba jaxp jaxws langtools jdk hotspot nashorn; do
@@ -50,7 +50,7 @@ git commit -m "merge sources"
 git remote add releases git@github.com:AdoptOpenJDK/openjdk-jdk8u.git
 git fetch --all
 # check if git diff is the same
-if [ `git diff releases/master | wc -l` -gt 0 ]; then
+if [ $(git diff releases/master | wc -l) -gt 0 ]; then
 	git checkout -b staging
 	git checkout master
 	git reset --hard releases/master
--- a/git-hg/git-hg.sh
+++ b/git-hg/git-hg.sh
@@ -50,7 +50,7 @@ git commit -m "merge sources"
 git remote add releases git@github.com:AdoptOpenJDK/openjdk-jdk8u.git
 git fetch --all
 # check if git diff is the same
-if [ $(git diff releases/master | wc -l) -gt 0 ]; then
+if [ "$(git diff releases/master | wc -l)" -gt 0 ]; then
 	git checkout -b staging
 	git checkout master
 	git reset --hard releases/master
--- a/git-hg/git-hg.sh
+++ b/git-hg/git-hg.sh
@@ -37,13 +37,13 @@ git init
 "$WORKSPACE"/fast-export/hg-fast-export.sh -r "$WORKSPACE"/openjdk_hg
 for i in corba jaxp jaxws langtools jdk hotspot nashorn; do
 	# copy the additional repos in
-	cp -r "$WORKSPACE"/openjdk_hg/$i $i
+	cp -r "${WORKSPACE}"/openjdk_hg/$i $i
 done
 git checkout
 # remove remaining mercurial stuff
 rm -rf .hg .hgignore .hgtags README-builds.html README get_source.sh
 # create new README
-cp "$WORKSPACE"/openjdk-build/git-hg/README.md .
+cp "${WORKSPACE}"/openjdk-build/git-hg/README.md .
 echo ".hg/" > .gitignore
 git add .
 git commit -m "merge sources"
--- a/git-hg/git-hg.sh
+++ b/git-hg/git-hg.sh
@@ -16,7 +16,7 @@
 git clone https://github.com/frej/fast-export.git
 hg clone http://hg.openjdk.java.net/jdk8u/jdk8u/ openjdk_hg
 
-cd openjdk_hg
+cd openjdk_hg || exit
 # get latest mercurial tag
 hgTag=`hg log -r "." --template "{latesttag}\n"`
 
@@ -24,26 +24,26 @@ bash ./get_source.sh
 for i in corba jaxp jaxws langtools jdk hotspot nashorn; do
    echo "cleaning up $i"
    # removes the mercurial stuff from the additional repos
-   cd $i
+   cd $i || exit
    rm -rf .hg .hgignore .hgtags
-   cd -
+   cd - || exit
 done
-cd $WORKSPACE
+cd "$WORKSPACE" || exit
 
 mkdir openjdk_git
-cd openjdk_git
+cd openjdk_git || exit
 git init
 # convert mercurial to github
-$WORKSPACE/fast-export/hg-fast-export.sh -r $WORKSPACE/openjdk_hg
+"$WORKSPACE"/fast-export/hg-fast-export.sh -r "$WORKSPACE"/openjdk_hg
 for i in corba jaxp jaxws langtools jdk hotspot nashorn; do
 	# copy the additional repos in
-	cp -r $WORKSPACE/openjdk_hg/$i $i
+	cp -r "$WORKSPACE"/openjdk_hg/$i $i
 done
 git checkout
 # remove remaining mercurial stuff
 rm -rf .hg .hgignore .hgtags README-builds.html README get_source.sh
 # create new README
-cp $WORKSPACE/openjdk-build/git-hg/README.md .
+cp "$WORKSPACE"/openjdk-build/git-hg/README.md .
 echo ".hg/" > .gitignore
 git add .
 git commit -m "merge sources"
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -27,18 +27,17 @@
 REPOSITORY=AdoptOpenJDK/openjdk-jdk8u
 OPENJDK_REPO_NAME=openjdk
 
-OS_KERNEL_NAME=$(uname) | awk '{print tolower($0)}'
-
-OS_MACHINE=$(uname -m)
+OS_KERNEL_NAME=$(uname | awk '{print tolower($0)}')
+OS_CPU_NAME=$(uname -m)
 
 JVM_VARIANT=server
-if [[ "$OS_MACHINE" == "s390x" ]] || [[ "$OS_MACHINE" == "armv7l" ]] ; then
+if [[ "$OS_CPU_NAME" == "s390x" ]] || [[ "$OS_CPU_NAME" == "armv7l" ]] ; then
  JVM_VARIANT=zero
 fi 
 
 BUILD_TYPE=normal
 
-BUILD_FULL_NAME=$OS_KERNEL_NAME-$OS_MACHINE-$BUILD_TYPE-$JVM_VARIANT-release
+BUILD_FULL_NAME=$OS_KERNEL_NAME-$OS_CPU_NAME-$BUILD_TYPE-$JVM_VARIANT-release
 
 USE_DOCKER=false
 WORKING_DIR=""
@@ -187,7 +186,7 @@ cloneOpenJDKGitRepo()
       echo "git clone -b ${BRANCH} git@github.com:${REPOSITORY}.git ${WORKING_DIR}/${OPENJDK_REPO_NAME}"
       git clone -b ${BRANCH} git@github.com:"${REPOSITORY}".git "${WORKING_DIR}/${OPENJDK_REPO_NAME}"
     else
-      echo "git clone -b ${BRANCH} https://github.com/"${REPOSITORY}".git "${WORKING_DIR}/${OPENJDK_REPO_NAME}""
+      echo "git clone -b ${BRANCH} https://github.com/${REPOSITORY}.git ${WORKING_DIR}/${OPENJDK_REPO_NAME}"
       git clone -b ${BRANCH} https://github.com/"${REPOSITORY}".git "$WORKING_DIR/$OPENJDK_REPO_NAME"
     fi
   fi
@@ -240,14 +239,14 @@ buildAndTestOpenJDKViaDocker()
   CONTAINER_ID=$(docker ps -a | awk '{ print $1,$2 }' | grep openjdk_container | awk '{print $1 }'| head -1)
 
   if [[ "${COPY_TO_HOST}" == "true" ]] ; then
-    echo "Copying to the host with docker cp "$CONTAINER_ID":/openjdk/jdk8u/OpenJDK.tar.gz $TARGET_DIR"
+    echo "Copying to the host with docker cp $CONTAINER_ID:/openjdk/jdk8u/OpenJDK.tar.gz $TARGET_DIR"
     docker cp "$CONTAINER_ID":/openjdk/jdk8u/OpenJDK.tar.gz "$TARGET_DIR"
   fi
 
   if [[ "${JTREG}" == "true" ]] ; then
     echo "Copying jtreg reports from docker"
-    docker cp $CONTAINER_ID:/openjdk/jdk8u/jtreport.zip $TARGET_DIR
-    docker cp $CONTAINER_ID:/openjdk/jdk8u/jtwork.zip $TARGET_DIR
+    docker cp "$CONTAINER_ID":/openjdk/jdk8u/jtreport.zip "$TARGET_DIR"
+    docker cp "$CONTAINER_ID":/openjdk/jdk8u/jtwork.zip "$TARGET_DIR"
   fi
 
   # Didn't specify to keep
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -287,7 +287,7 @@ buildAndTestOpenJDK()
 
 initialiseEscapeCodes
 sourceSignalHandler
-parseCommandLineArgs $@
+parseCommandLineArgs "$@"
 checkIfDockerIsUsedForBuildingOrNot
 checkInCaseOfDockerShouldTheContainerBePreserved
 setDefaultIfBranchIsNotProvided
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -28,15 +28,15 @@ REPOSITORY=AdoptOpenJDK/openjdk-jdk8u
 OPENJDK_REPO_NAME=openjdk
 
 OS_KERNEL_NAME=$(uname | awk '{print tolower($0)}')
-OS_CPU_NAME=$(uname -m)
+OS_MACHINE_NAME=$(uname -m)
 
 JVM_VARIANT=server
-if [[ "$OS_CPU_NAME" == "s390x" ]] || [[ "$OS_CPU_NAME" == "armv7l" ]] ; then
+if [[ "$OS_MACHINE_NAME" == "s390x" ]] || [[ "$OS_MACHINE_NAME" == "armv7l" ]] ; then
  JVM_VARIANT=zero
 fi 
 
 BUILD_TYPE=normal
-BUILD_FULL_NAME=$OS_KERNEL_NAME-$OS_CPU_NAME-$BUILD_TYPE-$JVM_VARIANT-release
+BUILD_FULL_NAME=${OS_KERNEL_NAME}-${OS_MACHINE_NAME}-${BUILD_TYPE}-${JVM_VARIANT}-release
 USE_DOCKER=false
 WORKING_DIR=""
 USE_SSH=false
@@ -142,10 +142,10 @@ setDefaultIfBranchIsNotProvided()
 setWorkingDirectoryIfProvided()
 {
   if [ -z "${WORKING_DIR}" ] ; then
-    echo "${info}WORKING_DIR is undefined so setting to $PWD${normal}."
+    echo "${info}WORKING_DIR is undefined so setting to ${PWD}${normal}."
     WORKING_DIR=$PWD
   else
-    echo "${info}Working dir is $WORKING_DIR${normal}."
+    echo "${info}Working dir is ${WORKING_DIR}${normal}."
   fi
 }
 
@@ -185,7 +185,7 @@ cloneOpenJDKGitRepo()
       git clone -b ${BRANCH} git@github.com:"${REPOSITORY}".git "${WORKING_DIR}/${OPENJDK_REPO_NAME}"
     else
       echo "git clone -b ${BRANCH} https://github.com/${REPOSITORY}.git ${WORKING_DIR}/${OPENJDK_REPO_NAME}"
-      git clone -b ${BRANCH} https://github.com/"${REPOSITORY}".git "$WORKING_DIR/$OPENJDK_REPO_NAME"
+      git clone -b ${BRANCH} https://github.com/"${REPOSITORY}".git "${WORKING_DIR}/${OPENJDK_REPO_NAME}"
     fi
   fi
   echo "${normal}"
@@ -230,7 +230,7 @@ buildAndTestOpenJDKViaDocker()
      echo "$normal"
   fi
 
-  docker run --privileged -t -v "$WORKING_DIR/$OPENJDK_REPO_NAME:/openjdk/jdk8u/openjdk" --entrypoint build.sh "$CONTAINER"
+  docker run --privileged -t -v "${WORKING_DIR}/${OPENJDK_REPO_NAME}:/openjdk/jdk8u/openjdk" --entrypoint build.sh "${CONTAINER}"
 
   testOpenJDKViaDocker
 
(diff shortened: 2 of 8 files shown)


git-hg/git-hg.sh
```diff
@@ -16,41 +16,41 @@
...
-cd openjdk_git
+cd openjdk_git || exit
 git init
 # convert mercurial to github
-$WORKSPACE/fast-export/hg-fast-export.sh -r $WORKSPACE/openjdk_hg
+"$WORKSPACE"/fast-export/hg-fast-export.sh -r "$WORKSPACE"/openjdk_hg
 for i in corba jaxp jaxws langtools jdk hotspot nashorn; do
 	# copy the additional repos in
-	cp -r $WORKSPACE/openjdk_hg/$i $i
+	cp -r "$WORKSPACE"/openjdk_hg/$i $i
```
neomatrix369: Why not use "${...} something ${i}" format? reads better and also safe

git-hg/git-hg.sh
```diff
@@ -16,41 +16,41 @@
...
 done
 git checkout
 # remove remaining mercurial stuff
 rm -rf .hg .hgignore .hgtags README-builds.html README get_source.sh
 # create new README
-cp $WORKSPACE/openjdk-build/git-hg/README.md .
+cp "$WORKSPACE"/openjdk-build/git-hg/README.md .
```
neomatrix369: same here as above - this will be in line with the guideline on https://github.com/AdoptOpenJDK/openjdk-build/pull/54

git-hg/git-hg.sh
```diff
@@ -16,41 +16,41 @@
...
 echo ".hg/" > .gitignore
 git add .
 git commit -m "merge sources"
 git remote add releases git@github.com:AdoptOpenJDK/openjdk-jdk8u.git
 git fetch --all
 # check if git diff is the same
-if [ `git diff releases/master | wc -l` -gt 0 ]; then
+if [ "$(git diff releases/master | wc -l)" -gt 0 ]; then
```
neomatrix369: :+1:

makejdk.sh
```diff
@@ -27,15 +27,16 @@
 REPOSITORY=AdoptOpenJDK/openjdk-jdk8u
 OPENJDK_REPO_NAME=openjdk
 
-OS_KERNEL_NAME=$(echo $(uname) | awk '{print tolower($0)}')
+OS_KERNEL_NAME=$(uname | awk '{print tolower($0)}')
+OS_CPU_NAME=$(uname -m)
```
neomatrix369: minor: `uname` terminology would be machine or machine name

makejdk.sh
```diff
@@ -141,67 +142,65 @@ setDefaultIfBranchIsNotProvided()
...
 }
 
 cloneOpenJDKGitRepo()
 {
-  echo $git
+  echo "${git}"
   if [ -d "${WORKING_DIR}/${OPENJDK_REPO_NAME}/.git" ] && [ "$REPOSITORY" == "AdoptOpenJDK/openjdk-jdk8u" ] ; then
     # It does exist and it's a repo other than the AdoptOpenJDK one
-    cd $WORKING_DIR/$OPENJDK_REPO_NAME
+    cd "${WORKING_DIR}/${OPENJDK_REPO_NAME}" || return
```
neomatrix369: :+1: I was referring to this convention above

makejdk.sh
```diff
@@ -141,67 +142,65 @@ setDefaultIfBranchIsNotProvided()
...
     echo "${info}Will reset the repository at $PWD in 10 seconds...${git}"
     sleep 10
     echo "${git}Pulling latest changes from git repo"
     git fetch --all
     git reset --hard origin/$BRANCH
-    echo $normal
-    cd $WORKING_DIR
+    echo "${normal}"
+    cd "${WORKING_DIR}" || return
```
neomatrix369: We had a discussion on one of the PR that "${}" is the same as ${}, i noted it for our guideline (see https://github.com/AdoptOpenJDK/openjdk-build/pull/54)

git-hg/git-hg.sh
```diff
@@ -16,41 +16,41 @@
 git clone https://github.com/frej/fast-export.git
```
neomatrix369: Any reason for not setting the `set -eu` at the first line, this will make the `|| exit` and `|| return` unnecessary if we want to script to fail-fast, i would vote for fail-fast. Unless you want a graceful exit for a good reason, there could be disadvantages to graceful exits in some cases.

sbin/signalhandler.sh
```diff
@@ -1,6 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# For terminal colors
+esc=$(echo -en "\033")
```
neomatrix369: We use these values repeatedly across multiple scripts, idea: extract into a single .sh file and source that file from all the scripts, will help maintain consistency and a central place to change or add new colour codes

sbin/signalhandler.sh
```diff
@@ -1,6 +1,27 @@
...
+error="${esc}[0;31m"
+normal=$(echo -en "${esc}[m\017")
+
+
 exit_script() {
     if [[ -z ${KEEP} ]] ; then
-      docker ps -a | awk '{ print $1,$2 }' | grep $CONTAINER | awk '{print $1 }' | xargs -I {} docker rm -f {}
+      docker ps -a | awk '{ print $1,$2 }' | grep "$CONTAINER" | awk '{print $1 }' | xargs -I {} docker rm -f {}
```
neomatrix369: maybe use {} instead of "", all depends on which convention we settle down for

makejdk.sh
```diff
@@ -27,15 +27,16 @@
...
-BUILD_TYPE=${BUILD_TYPE:-normal}
-
-BUILD_FULL_NAME=$OS_KERNEL_NAME-$OS_MACHINE-$BUILD_TYPE-$JVM_VARIANT-release
+JVM_VARIANT=server
+if [[ "$OS_CPU_NAME" == "s390x" ]] || [[ "$OS_CPU_NAME" == "armv7l" ]] ; then
+ JVM_VARIANT=zero
+fi 
 
+BUILD_TYPE=normal
+BUILD_FULL_NAME=$OS_KERNEL_NAME-$OS_CPU_NAME-$BUILD_TYPE-$JVM_VARIANT-release
```
neomatrix369: I used {} for each of the variables in my PR

makejdk.sh
```diff
@@ -141,67 +142,65 @@ setDefaultIfBranchIsNotProvided()
 setWorkingDirectoryIfProvided()
 {
   if [ -z "${WORKING_DIR}" ] ; then
-    echo "${info}WORKING_DIR is undefined so setting to $PWD${normal}"
+    echo "${info}WORKING_DIR is undefined so setting to $PWD${normal}."
```
neomatrix369: `{}` around `PWD` ==> `${PWD}`

makejdk.sh
```diff
@@ -141,67 +142,65 @@ setDefaultIfBranchIsNotProvided()
...
     WORKING_DIR=$PWD
   else
-    echo "${info}Working dir is $WORKING_DIR${normal}"
+    echo "${info}Working dir is $WORKING_DIR${normal}."
```
neomatrix369: same as above for WORKING_DIR

makejdk.sh
```diff
@@ -141,67 +142,65 @@ setDefaultIfBranchIsNotProvided()
...
+    docker run --privileged -t -v "${WORKING_DIR}/${OPENJDK_REPO_NAME}":/openjdk/jdk8u/openjdk --entrypoint jtreg.sh "${CONTAINER}"
   fi
 }
 
 buildAndTestOpenJDKViaDocker()
 {
-  PS_DOCKER=$(ps -ef | grep "docker" | wc -l)
-
-  if [ -z $(which docker) ] || [ ${PS_DOCKER} -lt 2 ]; then
+  if [ -z "$(which docker)" ] || [ -z "$(pgrep "docker")" ]; then
```
neomatrix369: :+1:

makejdk.sh
```diff
@@ -219,33 +218,33 @@ buildAndTestOpenJDKViaDocker()
...
-     echo $good
+     echo "$good"
      docker ps -a | awk '{ print $1,$2 }' | grep $CONTAINER | awk '{print $1 }' | xargs -I {} docker rm -f {}
      docker build -t $CONTAINER docker/jdk8u/x86_64/ubuntu
-     echo $normal
+     echo "$normal"
   fi
 
-  docker run --privileged -t -v $WORKING_DIR/$OPENJDK_REPO_NAME:/openjdk/jdk8u/openjdk --entrypoint build.sh $CONTAINER
+  docker run --privileged -t -v "$WORKING_DIR/$OPENJDK_REPO_NAME:/openjdk/jdk8u/openjdk" --entrypoint build.sh "$CONTAINER"
```
neomatrix369: `{}` for all the ENV variables in this line?

makejdk.sh
```diff
@@ -219,33 +218,33 @@ buildAndTestOpenJDKViaDocker()
...
 
   testOpenJDKViaDocker
 
   CONTAINER_ID=$(docker ps -a | awk '{ print $1,$2 }' | grep openjdk_container | awk '{print $1 }'| head -1)
 
   if [[ "${COPY_TO_HOST}" == "true" ]] ; then
-    echo "Copying to the host with docker cp $id:/openjdk/jdk8u/OpenJDK.tar.gz $TARGET_DIR"
-    docker cp $CONTAINER_ID:/openjdk/jdk8u/OpenJDK.tar.gz $TARGET_DIR
+    echo "Copying to the host with docker cp $CONTAINER_ID:/openjdk/jdk8u/OpenJDK.tar.gz $TARGET_DIR"
```
neomatrix369: {} for the ENV variables, instead of the ""

makejdk.sh
```diff
@@ -219,33 +218,33 @@ buildAndTestOpenJDKViaDocker()
...
+    docker cp "$CONTAINER_ID":/openjdk/jdk8u/OpenJDK.tar.gz "$TARGET_DIR"
   fi
 
   if [[ "${JTREG}" == "true" ]] ; then
     echo "Copying jtreg reports from docker"
-    docker cp $CONTAINER_ID:/openjdk/jdk8u/jtreport.zip $TARGET_DIR
-    docker cp $CONTAINER_ID:/openjdk/jdk8u/jtwork.zip $TARGET_DIR
+    docker cp "$CONTAINER_ID":/openjdk/jdk8u/jtreport.zip "$TARGET_DIR"
+    docker cp "$CONTAINER_ID":/openjdk/jdk8u/jtwork.zip "$TARGET_DIR"
```
neomatrix369: same for this and all the above changed lines

makejdk.sh
```diff
@@ -258,14 +257,14 @@ testOpenJDKInNativeEnvironmentIfExpected()
 {
   if [[ "$JTREG" == "true" ]];
   then
-      $WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JTREG_TEST_SUBSETS
+      "$WORKING_DIR"/sbin/jtreg.sh "$WORKING_DIR" "$OPENJDK_REPO_NAME" "$BUILD_FULL_NAME" "$JTREG_TEST_SUBSETS"
```
neomatrix369: {} instead of ""

makejdk.sh
```diff
@@ -283,7 +282,7 @@ buildAndTestOpenJDK()
 
 initialiseEscapeCodes
 sourceSignalHandler
-parseCommandLineArgs $@
+parseCommandLineArgs "$@"
```
neomatrix369: :+1:

makejdk.sh
```diff
@@ -258,14 +257,14 @@ testOpenJDKInNativeEnvironmentIfExpected()
...
-      $WORKING_DIR/sbin/jtreg.sh $WORKING_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JTREG_TEST_SUBSETS
+      "$WORKING_DIR"/sbin/jtreg.sh "$WORKING_DIR" "$OPENJDK_REPO_NAME" "$BUILD_FULL_NAME" "$JTREG_TEST_SUBSETS"
   fi
 }
 
 buildAndTestOpenJDKInNativeEnvironment()
 {
   echo "Calling sbin/build.sh $WORKING_DIR $TARGET_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JVM_VARIANT"
-  $WORKING_DIR/sbin/build.sh $WORKING_DIR $TARGET_DIR $OPENJDK_REPO_NAME $BUILD_FULL_NAME $JVM_VARIANT
+  "$WORKING_DIR"/sbin/build.sh "$WORKING_DIR" "$TARGET_DIR" "$OPENJDK_REPO_NAME" "$BUILD_FULL_NAME" "$JVM_VARIANT"
```
neomatrix369: same here

sbin/build.sh
```diff
@@ -89,9 +93,9 @@ checkingAndDownloadingAlsa()
   if [[ ! -z "${FOUND_ALSA}" ]] ; then
     echo "Skipping ALSA download"
   else
-    wget -nc ftp://ftp.alsa-project.org/pub/lib/alsa-lib-$ALSA_LIB_VERSION.tar.bz2
-    tar xvf alsa-lib-$ALSA_LIB_VERSION.tar.bz2
-    rm alsa-lib-$ALSA_LIB_VERSION.tar.bz2
+    wget -nc ftp://ftp.alsa-project.org/pub/lib/alsa-lib-"$ALSA_LIB_VERSION".tar.bz2
+    tar xvf alsa-lib-"$ALSA_LIB_VERSION".tar.bz2
+    rm alsa-lib-"$ALSA_LIB_VERSION".tar.bz2
```
neomatrix369: {} instead of {} for all three lines

sbin/build.sh
```diff
@@ -105,19 +109,20 @@ checkingAndDownloadingFreetype()
...
 
-    if [ $(uname -m) = "ppc64le" ]; then
+    if [ "$OS_CPU_NAME" = "ppc64le" ]; then
+      # shellcheck disable=SC1083
       PARAMS="--build=$(rpm --eval %{_host})"
     fi
 
     # We get the files we need at $WORKING_DIR/installedfreetype
-    bash ./configure --prefix=$WORKING_DIR/$OPENJDK_REPO_NAME/installedfreetype $PARAMS && make all && make install
+    bash ./configure --prefix="$WORKING_DIR"/"$OPENJDK_REPO_NAME"/installedfreetype "$PARAMS" && make all && make install
```
neomatrix369: {} in place of "" for all changed lines above

sbin/jtreg_prep.sh
```diff
@@ -116,10 +116,10 @@ fi
 # Step 3: Unpack Java if we need to.
 
 if [[ "$JAVA_SOURCE" == *\.tar\.gz ]]; then #If it's a tar file, unpack it.
-  cd "${JAVA_DESTINATION}"
-  tar xf *.tar.gz
+  cd "${JAVA_DESTINATION}" || exit
+  tar xf ./*.tar.gz
   echo "Java has been untarred."
-  cd "${WORKING_DIR}"
+  cd "${WORKING_DIR}" || exit
```
neomatrix369: "${VAR}" same as ${VAR} or "$VAR"

sbin/jtreg_prep.sh
```diff
@@ -95,7 +95,7 @@ fi
 if [ ! -d "${JAVA_DESTINATION}" ]; then
   mkdir -p "${JAVA_DESTINATION}"
 fi
-cd "${JAVA_DESTINATION}"
+cd "${JAVA_DESTINATION}"  || exit
```
neomatrix369: Convention has changed on this line and below, "${VAR}" same as ${VAR} or "$VAR"

sbin/jtreg_prep.sh
```diff
@@ -69,24 +69,24 @@ fi
...
   git reset --hard origin/$BRANCH
-  cd $WORKING_DIR
+  cd "$WORKING_DIR" || exit
 elif [ ! -d "${WORKING_DIR}"/$OPENJDK_REPO_NAME/.git ] ; then
   # If it doesn't exist, clone it
   echo "Didn't find any existing openjdk repository at WORKING_DIR (set to ${WORKING_DIR}) so cloning the source to openjdk"
   if [[ "${USE_SSH}" == true ]] ; then
     echo "git clone -b ${BRANCH} git@github.com:${REPOSITORY}.git $WORKING_DIR/$OPENJDK_REPO_NAME"
-    git clone -b ${BRANCH} git@github.com:${REPOSITORY}.git $WORKING_DIR/$OPENJDK_REPO_NAME
+    git clone -b ${BRANCH} git@github.com:"${REPOSITORY}".git "$WORKING_DIR/$OPENJDK_REPO_NAME"
```
neomatrix369: This line the next few changed lines do not need "{$VAR}", just ${VAR} or "$VAR" should be enough

makejdk.sh
tellison: But `${var}` is *not* the same as `"$var"`

See the [ref guide](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)

The parentheses are used to delimit the `var`  and the double quotes prevent globbing/word splitting of the var value.

So in this case we case we can use the quotes only, but if we don't put them in then the command will fail for a dir with spaces, e.g. try this
```
$ mkdir "test dir"
$ foo="test dir"
$ cd ${foo}
bash: cd: test: No such file or directory
$ cd "$foo"
[test dir]$ 
```

sbin/build.sh
```diff
@@ -287,7 +292,7 @@ createOpenJDKTarArchive()
 
 stepIntoTargetDirectoryAndShowCompletionMessage()
 {
-  cd $TARGET_DIR
+  cd $TARGET_DIR  || return
```
neomatrix369: No quotes around $TARGET?

sbin/jtreg.sh
```diff
@@ -105,32 +106,32 @@ packageTestResultsWithJCovReports()
...
   pwd
 
-  cd $WORKING_DIR/$OPENJDK_REPO_NAME/build/$BUILD_FULL_NAME/
+  cd "$WORKING_DIR/$OPENJDK_REPO_NAME/build/$BUILD_FULL_NAME/" || exit
  
-  artifact=${JOB_NAME}-testoutput-with-jcov-reports
+  artifact="${JOB_NAME}-testoutput-with-jcov-reports"
   echo "Tarring and zipping the 'testoutput' folder into artefact: $artifact.tar.gz" 
-  tar -cvzf $WORKING_DIR/$artifact.tar.gz   testoutput/
+  tar -cvzf "$WORKING_DIR/$artifact.tar.gz"   testoutput/
```
neomatrix369: `v` will enlist all the files being tarred, do we want to show this to the end-user or record it in the logs - in that case its fine

makejdk.sh
tellison: Changed, thanks

sbin/jtreg.sh
```diff
@@ -105,32 +106,32 @@ packageTestResultsWithJCovReports()
...
   pwd
 
-  cd $WORKING_DIR/$OPENJDK_REPO_NAME/build/$BUILD_FULL_NAME/
+  cd "$WORKING_DIR/$OPENJDK_REPO_NAME/build/$BUILD_FULL_NAME/" || exit
  
-  artifact=${JOB_NAME}-jcov-results-only
+  artifact="${JOB_NAME}-jcov-results-only"
   echo "Tarring and zipping the 'testoutput/../jcov' folder into artefact: $artifact.tar.gz" 
-  tar -cvzf $WORKING_DIR/$artifact.tar.gz   testoutput/*/JTreport/jcov/
+  tar -cvzf "$WORKING_DIR/$artifact.tar.gz"   testoutput/*/JTreport/jcov/
```
neomatrix369: same `v` option here (verbose)

sbin/signalhandler.sh
tellison: Yep, I look forward to the contribution ;-)  I took the lazy way out of fixing the undefined vars, but reuse would be better.

sbin/signalhandler.sh
tellison: No, I really meant to use double quotes here to avoid whitespace issues

makejdk.sh
tellison: Not sure it matters for a definition, but I've put them in.

neomatrix369:
Overall a good PR especially it will help us decide on conventions and once merged, will make PR #53 look lighter, but I noticed the linter didn't help with suggesting a convention (or has it?). I have pointed out the areas where even though it make the strings safe, was applied differently than another area in the script.

I think we will have to come up and agree on guidelines - it does not matter so much what I pointed out, if they are acceptable or not. My (proposed) guidelines were:
- use `"$VAR"` in if conditions
- use `${VAR}` if it was part of another string/variable i.e. "The source is ${VAR}" or `cd ${VAR1}/{VAR2}`
- use `$VAR` in all other circumstances

We can discuss this / correct it and apply them to our upcoming PRs. :100:

makejdk.sh
tellison: as above, really do need quotes.  If the house-style is {}'s too, that's fine I can add them.

makejdk.sh
tellison: Should `$JTREG_TEST_SUBSETS` be splitting on spaces?  Not sure whether this is a single arg or set of args (in which case it should be an array).

sbin/build.sh
tellison: Now you are being silly ;-)  lol
I'll add some parentheses too .

sbin/build.sh
tellison: I think `$PARAMS`  is a single value, correct?  If not we should be using arrays.

sbin/jtreg_prep.sh
tellison: As above, the dirs definitely need quoting to get through to the command.
Will add {} for house style.

sbin/jtreg.sh
tellison: I didn't change the `v` option?

tellison:
I agree that adopting a convention would be good, and admit that I may have been inconsistent myself as I tried to fix things up.  The main goal here was to fix lint errors/warnings.  A separate pass to apply a consistent style throughout would be a good thing.  Perhaps a discussion in a new issue would be the right thing here?

neomatrix369:
I think with this a great set of guidelines will come up - guided by the static analysis reports + our own decisions and due diligence

neomatrix369:
@tellison agreed lets get this PR merged and raise/re-raise a new PR, in fact PR #53 covers a good amount of it, i can merge master once your changes are merged and we will have a PR with consistency in it (hopefully) - we can discuss the guidelines via that PR.

sbin/jtreg.sh
neomatrix369: No you didnt just came to my notice
Can be a different PR

sbin/jtreg_prep.sh
neomatrix369: Ok lesson learn, will refer to the guide you mentioned

sbin/signalhandler.sh
neomatrix369: Done via PR #54

makejdk.sh
neomatrix369: Thanks for the ref guide, i'll add it to my PR.

neomatrix369:
LGTM is anyone else does not have any nit-picking interests on this PR

makejdk.sh
gibfahn: >We had a discussion on one of the PR that "${}" is the same as ${}, i noted it for our guideline (see #54)

I think the conclusion there was that `${VAR}` is the same as `$VAR`, quotes are a different thing.

So it could be `"${VAR}"` or `"$VAR"`, but I guess that's a question for #54

git-hg/git-hg.sh
gibfahn: Agreed that `-eu` is a nice default, it's better to explicitly say `|| true` if you want to ignore an error.

makejdk.sh
gibfahn: I think the house style is being decided in #54 , probably easiest to see how that goes and then update this (or land this as is and update in #54)...

gibfahn:
>LGTM is anyone else does not have any nit-picking interests on this PR

setting `-eu` at the top seems reasonable.

sbin/jtreg.sh
karianna: I removed the v from another script if we want to be consistent <shrug>

makejdk.sh
neomatrix369: I would say filter all discussions and then apply it on top of #54 and use that as our final guidelines.

PR #63, opened 2017-04-14
Fix errors introduced by config changes

tellison:
Roll-back two changes that cause build failures.

- config arguments are built up in a variable, which we then expect to be split into args again.  Should be using bash array.
- attempt to use more efficient string replacement failed.  Requires more investigation, so excluded checks for now.


neomatrix369:
LGTM - tested it personally so it works on two people's machines

PR #66, opened 2017-04-18
Extract args for any platform - ARM specific

neomatrix369:
- Reduced two env variables into one - make command line args env var...
- Moved arm specific environment variable setting to a common place
- Renamed script file to reflect functionality/purpose of the script
- Adjusted tab position of a line in makejdk.sh
- Make JVM_VARIANT overrideable when calling makejdk-any-platform.sh
- Renamed local variable CONFIGURE_CMD to CONFIGURE_ARGS
- Introducing a new environment variable CONFIGURE_ARGS_FOR_ANY_PLATFORM...
- Moved ARM specific configure settings from build.sh into makejdk-any-platform.sh

**Note:** to take advantage of these changes we should call `makejdk-all-platforms.sh` from all jenkin jobs.

Partially fulfills #51


neomatrix369:
There is more that needs to be done, but for the same of getting ARM to build, slowly pull off different platform specific settings into `makejdk-any-platform.sh` seems an okay idea.

neomatrix369:
Kicked off test job at https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk_build_arm_linux_mani_test/23/console

PR #67, opened 2017-04-19
Make JVM_VARIANT overrideable

neomatrix369:
Change config for the ARM such that JVM_VARIANT can be overriden via makejdk-any-platform


neomatrix369:
This helps with testing out server build which has been failing for sometime

neomatrix369:
Related to #66 (missed out in previous PR). Also part of #51.

PR #71, opened 2017-04-19
Replaced --with-num-cores with --with-jobs in makejdk-any-platform.sh

neomatrix369:
Fixes issue https://github.com/AdoptOpenJDK/openjdk-build/issues/70

### DIFF
--- a/makejdk-any-platform.sh
+++ b/makejdk-any-platform.sh
@@ -35,7 +35,7 @@ fi
 if [[ "$OS_MACHINE_NAME" == "armv7l" ]] ; then
    export JVM_VARIANT=${JVM_VARIANT:-zero}
    export MAKE_ARGS_FOR_ANY_PLATFORM=${MAKE_ARGS_FOR_ANY_PLATFORM:-"DEBUG_BINARIES=true images"}
-   export CONFIGURE_ARGS_FOR_ANY_PLATFORM=${CONFIGURE_ARGS_FOR_ANY_PLATFORM:-"--with-num-cores=4"}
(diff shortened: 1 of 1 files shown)


makejdk-any-platform.sh
```diff
@@ -35,7 +35,7 @@ fi
 if [[ "$OS_MACHINE_NAME" == "armv7l" ]] ; then
    export JVM_VARIANT=${JVM_VARIANT:-zero}
    export MAKE_ARGS_FOR_ANY_PLATFORM=${MAKE_ARGS_FOR_ANY_PLATFORM:-"DEBUG_BINARIES=true images"}
-   export CONFIGURE_ARGS_FOR_ANY_PLATFORM=${CONFIGURE_ARGS_FOR_ANY_PLATFORM:-"--with-num-cores=4"}
+   export CONFIGURE_ARGS_FOR_ANY_PLATFORM=${CONFIGURE_ARGS_FOR_ANY_PLATFORM:-"--with-jobs=4"}
```
gdams: please don't hard code the number (I suggest taking the value from `NUM_PROCESSORS` as this is already defined on all our build machines)

karianna: Darn it, I missed that in my review.  I'll raise an issue.

neomatrix369: \good catch @gdams , i'll fix it, in the next PR

neomatrix369: Thinking of it again, its not necessary incorrect, as the `CONFIGURE_ARGS_FOR_ANY_PLATFORM` can be overriden with a new value which uses the NUM_PROCESSORS

neomatrix369: with the new parameterising everything is dynamic and can be overriden, these values are not fixed

PR #92, opened 2017-05-11
Update git-hg

rochmadadhiyunianto:
#85 #82 Fix Shellcheck on git-hg

PR #94, opened 2017-05-14
Fix #89

### LINKED ISSUE #88: Issue with creating openjdk directory in docker build/test script
While reviewing and testing changes via https://github.com/AdoptOpenJDK/openjdk-build/pull/82, we found that running native build after building via docker is an issue.

How to reproduce it:
Clear everything and start from scratch
```
./makejdk-any-plattform.sh     <=== docker build
```
and then run the below:
```
./makejdk-any-plattform.sh    --source <some source> --destination <some destination>
```

Gives me this error: `cannot create directory: /openjdk`

Now clear everything and start from scratch
```
./makejdk-any-plattform.sh    --source <some source> --destination <some destination>
```
and then run the below:
```
./makejdk-any-plattform.sh     <=== docker build
```
Works fine no complains.

The above use case might not be common but nevertheless, it happens.

There might be some ownership/permission issue with the `openjdk` when the docker build creates it.

Discussed with @johnoliver while reviewing and test changes via PR #82

### LINKED ISSUE #89: Mount folder with build scripts for docker container
Currently, the docker builds copy the script into the docker image each time and any changes to the script mean rebuilding the image.

This can be error-prone and tedious, time-consuming as we have to rebuild the image each time.

A better option is to mount the folder as a volume at run-time (script it into the Dockerfile).

Note: the last time trying the above approach caused the container to prematurely terminate at run-time so it needs some investigation/testing, to find out why this happens and fix it.

Discussed with @johnoliver while reviewing and test changes via PR #82

johnoliver:
- Copy scripts into docker at build time, resolves #89.

- Convert docker to use a persistent volume to store and build the docker source, should resolve #88.

A data volume will be created called `openjdk-source-volume`. This has a checkout of the jdk, and the downloaded libraries needed to build. This data volume is persistent over restarts so can be used to build and test separately. This is an improvement over the current state as the build does not depend on the state of the local directory.

- Change /openjdk/jdk8u to /openjdk/build inside docker.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -16,8 +16,8 @@ The Dockerfiles and associated scripts found in this project are licensed under 
 	`docker build -t dockeropenjdk .`
 
  - Run:
-	`docker run -it -v <path to source>:/openjdk/jdk8u dockeropenjdk`
+	`docker run -it -v <path to source>:/openjdk/build dockeropenjdk`
 
  - Debug (to shell):
-	`docker run -it -v <path to source>:/openjdk/jdk8u --entrypoint /bin/bash dockeropenjdk`
+	`docker run -it -v <path to source>:/openjdk/build --entrypoint /bin/bash dockeropenjdk`
 
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -32,6 +32,10 @@ source "$SCRIPT_DIR/sbin/common-functions.sh"
 REPOSITORY=${REPOSITORY:-AdoptOpenJDK/openjdk-jdk8u}
 OPENJDK_REPO_NAME=${OPENJDK_REPO_NAME:-openjdk}
 
+DOCKER_SOURCE_VOLUME_NAME="openjdk-source-volume"
+CONTAINER=openjdk_container
+TMP_CONTAINTER_NAME=openjdk-copy-src
+  
 USE_DOCKER=false
 WORKING_DIR=""
 USE_SSH=false
@@ -39,6 +43,7 @@ TARGET_DIR=""
 BRANCH=""
 KEEP=false
 JTREG=false
+CLEAN_DOCKER_BUILD=false
 
 determineBuildProperties
 
@@ -78,6 +83,9 @@ parseCommandLineArgs()
 
       "--keep" | "-k" )
       KEEP=true;;
+    
+      "--clean-docker-build" | "-c" )
+      CLEAN_DOCKER_BUILD=true;;
 
       "--jtreg" | "-j" )
       JTREG=true;;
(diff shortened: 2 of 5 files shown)


makejdk.sh
```diff
@@ -185,7 +193,41 @@ testOpenJDKViaDocker()
...
+    echo "${info}Removing old volumes and containers${normal}"
+    docker rm -f $TMP_CONTAINTER_NAME || true
+    docker rm -f "$(docker ps -a | grep $CONTAINER | cut -d' ' -f1)" || true
+    docker volume rm "${DOCKER_SOURCE_VOLUME_NAME}" || true
+    
+    
+    echo "${info}Creating volume${normal}"
+    docker volume create --name "${DOCKER_SOURCE_VOLUME_NAME}"
+    docker run -v "${DOCKER_SOURCE_VOLUME_NAME}":/openjdk/build --name $TMP_CONTAINTER_NAME ubuntu:14.04 /bin/bash
+    docker cp openjdk $TMP_CONTAINTER_NAME:/openjdk/build/
```
karianna: ubuntu:14.04 seems very specific

makejdk.sh
```diff
@@ -197,24 +239,15 @@ buildAndTestOpenJDKViaDocker()
...
-  CONTAINER=openjdk_container
-
-  # Copy our script for usage inside of the container
-  rm docker/jdk8u/x86_64/ubuntu/colour-codes.sh
-  cp "${SCRIPT_DIR}/sbin/colour-codes.sh" docker/jdk8u/x86_64/ubuntu 2>/dev/null
 
-  rm docker/jdk8u/x86_64/ubuntu/build.sh
-  cp "${SCRIPT_DIR}/sbin/build.sh" docker/jdk8u/x86_64/ubuntu 2>/dev/null
+  # Copy our scripts for usage inside of the container
+  rm -r docker/jdk8u/x86_64/ubuntu/sbin
```
karianna: jdk8u probably needs to be parameterised later, but that can be a future PR I guess.

karianna: Not sure about hardcoding of ubuntu either.

neomatrix369:
@johnoliver: @tellison had made a point about delivering portable images that already contained the scripts in it, does this work also take into account that consideration?

johnoliver:
yes

neomatrix369:
@karianna I think the references @johnoliver makes to `linux` and `ubuntu` terms might be okay to hang on to for some time, we won't be changing the OS of the image that quickly although references to `jdk` could be parameterised so to re-use with different variants of `jdk`, or give it a neutral name so its transparent across different variants /cc @johnoliver

neomatrix369:
@johnoliver I will test this out on my local machine, will be back with a response soon

docker/README.md
```diff
@@ -16,8 +16,8 @@ The Dockerfiles and associated scripts found in this project are licensed under
 	`docker build -t dockeropenjdk .`
 
  - Run:
-	`docker run -it -v <path to source>:/openjdk/jdk8u dockeropenjdk`
+	`docker run -it -v <path to source>:/openjdk/build dockeropenjdk`
```
neomatrix369: :+1:

makejdk.sh
```diff
@@ -32,13 +32,18 @@ source "$SCRIPT_DIR/sbin/common-functions.sh"
 REPOSITORY=${REPOSITORY:-AdoptOpenJDK/openjdk-jdk8u}
 OPENJDK_REPO_NAME=${OPENJDK_REPO_NAME:-openjdk}
 
+DOCKER_SOURCE_VOLUME_NAME="openjdk-source-volume"
+CONTAINER=openjdk_container
+TMP_CONTAINTER_NAME=openjdk-copy-src
```
neomatrix369: TMP_CONTAINTER_NAME ==> TMP_CONTAINER_NAME

makejdk.sh
```diff
@@ -185,7 +193,41 @@ testOpenJDKViaDocker()
...
+  docker volume inspect $DOCKER_SOURCE_VOLUME_NAME > /dev/null 2>&1
+  DATA_VOLUME_EXISTS=$?
+  
+  if [[ "$CLEAN_DOCKER_BUILD" == "true" || "$DATA_VOLUME_EXISTS" != "0" ]]; then
+  
+    echo "${info}Removing old volumes and containers${normal}"
+    docker rm -f $TMP_CONTAINTER_NAME || true
+    docker rm -f "$(docker ps -a | grep $CONTAINER | cut -d' ' -f1)" || true
+    docker volume rm "${DOCKER_SOURCE_VOLUME_NAME}" || true
+
```
neomatrix369: extra blank line?

makejdk.sh
```diff
@@ -197,24 +239,15 @@ buildAndTestOpenJDKViaDocker()
   fi
 
   echo "${info}Using Docker to build the JDK${normal}"
+  
+  createPersistentDockerDataVolume
```
neomatrix369: minor: createPersistentDockerDataVolume => createPersistentDataVolume
It contains scripts that will be used inside the container so no Docker related data

neomatrix369:
Currently building container from scratch

karianna:
@neomatrix369 FYI - the Shellcheck failures are for the referenced files which aren't part of this PR.

johnoliver:
Yeah all the shellcheck errors are addressed by #92

neomatrix369:
Let's skip the Shellcheck error messages. I think the build test passed last night and this morning so we should be good to merge once other pending PR comments are resolved. I have tried both testing approaches mentioned in #88.

The only difference is that the two methods create the output file with a different name, ie. `OpenJDK.tar.gz` and `OpenJDK-xxxxx-xxxx-.tar.gz`

There are a few incorrect namings but that is for another PR. One me suggestion: changes to #88 and #89 could have been changed chained one behind the other, they have slightly different goals.

neomatrix369:
@johnoliver Merge conflicts after merging previous PR

neomatrix369:
Merging as all looks good, lets look at pending comments in another series of PRs

PR #97, opened 2017-05-17
Modify version strings closes #93

karianna:

Issue #93, opened 2017-05-12
Change version block to highlight update, build number and community information


karianna:
There's a challenge involved in making version numbers give credit to the produce for Java.  The following issue discusses this: https://bugs.openjdk.java.net/browse/JDK-8189761 and a fix was applied for Java 10 and 11 which seems to also be back ported to 9.  However, this is not backported to Java 8, so we have some back porting work to do

---

Our existing 8 builds produce a Java 8 version like this:

```
openjdk version "1.8.0_181"
OpenJDK Runtime Environment (AdoptOpenJDK) (build 1.8.0_181-13)
OpenJDK 64-Bit Server VM (build 25.181-b13, mixed mode)
```

However, this is done using a hack which also overrides the `java.runtime.name` property, which we don't want to do as there is tooling in the wild that relies on it.

---

Now onto what Oracle produces and what we want to produce.  Oracle currently produces this:

_Java 8:_

```
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)
```

_Java 9:_

```
java version "9.0.4"
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)
```

_Java 10:_

Interesting how they add the date here

```
java version "10.0.2" 2018-07-17
Java(TM) SE Runtime Environment 18.3 (build 10.0.2+13)
Java HotSpot(TM) 64-Bit Server VM 18.3 (build 10.0.2+13, mixed mode)
```

---

For releases I think AdoptOpenJDK should produce something similar to:

_**Java 8**_

```
openjdk version "1.8.0_181"
OpenJDK Runtime Environment (build 1.8.0_181-b13-adoptopendk)
OpenJDK 64-Bit Server VM (build 25.181-b13, mixed mode)
```

_**Java 9**_

Note patch number is specified in JEP 223 but is rarely used.

```
java version "9.0.4.<patch number>"
OpenJDK Runtime Environment (build 9.0.4+11-adoptopenjdk)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)
```

_**Java 10**_

Note date format matches Oracle's JDK.  Note patch number is specified in JEP 223 but is rarely used.

```
java version "10.0.2.<patch number>" 2018-07-17
[...]

PR #103, opened 2017-05-19
Echo version string into console after build completion

neomatrix369:
- check if the build has produced the JDK image in the right folder and contains the `java` launcher
- run the `java -version` command to produce the version string onto the console
- fixed the issue with `BUILD_FULL_NAME` not being passed into all the sub-scripts

Created on the back of PRs #93, #97 and #101


neomatrix369:
Kicked off build https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk_build_ppc64le_linux_mani_test/3/console - sanity check if the string appears in the console as expected

sbin/common-functions.sh
```diff
@@ -28,7 +28,7 @@ determineBuildProperties() {
 
     BUILD_TYPE=normal
     DEFAULT_BUILD_FULL_NAME=${OS_KERNEL_NAME}-${OS_MACHINE_NAME}-${BUILD_TYPE}-${JVM_VARIANT}-release
-    BUILD_FULL_NAME=${BUILD_FULL_NAME:-"$DEFAULT_BUILD_FULL_NAME"}
+    export BUILD_FULL_NAME=${BUILD_FULL_NAME:-"$DEFAULT_BUILD_FULL_NAME"}
```
neomatrix369: Updating the env variable here I guess this isn't a bad thing - this has gone unnoticed i think hope other types of builds work fine

neomatrix369:
When successful on a platform, prints this at the bottom of the logs:
```
'/home/jenkins/workspace/openjdk_build_x86-64_linux_johno_test/openjdk/build/linux-x86_64-normal-server-release/images/j2sdk-image' found

openjdk version "1.8.0_152-adoptopenjdk"
OpenJDK Runtime Environment (build 1.8.0_152-adoptopenjdk-builddateb00-b00)
OpenJDK 64-Bit Server VM (build 25.152-b00, mixed mode)
```

neomatrix369:
Merged changes to version string by @karianna via PR #101

neomatrix369:
https://ci.adoptopenjdk.net/view/work%20in%20progress/ - most builds are successful (ignore Windows as it does not yet use the makejdk scripts). Two builds still in progress.

neomatrix369:
All checks have failed -  shellcheck failiures due known reasons

sbin/build.sh
```diff
@@ -197,6 +197,23 @@ buildOpenJDK()
   echo "${normal}"
 }
 
+
+printJavaVersionString()
```
karianna: some white space indentation issues in this function?

sbin/common-functions.sh
karianna: I assume this is the only place it is declared?

neomatrix369: One of the entry points - will check, fairly sure it does not affect anything, all the builds I ran this morning across all our platforms work fine despite this change, it takes the existing value and updates the env variable

neomatrix369: Just checked `makejdk-any-platform.sh` is the entry point, but if its not set there, it never gets set, so `export` keyword is important for setting a default value

PR #105, opened 2017-05-19
Shallow git clone

neomatrix369:
To speed up cloning of openjdk repo on local env and VMs

### DIFF
--- a/makejdk.1
+++ b/makejdk.1
@@ -32,3 +32,6 @@ select one or more jtreg tests to run
 .BR \-S ", " \-\-ssh
 use ssh when cloning git
 .TP
+.BR \-dsgc ", " \-\-disable-shallow-git-clone
+disable shallow cloning of git repo(s) using the --depth=1 CLI option
+.TP
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -31,7 +31,7 @@ source "$SCRIPT_DIR/sbin/common-functions.sh"
 
 REPOSITORY=${REPOSITORY:-AdoptOpenJDK/openjdk-jdk8u}
 OPENJDK_REPO_NAME=${OPENJDK_REPO_NAME:-openjdk}
-SHALLOW_CLONE_OPTION="--depth 1"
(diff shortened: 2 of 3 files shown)


gdams:
@neomatrix369 does this enable shallow clone by default ? or do you have to specify it?

neomatrix369:
@gdams shallow clones by default, git flag is set in the header of the script

gdams:
okay I would argue that we should not shallow clone by default. My reason being that the idea of makejdk.sh is that you can clone the git repo and then make changes and then test builds. If we don't clone the git history in by default it will make creating Pull Requests much harder. CC @tellison for his opinion here.

neomatrix369:
I can make this into an environment variable but then we will have to manually set it up in each environment or in your ansible scripts - is that okay? by default it will be blank once I change it - works for you?

gdams:
can we not just make it a command line variable ?

neomatrix369:
@gdams thats another way to make it available - lets see what I can do

karianna:
@neomatrix369 Is there more for you to do on this?

neomatrix369:
@karianna @gdams and I are discussing what the default value if this option (git shallow) should be and whether to use a cmd-line arg. or env. variable to set this value.

neomatrix369:
My thoughts here are:
- by default always have shallow git clone (due to low usage always cloning shallow is harmless)
- override this setting via a command-line switch --disable-shallow-git-clone (although I would prefer an environment variable but due to low usage command-line switch is good, cmd-line switch is fine too, but if number of scripts increases then env variable would be a better choice. As amending multiple script to maintain consistency and check for flag, etc... can become tedious with cmd-line switch)

I'll send out an updated PR soon.

neomatrix369:
@gdams just about doing that :+1:

neomatrix369:
Also kicked off https://ci.adoptopenjdk.net/view/OpenJDK%20PR/job/openjdk_pr_pipeline/8/console with git shallow clone option enabled

PR #117, opened 2017-06-09
Specify freetype dir for the FreeType Font library to build script

neomatrix369:
- Corrected spelling of a variable name
- Accept a new command-line param to specify the location of the pre-built FreeType Font directory, instead of building one each time, for e.g,

```
./makejdk-any-platform.sh [other params] --freetype-dir=/path/to/new/installfreetype/
```

Look for the below message on the console or build.log:
```
---> Skipping the process of checking and downloading the FreeType Font dependency, a pre-built version provided at /path/to/new/installfreetype/ <---
```

Note: The way this is written, if an environment variable by the name 
`FREETYPE_DIRECTORY` is created with the correct location, the build scripts will pick it up and use it.

Also added some performance checks around the dependency downloads for future references, metrics will be written to both console and build log file. Look out for these:

```
real	1m9.623s
user	0m14.213s
sys	0m2.752s
```

### DIFF
--- a/makejdk.1
+++ b/makejdk.1
@@ -29,6 +29,9 @@ run jtreg after building
 .BR \-js ", " \-\-jtreg-subsets
 select one or more jtreg tests to run
 .TP
+.BR \-ftd ", " \-\-freetype-dir
+specify the location of an existing FreeType library that can be used for the OpenJDK build process
+.TP
 .BR \-S ", " \-\-ssh
 use ssh when cloning git
 .TP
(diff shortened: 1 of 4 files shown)


neomatrix369:
dKicked off a build job - https://ci.adoptopenjdk.net/view/OpenJDK%20PR/job/openjdk_pr_ppc64le_linux/21/consoleFull - default functionality when the  `--freetype-directory=/path/to/font/library` is not passed

neomatrix369:
Kicked off a build job - https://ci.adoptopenjdk.net/view/OpenJDK%20PR/job/openjdk_pr_ppc64le_linux/24/console - where the  `--freetype-directory=/path/to/font/library` is passed, and the logs show the information the flag has been used (build fails as the location provided does not exist)

gdams:
this is perfect, @neomatrix369 could you perhaps write us an easy script called setupFreetype.sh or something? The idea being that it does the normal freetype setup but stores it in the users home directory? It would make it much easier for us to deploy this new feature across all our machines

PR #124, opened 2017-06-13
Remove hint about jdk boot dir when configure fails

a-roberts:
When configure fails it'd give us the cause anyway which may include the JDK_BOOT_DIR not being properly set

By always providing this hint this may falsely lead people to check their JDK_BOOT_DIR when there is a different problem

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -171,9 +171,6 @@ runTheOpenJDKConfigureCommandAndUseThePrebuildConfigParams()
     if [ $? -ne 0 ]; then
       echo "${error}"
       echo "Failed to configure the JDK, exiting"
-      echo "Did you set the JDK boot directory correctly? Override by exporting JDK_BOOT_DIR"
-      echo "For example, on RHEL you would do export JDK_BOOT_DIR=/usr/lib/jvm/java-1.7.0-openjdk-1.7.0.131-2.6.9.0.el7_3.x86_64"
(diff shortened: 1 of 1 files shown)


gdams:
can we be more clever about this? we could perform our own check for the java boot jdk at the start of makejdk.sh. Should be a case of checking for `javac`.

a-roberts:
Yeah that'd be quite easy to do: check the return code of java -version at JDK_BOOT_DIR (as it's a **JDK** boot dir we can also do a javac -version to rule out it just being a JRE)

We can also add a check that it returns a valid *previous version* string (so folks should not be building 8 with 8 as a bootstrap, or 6 for that matter).

gdams:
yeah I would personally just check for `javac --version`. If `javac` doesn't exist the build won't work anyway

neomatrix369:
@gdams an example of checking java for version:
```
java -version 2>&1 | grep -q '"9' || (echo "**** Fix your PATH! ****" && java -version && exit 1)
```

neomatrix369:
@gdams that was just an example - so you are right it can be used to check for `javac` and `7` - is what we should explicitly check for

karianna:
@a-roberts Are you able to make the changes suggested by @neomatrix369 and we can get this merged.

tellison:
This PR has been abandoned.  Please open a new PR if required.

PR #133, opened 2017-06-19
platform: add aix support

gdams:
Okay so this PR puts together all the required features and tweaks for the aix build

### DIFF
--- a/sbin/common-functions.sh
+++ b/sbin/common-functions.sh
@@ -21,7 +21,9 @@ FREETYPE_FONT_VERSION=${FREETYPE_FONT_VERSION:-2.4.0}
 
 determineBuildProperties() {
 
+    export OS_KERNEL_NAME=""
     OS_KERNEL_NAME=$(uname | awk '{print tolower($0)}')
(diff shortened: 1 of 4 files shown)


neomatrix369:
LGTM - would be good to run the PR via the Jenkins PR jobs.

sbin/common-functions.sh
```diff
@@ -21,8 +21,8 @@ FREETYPE_FONT_VERSION=${FREETYPE_FONT_VERSION:-2.4.0}
 
 determineBuildProperties() {
 
-    OS_KERNEL_NAME=$(uname | awk '{print tolower($0)}')
-    OS_MACHINE_NAME=$(uname -m)
+    export OS_KERNEL_NAME=$(uname | awk '{print tolower($0)}')
+    export OS_MACHINE_NAME=$(uname -m)
```
neomatrix369: Minor: `OS_MACHINE_NAME` is the value used in the global context, otherwise might not be necessary to make it a global var

gdams: I think that both of these vars should be globally readable throughout the build process

PR #134, opened 2017-06-20
build: fixups

gdams:
1st fix: removes error fatal: No tags can describe ''.
2nd fix: renames j2sdk-image to be the tag name making it clearer once
extracted
3rd fix: remove case sensitive dependancy for checking repo name
4th fix: shallow clone the `git fetch --tags` command

### DIFF
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -29,14 +29,15 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=sbin/common-functions.sh
 source "$SCRIPT_DIR/sbin/common-functions.sh"
 
-REPOSITORY=${REPOSITORY:-AdoptOpenJDK/openjdk-jdk8u}
+REPOSITORY=${REPOSITORY:-adoptopenjdk/openjdk-jdk8u}
+REPOSITORY="$(echo "${REPOSITORY}" | awk '{print tolower($0)}')"
 OPENJDK_REPO_NAME=${OPENJDK_REPO_NAME:-openjdk}
 SHALLOW_CLONE_OPTION="--depth=1"
 
 DOCKER_SOURCE_VOLUME_NAME="openjdk-source-volume"
 CONTAINER=openjdk_container
 TMP_CONTAINER_NAME=openjdk-copy-src
-  
+
 USE_DOCKER=false
 WORKING_DIR=""
 USE_SSH=false
@@ -86,7 +87,7 @@ parseCommandLineArgs()
 
       "--keep" | "-k" )
       KEEP=true;;
-    
+
       "--clean-docker-build" | "-c" )
       CLEAN_DOCKER_BUILD=true;;
 
@@ -174,7 +175,7 @@ setTargetDirectoryIfProvided()
 cloneOpenJDKGitRepo()
 {
   echo "${git}"
-  if [ -d "${WORKING_DIR}/${OPENJDK_REPO_NAME}/.git" ] && [ "$REPOSITORY" == "AdoptOpenJDK/openjdk-jdk8u" ] ; then
+  if [ -d "${WORKING_DIR}/${OPENJDK_REPO_NAME}/.git" ] && [ "$REPOSITORY" == "adoptopenjdk/openjdk-jdk8u" ] ; then
     # It does exist and it's a repo other than the AdoptOpenJDK one
     cd "${WORKING_DIR}/${OPENJDK_REPO_NAME}" || return
     echo "${info}Will reset the repository at $PWD in 10 seconds...${git}"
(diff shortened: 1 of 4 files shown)


makejdk.sh
```diff
@@ -215,9 +215,10 @@ getOpenJDKUpdateAndBuildVersion()
     # It does exist and it's a repo other than the AdoptOpenJDK one
     cd "${WORKING_DIR}/${OPENJDK_REPO_NAME}" || return
     echo "${git}Pulling latest tags and getting the latest update version"
-    git fetch --tags
-    OPENJDK_UPDATE_VERSION=$(git describe --abbrev=0 --tags | cut -d'u' -f 2 | cut -d'-' -f 1)
-    OPENJDK_BUILD_NUMBER=$(git describe --abbrev=0 --tags | cut -d'b' -f 2 | cut -d'-' -f 1)
+    git fetch -q --tags "${GIT_CLONE_ARGUMENTS[@]}"
```
neomatrix369: GIT_CLONE_ARGUMENTS ==> GIT_FETCH_ARGUMENTS?

makejdk.sh
```diff
@@ -215,9 +215,10 @@ getOpenJDKUpdateAndBuildVersion()
...
+    export OPENJDK_TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
```
neomatrix369: OPENJDK_TAG ==> OPENJDK_REPO_TAG - more specific to what it contains

makejdk.sh
gdams: no because the same arguments apply right? I assume that's where the shallow setting comes from

karianna:
Will merge after CI passes

karianna:
Hmm build failed

gdams:
for some reason I am not seeing ` echo "${git}Pulling latest tags and getting the latest update version using git fetch -q --tags ${SHALLOW_CLONE_OPTION}"` in the jenkins runs but am when I run it locally... almost like the if statement doesn't work in jenkins

gdams:
CI run https://ci.adoptopenjdk.net/view/OpenJDK%20PR/job/openjdk_pr_pipeline/22/

gdams:
@neomatrix369 / @tellison any idea how to fix https://travis-ci.org/AdoptOpenJDK/openjdk-build/builds/245674305?utm_source=github_status&utm_medium=notification ?

neomatrix369:
@gdams can we check if we are running the same version of `shellcheck` on Travis as on our local machine (and the config we have applied to `shellcheck` on Travis), I'll to look for something quickly for you and come back

makejdk.sh
```diff
@@ -211,15 +212,24 @@ cloneOpenJDKGitRepo()
...
-    echo "${OPENJDK_UPDATE_VERSION} ${OPENJDK_BUILD_NUMBER}"
-    cd "${WORKING_DIR}" || return
+  if [ -d "${WORKING_DIR}/${OPENJDK_REPO_NAME}/.git" ]; then
+    case "${REPOSITORY}" in
+      *openjdk-jdk8u)
+        # It does exist and it's a repo other than the AdoptOpenJDK one
+        cd "${WORKING_DIR}/${OPENJDK_REPO_NAME}" || return
+        echo "${git}Pulling latest tags and getting the latest update version using git fetch -q --tags ${SHALLOW_CLONE_OPTION}"
+        git fetch -q --tags "${SHALLOW_CLONE_OPTION}"
+        export OPENJDK_REPO_TAG="$(git describe --tags $(git rev-list --tags --max-count=1))"
```
neomatrix369: I think it's better to place this `export OPENJDK_REPO_TAG` on a separate line and then assign it separately here.

neomatrix369: The way to disable the error would be:
```
THE_FIRST_TAG="$(git rev-list --tags --max-count=1)"
# shellcheck disable=SC2155
export OPENJDK_REPO_TAG="$(git describe --tags $THE_FIRST_TAG)"
```

gdams:
okay @neomatrix369 updated PTAL

gdams:
https://ci.adoptopenjdk.net/view/OpenJDK%20PR/job/openjdk_pr_pipeline/24/

karianna:
Looks like the git tags command is still not taking hold correctly?

gdams:
@karianna it worked on those builds?

``` bash
Pulling latest tags and getting the latest update version using git fetch -q --tags --depth=1
Version: 162 00
```

karianna:
@gdams right you are - the version string is not being set correctly however.  Looking at my original impl to see where it might be going wrong.

karianna:
@gdams - in `configuringVersionStringParameter` can you add the following to the bottom?:

`  echo "Completed configuring the version string parameter, config args are now: ${CONFIGURE_ARGS}"`

gdams:
@karianna updated PTAL

karianna:
```
fatal: No tags can describe '29c2c69bd3c2d03c954bfd54c97cba62edfb8424'.
Try --always, or create some tags.
fatal: No tags can describe '29c2c69bd3c2d03c954bfd54c97cba62edfb8424'.
Try --always, or create some tags.
```

karianna:
I think merging https://github.com/AdoptOpenJDK/openjdk-build/pull/114/ will resolve that

gdams:
@karianna this PR already fixes that issue

karianna:
@gdams I landed 114 so you'll need to rebase and push

PR #135, opened 2017-06-26
add disable colour option

gdams:
add the option to disable colour output


makejdk.sh
```diff
@@ -347,9 +350,11 @@ buildAndTestOpenJDK()
 
 ##################################################################
 
-sourceFileWithColourCodes
 sourceSignalHandler
 parseCommandLineArgs "$@"
+if [ "${COLOUR}" != false ] ; then
```
karianna: Prefer == true to avoid double negatives

gdams:
@karianna updated PTAL

PR #141, opened 2017-07-03
Build freetype font library via an external script

neomatrix369:
Further development work, on the back of the PR https://github.com/AdoptOpenJDK/openjdk-build/pull/117


gdams:
It would be nice to automatically run this if I specify a freetype directory that doesn't exist?

neomatrix369:
Just to understand what you say - you like me to put a check in there, that if the destination folder specified as path does not exists to run the script otherwise just return.

Where would we be using this script (out of curiosity):
- manual execution
- via ansible scripts
- via Jenkins jobs
- via the makejdk / build scripts?

PR #143, opened 2017-07-04
Update to docker runtime image creation to enable pushing to docker hub

spoole167:
Changed the process for creating a docker image from a standalone approach to one suitable to be driven from Jenkins. Relies on docker push capabilities provided by Jenkins Docker plugins.

Prototype process is working on jenkins
using my repo See https://ci.adoptopenjdk.net/view/all/job/openjdk_docker_jdk8_ubuntu_x86_64_image/

Images are being pushed to Dockerhub - https://hub.docker.com/r/adoptopenjdk/openjdk8/

### DIFF
--- a/installer/docker/ubuntu/jdk/build.sh
+++ b/installer/docker/ubuntu/jdk/build.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Script to create a Docker image using Ubuntu and latest AdoptOpenJDK  linux binary
+# This script works in conjunction with jenkins plugins
+# The script assumes that Jenkins has placesd the tar.gz of the binary
+# in the root of the workspace. (WORKSPACE envar is set by jenkins)
+#
+# REPO defines the target docker repository for tagging images with
+# ARCH defines the computer architecture tag to be used for docker image tags
+# that are machine specific.
+#
+# This script builds the docker image but does not push to docker hub.
+# The docker plugin could build the image but is not dynamically configurable
+# so is only used to push the created image.
+#
+
+: "${WORKSPACE:=$PWD}"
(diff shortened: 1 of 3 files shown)


installer/docker/ubuntu/jdk/build.sh
```diff
@@ -0,0 +1,60 @@
...
+done
+
+# create place to hold image to be dockerized
+rm -rf docker_image
+mkdir -p docker_image
+
+# add contents to image
+cp Dockerfile docker_image
+cd docker_image || exit
+mv "$WORKSPACE"/OpenJDK8*.tar.gz .
```
karianna: OpenJDK8 hard coded?

spoole167: Well my assumption is that this particular process is for 8 only. Docker runtime images for 9+ will be generated differently due to the new wonders of the module system.

karianna: OK, I guess there will be a refactoring of names/paths that these files sit under when 9 comes in.  I'd expect that there should be a fair amount of reuse as well....

spoole167: Yes - I expect as we expand the docker images  scope (for instance when we do alpine or jres) then this work will be revisited.

installer/docker/ubuntu/jdk/build.sh
```diff
@@ -0,0 +1,60 @@
...
+# so is only used to push the created image.
+#
+
+: "${WORKSPACE:=$PWD}"
+: "${REPO:=adoptopenjdk/openjdk8}"
+: "${ARCH:=x86_64}"
+
+# remove any locally built images before building
+
+images=$(docker images | grep "^$REPO"  | awk '{print $3}' | uniq)
```
johnoliver: unless this list comes out pre-sorted, you will need to pipe into sort before uniq

installer/docker/ubuntu/jdk/build.sh
```diff
@@ -0,0 +1,60 @@
...
+
+for i in $images
+do
+   docker rmi -f "$i"
+done
+
+# create place to hold image to be dockerized
+rm -rf docker_image
```
johnoliver: may want to do 
```
rm -rf docker_image || true
```
since first time this is executed the dir may not exist

gdams: @johnoliver `rm -rf` will continue if the directory doesn't exist anyway, there should be no need to add `|| true`

installer/docker/ubuntu/jdk/build.sh
spoole167: I couldn't find any canonical info about output order for docker images so I will add sort

spoole167:
I'm going to merge this code as the travis failure is unrelated

PR #145, opened 2017-07-07
Move aix related checks to main script

neomatrix369:
At the moment we are checking if we are running in AIX and then choosing to run `gmake` to build OpenJDK source, this can already been predetermined and set in the `makejdk-any-platform.sh` script like we do for the other platforms checks.

There is still some more such checks in the different scripts but that's for subsequent PRs.

### DIFF
--- a/makejdk-any-platform.sh
+++ b/makejdk-any-platform.sh
@@ -27,6 +27,7 @@
 set -eu
 
 OS_MACHINE_NAME=$(uname -m)
+OS_KERNEL_NAME=$(uname | awk '{print tolower($0)}')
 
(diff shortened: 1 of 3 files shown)


neomatrix369:
Passed jenkins PR build - https://ci.adoptopenjdk.net/view/OpenJDK%20PR/job/openjdk_pr_pipeline/31/console

neomatrix369:
Anyone else - please review before we merge this PR

neomatrix369:
TravisCI is complaining about changes not part of this PR - not sure why. Other PRs are also affected by it.

PR #147, opened 2017-07-11
Resolve one of the Shellcheck warnings reported by TravisCI for PR #145

neomatrix369:
Added shellcheck disable notation to not warn about `SC2086` - the variable will contain values with spaces and its okay

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -201,6 +201,7 @@ buildOpenJDK()
   fi
 
   echo "Building the JDK: calling ${MAKE_COMMAND_NAME} ${MAKE_ARGS_FOR_ANY_PLATFORM}"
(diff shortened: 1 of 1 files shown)


neomatrix369:
@gdams the parameters must be separated by spaces and not sent into the make command as one array of strings for e.g.

`make test TEST="hotspot"`

and not as 

`make "test TEST="hotspot""`   <=== this one will have double quotes surround all the parameters and `make` won't be able to parse it properly, we will get an error like this from `make`:
```
make: *** No rule to make target 'test TEST=hotspot'. Stop.
```

neomatrix369:
@gdams Do you have anything else on this PR, can I pls merge it?

PR #154, opened 2017-08-17
WIP: add support for OpenJDK9

gdams:
This is the initial grunt of work to get docker builds working for OpenJDK, I have currently commented out the jdk8u stuff as we need to come up with some sort of switch between versions (e.g `--version=9`). There is also an issue with copying out the binary from docker, I'm starting to question whether this works for 8u though. FYI @tellison @EnriqueL8


makejdk.sh
```diff
@@ -29,7 +29,8 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=sbin/common-functions.sh
 source "$SCRIPT_DIR/sbin/common-functions.sh"
 
-REPOSITORY=${REPOSITORY:-adoptopenjdk/openjdk-jdk8u}
+#REPOSITORY=${REPOSITORY:-adoptopenjdk/openjdk-jdk8u}
```
neomatrix369: `jdk8` and `jdk9` seem to be common factors all across, why not extract them into an environment variabel `JDK_VERSION`, that can be reused across the script without having to modify it

PR #165, opened 2017-09-06
add osx codesign option

gdams:
This PR allows users to specify `--sign "certificate name"` to sign all binaries produced in the build stage. This obviously requires the certificate to be installed in the keychain of the build mac


sbin/build.sh
```diff
@@ -263,22 +263,33 @@ removingUnnecessaryFiles()
...
+        # Sign .dll files
+        FILES=$(find "${OPENJDK_REPO_TAG}" -type f -name '*.dll')
+        for f in $FILES; do
+          "$signToolPath" sign /f "$CERTIFICATE" /p "$SIGN_PASSWORD" /fd SHA256 /t http://timestamp.verisign.com/scripts/timstamp.dll "$f"
+        done
+      ;;
+      "darwin"*)
+        echo "Signing OSX release"
+        # Sign all executables
+        FILES=$(find "${OPENJDK_REPO_TAG}" -perm +111 -type f)
```
gibfahn: This will sign anything marked executable (e.g. python/shell scripts). is that intended?

Probably worth doing:

```bash
FILES=$(find "${OPENJDK_REPO_TAG}" -perm +111 -type f || find "${OPENJDK_REPO_TAG}" -perm /111 -type f)
```

to handle people who've `brew install`'d a GNU version of `find` (as `-perm +` has been deprecated since 2005 and long since removed).

sbin/build.sh
```diff
@@ -263,22 +263,33 @@ removingUnnecessaryFiles()
...
+        for f in $FILES; do
```
gibfahn: I don't think this handles spaces in paths (`./foo bar` becomes `./foo` and then `bar`), you need:

```bash
for f in "$FILES"; do
```

gdams: if you do `for f in "$FILES"; do` then it will only loop once as it treats it as a string. I think that we're lucky enough that there aren't any spaces in any of the paths

gibfahn: Whoops, I could have sworn that worked yesterday. Okay, POSIX way of doing it is:

```bash
echo "$FILES" | while read f; do codesign -s "$CERTIFICATE" "$f"; done
```

sbin/build.sh
neomatrix369: is it possible to use a constant for `+111` so we know what files and which types of permissions are being filtered?

neomatrix369:
Why is this not yet merged? Looks good to me.

karianna:
@neomatrix369 Because neither of the 2 reviewers have set their review to Approved :-)

sbin/build.sh
```diff
@@ -263,22 +263,27 @@ removingUnnecessaryFiles()
...
+        # Sign .exe files
+        FILES=$(find "${OPENJDK_REPO_TAG}" -type f -name '*.exe')
+        echo "$FILES" | while read -r f; do "$signToolPath" sign /f "$CERTIFICATE" /p "$SIGN_PASSWORD" /fd SHA256 /t http://timestamp.verisign.com/scripts/timstamp.dll "$f"; done
+        # Sign .dll files
+        FILES=$(find "${OPENJDK_REPO_TAG}" -type f -name '*.dll')
+        echo "$FILES" | while read -r f; do "$signToolPath" sign /f "$CERTIFICATE" /p "$SIGN_PASSWORD" /fd SHA256 /t http://timestamp.verisign.com/scripts/timstamp.dll "$f"; done
+      ;;
+      "darwin"*)
+        echo "Signing OSX release"
+        # Sign all executables
```
gibfahn: So @neomatrix369 's comment probably got missed (https://github.com/AdoptOpenJDK/openjdk-build/pull/165#discussion_r139806299).

>is it possible to use a constant for +111 so we know what files and which types of permissions are being filtered?

I don't think there is a constant that can be used here, but I think changing the comment to:

```bash
# Sign all files with the executable permission bit set.
```

Might be clear enough.

sbin/build.sh
neomatrix369: I was something like this could work:
```
CAN_EXECUTE_BUT_NOT_READ="+111"
FILES=$(find "${OPENJDK_REPO_TAG}" -perm "${CAN_EXECUTE_BUT_NOT_READ}" -type f)
```

gibfahn: Hmm, I think I'd find that extra indirection less clear than a comment that specified that we're checking for the executable bit (i.e https://github.com/AdoptOpenJDK/openjdk-build/pull/165#discussion_r140487777).

sbin/build.sh
tssva: Using the permission symbolic names might also make it clearer. So `+u+x,g+x,o+x` and `/u+x,g+x,o+x` instead of `+111` and `/111`.

gdams:
In order to use this, each build must have the following flag added: `--sign "Developer ID Application: London Jamocha Community CIC"`

gdams:
First run is here: https://ci.adoptopenjdk.net/job/openjdk10_build_x86-64_macos/

PR #144, opened 2017-07-07
Build freetype font library via an external script

neomatrix369:
Further development work, on the back of the PR #117

Conversation history on https://github.com/AdoptOpenJDK/openjdk-build/pull/141, closed due to issues with some changes made not reflected in the PR

PR #166, opened 2017-09-07
Support for OpenJDK9 builds with OpenJ9

mstoodle:
Changes to support building OpenJDK9 with Eclipse OpenJ9.

The primary changes are to:
   1) add a dockerfile to provide the correct Linux x86 build environment for Eclipse OpenJ9
   2) add a --variant option to makejdk.sh which sets BUILD_VARIANT used to do OpenJ9-specific work. BUILD_VARIANT is temporarily needed because OpenJ9 does not currently rely on the JVM variants in OpenJDK build scripts. Once openj9 becomes a JVM variant, BUILD_VARIANT can change to BUILD_VARIANT
   3) download freemarker dependency needed for OpenJ9 (akin to ALSA, FreeType, and cacerts). Eclipse OpenJ9 is currently doing work to eliminate this additional dependency, so I can also remove this part once that work is complete.

Until Eclipse OpenJ9 is available in the open, this PR won't be testable, so I've marked it as WIP so it can be reviewed ahead of time. I don't recommend merging until these changes can be tested using open repositories, but I have tested the changes using repositories I can access using a command line like:

`./makejdk-any-platform.sh -c --version jdk9 -r <openjdk repo name> -b <branch> --variant openj9`


makejdk.sh
```diff
@@ -338,22 +344,32 @@ buildAndTestOpenJDKViaDocker()
...
+             echo Building variant $ALTERNATE_VARIANT
+             docker build -t "$CONTAINER" -f "$PATH_BUILD/Dockerfile-$ALTERNATE_VARIANT" "$PATH_BUILD"
+         fi
      fi
   else
      echo "${info}Building as you've not specified -k or --keep"
      echo "$good"
      docker ps -a | awk '{ print $1,$2 }' | grep $CONTAINER | awk '{print $1 }' | xargs -I {} docker rm -f {}
      docker build -t "${CONTAINER}" "${PATH_BUILD}" --build-arg OPENJDK_VERSION="${OPENJDK_VERSION}"
+     if [[ "$ALTERNATE_VARIANT" != "" && -f $PATH_BUILD/Dockerfile-$ALTERNATE_VARIANT ]]; then
```
karianna: Not sure if this small duplicate chunk can be extracted?

sbin/build.sh
```diff
@@ -130,6 +130,7 @@ buildingTheRestOfTheConfigParameters()
   CONFIGURE_ARGS="${CONFIGURE_ARGS} --with-jvm-variants=${JVM_VARIANT}"
   CONFIGURE_ARGS="${CONFIGURE_ARGS} --with-cacerts-file=${WORKING_DIR}/cacerts_area/security/cacerts"
   CONFIGURE_ARGS="${CONFIGURE_ARGS} --with-alsa=${WORKING_DIR}/alsa-lib-${ALSA_LIB_VERSION}"
+  CONFIGURE_ARGS="${CONFIGURE_ARGS} --with-freemarker-jar=${WORKING_DIR}/freemarker-${FREEMARKER_LIB_VERSION}/lib/freemarker.jar"
```
karianna: Can this clash with the freetype if building the 'regular' OpenJDK?  I assume not.

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
```diff
@@ -0,0 +1,39 @@
...
+MAINTAINER Mark Stoodley <mstoodle@ca.ibm.com>
+
+# Install required OS tools
+RUN apt-get update \
+  && apt-get install -qq -y --no-install-recommends \
+    autoconf \
+    bison \
+    libffi-dev \
+    libnuma-dev \
+    g++-4.8 \
```
gdams: Can you perhaps move all of this gcc stuff up to the main dockerfile. It currently fetches 4.8 already I think but we should be hard coding it anyway

sbin/build.sh
gdams: I think that configure normally ignores unknown arguments, for safety though we should wrap this in an if statement to only use if we're building openj9

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
```diff
@@ -0,0 +1,39 @@
...
+RUN rm /usr/bin/gcc \
+    && rm /usr/bin/g++ \
+    && rm /usr/bin/c++ \
+    && ln -s /usr/bin/gcc-4.8 /usr/bin/gcc \
+    && ln -s /usr/bin/g++-4.8 /usr/bin/g++ \
+    && ln -s /usr/bin/g++-4.8 /usr/bin/c++
+
+ARG OPENJDK_VERSION
+ENV OPENJDK_VERSION=$OPENJDK_VERSION
+ENV JDK_PATH=jdk
```
gdams: I like how you've done this. @EnriqueL8 just added this [PR](https://github.com/AdoptOpenJDK/openjdk-build/pull/164) to do something similar but instead of passing to the docker file, we can ofcourse hard code it. Do you want to move `ENV JDK_PATH=jdk` to the parent dockerfile? We should also add `ENV JDK_PATH=j2sdk-image` to the jdk8u dockerfile

gdams:
also you've got a few linter errors that need fixing up https://travis-ci.org/AdoptOpenJDK/openjdk-build/builds/272779777?utm_source=github_status&utm_medium=notification

gdams:
Also, I know that we're obviously waiting for for the OpenJ9 code to be put in the public but you're going to need to add some switch to override the default branch and repo when openj9 is specified

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
```diff
@@ -0,0 +1,39 @@
+# Copyright IBM Corporation 2017.
```
tellison: Please keep the copyright notices in the NOTICE file as described in [the contributing guide](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/CONTRIBUTING.md).
(I realize it needs removing from the existing Dockerfile src too!)

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
tellison: @gdams I agree, good idea to be explicit about the gcc/g++ versions used in the Docker builds for all OpenJDK.

makejdk.sh
```diff
@@ -47,6 +47,8 @@ TARGET_DIR=""
 BRANCH=""
 KEEP=false
 JTREG=false
+JVM_VARIANT=""
+ALTERNATE_VARIANT=""
```
tellison: _Minor comment:_  I'm not wild about this name.  If we are having `<something>_VARIANT` then the `<something>` should be a noun.  Consider `BUILD_VARIANT`?  Just a bit more descriptive.

sbin/common-functions.sh
```diff
@@ -49,6 +50,22 @@ checkingAndDownloadingAlsa()
   fi
 }
 
+# Freemarker for OpenJ9
+checkingAndDownloadingFreemarker()
+{
+  echo "Checking for FREEMARKER"
+
+  FOUND_FREEMARKER=$(find "${WORKING_DIR}" -name "freemarker-${FREEMARKER_LIB_VERSION}")
```
tellison: _Minor comment:_  consider adding `-type d`  to the find criteria?

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
```diff
@@ -0,0 +1,39 @@
...
+# limitations under the License.
+
+FROM openjdk_container
+
+MAINTAINER Mark Stoodley <mstoodle@ca.ibm.com>
+
+# Install required OS tools
+RUN apt-get update \
+  && apt-get install -qq -y --no-install-recommends \
+    autoconf \
```
tellison: `autoconf` will be in the base image.  I don't think it needs to be explicitly installed here.

sbin/build.sh
mstoodle: Yes, that's a good point. Thanks for the feedback, I'll add an if statement.

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
mstoodle: I could do if you would prefer. I was trying to isolate the changes that were needed for OpenJ9 to the -openj9 dockerfile. I didn't think OpenJDK required gcc 4.8 (or else your Dockerfile is broken already). By default, I believe you're getting gcc 5.something .

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
mstoodle: Sure, I can do that.

mstoodle:
I am expecting OpenJ9 builds to be performed by a separate Jenkins job which will probably contain, at its heart, the command line at the end of my first comment (which does provide the options to override the repository and the branch, as you say :) ).

mstoodle:
I am expecting OpenJDK with Eclipse OpenJ9 builds to be performed by a separate Jenkins job which will probably contain, at its heart, the command line at the end of my first comment (which does provide the options to override the repository and the branch, as you say :) ).

mstoodle:
I updated the description to indicate that the freemarker.jar dependency is a point-in-time requirement that Eclipse OpenJ9 is currently working to eliminate.

makejdk.sh
mstoodle: Good feedback, thanks. i can extract into a "buildDockerContainer" function that avoids most of the duplication. will add a commit to show that (after I've tested it :) ) for review

makejdk.sh
mstoodle: I admit, the 8 milliseconds of thought I put into it probably wasn't enough :) . I can change it to BUILD_VARIANT.

I'm hoping this won't be a long lived variable name, but I'm happy to change it to something you find more descriptive :) .

sbin/common-functions.sh
mstoodle: yep, will do. thanks!

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
mstoodle: I don't believe that's true. At least, when I removed this line, openj9 failed to build in a fairly catastrophic way. Note that autoconf isn't needed to run the configure scripts; it's needed to generate such a script.

mstoodle:
My apologies; I was trying to keep my history intact so the review comments would stay relevant, but one ill-considered force push pretty much wrecked it :( .

I believe I have addressed all the suggestions from the reviewers.

Two things I would like to point out:
   1. Dockerfile-openj9 is fairly minimal at this point. I'm happy to leave it as-is, but it might be simpler to just merge the few extra dependencies into the main Dockerfile and then even more of my change melt away.
   2. I was unable to eliminate one linter warning in makejdk.sh, so I had to add a shellcheck comment to disable C2086 (double quotes to prevent globbing/word splitting) on one particular line with nested double quoting. It looks right to me, but the tool still doesn't like it. I would appreciate additional eyes on this line (line 358) to make sure the shellcheck comment is really necessary.

Of course, if there are additional recommendations, I'll be happy to work them in.

mstoodle:
Folding the docker files together gets even more attractive: it turns out only autoconf and libnuma-dev are the only additional requirements. I pushed an additional commit to remove bison and libffi.

docker/jdk9/x86_64/ubuntu/Dockerfile
```diff
@@ -40,8 +39,18 @@ RUN apt-get update \
     git \
     unzip \
     realpath \
+    g++-4.8 \
+    gcc-4.8 \
 && rm -rf /var/lib/apt/lists/*
 
+# Make sure build uses 4.8
+RUN rm /usr/bin/gcc \
+    && rm /usr/bin/g++ \
```
gdams: will this fail if g++ doesn't already exist? perhaps add `-f` to be safe

mstoodle: the base image installs g++, so the original is there "by construction"

makejdk.sh
```diff
@@ -298,13 +303,14 @@ createPersistentDockerDataVolume()
...
     docker volume rm "${DOCKER_SOURCE_VOLUME_NAME}" || true
 
     echo "${info}Creating volume${normal}"
     docker volume create --name "${DOCKER_SOURCE_VOLUME_NAME}"
     docker run -v "${DOCKER_SOURCE_VOLUME_NAME}":/openjdk/build --name $TMP_CONTAINER_NAME ubuntu:14.04 /bin/bash
     docker cp openjdk $TMP_CONTAINER_NAME:/openjdk/build/
+
     ls $TMP_CONTAINER_NAME:/openjdk/build/
     echo "${info}Updating source${normal}"
     docker exec $TMP_CONTAINER_NAME "cd /openjdk/build/openjdk && sh get_source.sh"
```
gdams: This get_source command only applies to openj9 so should be in an if statement

mstoodle:
@gdams addressed your comment about needing to run get_source.sh . I'm not sure how that didn't get into my original PR (other than the obvious and sad -- for me -- answer).

mstoodle:
verified that "plain" java9 builds still work `makejdk-any-platform.sh -c --version jdk9`

mstoodle:
I'll wait for instructions from @gdams, at least, before squashing some of the minor commits (unless you really don't want me to do that) and removing the "WIP" from the title (since the repositories are all externally accessible now and can be tested and built from github.com).

mstoodle:
Actually, I'd like to put a vote in for just merging the two Dockerfiles together which eliminates the whole BUILD_VARIANT complexity of this pull request. maybe I'll put together a second PR just to show what that looks like for comparison.

gdams:
okay so there is still one pending issue here....

1.  I'm still having to pass it `-r ibmruntimes/openj9-openjdk-jdk9 -b openj9`

mstoodle:
@gdams You need those options so that it clones from a version of OpenJDK that has the patches needed to incorporate Eclipse OpenJ9. Those options specify an open repository on Github...

gdams:
@mstoodle sure but we can hard code them to `-r ibmruntimes/openj9-openjdk-jdk9 -b openj9` if you specify openj9 right ?

mstoodle:
Hm, I guess my operating model was different: make as few changes as possible that start with "if this is openj9 then..." but rather have a generic option provide the information. But that has the down side that the openj9-specific information has to spread to the Jenkins jobs, and that also seems less than ideal.

Ok, I'll fold these settings into the scripts.

mstoodle:
Just curious, does anyone actually use the docker container based approach to building OpenJDK using these scripts? Am I making something work with OpenJ9 that doesn't really need to?

mstoodle:
Ok, I think I am done with this PR, barring more feedback. I believe OpenJDK9 with OpenJ9 builds are now working on the three primary platforms it was meant to enable (Linux x86-64, Linux ppc64le, and S390x) with builds still working for OpenJDK9 with HotSpot and OpenJDK8u with HotSpot.

If there's no additional cleanup, I can reorder/squash the commits that make sense in preparation for Merge readiness. Just let me know @gdams @neomatrix369 !

neomatrix369:
Yay! OpenJ9!

README.md
```diff
@@ -27,6 +27,7 @@ Options:
   -k, --keep                 reuse docker container (prevents deleting)
   -j, --jtreg                run jtreg after building
   -S, --ssh                  use ssh when cloning git
+  --variant <name>           specify a build variant name, e.g. openj9
```
neomatrix369: minor niggle: we usually have a short and long param option, if `-v` or `-V` is available would be good to use

docker/jdk8u/x86_64/ubuntu/Dockerfile
```diff
@@ -56,3 +56,4 @@ CMD ["images"]
 
 ARG OPENJDK_VERSION
 ENV OPENJDK_VERSION=$OPENJDK_VERSION
+ENV JDK_PATH=j2sdk-image
```
neomatrix369: Just a naming thing, this isn't really JDK_PATH but more the path to the JDK image created by the OpenJDK build process, so `JDK_IMAGE_PATH` or in those lines would be helpful, otherwise makes me think its like `JAVA_HOME` in a way.

docker/jdk9/x86_64/ubuntu/Dockerfile
```diff
@@ -59,3 +58,4 @@ CMD ["images"]
 
 ARG OPENJDK_VERSION
 ENV OPENJDK_VERSION=$OPENJDK_VERSION
+ENV JDK_PATH=jdk
```
neomatrix369: Same here

makejdk.sh
```diff
@@ -47,6 +47,8 @@ TARGET_DIR=""
 BRANCH=""
 KEEP=false
 JTREG=false
+JVM_VARIANT=""
```
neomatrix369: Do we have a list of these variants mentioned somewhere, our wiki or some README. I know these variants may vary a bit, good to have them to hand.

makejdk.sh
```diff
@@ -120,11 +122,23 @@ parseCommandLineArgs()
...
+      export BUILD_VARIANT=$1; shift;;
+
       *) echo >&2 "${error}Invalid option: ${opt}${normal}"; man ./makejdk-any-platform.1; exit 1;;
      esac
   done
 }
 
+doAnyBuildVariantOverrides()
+{
+  if [[ "${BUILD_VARIANT}" == "openj9" ]]; then
```
neomatrix369: Is this `BUILD_VARIANT` overlapping with OpenJDK's own variant terminologies (borderline)?

sbin/common-functions.sh
```diff
@@ -49,6 +50,22 @@ checkingAndDownloadingAlsa()
   fi
 }
 
+# Freemarker for OpenJ9
```
neomatrix369: We have a PR related to this, will the below be also integrated there?

neomatrix369:
Any reasons this is still not merged? @gdams @karianna

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
```diff
@@ -20,8 +20,18 @@ RUN apt-get update \
...
+    gcc-4.8 \
 && rm -rf /var/lib/apt/lists/*
 
+# Make sure build uses 4.8
+RUN rm /usr/bin/gcc \
+    && rm /usr/bin/g++ \
+    && rm /usr/bin/c++ \
+    && ln -s /usr/bin/gcc-4.8 /usr/bin/gcc \
+    && ln -s /usr/bin/g++-4.8 /usr/bin/g++ \
+    && ln -s /usr/bin/g++-4.8 /usr/bin/c++
```
neomatrix369: ` ln -s /usr/bin/g++-4.8 /usr/bin/c++` is this line correct? it appears the same as ` ln -s /usr/bin/g++-4.8 /usr/bin/g++ ` `c++` and `g++` is mapped to to the same binary executable?

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
```diff
@@ -20,8 +20,18 @@ RUN apt-get update \
   && apt-get install -qq -y --no-install-recommends \
     autoconf \
     libnuma-dev \
+    g++-4.8 \
```
neomatrix369: minor: `gcc-4.8` and `g++-4.8`could be replaced with constants or labels - makes it easier to manage versioning of dependencies

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
mstoodle: um, yes, i'm not super happy about it either, but I found some commands in the build run c++ and some run g++. I doubted that they are so particular about what they want, so I just made them the same. Does that sound reasonable to you?

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
mstoodle: @neomatrix369  sorry, I'm not quite sure what you mean here but, ignoring that, it sounds like a very reasonable suggestion (I know how silly that sounds :) ).

can you please jump start my poor brain with a couple more details on what you're looking for here?

makejdk.sh
mstoodle: No, it's separate because OpenJ9's build system does not (yet) rely on OpenJDK's variant stuff (sorry, we know that's not good). Our plan is to fix that, but for now we needed this extra little bit. My hope is that, in future, we can remove the option and change BUILD_VARIANT to JVM_VARIANT.

docker/jdk8u/x86_64/ubuntu/Dockerfile
mstoodle: Yeah, JDK_PATH existed before this change, I just put it into the Dockerfile rather than passing it in from the scripts above (which have to set it differently depending on jdk8 or jdk9). I found it easier to just set it directly in the Dockerfile because it knows which jdk it's building.

sbin/common-functions.sh
mstoodle: You're referring to #144 for freetype, I guess? I hadn't noticed that one until now. Would you be ok with doing that work as a follow-on enhancement? I'd really like to get this one folded into the master branch, but if you feel strongly about it, I will look into applying the #144 approach for freemarker here.

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
neomatrix369: Ok, thanks for sharing this.

docker/jdk9/x86_64/ubuntu/Dockerfile-openj9
neomatrix369: I meant something like this https://docs.docker.com/engine/reference/builder/#arg helps also give a view on the parametric nature of an entity in this case the Dockerfile itself

sbin/common-functions.sh
neomatrix369: Yes exactly, feel free to use the script(s) there and we can work together to iron it out. That way there would be consistency across the different scripts. Happy to help you out, if you get stuck or need any explanation on what we are doing. 

And doing it as a follow-on work is fine, as long it does not fall thru. :+1:

makejdk.sh
neomatrix369: Ok cool thnx

docker/jdk8u/x86_64/ubuntu/Dockerfile
neomatrix369: Okay no worries, one of these days I'll clean sweep such namings, as there are so many terms to remember in the OpenJDK world, easier to remember the better.

gdams:
okay no further objections so landing this now

PR #168, opened 2017-09-14
OpenJDK9 with Eclipse OpenJ9 build enablement (single Dockerfile)

mstoodle:
An alternative implementation for #166 simplified by not introducing a new docker file (since there are only two additional packages required to be installed and they aren't large). I prefer this pull request because I think it's simpler. If there is agreement, I will close #166.

Two new command-line options are added to support building OpenJDK9 with Eclipse OpenJ9 : `--freemarker` (downloads one additional dependency like ALSA or freetype), and `--get-source` (runs `bash get_source.sh` after cloning the openjdk repository). The Eclipse OpenJ9 project is actively working to eliminate the freemarker dependency as we're moving to a cmake based build system which no longer requires freemarker, so we should be able to remove that one at some hopefully not too far distant future.

This pull request also sets up `gcc` and `g++` 4.8 as the supported build environment for OpenJDK (currently using 5.4), and removes the copyright statements from the project's two Dockerfile's to bring them in line with this project's contribution guidelines.

With these changes, one can build OpenJDK9 with Eclipse OpenJ9 with the following command:
`makejdk-any-platform.sh -c --version jdk9 -r ibmruntimes/openj9-openjdk-jdk9 -b openj9 --freemarker --get-source` .

I have verified that OpenJDK9 with HotSpot can still be built with `makejdk-any-platform.sh -c --version jdk9`.

For the record, `github.com/ibmruntimes/openj9-openjdk-jdk9` is a git mirror of the OpenJDK jdk9 forest that does not include HotSpot. The `openj9` branch of this repository includes a few patches needed to build with Eclipse OpenJ9, which is now located at `github.com/eclipseopenj9`

This pull request is not WIP because anyone should be able to test this command now to verify the build works.


docker/jdk9/x86_64/ubuntu/Dockerfile
```diff
@@ -40,8 +40,18 @@ RUN apt-get update \
     git \
     unzip \
     realpath \
+    g++-4.8 \
+    gcc-4.8 \
 && rm -rf /var/lib/apt/lists/*
 
+# Make sure build uses 4.8
+RUN rm /usr/bin/gcc \
```
karianna: Isn't this forcing all builds to use the older gcc?

makejdk-any-platform.1
```diff
@@ -53,3 +53,9 @@ use ssh when cloning git
 .BR \-dsgc ", " \-\-disable-shallow-git-clone
 disable shallow cloning of git repo(s) using the --depth=1 CLI option
 .TP
+.BR \-fm ", " \-\-freemarker
+fetch freemarker jar file and add --with-freemarker-jar option to configure command
+.TP
+.BR \-gs ", " \-\-get-source
```
karianna: Perhaps add a comment that this will get removed later and maybe raise a GitHub issue?

sbin/build.sh
```diff
@@ -131,6 +131,11 @@ buildingTheRestOfTheConfigParameters()
   CONFIGURE_ARGS="${CONFIGURE_ARGS} --with-cacerts-file=${WORKING_DIR}/cacerts_area/security/cacerts"
   CONFIGURE_ARGS="${CONFIGURE_ARGS} --with-alsa=${WORKING_DIR}/alsa-lib-${ALSA_LIB_VERSION}"
 
+  # Freemarker dependency?
```
karianna: remove '?'

docker/jdk9/x86_64/ubuntu/Dockerfile
tellison: Yeah, I'm inclined to agree with @karianna's implication that this is not helpful to all builds, so should be kept separate.

I can imagine (but have no evidence) that down-leveling the compiler will hurt performance for code such as Hotspot, and our builds should be optimal for the included components.

tellison:
Although the code differences are modest, they have the potential to impact our current binaries; so on balance I think we should keep them separated as an OpenJ9-specific build environment.

docker/jdk9/x86_64/ubuntu/Dockerfile
mstoodle: In #166 I was specifically asked to move this code into the base Docker image so I also did that here. Note that OpenJDK8 documents gcc-4.8 (granted not on Ubuntu 16.04) as the build tool chain here https://wiki.openjdk.java.net/display/Build/Supported+Build+Platforms .

makejdk-any-platform.1
mstoodle: i can do that, sure.

sbin/build.sh
mstoodle: i put a ? because it was only done conditionally, but I can appreciate it makes the comment sound unsure of itself :) . I'll remove it.

mstoodle:
I think I agree with @tellison 's direction. I think the better solution is offered by #166 at this point.

karianna:
Closing in favour of #166

Issue #174, opened 2017-09-28
Bash fails on `if [...` and `if [[...` on local environment and Jenkins respectively


neomatrix369:
I have noticed lines like this:
`if [...` fails on my local machine
and 

`if [[...` fails on Jenkins

See `sbin/common-functions.sh: line 103`, to see the block of code I'm talking about, there are a couple more such implementations.

Anyone has any idea whats happening here?
Bash version issue?

```
bash --version
GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```
From my local env.

sxa555:
That syntax looks very strange to me - basically it's:

`if [ $(bash ./configure ... && $MAKE all && $MAKE install) -ne 0 ]; then`

The result of the $(...) is going to be the whole text output, not the return code, so comparing it to zero doesn't strike me as making a lot of sense. I was going to suggest running the command in the line before then `if [ $? -ne 0 ]` ot similar but it looks like you have explicitly switched it away from that in https://github.com/AdoptOpenJDK/openjdk-build/commit/5a311d824d25989e0e9ca110829f4e52f5180a00 @neomatrix369 - what was the reasoning for that?

gibfahn:
So maybe change:

```bash
if [ $(bash ./configure ... && $MAKE all && $MAKE install) -ne 0 ]; then
```

to

```bash
if ! $(bash ./configure ... && $MAKE all && $MAKE install); then
```

neomatrix369:
@sxa555 happened when I was battling with local env, docker, jenkins and shellcheck - but i'll give @gibfahn approach a try and see what happens.

neomatrix369:
Fixed via PR #175

neomatrix369:
Thanks @sxa555 and @gibfahn for your feedback.

PR #175, opened 2017-10-02
Bash if condition fix in a few scripts

neomatrix369:
Replaced all 'if [ $(...) -ne 0 ]; then' style if conditions with 'if ! (...) ; then' across a number of bash scripts.

Fixes issue #174

### DIFF
--- a/sbin/common-functions.sh
+++ b/sbin/common-functions.sh
@@ -75,7 +75,7 @@ checkingAndDownloadingFreeType()
 
     # We get the files we need at $WORKING_DIR/installedfreetype
     # shellcheck disable=SC2046
-    if [ $(bash ./configure --prefix="${WORKING_DIR}"/"${OPENJDK_REPO_NAME}"/installedfreetype "${FREETYPE_FONT_BUILD_TYPE_PARAM}" && $MAKE all && $MAKE install) -ne 0 ]; then
+    if ! (bash ./configure --prefix="${WORKING_DIR}"/"${OPENJDK_REPO_NAME}"/installedfreetype "${FREETYPE_FONT_BUILD_TYPE_PARAM}" && $MAKE all && $MAKE install); then
       # shellcheck disable=SC2154
       echo "${error}Failed to configure and build libfreetype, exiting"
       exit;
@@ -100,7 +100,7 @@ checkingAndDownloadCaCerts()
   git clone https://github.com/AdoptOpenJDK/openjdk-build.git cacerts_area
   echo "cacerts should be here..."
   # shellcheck disable=SC2046
-  if [ $(file "${WORKING_DIR}/cacerts_area/security/cacerts") -ne 0 ]; then
+  if ! (file "${WORKING_DIR}/cacerts_area/security/cacerts"); then
     echo "Failed to retrieve the cacerts file, exiting..."
     exit;
   else
--- a/sbin/jtreg.sh
+++ b/sbin/jtreg.sh
@@ -55,7 +55,7 @@ downloadJtregAndSetupEnvironment()
    JTREG_BINARY_FILE="jtreg-${JTREG_VERSION}.tar.gz"
 
    # shellcheck disable=SC2046
-   if [ $(cd "$TMP_DIR" && wget https://ci.adoptopenjdk.net/job/jtreg/lastSuccessfulBuild/artifact/"$JTREG_BINARY_FILE") -ne 0 ]; then
+   if ! (cd "$TMP_DIR" && wget https://ci.adoptopenjdk.net/job/jtreg/lastSuccessfulBuild/artifact/"$JTREG_BINARY_FILE"); then
      echo "Failed to retrieve the jtreg binary, exiting"
      exit
    fi
--- a/sbin/jtreg_prep.sh
+++ b/sbin/jtreg_prep.sh
@@ -127,7 +127,7 @@ downloadJDKArtifact()
     #If it's a http location, use wget.
     if [[ "${JAVA_SOURCE}" == http* ]]; then
       # shellcheck disable=SC2046
-      if [ $(wget "${JAVA_SOURCE}") -ne 0 ]; then
+      if ! (wget "${JAVA_SOURCE}"); then
         echo "Failed to retrieve the JDK binary from ${JAVA_SOURCE}, exiting"
         exit 1
       fi


neomatrix369:
Kicked off two Jenkins jobs (8 and 9):
- https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk8_x86-64_linux_mani_pr_175/2/console
- https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk9_x86-64_linux_mani_pr_175/2/console

sbin/common-functions.sh
```diff
@@ -100,7 +100,7 @@ checkingAndDownloadCaCerts()
   git clone https://github.com/AdoptOpenJDK/openjdk-build.git cacerts_area
   echo "cacerts should be here..."
   # shellcheck disable=SC2046
-  if [ $(file "${WORKING_DIR}/cacerts_area/security/cacerts") -ne 0 ]; then
+  if ! (file "${WORKING_DIR}/cacerts_area/security/cacerts"); then
```
gibfahn: The brackets aren't _strictly_ necessary here, these could all just be 

```bash
   if ! file "${WORKING_DIR}/cacerts_area/security/cacerts"; then 
```

For the longer ones like:

```bash
     if ! (bash ./configure --prefix="${WORKING_DIR}"/"${OPENJDK_REPO_NAME}"/installedfreetype "${FREETYPE_FONT_BUILD_TYPE_PARAM}" && $MAKE all && $MAKE install); then 
```

I think the brackets probably make things clearer (especially for people not familiar with bash), but if it's a simple command it can just be `if command; then`

neomatrix369: Yes agree and kinda marks demarkation between the condition and if ... then syntax

neomatrix369:
##### OpenJDK8 successful

```
...
openjdk version "1.8.0-internal"
OpenJDK Runtime Environment (build 1.8.0-internal-jenkins_2017_10_02_19_12-b00)
OpenJDK 64-Bit Server VM (build 25.71-b00, mixed mode)


Removing unnecessary files now...
Fetching the first tag from the OpenJDK git repo...
Currently at '/home/jenkins/workspace/openjdk8_x86-64_linux_mani_pr_175/openjdk/build/linux-x86_64-normal-server-release/images'
moving j2sdk-image to jdk8u144-b01
Finished removing unnecessary files from jdk8u144-b01
Archiving the build OpenJDK image...
OpenJDK repo tag is jdk8u144-b01
Your final .tar.gz was created at /home/jenkins/workspace/openjdk8_x86-64_linux_mani_pr_175/openjdk/build/linux-x86_64-normal-server-release/images
Moving the artifact to /home/jenkins/workspace/openjdk8_x86-64_linux_mani_pr_175/OpenJDK8_x64_Linux_201702101909.tar.gz
All done!
Triggering projects: openjdk8_test_x86-64_linux
Archiving artifacts
Recording test results
None of the test reports contained any result
Finished: SUCCESS
```

neomatrix369:
##### OpenJDK9 successful

```

'/mnt/jenkins/workspace/openjdk9_x86-64_linux_mani_pr_175/openjdk/build/linux-x86_64-normal-server-release/images/jdk' found

openjdk version "9-internal"
OpenJDK Runtime Environment (build 9-internal+0-adhoc.jenkins.openjdk)
OpenJDK 64-Bit Server VM (build 9-internal+0-adhoc.jenkins.openjdk, mixed mode)


Removing unnecessary files now...
Fetching the first tag from the OpenJDK git repo...
Currently at '/mnt/jenkins/workspace/openjdk9_x86-64_linux_mani_pr_175/openjdk/build/linux-x86_64-normal-server-release/images'
moving jdk to jdk-9+181
Finished removing unnecessary files from jdk-9+181
Archiving the build OpenJDK image...
OpenJDK repo tag is jdk-9+181
Your final .tar.gz was created at /mnt/jenkins/workspace/openjdk9_x86-64_linux_mani_pr_175/openjdk/build/linux-x86_64-normal-server-release/images
Moving the artifact to /mnt/jenkins/workspace/openjdk9_x86-64_linux_mani_pr_175/OpenJDK9_x64_Linux_201702101909.tar.gz
All done!
Archiving artifacts
Recording test results
None of the test reports contained any result
Finished: SUCCESS
```

neomatrix369:
@gibfahn it worked on Jenkins but it didn't work on local environment (for me at least - guessing might have been the same for others).

But we rarely build locally - is that right?

gibfahn:
>@gibfahn it worked on Jenkins but it didn't work on local environment (for me at least - guessing might have been the same for others).

Okay, it seems like you get an error (on my shell), but the if condition doesn't run, so I guess that explains why it didn't break Jenkins.

```bash
$ if [ a -eq 0 ]; then echo true; else echo false; fi
[: integer expression expected: a
false
```

neomatrix369:
Yes thats how I remember seeing in the log files, when I noticed the issue

sxa555:
@neomatrix369 I'd strongly suggest putting it back the way it was, with the command on a previous line and testing $? as it's vastly more readable.

neomatrix369:
@sxa555 I'll revert it to the original lines, as mentioned earlier I only changed it as shellcheck was raising warnings on those lines.

@gibfahn What is your view on the original implementation?

gibfahn:
>I'd strongly suggest putting it back the way it was, with the command on a previous line and testing $? as it's vastly more readable.

I disagree that it's more readable. Using `$?` is arguably _slightly_ more readable for those familiar with bash, but less readable for people familiar with any other programming language.

>@gibfahn What is your view on the original implementation?

I prefer it as it is now. If the worry is long lines, then break it up like:

```bash
     if ! (bash ./configure --prefix="${WORKING_DIR}"/"${OPENJDK_REPO_NAME}"/installedfreetype "${FREETYPE_FONT_BUILD_TYPE_PARAM}" && $MAKE all && $MAKE install); then 
```

becoming:

```bash
if ! (./configure --prefix="${WORKING_DIR}"/"${OPENJDK_REPO_NAME}"/installedfreetype "${FREETYPE_FONT_BUILD_TYPE_PARAM}" && 
  $MAKE all && $MAKE install); then 
```

I don't see how this

```bash
./configure --prefix="${WORKING_DIR}"/"${OPENJDK_REPO_NAME}"/installedfreetype "${FREETYPE_FONT_BUILD_TYPE_PARAM}" && 
  $MAKE all && $MAKE install
if [ $? -ne 0 ]; then 
```

is cleaner.

For shorter conditions I think the `if ! (command)` is much cleaner, i.e.

```bash
if ! (wget "${JAVA_SOURCE}"); then
```

rather than 

```bash
wget "${JAVA_SOURCE}"
if [ $? -ne 0 ]; then 
```

I'd be interested to hear what others think though.

neomatrix369:
I see the same clarity with the ` if ! (command)`, although now our scripts have two different conventions, so best to agree to one and I'll be happy to steer it to the agreed one.

neomatrix369:
@sxa555 @gibfahn Should we do a drive to standardise/unify this implementation across all our scripts, once we have come to an agreement? I#ll be happy do change it for us.

gibfahn:
SGTM

PR #180, opened 2017-10-10
Fix failing s390x openjdk9 build on Jenkins

neomatrix369:
Ensuring the JVM_VARIANT variable is being populated with the value from the appropriate value from environment


neomatrix369:
Building fix via https://ci.adoptopenjdk.net/job/openjdk8_build_s390x_linux/51/console  (openjdk8) and https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk9_build_s390x_linux_mani_pr_180/1/console (openjdk9)

neomatrix369:
Both openjdk8 and openjdk9 builds are successful!

PR #187, opened 2017-11-20
Enable openjdk8 openj9 build

adamfarley:
These changes should allow:

1) Someone to build Openjdk8 on Openj9

and

2) Tolerance for jdk#u's. Previously the permissible version strings were jdk8u and jdk9. However, since jdk9u is in the cards, and the openj9 extensions repository doesn't use the "u" suffix, I've added code so the "u" is discarded on openj9 builds, and optional on the hotspot builds (so long as there's a mercurial repository with/without the u you are/aren't using).


karianna:
Hmm, if this is a OpenJ9 vs OpenJDK naming thing, then perhaps better to have a variable, say `OPENJ9` explicitly set to swap between the two?  Else this looks a little like arbitrary swapping of 'u' and not 'u'...

docker/jdk8/x86_64/ubuntu/Dockerfile-openj9
```diff
@@ -0,0 +1,38 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM openjdk_container
+
+MAINTAINER Mark Stoodley <mstoodle@ca.ibm.com>
```
tellison: I assume @mstoodle is aware that you are naming him as the maintainer?!
We should consider a community e-mail address to redirect any contact made via this address to the team, rather than burdening an individual.

karianna:
Thanks for the PR! I'm OK with the code at a nit picking detail level, but something doesn't feel right about the approach, although that could be me not liking that we even have jdk8, 9 etc hard coded in here.  @neomatrix369 - further thoughts?

adamfarley:
I propose accepting this code (if you're happy with it), and I'm happy to raise a further issue proposing a pattern-match of some kind (e.g. "jdk[1-9]+[0-9]*") to replace the validation bits.

makejdk-any-platform.sh
```diff
@@ -40,33 +41,40 @@ for i in "$@"; do
...
 
 if [ "$OPENJDK_VERSION" == "jdk9" ]; then
   export JDK_PATH="jdk"
   export CONFIGURE_ARGS_FOR_ANY_PLATFORM=${CONFIGURE_ARGS_FOR_ANY_PLATFORM:-"--disable-warnings-as-errors"}
-elif [ "$OPENJDK_VERSION" == "jdk8u" ]; then
+elif [ "$OPENJDK_VERSION" == "jdk8" ]; then
   export JDK_PATH="j2sdk-image"
 else
-  echo "Please specify a version with --version or -v , either jdk9 or jdk8u"
+  echo "Please specify a version with --version or -v , either jdk9 or jdk8, with or without a \'u\' suffix."
```
tellison: Since OpenJDK [jdk8](http://hg.openjdk.java.net/jdk8) and [jdk8u](http://hg.openjdk.java.net/jdk8u/jdk8u/) are different code repos, I find it confusing to make `--version jdk8u` a synonym for `--version jdk8`.  What's the rationale for not requiring the trailing 'u'?

docker/jdk8/x86_64/ubuntu/Dockerfile-openj9
adamfarley: The hazards of copy, paste, and edit. Can we use a placeholder until we get a group email, or do we need an actual person?

...Do we need specific file maintainer at all?

makejdk-any-platform.sh
adamfarley: jdk9 is soon to become 9u, and extensions for jdk 8 and 9 don't have a "u" in their repository name.

I thought this was simpler, solved the future 9u problem, and also allows the user to decide if they want a u while not totally relinquishing control.

adamfarley: Plus, if someone just copies the build script arguments for jdk8u and sticks "-variant=openj9" in there, it removes this problem for them.

docker/jdk8/x86_64/ubuntu/Dockerfile-openj9
mstoodle: I am ok with my name being used for the moment. The other Dockerfiles at the repo are maintained by @tellison and @gdams, so I'm in very good company ;) .

Nonetheless, I think it does make sense to have a more community focused maintenance story.

makejdk-any-platform.sh
mstoodle: @tellison raises a reasonable point. Since this "not having a u" thing is specific to the OpenJ9 Extensions repositories, maybe the variables could be constructed differently to reflect that.

Perhaps OPENJDK_VERSION should remain the actual full "forest" name (jdk8, jdk8u, jdk9, jdk9u, etc.) and we introduce an OPENJDK_VERSION_BASELEVEL or something like that, to denote the corresponding jdk8, jdk8, jdk9, jdk9 (respectively), and we use that variable to set the OpenJ9 extensions repository?

That's just a suggestion; @tellison @gdams @karianna @neomatrix369 any opinions?

karianna: That makes more sense to me yes. Making an implicit derivation of an environment variable into its own explicit declaration matches the convention of the code in this repo.

@adamfarley Thanks again for your patience on this, appreciate it's frustrating have good functionality held up in review!.

adamfarley: Thanks karianna.

If we don't seperate the "u" like this, I worry we'll be in for another problem when jdk9u (mercurial) happens, not to mention jdk10->jdk10u, jdk11->jdk11u, etc.

adamfarley: We could separate that into another issue I guess. Trying to solve too many problems at once, perhaps.

docker/jdk8/x86_64/ubuntu/Dockerfile-openj9
tellison: I've opened #188 for this.  It's certainly not a blocker for this PR.

makejdk-any-platform.sh
tellison: @adamfarley yep, let's think ahead, but not try to fix the future just yet :-)  Just make the current logic as clear as we can so it's easy to change for the next person.

So as @mstoodle wrote, defining extra variables with clear names that denote what they are will be better than manipulating a smaller set with append/pre-pend knowledge.
Though IMHO
> OPENJDK_VERSION should remain the actual full "forest" name" 

it may be better to rename the variable `OPENJDK_FOREST_NAME` etc if we go that route to maintain maximum readability.

adamfarley: Ok, done and comitted.

mstoodle: I'm going to quibble a bit on the choice of OPENJDK_FOREST_NAME, just because in JDK10 the "forest" has actually been collapsed (that's why I put it in quotes originally :) ). But I don't care enough to block it because I can still squint and accept it; I'm just registering I don't think that name is 100% accurate either, so if we're going to change it, I'd suggest changing it to OPENJDK_REPO_NAME (sorry @adamfarley :) ). I am willing to leave that to the JDK10 support work so this can move forward.

karianna: @mstoodle Can you add that to #189 ?

PR #196, opened 2017-12-08
fixed stripping of \u\ char for jdk8

stevewallin:
cloning of jdk8u failed when running with bash on macos


gdams:
yeah we have seen this error, it's to due with the fact that macOS uses bash 3 and this syntax only works on 4.

PR #197, opened 2017-12-10
fix docker building of jdk8u

stevewallin:
docker file path incorrectly generated when specifying jdk8u


gdams:
issue reported here: https://github.com/AdoptOpenJDK/openjdk-build/issues/194

PR #199, opened 2017-12-13
fix finding existing containers

stevewallin:
when looking for containers to clean up the name was truncated in the docker ps command.  using --no-trunc avoids truncations and allows the name to be found.


karianna:
@gdams I thought you've moved all the docker code over to `openjdk docker`, or does some still remain here?

gdams:
This is the docker based build rather than the docker images that we create

PR #200, opened 2017-12-14
Add jdk10 docker support

stevewallin:


gdams:
test build: https://ci.adoptopenjdk.net/job/openjdk10_build_docker/3/console

makejdk-any-platform.sh
```diff
@@ -59,13 +59,14 @@ for i in "$@"; do
   esac
 done
 
-if [ "$OPENJDK_CORE_VERSION" == "jdk9" ]; then
+echo $OPENJDK_CORE_VERSION
```
gdams: Did you mean to leave this here? It either needs to be removed or double quoted

docker/jdk10/x86_64/ubuntu/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+&& rm -rf /var/lib/apt/lists/*
+
+# Pick up build instructions
+RUN mkdir -p /openjdk/target
+
+ADD sbin /openjdk/sbin
+
+WORKDIR /openjdk/build/
+
+ENV JDK_BOOT_DIR=/usr/lib/jvm/java-1.8.0-openjdk-amd64
```
gdams: We are going to need a jdk9 boot dir here... I would suggest using `zulu-9-amd64` as `openjdk-9-jdk` doesn't currently work on ubuntu :)

gdams:
I've added a commit to your PR that adds zulu-9 as we need a jdk9 boot java

gdams:
okay so it took a few commits but we are there! zulu-9 installed https://ci.adoptopenjdk.net/job/openjdk10_build_docker/8/console.

Seeing the following failure though:

```bash
Not reconfiguring due to the presence of config.status in /openjdk/build
Building the JDK: calling 'make images'

/bin/mkdir: missing operand
Try '/bin/mkdir --help' for more information.
make: *** No rule to make target 'images'.  Stop.
/openjdk/sbin/build.sh: line 224: [: : integer expression expected
```

It's worth noting that we are seeing this on our jdk9 docker builds too so it is probably a bug in our existing code somewhere

neomatrix369:
@gdams I have had similar issue, this isn't an issue when we run it native but inside the Docker container, try checking the version of bash between the two environments: it might be expecting that line to read as `if [[... ]];...`, see https://github.com/AdoptOpenJDK/openjdk-build/issues/174

makejdk-any-platform.sh
stevewallin: good spot - I'll remove this

stevewallin:
Thinking about the boot JDK.  Did we want to use an LTS release for the boot ? 

Using 8 for 9, 10 and 11... and then using 11 for 12, 13, 14, 15, 16 & 17 ?

The LTS will be the only versions with security fixes applied for sustained period, so using the previous version for 11 will currently be using 10 which has no support for the entire life of 11. where as 8 will have security fixes until at least 2025. Thoughts ?

gdams:
That’s an interesting point that you make.....

karianna:
@stevewallin Probably no.  This is because a version (say 13) may need to rely on a feature in the previous non LTS version (say 12).

stevewallin:
I'm not sure how good an idea it is to keep insecure JVMs around for 8 years+ on build machines.  Given it's mostly used to bootstrap the javac compile stage - and I don't think it heavily uses new features the LTS releases might be better ?
Would be a good topic to discuss further at OpenJDK ?

gdams:
AFAIK the configure step currently fails for jdk10 builds unless you are using a jdk9 boot jdk

gdams:
are we okat to land this and we can discuss boot JDK's in a separate issue?

stevewallin:
Good with me

karianna:
@stevewallin I'm guessing that with some of the new bytecodes they're introducing in non LTS releases that javac may need to rely on on non LTS release bootstrap JDKs, but lets check over there.

PR #209, opened 2018-01-23
Updating the new hg repo for jdk 10

neomatrix369:
Changing the current source of changesets for JDK 10 from jdk10/jdk10 (http://hg.openjdk.java.net/jdk10/jdk10) to jdk/jdk (http://hg.openjdk.java.net/jdk/jdk/)

As per @Shura, the new announcement points to this change - http://mail.openjdk.java.net/pipermail/jdk-dev/2017-December/000388.html

Will need to change https://ci.adoptopenjdk.net/view/git-hg/job/git-diff-10/ to pass in the extra parameters to the new repo `jdk/jdk`

After merging we run Jenkins job: https://ci.adoptopenjdk.net/view/git-hg/job/git-hg-10/ followed by https://ci.adoptopenjdk.net/view/git-hg/job/git-diff-10/

Scripts impacted:
- /openjdk-build/git-hg/update.sh <=== might not be impacted, we will see once this PR is merged
- ~/openjdk-build/git-hg/diff.sh~

*Usage:*
```
$ diff.sh jdk10 jdk jdk
```
_Old usages will continue to work._

PR #210, opened 2018-01-28
Updating the new hg for jdk 10 repo

neomatrix369:
Follow on from the PR #209 and related to the change, amending the `update.sh` script to make it parameterised

Split the parameter passed, into the root forest name and the repo name.

This will help us amend the script in https://ci.adoptopenjdk.net/view/git-hg/job/git-hg-10/ the jenkins job to accept the new repo name, although `setup.sh` needs to be run before we can run `update.sh`.


neomatrix369:
@karianna I have updated the Jenkins job to update `jdk/jdk` forest/repo, although we need to have run the `setup10.sh` script first before the updates can happen.

@gdams when's a good time to run `setup10.sh` - not sure if it needs an empty repo or will it purge the existing repo and start from scartch?

neomatrix369:
git-diff jobs are failing for `jdk10` - https://ci.adoptopenjdk.net/view/git-hg/job/git-diff-10/125/console, so we do need to run the `setup10.sh` script for `jdk10` immediately.

PR #212, opened 2018-02-01
Updating the new hg for jdk 10 repo - fixing JDK10 repo name

neomatrix369:
After a conversation with Stuard and Tim on Slack, indeed we have been pointing to the incorrect repo also our scripts need to be able to accept parameters rather than have OpenJDK project repo names hard-coded in them.

Correction from `jdk/jdk` ==> `jdk/jdk10`

New PR is on its way to just do that. See #213 

This PR affects these jobs:
- https://ci.adoptopenjdk.net/view/git-hg/job/git-hg-10/
- https://ci.adoptopenjdk.net/view/git-hg/job/git-diff-10/


neomatrix369:
Will test this script once merged, currently we are downloading an incorrect hg repo for JDK10

PR #215, opened 2018-02-01
Add support for SapMachine bulids

sxa555:
Make build process accept SapMachine variant to build from github.com/SAP/sapmachine

I am assuming the default branch of `sapmachine`is the correct one to use.

I've also got a fix for the fact that a git error is currently thrown when the option not to do a shallow clone is specified, as (at present) that option is required to get around another issue in the SAP build (it tries to do a `git log HEAD^` which doesn't work on a shallow clone

Note to reviewers: I've made the variant option `SapMachine` - with hindsight perhaps it's better to leave it all lower case as `sapmachine` instead

Note 2 to reviewers: The two if statements for the variants could be replaced with a case ...

Note 3: This has been tested through the following job: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/SAPmachine10_build_x86-64_linux/ which I can move into the main build tab once this is accepted and the job is modified not to pull from my fork of this repository :-)


neomatrix369:
Just a side comment, I think trying to invoke `--jtreg` via the build script might not be a good think going forwards, we should make that an independent process.

In a way its making the build script brittle and we have to maintain this link each time we make changes of this nature.

`jtreg.sh` maintained on its own is more beneficial than via a call from within the build script which we may have intended in the past but never ever have used it in this effect.

What does everyone think about this, in favour of testing and maintaining both build and test scripts.

smlambert:
We are running the openjdk regression tests (https://ci.adoptopenjdk.net/view/OpenJDK%20tests/) as a separate test build where we call jtreg directly per test group jdk_io, jdk_lang, jdk_beans, etc, so its already independent in that respect.  

The one downside is that we do not have test machines for all platforms, once this is addressed, you may wish to not run the --jtreg option as part of the build as we duplicate testing, and increase build time.  I have created this test build https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk10_sap_openjdktest_x86-64_linux/ which will be triggered by the SAPmachine build that @sxa555 just added.  It will run all of the openjdk regression tests (minus the ones we have excluded due to known issues).

smlambert:
Would also prefer the build job name follow the same naming convention as other builds (as there seems to be one), so expected it to be named openjdk10_sap_build_x86-64_linux

neomatrix369:
@smlambert Thanks for sharing this with us, and yes let's do as you suggest
@sxa555 is it possible to apply the goodness of your work to other build types, platforms and OSes?

sxa555:
@neomatrix369 The changes were pretty straightforward and will work for any platform we choose to build on. The only issue (which may be a non-issue anyway!) is that the `applyingJCovSettingsToMakefileForTests()` function in `jtreg.sh` doesn't work as they don't have a jdk/test directory in their v10 source (ours does). So I'm going to leave that "broken" for now as it's not a prereq to getting a build out.

FYI Have created https://github.com/AdoptOpenJDK/openjdk-build/issues/218 for any further discussions since that's a better place for it than a PR for tracking purposes. I'll make my job available in the main build tab on jenkins as `openjdk10_sap_build_x86-64_linux` as @smlambert suggests for now and anyone can object in that issue if they so desire :-)

PR #220, opened 2018-02-13
Allow building from a specific git tag in the repository

sxa555:
Partial resolution for https://github.com/AdoptOpenJDK/openjdk-build/issues/214


karianna:
shell check fails.

sxa555:
Fixed

makejdk.sh
```diff
@@ -224,6 +228,9 @@ checkOpenJDKGitRepo()
        showShallowCloningMessage "fetch"
        git fetch --all ${SHALLOW_CLONE_OPTION}
        git reset --hard origin/$BRANCH
```
neomatrix369: A tag and branch are synonymous in git (possibly in hg as well) - maybe introducing on switch that take a commit point would be better, a commit point can be a commit sha, branch name or tag, the checkout command understands all of them.

`git reset --hard origin/$BRANCH`

can also 

`git reset --hard origin/$TAG`

or

`git reset --hard origin/[some sha]`

We might be introducing additional switches that might be doing the same thing

sxa555: Possibly, although we'd have to think about what's the right thing to do in light of https://github.com/AdoptOpenJDK/openjdk-build/pull/223 which uses the "tag" name to override the top level directory in the tarball from the default of "the last tag in the repo".

Perhaps having the "commit point" option as you say and a "directory name override" would work better going forward.

neomatrix369: Okay makes sense, lets use this one to fix the current issue(s) and then work on a better fix with a new PR,

PR #221, opened 2018-02-14
git-hg/update.sh: Ensure jenkins jobs fail if the push of the git tags fails

sxa555:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/211


git-hg/update.sh
```diff
@@ -76,7 +76,7 @@ do
 
     echo "Push"
 
-    git push github master --tags
+    git push github master --tags || exit 1
```
neomatrix369: Just got a chance to look at this, correct me if I'm wrong, this might only be a partial fix as if there is a failure elsewhere in the script the job might not record a failure

karianna: I think we're OK, looks like exit 1's are placed correctly elswhere

neomatrix369: Have a look at https://github.com/sxa555/openjdk-build/blob/385fba0bf6976bb96aa8496a7bdbb08fa0e2621d/git-hg/update.sh#L28 the blocks there don't have a error handling

Not trying to find faults but these changes will come back to us.

Also adding `|| exit 1` means we have compromised our general error handling somewhere.

Could help if we added a `set -e` at the top of the script - lets do that as well, whoever gets there first

PR #228, opened 2018-02-26
force openj9 builds to use gcc4.8

gdams:
OpenJ9 builds don't work on gcc4.8 or later. We currently have to use separate build machines in order to handle this but we can use environment variables to manage this.


sxa555:
I'm going to revert this to see if it's the cause of the build failures across the board at present. It's not obvious why though as this didn't happen when I set the CC/CXX vairables in the jobs last week. If it doesn't resolve it I'll reinstate this change (although bear in mind that by doing this we could be lowering the version of the compiler on non-X64 architectures where the later ones work ok ...)

PR #231, opened 2018-02-27
Allow user to pass bespoke configure parameters to configure.sh

adamfarley:

### DIFF
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -49,6 +49,7 @@ TAG=""
 KEEP=false
 JTREG=false
 BUILD_VARIANT=${BUILD_VARIANT-:""}
+USER_SUPPLIED_CONFIGURE_ARGS=""
 
 JVM_VARIANT=${JVM_VARIANT-:server}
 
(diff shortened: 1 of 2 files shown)


makejdk.sh
```diff
@@ -129,6 +130,9 @@ parseCommandLineArgs()
       "--variant"  | "-bv" )
       export BUILD_VARIANT="$1"; shift;;
 
+      "--config-args"  | "-ca" )
```
sxa555: I would change this to `--configure-args` to make it consistent with the variable names and the name of the `configure` script

sbin/build.sh
```diff
@@ -25,9 +25,10 @@ WORKING_DIR=$1
 TARGET_DIR=$2
 OPENJDK_REPO_NAME=$3
 JVM_VARIANT=${JVM_VARIANT:-server}
-OPENJDK_UPDATE_VERSION=$5
```
sxa555: What is the reason for putting this in as the fifth parameter instead of adding it to the end? While `build.sh` isn't often called outside `makejdk.sh` this would cause a backwards compatibility issue for anyone who is invoking it directly

adamfarley: I was concerned that blank variables weren't showing up on the command line in Jenkins, and that they might throw off the strict ordering requirement build.sh has. Missed the fact that the echo statement lacks the speech marks that act as placeholders in the actual command.

adamfarley:
@sxa555 - Both comments were addressed.

sxa555:
LGTM - Will merge and check

adamfarley:
Thanks @sxa555 :)

PR #233, opened 2018-02-28
Revert "force openj9 builds to use gcc4.8 (#228)"

sxa555:
This reverts commit 4f9e3014ed3d72f9669d819d9d66a90412df6258.


sxa555:
I'm going to merge this directly without secondary approval as it may be the cause of a build break - see my comment on https://github.com/AdoptOpenJDK/openjdk-build/pull/228

PR #234, opened 2018-02-28
Update branch of openjdk8-openj9 to openj9-0.8

sxa555:


sxa555:
Need to get this in as the --branch parameter doesn't appear to be working

PR #235, opened 2018-02-28
Correctly align the arguments for build.sh, and accomodate milestone

adamfarley:
Because if we pass the with-milestone config arg in on the command
line, we don't want to set it again internally.


sxa555:
Shouldn't cause any extra problems, therefore approving

adamfarley:
Thanks @sxa555

PR #236, opened 2018-02-28
Improve build script argument handling

adamfarley:
This improvement allows the build script to take arguments in any
order, and also prevents conflicts between user-specified configure
args and configure args we add as part of the script.

### DIFF
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -426,8 +426,22 @@ testOpenJDKInNativeEnvironmentIfExpected()
 
 buildAndTestOpenJDKInNativeEnvironment()
 {
-  echo "Calling sbin/build.sh $WORKING_DIR $TARGET_DIR $OPENJDK_REPO_NAME $JVM_VARIANT $OPENJDK_UPDATE_VERSION $OPENJDK_BUILD_NUMBER $TAG"
-  "${SCRIPT_DIR}"/sbin/build.sh "${WORKING_DIR}" "${TARGET_DIR}" "${OPENJDK_REPO_NAME}" "${JVM_VARIANT}" "${OPENJDK_UPDATE_VERSION}" "${OPENJDK_BUILD_NUMBER}" "${TAG}" "${USER_SUPPLIED_CONFIGURE_ARGS}"
+  BUILD_ARGUMENTS=""
+  declare -a BUILD_ARGUMENT_NAMES=("--source" "--destination" "--repository" "--variant" "--update-version" "--build-number" "--repository-tag" "--configure-args")
+  declare -a BUILD_ARGUMENT_VALUES=("${WORKING_DIR}" "${TARGET_DIR}" "${OPENJDK_REPO_NAME}" "${JVM_VARIANT}" "${OPENJDK_UPDATE_VERSION}" "${OPENJDK_BUILD_NUMBER}" "${TAG}" "${USER_SUPPLIED_CONFIGURE_ARGS}")
+
+  BUILD_ARGS_ARRAY_INDEX=0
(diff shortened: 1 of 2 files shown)


karianna:
Travis fails with 

```
<<<<<<< HEAD
^-- SC1073: Couldn't parse this here string.
^-- SC1009: The mentioned parser error was in this simple command.
    ^-- SC1072:  Fix any mentioned problems and try again.
```

adamfarley:
I know what happened. At one point Eclipse tried to mess with my code during a merge. Thought I'd removed it all. Apparently not. 

I will fix this.

adamfarley:
Still broken. How did this ever work locally? Will fix.

adamfarley:
Ok, problem fixed.

makejdk.sh
```diff
@@ -426,8 +426,22 @@ testOpenJDKInNativeEnvironmentIfExpected()
...
+  BUILD_ARGS_ARRAY_INDEX=0
+  while [[ ${BUILD_ARGS_ARRAY_INDEX} < ${#BUILD_ARGUMENT_NAMES[@]} ]]; do
+    if [[ ${BUILD_ARGUMENT_VALUES[${BUILD_ARGS_ARRAY_INDEX}]} != "" ]];
+    then
+        BUILD_ARGUMENTS="${BUILD_ARGUMENTS}${BUILD_ARGUMENT_NAMES[${BUILD_ARGS_ARRAY_INDEX}]} ${BUILD_ARGUMENT_VALUES[${BUILD_ARGS_ARRAY_INDEX}]} "
+    fi
+    ((BUILD_ARGS_ARRAY_INDEX++))
+  done
+  
+  echo "Calling sbin/build.sh ${BUILD_ARGUMENTS}"
```
karianna: update the echo to include ${SCRIPT_DIR)

sbin/build.sh
```diff
@@ -21,14 +21,49 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
+    WORKING_DIR="$1"; shift;;
+
+    "--destination" | "-d" )
+    TARGET_DIR="$1"; shift;;
+
+    "--repository" | "-r" )
+    OPENJDK_REPO_NAME="$1"; shift;;
+
+    "--variant"  | "-jv" )
+    JVM_VARIANT="$1"; shift;;
```
karianna: I think this means the -server default always gets overridden correct?

sbin/build.sh
```diff
@@ -101,7 +144,7 @@ configuringBootJDKConfigureParameter()
 
   echo "Boot dir set to ${JDK_BOOT_DIR}"
 
-  CONFIGURE_ARGS="${CONFIGURE_ARGS} --with-boot-jdk=${JDK_BOOT_DIR}"
+  addConfigureArg "--with-boot-jdk=" "${JDK_BOOT_DIR}"
```
karianna: is the space between correct? `--with-boot-jdk=" "${JDK_BOOT_DIR}"`, I'm not sure that would pass through correctly?

makejdk.sh
adamfarley: We didn't use SCRIPT_DIR there before my change, but sure. While we're at it. :)

sbin/build.sh
adamfarley: If it gets passed in as an argument, which is the case the case when you're using makejdk.sh to call it, yes.

sbin/build.sh
adamfarley: They are two separate arguments to the addConfigureArg function. They get concatenated with no space in the function. I have to do it split like that so we can check if a conflicting config parameter of the same name (but potentially different value) has been passed in by the user via the command line or environment variable.

adamfarley:
Thanks @karianna :)

PR #240, opened 2018-03-02
Repair JVM_VARIANT bug

adamfarley:
The old form of build.sh caused it to ignore the command line when
setting jvm_variant. This has been fixed, so a bug in makejdk.sh
is now causing build failures (makejdk.sh used :server, which is
wrong). This is the fix for that, so makejdk.sh now passes the correct
jvm_variant value.


sbin/build.sh
```diff
@@ -24,7 +24,7 @@ source "$SCRIPT_DIR/common-functions.sh"
 WORKING_DIR=""
 TARGET_DIR=""
 OPENJDK_REPO_NAME=""
-JVM_VARIANT="-server"
+JVM_VARIANT="JVM_VARIANT:-server"
```
sxa555: I believe this should be `JVM_VARIANT=${JVM_VARIANT:-server}` otherswwise it'll be set to the literal string `JVM_VARIANT:-server`

adamfarley: You're right. Will fix.

sxa555:
fixes https://github.com/AdoptOpenJDK/openjdk-build/pull/240

PR #246, opened 2018-03-06
Add isEmpty checks to some configure parameters

adamfarley:
If some parameters are empty, we don't want to use the configure
options that set those parameters. So is $hat is empty, we don't
want to use "--with-hat=$hat" .

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -89,6 +89,14 @@ addConfigureArg()
   fi
 }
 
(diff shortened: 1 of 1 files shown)


sxa555:
Looks like it'll fix the problem, therefore approving (hopefully there won't be any unintended side effects if some of those values are not specified!)

Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/244

PR #247, opened 2018-03-06
Undo build.sh typo

adamfarley:
Argument 1 will never be empty because we always pass hard-coded
strings. It is argument 2 we should be worried about, as that's the
one containing potentially-empty variables.


sxa555:
LGTM

PR #256, opened 2018-03-12
update version tags for jdk9 and later

gdams:
This changes the version string from:
`* Version string: 9-internal+0-adhoc.jenkins.openjdk (9-internal)`
to:
`* Version string: 9-adoptopenjdk+181 (9-adoptopenjdk)`


karianna:
This fixes an old issue or two.  Does it work OK for java 1.8 as well?

PR #260, opened 2018-03-15
Revert "Update branch of openjdk8-openj9 to openj9-0.8"

sxa555:
This reverts commit d0e331732423bcb8d9086975b7191dd4d04badb6.

### DIFF
--- a/makejdk.sh
+++ b/makejdk.sh
@@ -143,11 +143,7 @@ doAnyBuildVariantOverrides()
   if [[ "${BUILD_VARIANT}" == "openj9" ]]; then
     # current (hoping not final) location of Extensions for OpenJDK9 for OpenJ9 project
     REPOSITORY="ibmruntimes/openj9-openjdk-${OPENJDK_CORE_VERSION}"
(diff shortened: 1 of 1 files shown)


karianna:
Why does this need reverting?

sxa555:
@karianna The switch was made specifically to build the release off the separate branch that the repository under ibmruntimes is using. This switches it back to the development branch now that the release has been done.

karianna:
Hmmm, are we going to have to repeat this each time ibmruntimes puts out a version?

sxa555:
@karianna 

Short answer: Not necessarily ... 

Long answer: I was initially envisaging using the tag parameter for this (as from a github perspective checking out a tag and branch are the same) but while that worked it left us with the branch name (openj9-0.8.0) as the top level directory in the resultant tgz file, so as a "quick fix" I modified the script which also has the advantage of not risking forgetting to set the parameter to the branch. Going forward, Adding a separate "branch" parameter which won't cause an override in the top level directory name in the tgz is almost certainly the better option and would remove any need to make changes such as these.

gdams:
+1 on adding a branch parameter

PR #261, opened 2018-03-15
Add pipeline code

karianna:
This initial code is directly copied and pasted from the existing Jenkins pipeline jobs.


pipelines/openjdk10_nightly_pipeline.groovy
```diff
@@ -0,0 +1,42 @@
+println "building OpenJDK10"
+stage 'build OpenJDK'
+def Platforms = [:]
+Platforms["Linux"] = {
+    node {
+        build job: 'openjdk10_build_x86-64_linux', propagate: false
```
AdamBrousseau: It is not required, nor recommended to occupy a node (executor) while launching a downstream build as this will "hold" the node while the downstream job runs.

pipelines/openjdk10_nightly_pipeline.groovy
```diff
@@ -0,0 +1,42 @@
+println "building OpenJDK10"
+stage 'build OpenJDK'
```
AdamBrousseau: I would recommend using braces around a stage to clearly identify what the stage encapsulates
```
stage ('build OpenJDK') {
    parallel Platforms
}
```

pipelines/openjdk10_nightly_pipeline.groovy
```diff
@@ -0,0 +1,42 @@
...
+}
+parallel Platforms
+stage 'checksums'
+node {
+    def job = build job: 'openjdk10_build_checksum'
+}
+stage 'publish release'
+node {
+    def job = build job: 'openjdk_release_tool', parameters: [string(name: 'REPO', value: 'nightly'), string(name: 'TAG', value: "${JDK_TAG}"), string(name: 'VERSION', value: 'jdk10')]
+}
```
AdamBrousseau: no new line at end of file

pipelines/openjdk10_nightly_pipeline.groovy
```diff
@@ -0,0 +1,42 @@
...
+Platforms["aarch64"] = {
+    node {
+        build job: 'openjdk10_build_arm64_linux'
+    }
+}
+parallel Platforms
+stage 'checksums'
+node {
+    def job = build job: 'openjdk10_build_checksum'
```
AdamBrousseau: Is there a reason why the previous builds are launched `build job:` and the last 2 are stored in a variable `def job = build job:`?

pipelines/openjdk10_nightly_pipeline.groovy
AdamBrousseau: Is there a reason why the x86-64 build is the only one with `propagate: false`?

karianna:
Added a description and you can Squash and Merge on accepting the PR.  I'll tackle the rest of the Q's as well.

pipelines/openjdk10_nightly_pipeline.groovy
karianna: Should that node not be held whilst the job runs though?  Not sure I follow :-)

AdamBrousseau: The downstream job will grab it's own node

pipelines/openjdk10_nightly_pipeline.groovy
karianna: Fixed (will push shortly)

pipelines/openjdk10_nightly_pipeline.groovy
karianna: Fixed for all scripts

pipelines/openjdk10_nightly_pipeline.groovy
karianna: I couldn't tell you why sorry (don't know the intent of the original author here)

pipelines/openjdk10_nightly_pipeline.groovy
karianna: I'm actually not sure, reading the docs it says that setting to false lets the next build step execute even if this one was unstable.  I can't see why we'd want that so I've removed it.

AdamBrousseau: Right so, as the code reads now, if any builds fails, except for x86-64, don't execute the `checksums` stage

karianna: I'll confess I'm now in "I don't know how to fix this land".  I assume if I remove the node { ... } wrapper that the build job will still execute but not lock down that node?

AdamBrousseau: Correct

AdamBrousseau: Technically code not in a node block will execute on `master` but will not occupy an executor. All "real" work should always be wrapped in a node block. The part executing on master here is just the trigger of the downstream job.

karianna: Cool, I've pushed again and removed the node blocks for non master downstream jobs

karianna:
@gdams We had propagate: false set on most of the Linux builds, can you remember why?

pipelines/openjdk10_nightly_pipeline.groovy
```diff
@@ -0,0 +1,33 @@
...
+    }
+    Platforms["aarch64"] = {
+        build job: 'openjdk10_build_arm64_linux'
+    }
+    parallel Platforms
+}
+stage('checksums') {
+    node {
+        def job = build job: 'openjdk10_build_checksum'
+    }
```
AdamBrousseau: Not required to have any job wrapped in a node block. Any/all downstream jobs will get their own nodes.

smlambert:
If you wish to just merge what you have here, we (@sophiag and I) will follow-up with a 2nd PR to update with Adam's suggestions.

pipelines/openjdk10_nightly_pipeline.groovy
karianna: @AdamBrousseau Thanks for being patient with me :-).  I understand why we don't want to lock the master node when we're firing off downstream jobs that are running on other nodes. But in this case checksum and release tools are running on master.  Or is it the case that you're saying remove this anyway incase we shifted those jobs off master?

AdamBrousseau: Every job will grab its own node, whether it's master or any other node. Therefore you don't need to wrap any `build job:` step in  a node block.

sophia-guo: @karianna if you check the configure of job openjdk10_build_checksum you will see the job is restricted to node with master label https://ci.adoptopenjdk.net/job/openjdk10_build_checksum/configure 

screen shot 2018-03-21 at 2 01 22 pm

karianna: Right, so it's safe to remove the node{} block because the job will grab its own node (in this case master because that's what we've restricted it to).

Also as a side note I assume I can remove the `def job =` declaration as we don't reuse that var?

sophia-guo: Yes, for now we can remove the def job = .

karianna:
OK so I think I've sorted everything except for finding out why we propagate (or don't) - paging @gdams

karianna:
@AdamBrousseau That's awesome (pulling in the code directly) I'll wait on the propagate decision and then remove the comments and do just that.  Thanks for all of your help and patience (and @sophia-guo as well).

PR #275, opened 2018-03-23
Replace 'file' use with '-r' check for CaCerts

sxa555:
Should fix https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/326


sxa555:
Yeah standard sh syntax so should be fine everywhere

PR #278, opened 2018-03-26
Fix broken docker build for OpenJ9

dinogun:
Pass the right OPENJDK_CORE_VERSION to the docker build and update the Dockerfile-openj9 to add additional dependencies.

### DIFF
--- a/docker/jdk10/x86_64/ubuntu/Dockerfile-openj9
+++ b/docker/jdk10/x86_64/ubuntu/Dockerfile-openj9
@@ -17,12 +17,22 @@ MAINTAINER Mark Stoodley <mstoodle@ca.ibm.com>
 
 # Install required OS tools
 RUN apt-get update \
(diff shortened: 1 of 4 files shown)


karianna:
LMK if this PR is done and I'll happily merge

dinogun:
@karianna Please review. @mstoodle FYI.

dinogun:
I hadn't looked at the ansible scripts, thanks for pointing that out. I do see quite a few more packages getting installed in ansible.

bison
build-essential
cpanminus
curl
flex
gettext
libcurl4-openssl-dev
libexpat1-dev
libffi-dev
libfontconfig1-dev
libgmp3-dev
libmpfr4
libmpfr4-dbg
libmpfr-dev
libmpfr-doc
libwww-perl
ntp
openjdk-8-jdk

But the builds went through fine without these in my local docker build. So not sure if these are really needed.

PR #291, opened 2018-04-04
Set nightly builds deployment per platform

sophia-guo:
Set nightly builds deployment per platform, keep release as before. This will need co-operation with configuration updates of build openjdk_{JVMVERSION}_build_checksum and openjdk_release_tool

### DIFF
--- a/pipelines/openjdk10_nightly_pipeline.groovy
+++ b/pipelines/openjdk10_nightly_pipeline.groovy
@@ -17,26 +17,35 @@ for ( int i = 0; i < buildPlatforms.size(); i++ ) {
 	def archOS = buildMaps[platform].ArchOSs
 	jobs[platform] = {
 		def buildJob
+		def buildJobNum
+		def checksumJob
 		stage('build') {
 			buildJob = build job: "openjdk10_build_${archOS}"
+			buildJobNum = buildJob.getNumber()
 		}
 		if (buildMaps[platform].test) {
 			stage('test') {
 				typeTests.each {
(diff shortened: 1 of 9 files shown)


smlambert:
To clarify, this is a change to allow for deploying individual platform builds, even if some platforms fail to build.  

As @sophia-guo notes, the merge of this PR needs to be co-ordinated with minor updates to the openjdk_xxx_build_checksum jobs and the openjdk_release_tool job in Jenkins.

sophia-guo:
Ideally if all those openjdk_xxx_build_checksum jobs and the openjdk_release_tool job can also be contributed to build repo this kind of update job will be much easier.

karianna:
@sophia-guo Are those jobs not in source control?

sophia-guo:
Job openjdk_xxx_build_checksum jobs and the openjdk_release_tool job are not in source control as far as I know.

karianna:
OK let me see if I can add those, PR hopefully soon.

karianna:
@sophia-guo So the only code I think I can extract out of the checksum jobs are a small piece of shell script that executes:

```
for FILE in *; do
  case $FILE in
  *.sha256.txt) echo Skipping $FILE ;;
  *) sha256sum $FILE > $FILE.sha256.txt ;;
  esac
done
```

Is that what you want stored in source control?  Happy to oblige, but not 100% of the value.

karianna:
@sophia-guo and the openjdk release tool is sourced from https://github.com/AdoptOpenJDK/openjdk-website-backend.git

PR #293, opened 2018-04-06
Remove stray ')' from pipeline script

smlambert:
Fix pipeline failure due to extra )


smlambert:
FYI @sophia-guo

smlambert:
Merging without review as it is a minor typo, and need to unblock the builds.

PR #308, opened 2018-04-19
Handle and check exit code from make properly

sxa555:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/302
Logic to obtain the exit code was incorrect


sxa555:
@karianna Unfortunately that seems to be a travis issue rather than anything related to my PR as it's failing doing updates while installing the shellcheck tool as per issue 310 :-( I don't really want to push this one through til I've verified that shellcheck passes. I was kinda hoping it'd be a temporary issue but if we don't get a resolution I'll just run shellcheck myself and update based on the results :-)

PR #318, opened 2018-05-07
Remove systemtest from arm64_linux test list

smlambert:
No system tests running on arm64 at present.
- added individual test lists per platform and removed systemtest from the test list for arm.   
- fixes a pipeline failure that was causing the deploy phase of the pipeline from proceeding

Signed-off-by: smlambert <slambert@gmail.com>


pipelines/openjdk9_nightly_pipeline.groovy
```diff
@@ -2,14 +2,13 @@ println "building ${JDK_VERSION}"
 
 def buildPlatforms = ['Mac', 'Linux', 'zLinux', 'ppc64le', 'Windows', 'AIX', 'arm64']
 def buildMaps = [:]
-buildMaps['Mac'] = [test:true, ArchOSs:'x86-64_macos']
+buildMaps['Mac'] = [test:['openjdktest', 'systemtest'], ArchOSs:'x86-64_macos']
 buildMaps['Windows'] = [test:false, ArchOSs:'x86-64_windows']
-buildMaps['Linux'] = [test:true, ArchOSs:'x86-64_linux']
-buildMaps['zLinux'] = [test:true, ArchOSs:'s390x_linux']
-buildMaps['ppc64le'] = [test:true, ArchOSs:'ppc64le_linux']
+buildMaps['Linux'] = [test:['openjdktest', 'systemtest', 'perftest', 'externaltest'], ArchOSs:'x86-64_linux']
```
sophia-guo: I think we don't have perftest, externalTest for openjdk9 hotspot? Probably just remove those two test types.

smlambert: Yes, I can separate into different PR (I will create those builds, so we can run those types of types on Java9 builds shortly though).

smlambert:
@sophia-guo updated (removed perf/system for now).

PR #329, opened 2018-05-16
Add triggers for openjdk9-j9/hs and openjdk10-j9/hs for External Tests

Mesbah-Alam:
Add triggers for openjdk9 and openjdk10 builds on both j9 and hs for External tests.

Signed-off-by: Mesbah <Mesbah_Alam@ca.ibm.com>


Mesbah-Alam:
@sophia-guo - please review this PR.

smlambert:
I think we should also try to get docker installed on other machines to make use of the test machines that we have, see https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/347

Please add a comment to that infra issue, if you agree, and could benefit.

This way, even as we add more tests, and increase execution time, we are utilizing the lab better.

Mesbah-Alam:
@sophia-guo - Java9 and Java 10 builds should take similar time as taken by the Java8 counterparts. As you can see in `openjdk8_j9_externaltest` build (https://ci.adoptopenjdk.net/view/External%20tests/job/openjdk8_j9_externaltest_x86-64_linux/), the average time taken is approx ~ 2 hours.  
`openjdk8_hs_externaltest` test (https://ci.adoptopenjdk.net/view/External%20tests/job/openjdk8_hs_externaltest_x86-64_linux/) take longer, approx ~ 3 hours on average. 

We intend to parallellize these builds so different tests under them run in parallel. This should reduce the total time taken.

PR #352, opened 2018-05-30
Update release tags for AO8/J9 and AO9/J9

sxa555:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/351


sxa555:
@pshipton As per the comment I added to 351 I've made a change in the jenkins job to cover 10 which is currently done differently.

gdams:
I'm still not happy with the way that we are hard coding tags but this will have to do for now

PR #358, opened 2018-06-06
Refactor updateforesttags to use functions

karianna:


git-hg/movefile.awk
```diff
@@ -0,0 +1,11 @@
+#!/bin/awk -f
```
johnoliver: Unused file

karianna: removed

PR #360, opened 2018-06-06
Ensure tag from the source repo instead of openjdk-build is used

sxa555:
Gah ... This is kinda why I don't use functions in my own shell scripts - can be hard to trace the flow, but not to worry.

This change ensures is changes into the source repo before trying to determine the tag (otherwise it picks up the tag from the `openjdk-build` repo apparently! Hopefully this won't result in any issues further down the line.

I've accidentally created this branch in the AdoptOpenJDK repo instead of my own so when merged please feel free to delete the branch.


sxa555:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/359

PR #368, opened 2018-06-12
WIP: Update arm64 name to be aarch64

smlambert:
This change needs to be co-ordinated with updating Jenkins job names, and fixes the problem of no aarch64/arm64 builds showing up in the nightlies on the website.

### DIFF
--- a/pipelines/openjdk10_nightly_pipeline.groovy
+++ b/pipelines/openjdk10_nightly_pipeline.groovy
@@ -1,6 +1,6 @@
 println "building ${JDK_VERSION}"
 
-def buildPlatforms = ['Mac', 'Windows', 'Linux', 'zLinux', 'ppc64le', 'AIX', 'arm64']
+def buildPlatforms = ['Mac', 'Windows', 'Linux', 'zLinux', 'ppc64le', 'AIX', 'aarch64']
 def buildMaps = [:]
 buildMaps['Mac'] = [test:['openjdktest', 'systemtest'], ArchOSs:'x86-64_macos']
 buildMaps['Windows'] = [test:['openjdktest'], ArchOSs:'x86-64_windows']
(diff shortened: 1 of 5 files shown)


smlambert:
@sophia-guo - could you review please?

pipelines/openjdk10_nightly_pipeline.groovy
```diff
@@ -1,14 +1,14 @@
 println "building ${JDK_VERSION}"
 
-def buildPlatforms = ['Mac', 'Windows', 'Linux', 'zLinux', 'ppc64le', 'AIX', 'arm64']
+def buildPlatforms = ['Mac', 'Windows', 'Linux', 'zLinux', 'ppc64le', 'AIX', 'aarch64']
```
sophia-guo: This value is used as index. Either arm64 or aarch64 is ok.

pipelines/openjdk8_nightly_pipeline.groovy
```diff
@@ -1,13 +1,14 @@
 println "building ${JDK_VERSION}"
 
-def buildPlatforms = ['Mac', 'Windows', 'Linux', 'zLinux', 'ppc64le', 'AIX']
+def buildPlatforms = ['Mac', 'Windows', 'Linux', 'zLinux', 'ppc64le', 'AIX', 'aarch64']
```
sophia-guo: I'm not sure if we should add arm64 job to pipeline as it's still unstable? https://ci.adoptopenjdk.net/job/openjdk8_build_arm64_linux/

pipelines/openjdk10_nightly_pipeline.groovy
smlambert: I think I will change it anyway (even if change not needed, reads better).

pipelines/openjdk8_nightly_pipeline.groovy
smlambert: in which case, I can remove this change.  I would keep the java8 release pipeline change as its already added there

gdams: It should be stable now :)

smlambert:
Ok, I just saw @gdams note about the openjdk8_aarch64 builds being stable now, so will add to the openjdk8 nightly pipeline too.

smlambert:
If I can get one more review, then I will remove the WIP and merge when I rename the builds.

smlambert:
I believe all the build and test jobs are now renamed (thanks to whoever helped!), and also the 3 relevant checksum jobs updated.

PR #374, opened 2018-06-14
Use escape backslash for ABS_PLATFORM_BUILD_ROOT to enable Jcov result

sophia-guo:


karianna:
@sophia-guo I'm happy with this, do you need to test it before I merge?

sophia-guo:
@karianna I have added another change for java9 and up. Testing now. Will update when it's done.

sophia-guo:
Revert changes for java9 and up. Will work on java8 first.

sophia-guo:
@karianna it's good to go. Thanks.

PR #379, opened 2018-06-18
Pipelines allow alternate branch

sxa555:
At the moment we can't pass in an alternate parameter to the pipelines for the openj9 release build process. Previously we've done temporary modifications to the build jobs. Now that we have the branch able to be passed as a parameter to the build jobs, it makes sense to allow this to be passed in from the pipelines.

I've done this across all pipelines (not just the openj9 ones) for consistency, although we're only likely to use it for the openj9 builds. I've modified the pipelines to take an optional `ALT_BRANCH` parameter

### DIFF
--- a/pipelines/openjdk10_nightly_pipeline.groovy
+++ b/pipelines/openjdk10_nightly_pipeline.groovy
@@ -20,7 +20,8 @@ for ( int i = 0; i < buildPlatforms.size(); i++ ) {
 		def buildJobNum
 		def checksumJob
 		stage('build') {
-			buildJob = build job: "openjdk10_build_${archOS}"
+			buildJob = build job: "openjdk10_build_${archOS}",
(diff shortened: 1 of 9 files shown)


smlambert:
This is similar to the change I am working on to pass a central timestamp to builds, so shortly builds will take 2 parameters.

smlambert:
Where is $ALT_BRANCH set (in Jenkins config?) and is there a default value?

sxa555:
Default is whatever the scripts have hard coded into them. `master` for adoptopenjdk, `openj9` for Eclipse openj9 and I think `SAPMachine` for the SAP builds (See the `BRANCH` variable in https://github.com/AdoptOpenJDK/openjdk-build/blob/master/makejdk.sh which this provides an override for). This parameter is only needed if that needs to be modified, as openj9 do when they are releasing.

PR #382, opened 2018-06-20
Add timestamps to top-level pipelines

smlambert:
This is to pass PIPELINE_TIMESTAMP to build jobs (so they can be same across platforms, which will ensure they will be part of the same 'release' set on the website).

Signed-off-by: smlambert <slambert@gmail.com>

### DIFF
--- a/pipelines/openjdk10_nightly_pipeline.groovy
+++ b/pipelines/openjdk10_nightly_pipeline.groovy
@@ -2,7 +2,7 @@ println "building ${JDK_VERSION}"
 
 def buildPlatforms = ['Mac', 'Windows', 'Linux', 'zLinux', 'ppc64le', 'AIX', 'aarch64']
 def buildMaps = [:]
-def PIPELINE_TIMESTAMP = new Date(currentBuild.startTimeInMillis).format("yyyyddMMHHmm")
+def PIPELINE_TIMESTAMP = new Date(currentBuild.startTimeInMillis).format("yyyyMMddHHmm")
 
 buildMaps['Mac'] = [test:['openjdktest', 'systemtest'], ArchOSs:'x86-64_macos']
 buildMaps['Windows'] = [test:['openjdktest'], ArchOSs:'x86-64_windows']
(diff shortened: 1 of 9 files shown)


smlambert:
Tested out "live" with this pipeline build:
https://ci.adoptopenjdk.net/view/Pipelines/job/openjdk8_openj9_nightly_pipeline_scm/68/console

screen shot 2018-06-19 at 11 07 46 pm

Verified that build & test jobs work, and the artifacts are properly timestamped (with the exception of the win32 build, as I forgot to update the Jenkins job config for it, updated now).  

This fixes the problem we had been seeing where the api link for nightly/latest/binaries would not always return a result (because the nightly builds were not contained in the same github release object, since we changed the pipeline logic to not wait to publish nightlies unless all platforms built).  By matching the timestamps, the build artifacts are now all added to the same github release again.

When this is merged, we need to finish updating the build job configuration in Jenkins (to take the PIPELINE_TIMESTAMP as a parameter, rather than independently set the timestamp in each build job.

smlambert:
Updated to use yyMMdd (as per https://github.com/AdoptOpenJDK/openjdk-website-backend/pull/12).

smlambert:
I have updated all of the build jobs with the extra PIPELINE_TIMESTAMP parameter.

smlambert:
Yes, I was thinking that (common pipeline code) as I created this PR... 

I am very certain we can reduce this set (version & impl can be data to these), and I also considered creating a utilities.groovy script where we can put methods like getTimeStamp

gdams:
@smlambert updating the jobs has caused an issue... we won't be able to kick off jobs on their own anymore as the timestamp won't be generated... 

each job will need the following code added at the top to fix this:

```bash
[ -z $PIPELINE_TIMESTAMP ] && PIPELINE_TIMESTAMP="$(date +'%Y%d%m%H%M')"
```

smlambert:
Thanks George, I forgot about the desire to run these jobs independently of the pipeline trigger, I will proceed to update the job configurations as suggested.

PR #383, opened 2018-06-20
Enable openjdk code coverage stats

sophia-guo:


sophia-guo:
@karianna all of the jtreg testing is done via scripts in openjdk-tests repo. Currently the easiest and quickest  way to get code coverage stats is from openjdk-buiild repo. My next step is to move it to test repo. Also deliver to openjdk-build is to record those chages, which should also be useful when moving to test repo.

Example:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk8_jcov/

karianna:
OK, cool - if there's a plan in place I'm happy :-)

PR #389, opened 2018-06-25
Fix empty string PIPELINE_TIMESTAMP issue

sophia-guo:
Recent openjdk8 build timestamp is either empty or use the build timestamp itself. It should use the timestamp generated by upstream pipeline job, which is passed in the jdk build job.

### DIFF
--- a/pipelines/openjdk8_nightly_pipeline.groovy
+++ b/pipelines/openjdk8_nightly_pipeline.groovy
@@ -22,7 +22,7 @@ for ( int i = 0; i < buildPlatforms.size(); i++ ) {
 		def buildJobNum
 		def checksumJob
 		stage('build') {
-			buildJob = build job: "openjdk8_build_${archOS}"
+			buildJob = build job: "openjdk8_build_${archOS}",
 					parameters: [string(name: 'BRANCH', value: "$ALT_BRANCH"),
 					string(name: 'PIPELINE_TIMESTAMP', value: "${PIPELINE_TIMESTAMP}")]
 			buildJobNum = buildJob.getNumber()


sophia-guo:
empty timestamp:
https://ci.adoptopenjdk.net/job/openjdk8_build_x86-64_linux/397/

jdk build timestamp:
https://ci.adoptopenjdk.net/job/openjdk8_build_x86-64_linux/399/

sophia-guo:
@smlambert  really tricky one, no any clue from jenkins output.

smlambert:
I will merge your PR, to keep it the same as other scripts, but I think the empty string problem happens when someone runs the build job directly (not triggered from pipeline), and if I failed to add a line to the top of the build script in Jenkins to account for the manual build case.  The Mac build was missing this line (which I have now added:

[ -z $PIPELINE_TIMESTAMP ] && PIPELINE_TIMESTAMP="$(date +'%Y%d%m%H%M')"

sophia-guo:
Yes, adding the zero check can void the empty string when triggered manually. 

The hidden issue is if you see the jobs on [ppcle](https://ci.adoptopenjdk.net/job/openjdk8_build_ppc64le_linux/360/) and on [linux](https://ci.adoptopenjdk.net/job/openjdk8_build_x86-64_linux/399/) have different timestamps(OpenJDK8_ppc64le_Linux_**201825061558**.tar.gz, OpenJDK8_x64_Linux_**201825061600**.tar.gz). while job on [mac](https://ci.adoptopenjdk.net/job/openjdk8_build_x86-64_macos/386/) has empty string.
 These three  jobs were all  triggered by same [pipeline job](https://ci.adoptopenjdk.net/view/Pipelines/job/openjdk8_nightly_pipeline_scm/67/console) . That is because no  PIPELINE_TIMESTAMP passed in. Jobs are either using its own timestamp if there is the line ```[ -z $PIPELINE_TIMESTAMP ] && PIPELINE_TIMESTAMP="$(date +'%Y%d%m%H %M')" ```  added or using empty string if no timestamp generated in shell script.

PR #436, opened 2018-08-10
Fix odd character at end of a line and add Win32/xLinuix-LH to AO8/J9…

sxa555:
… release pipeline


sxa555:
Taking an executive decision to force this through since it just plain doesn't work just now and can't really get any worse ;-)

PR #450, opened 2018-08-20
Fix timestamp pattern for consistency

keithc-ca:
This is the same change as https://github.com/AdoptOpenJDK/openjdk-build/pull/427, applied to the 'new_build_scripts' branch.


karianna:
@johnoliver - for your review (are you still on that branch?)

keithc-ca:
The new build scripts seem to use the pattern `YYYY-MM-DD-HH-mm` so this changes would appear unnecessary: closing.

PR #495, opened 2018-09-11
ppcle for openj9 needs an ubuntu to run

karianna:


sxa555:
This is something we really need to fix for in order to complete https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/256 and has shown up on my radar again as a result of setting up the other ppc64le machine at https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/408.

PR #505, opened 2018-09-13
Try to fix some java 11 build issues

johnoliver:


pipelines/build/openjdk11_pipeline.groovy
```diff
@@ -55,7 +55,7 @@ def buildConfigurations = [
 
         arm32Linux    : [
                 os                  : 'linux',
-                arch                : 'arm32',
+                arch                : 'arm&&sxa',
```
karianna: Might want a comment around that.  Using Stuart's is very specific :-)

PR #513, opened 2018-09-15
 Confirm test exists before running it

johnoliver:


AdamBrousseau:
I think I understand the point of this change, but doesn't it cover up the real issue? In my opinion if a job doesn't exist either it should be created or the code shouldn't be trying to run it. Maybe I'm missing something from the larger picture?

johnoliver:
Yeah you can take the view that we need to create the jobs, think the biggest issue is there are quite a few java 11 tests missing. So until those are created (or this pull request goes in) they will block 11 nightlys

PR #553, opened 2018-09-21
 add windows 32 java 8

johnoliver:


pipelines/build/openjdk8_pipeline.groovy
```diff
@@ -36,13 +36,22 @@ def buildConfigurations = [
...
                 ],
                 test                : ['openjdktest']
         ],
 
+        x32Windows: [
+                os                  : 'windows',
+                arch                : 'x86-32',
+                additionalNodeLabels: [
+                        hotspot: 'win2008',
+                        openj9:  'win2012&&build-softlayer-win2012r2-x64-1'
```
sxa555: `build-softlayer-win2012r2-x64-1` should be unnecessary in here since that should be the only machine that has the `x86-32` tag

build-farm/platform-specific-configurations/windows.sh
```diff
@@ -24,7 +24,14 @@ export LANG=C
 export JAVA_HOME=$JDK_BOOT_DIR
 export BUILD_ARGS="--tmp-space-build ${BUILD_ARGS}"
 
-
+if [ "${ARCHITECTURE}" == "x86-32" ] && [ "${VARIANT}" == "hotspot" ]
```
sxa555: Should probably enable this for OpenJ9+JDK8 as well since that works ok :-)

sxa555:
LGTM - approved

PR #557, opened 2018-09-24
HTTPS everywhere

karianna:


sbin/prepareWorkspace.sh
```diff
@@ -170,7 +170,7 @@ checkingAndDownloadingFreemarker()
     echo "Skipping FREEMARKER download"
   else
 
-    wget -nc --no-check-certificate "http://www.mirrorservice.org/sites/ftp.apache.org/freemarker/engine/${FREEMARKER_LIB_VERSION}/binaries/apache-freemarker-${FREEMARKER_LIB_VERSION}-bin.tar.gz"
+    wget -nc --no-check-certificate "https://www.mirrorservice.org/sites/ftp.apache.org/freemarker/engine/${FREEMARKER_LIB_VERSION}/binaries/apache-freemarker-${FREEMARKER_LIB_VERSION}-bin.tar.gz"
```
dbaxa: @karianna I'd suggest verifying the `sha512` hash of this download, which is currently `534cb51b781e83b2109f43a9186a030a1c3d6c5c13117bd5a6168b10c9a3cbd010e6ce806fdc6fce66e4b8a59d8a752f0552758ec769795e2bd3b66c09c0442a` - as per https://www.apache.org/dist/freemarker/engine/2.3.28/binaries/apache-freemarker-2.3.28-bin.tar.gz.sha512.

karianna: Good idea - in fact we should SHA check all downloads across our scripts.  I'm adding a security issue for this

PR #564, opened 2018-09-26
Bring inline with master

johnoliver:


karianna:
Merged because we wanted the commits to be inline for our merge back into master

PR #565, opened 2018-09-26
Make some freetype downloading improvements. Add release table generation script

johnoliver:


README.md
```diff
@@ -1,3 +1,30 @@
+# This Branch is a legacy branch!!
```
karianna: I don't think this readme should be in this PR?

PR #601, opened 2018-10-02
Add perftest category  to all enabled platforms for jdk11

sophia-guo:
Add perftest category to all enabled platforms for jdk11. So when the test build is created it will be triggered automatically. No need to update those build script.
 
### DIFF
--- a/pipelines/build/openjdk11_pipeline.groovy
+++ b/pipelines/build/openjdk11_pipeline.groovy
@@ -17,7 +17,7 @@ def buildConfigurations = [
                 os                  : 'mac',
                 arch                : 'x64',
                 additionalNodeLabels: 'build-macstadium-macos1010-1',
-                test                : ['openjdktest', 'systemtest']
+                test                : ['openjdktest', 'systemtest', 'perftest']
         ],
 
         x64Linux  : [
@@ -35,7 +35,7 @@ def buildConfigurations = [
                 os                  : 'windows',
                 arch                : 'x64',
                 additionalNodeLabels: 'win2012',
-                test                : ['openjdktest']
+                test                : ['openjdktest', 'perftest']
         ],
 
         x32Windows: [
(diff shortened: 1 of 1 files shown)


pipelines/build/openjdk11_pipeline.groovy
```diff
@@ -55,27 +55,27 @@ def buildConfigurations = [
...
                 additionalNodeLabels: 'ubuntu',
-                test                : ['openjdktest', 'systemtest']
+                test                : ['openjdktest', 'systemtest', 'perftest']
         ],
 
         arm32Linux    : [
                 os                  : 'linux',
                 arch                : 'arm',
-                test                : ['openjdktest']
+                test                : ['openjdktest', 'perftest', 'perftest']
```
karianna: perftest twice

pipelines/build/openjdk11_pipeline.groovy
```diff
@@ -55,27 +55,27 @@ def buildConfigurations = [
                 os                  : 'linux',
                 arch                : 's390x',
                 additionalNodeLabels: 'build-marist-rhel74-s390x-2',
-                test                : ['openjdktest', 'systemtest']
+                test                : ['openjdktest', 'systemtest', 'perftest']
```
karianna: has this been run on win32 before?

sophia-guo: I don't think so @smlambert may confirm this? Though eventually we should run perf tests on all platforms.

smlambert: Certainly never run on win32 machines in the open.  

Also, some benchmarks are marked as not applicable for windows, so they will simply be skipped, example:
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/performance/idle_micro/playlist.xml#L27  (note: the ^os.win, not windows).

karianna: We can fix as need be - a good goal is to have all platforms running the same test suite and just excluding tests that are irrelevant or need fixing.

PR #602, opened 2018-10-03
Fix unbound variable error for --help arg

ParkerM:
Add check to prevent `--help` arg throwing unbound variable error (example below) when no java version is provided.

```bash
./makejdk-any-platform.sh --help
Starting ./makejdk-any-platform.sh to configure, build (Adopt)OpenJDK binary
Parsing opt: --help
/myworkspace/sbin/common/config_init.sh: line 158: $1: unbound variable
```

Signed-off-by: Parker Mauney <parkermauney@gmail.com>

### DIFF
--- a/sbin/common/config_init.sh
+++ b/sbin/common/config_init.sh
@@ -155,7 +155,10 @@ function parseConfigurationArguments() {
       shift;
 
       echo "Parsing opt: ${opt}"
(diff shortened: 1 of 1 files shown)


karianna:
@johnoliver I'm not sure the fix is in the right area, is that echo part of the debug we were doing earlier?

PR #606, opened 2018-10-03
WIP - Change variable names to keep consistent between jenkins and build scripts

johnoliver:
Probably want to hold off till after releases.

Renames:
javaVersion/javaToBuild -> forestName
target_os/os -> operatingSystem


build-farm/make-adopt-build-farm.sh
```diff
@@ -22,12 +22,12 @@ PLATFORM_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
 TIMESTAMP="$(date +'%Y-%m-%d-%H-%M')"
 
+OPERATING_SYSTEM=$(echo "${OPERATING_SYSTEM}" | tr '[:upper:]' '[:lower:]')
 export OPERATING_SYSTEM
-OPERATING_SYSTEM=$(echo "${TARGET_OS}" | tr '[:upper:]' '[:lower:]')
 
 
 echo "BUILD TYPE: "
-echo "VERSION: ${JAVA_TO_BUILD}"
+echo "VERSION: ${FOREST_NAME}"
```
karianna: Not sure if forest name is the right name, does openj9 use the idea of forest?

pipelines/build/build_base_file.groovy
```diff
@@ -39,16 +39,16 @@ def toBuildParams(enableTests, params) {
...
 
     def buildParams = [
-            JAVA_TO_BUILD: javaToBuild,
-            NODE_LABEL   : "${additionalNodeLabels}&&${configuration.os}&&${configuration.arch}",
+            FOREST_NAME: forestName,
+            NODE_LABEL   : "${additionalNodeLabels}&&${configuration.operatingSystem}&&${configuration.arch}",
             VARIANT      : variant,
             ARCHITECTURE : configuration.arch,
-            TARGET_OS    : configuration.os
+            OPERATING_SYSTEM    : configuration.operatingSystem
```
karianna: nitpick, nice to the have `:` all lined up

johnoliver:
I am not a fan of the naming either. We have never had a concept of just "VERSION" we have had OPENJDK_FOREST_NAME that is of the form "jdk8u" which most closely matches what we use in the Jenkins code hence why I mirrored the name in Jenkins. We also have OPENJDK_CORE_VERSION that is of the form "jdk8", the majority of the scripts use this however -v/--version argument typically takes what the scripts would call a OPENJDK_FOREST_NAME and not a CORE_VERSION.

Think your options here are redefine/rename OPENJDK_FOREST_NAME in the existing scripts, or start using CORE_VERSION jenkins side then add code to add the "u" if needed to build the FOREST_NAME.

karianna:
Thanks for the details! `OPENJDK_FOREST_NAME` can also be interesting values like `amber`.  I wonder if it's better-called something like OPENJDK_SOURCE.  I've set a reminder to myself to look at this properly in the morning after coffee

PR #607, opened 2018-10-04
Try to fix cacerts on windows

johnoliver:


johnoliver:
so it is not skipped when run on windows

PR #609, opened 2018-10-05
build: change extracted image for macOS

gdams:
THIS IS A BREAKING CHANGE

This will change the directory layout of the binaries to be the apple layout which is built at the same time as the main `JDK` and `JRE` images

### DIFF
--- a/configureBuild.sh
+++ b/configureBuild.sh
@@ -127,10 +127,22 @@ setVariablesForConfigure() {
     [ "$openjdk_core_version" == "${JDKHEAD_CORE_VERSION}" ]; then
     local jdk_path="jdk"
     local jre_path="jre"
+    case "${BUILD_CONFIG[OS_KERNEL_NAME]}" in
+    "darwin")
+      local jdk_path="jdk-bundle/jdk-*.jdk"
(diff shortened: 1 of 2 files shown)


configureBuild.sh
```diff
@@ -127,10 +127,22 @@ setVariablesForConfigure() {
     [ "$openjdk_core_version" == "${JDKHEAD_CORE_VERSION}" ]; then
     local jdk_path="jdk"
     local jre_path="jre"
+    case "${BUILD_CONFIG[OS_KERNEL_NAME]}" in
```
karianna: Is that correct for Java 9+?

configureBuild.sh
```diff
@@ -127,10 +127,22 @@ setVariablesForConfigure() {
...
+    "darwin")
+      local jdk_path="jdk-bundle/jdk-*.jdk"
```
karianna: I assume `local` in bash is local to the scope it's in (i.e. the if statement)

configureBuild.sh
gdams: yup

configureBuild.sh
gdams: yup

PR #626, opened 2018-10-09
Use wget to pull JEP319 cacerts on Windows

sxa555:


karianna:
For reference #545 was the original PR that use curl across all platforms - @johnoliver I can't remember the history as to why.

sxa555:
See also https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/506

PR #644, opened 2018-10-15
Remove debug files from windows build

johnoliver:


karianna:
For future support see https://github.com/AdoptOpenJDK/openjdk-build/issues/146

PR #675, opened 2018-10-20
Start building JDK11 for windows x86

ali-ince:
Fix compilation failures by adding `--disable-warnings-as-errors` configure option and setting toolset version to 2013 for windows x86-32 builds.


ali-ince:
@karianna I've tested the changes with https://ci.adoptopenjdk.net/view/work%20in%20progress/job/jdk11-windows-x86-32-hotspot-wip/2/ and it seems ok.

PR #711, opened 2018-11-05
Set vendor information for java 11

johnoliver:


sbin/build.sh
```diff
@@ -191,6 +191,10 @@ configuringVersionStringParameter()
     addConfigureArg "--without-version-pre" ""
     addConfigureArgIfValueIsNotEmpty "--with-version-build=" "${buildNumber}"
     addConfigureArg "--with-vendor-version-string=" "AdoptOpenJDK"
+    addConfigureArg "--with-vendor-url=" "https://adoptopenjdk.net/"
+    addConfigureArg "--with-vendor-name=" "AdoptOpenJDK"
+    addConfigureArg "--with-vendor-vm-bug-url=" "/dev/null"
```
karianna: I think folks should be able to report issues to openjdk-build - we then triage or pass upstream as needed.  Or is this property an automated bug reporter?

PR #727, opened 2018-11-09
Update to set new repository var format for OpenJ9/SAP builds

sxa555:
Fixes `repository` values after a change to what it's set to being introduced in https://github.com/AdoptOpenJDK/openjdk-build/commit/23f3c289306bc10548d451716d18b326c7e945fb


karianna:
Probably superseded by #728 - but lets see

PR #729, opened 2018-11-09
Try bumping freetype to 2.9.1

johnoliver:


johnoliver:
I think we only override the version for aarch

PR #752, opened 2018-11-18
Add linux+osx Travis builds

skybber:
Add better integration with Travis. It builds OSX and Linux images (support of Windows on Travis is still in early stages) . Target images are not uploaded to any destination, it is left on users. Travis integration is inteded as a simplification of customized AdoptOpenJDK's builds, can be used for build script development, tuning, experiments etc. Build reseults at https://travis-ci.org/skybber/openjdk-build/builds/456612289

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
  
 install:
  - choco upgrade msys2
- - ps a
+ - ps axu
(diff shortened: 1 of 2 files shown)


karianna:
I'm not sure this is the right thing to do in this case.  We lose quite a bit if we drop shellcheck and the jenkins farm replicates this facility but on more platforms.

PR #756, opened 2018-11-22
WIP initial guess at mac installer build

johnoliver:


gdams:
for major version we can probably do the following:

```groovy
String text = "jdk8u";
String digits = text.replaceAll("[^0-9.]", "");
System.out.println(digits);
```
prints
```
8
```

PR #758, opened 2018-11-22
Do not skip freetype on ppc64le and aarch64

sxa555:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/754

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -61,12 +61,6 @@ then
   export CONFIGURE_ARGS_FOR_ANY_PLATFORM="--with-jobs=4 --with-memory-size=2000 --disable-warnings-as-errors"
 fi
 
-# Skip Building Freetype for certain platforms
-if [ "${ARCHITECTURE}" == "aarch64" ] || [ "${ARCHITECTURE}" == "ppc64le" ]
-then
-  export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
-fi
-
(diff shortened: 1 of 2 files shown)


sxa555:
Not sure why but `ft2build.h` wasn't being copied to `installedfreetype/include` during `make install` (I tested standalone with freetype 2.9.1 and it seemed not to have the problem) ...

Hmmm looks like there is some funky stuff to compensate for what freetype 2.5.2 does at https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/prepareWorkspace.sh#L219 but for whatever reason that isn't working properly on the aarch64 boxes.

Wiping `installedfreetype` in the workspace appears to have allowed it to progress on the next run so I'm going to assume this was a transient error and shouldn't hold up merging this PR

sxa555:
OK The ppc64le boxes were pulling freetype 2.9.1, aarch64 pulling 2.5.2 ... Looks like we're explicitly locking it to 2.5.2 on aarch64 (https://github.com/AdoptOpenJDK/openjdk-build/blob/master/configureBuild.sh#L242 - which was clearly rather pointless since we have it disabled entirely...) WIll push another commit to remove that setting

sxa555:
Reasons for the above are probably lost in time as they were both introduced as part of a refactor under https://github.com/AdoptOpenJDK/openjdk-build/pull/393

sxa555:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/SXA-freetype_aarch64/5/ is now running through OK with freetype 2.9.1 consistent with the other platforms

sxa555:
aarch64 build passed - removing WIP - now ready to merge

PR #759, opened 2018-11-22
Do not force /usr/bin to start of PATH on Linux/s390x/J9

sxa555:
Fixes https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/595

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -29,8 +29,6 @@ then
 
   if [ "${VARIANT}" == "openj9" ]
   then
-    export PATH="/usr/bin:$PATH"
-
     if [ "${JAVA_TO_BUILD}" == "${JDK8_VERSION}" ] || [ "${JAVA_TO_BUILD}" == "${JDK9_VERSION}" ] || [ "${JAVA_TO_BUILD}" == "${JDK10_VERSION}" ]
     then
       if which g++-4.8; then


sxa555:
@karianna /usr/bin should still be in there (otherwise not a lot would work I suspect!) just not at the front. I've run a build using the changes in this PR and it appears to have come out ok. I presume this was set exclusively on Linux/s390x at some point to fix a transient problem, but it does not appear to cause an issue now (See https://ci.adoptopenjdk.net/view/work%20in%20progress/job/SXA-j9s390xpath/1 if you're nervous but that seems to have got most of the way through now) and this additional path entry has not been in place on other platforms as this was within a s390x/j9 specific section so this restores consistency. If nothing else if this does cause a break somewhere we'll be able to discover and document why it was put in :-)

karianna:
> @karianna /usr/bin should still be in there (otherwise not a lot would work I suspect!) just not at the front. I've run a build using the changes in this PR and it appears to have come out ok. I presume this was set exclusively on Linux/s390x at some point to fix a transient problem, but it does not appear to cause an issue now (See https://ci.adoptopenjdk.net/view/work%20in%20progress/job/SXA-j9s390xpath/1 if you're nervous but that seems to have got most of the way through now) and this additional path entry has not been in place on other platforms as this was within a s390x/j9 specific section so this restores consistency. If nothing else if this does cause a break somewhere we'll be able to discover and document why it was put in :-)

I'm convinced - thank you :-).

PR #761, opened 2018-11-23
Fix using clean workspace

johnoliver:


sxa555:
Just to be clear - this change is just going to enable the option to clean the workspace after the build (since it's still calling the jenkins `cleanWs` operation) and not actually clean before building - is that correct?

karianna:
> Just to be clear - this change is just going to enable the option to clean the workspace after the build (since it's still calling the Jenkins `cleanWs` operation) and not actually clean before building - is that correct?

We'll soon find out - going to test this today.

PR #768, opened 2018-11-26
disable Deferred Wipeout

johnoliver:


AdamBrousseau:
Have you verified this works?
We also had tried this option at OpenJ9 and it was broken.
See https://issues.jenkins-ci.org/browse/JENKINS-54225

PR #779, opened 2018-11-29
Add a releaseApproval check for releases

karianna:


karianna:
@johnoliver Perhaps review this after we land your beast.

PR #783, opened 2018-11-30
Only select tags prefixed 'jdk'

sxa555:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/748

Marking as WIP because I want unanimous consensus across reviewers on the solution - As per a comment I've made in there we could do the following as alternatives:

- Knock out anything with `openj9` in it
- Use a more strict patter to match the JDK8 and JDK11 formats rather than just a `jdk` prefix
- Adjust the rather arbitrary 100 figure (I'm assuming people won't be creating that many tags in their repositories between official releases!)


sbin/build.sh
```diff
@@ -478,9 +478,9 @@ makeACopyOfLibFreeFontForMacOSX() {
 getFirstTagFromOpenJDKGitRepo()
 {
     git fetch --tags "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
-    justOneFromTheRevList=$(git rev-list --tags --max-count=1)
-    tagNameFromRepo=$(git describe --tags "$justOneFromTheRevList")
-    echo "$tagNameFromRepo"
+    RevList=$(git rev-list --tags --max-count=100)
```
karianna: style issue with this var name

karianna: Not sure on the performance impact of pulling down 100

karianna: 100 is a magic number

sbin/build.sh
```diff
@@ -478,9 +478,9 @@ makeACopyOfLibFreeFontForMacOSX() {
...
+    firstMatchingNameFromRepo=$(git describe --tags "$RevList" | grep ^jdk | head -1)
```
karianna: Hmmmm, fairly big assumption to make.  I'd want to see this get extracted to a labeled const RELEASE_TAG_PREFIX or similar.

We do also have the following release tag in aarch64 `aarch64-jdk8u181-b13` - although we deliberately did rewrite that as `jdk8u181-b13` - see https://github.com/AdoptOpenJDK/openjdk-aarch64-jdk8u/releases

sbin/build.sh
sxa555: When I tried it on the command line is was close enough to instantaneous not to be a concern

sbin/build.sh
sxa555: Have relaxed it to ensure it has "jdk" anywhere but excludes `openj9` ... I could have just excluded `openj9` but for now this felt better in case we get and `jdk8u192-openj9` or similar tags that we want to avoid

sbin/build.sh
```diff
@@ -475,12 +475,13 @@ makeACopyOfLibFreeFontForMacOSX() {
 
 
 # Get the first tag from the git repo
+# Excluding "openj9" tags as they have other ones for milestones etc. that get in the way
```
karianna: other tag names you mean?

sbin/build.sh
```diff
@@ -475,12 +475,13 @@ makeACopyOfLibFreeFontForMacOSX() {
...
 getFirstTagFromOpenJDKGitRepo()
 {
     git fetch --tags "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
-    justOneFromTheRevList=$(git rev-list --tags --max-count=1)
-    tagNameFromRepo=$(git describe --tags "$justOneFromTheRevList")
-    echo "$tagNameFromRepo"
+    revList=$(git rev-list --tags --max-count=100)
```
karianna: 100 is a magic number can you extract that to a named const?

sbin/build.sh
sxa555: Comment adjusted

sbin/build.sh
sxa555: Extracted into a variable in the constants file

PR #802, opened 2018-12-13
Disable ccache on JDK11+OpenJ9 on x/p-Linux

sxa555:
The changes to enable openssl on JDK11 seem to have caused problems when CCACHE is enabled.
This doesn't appear to occur outside jenkins so at least for now I'm going to disable it


sxa555:
Yep temporary (I hope!) related issue will be kept around to cover a subsequent better fix

PR #804, opened 2018-12-14
JDK11 with OpenJ9 - enable openssl building from source on macos and Linux/s390x

sxa555:


sxa555:
(Merging myself as I don't see John or Martijn online just now and this is currently a build break on Linux/s390x and I will be testing myself immediately)

PR #811, opened 2018-12-20
initial solaris solaris suppport

gdams:
test build passing here https://ci.adoptopenjdk.net/job/jdk8u-solaris-sparcv9-hotspot-test/

### DIFF
--- a/sbin/common/config_init.sh
+++ b/sbin/common/config_init.sh
@@ -280,7 +280,7 @@ function setBranch() {
 
 # Set the config defaults
 function configDefaults() {
- 
+
   # The OS kernel name, e.g. 'darwin' for Mac OS X
   BUILD_CONFIG[OS_KERNEL_NAME]=$(uname | awk '{print tolower($0)}')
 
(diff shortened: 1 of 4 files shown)


sbin/common/config_init.sh
```diff
@@ -309,7 +309,12 @@ function configDefaults() {
   BUILD_CONFIG[FREETYPE_FONT_VERSION]="2.9.1"
   BUILD_CONFIG[FREETYPE_FONT_BUILD_TYPE_PARAM]=""
 
-  BUILD_CONFIG[MAKE_COMMAND_NAME]="make"
+  if [ "${BUILD_CONFIG[OS_KERNEL_NAME]}" == "aix" ] || [ "${BUILD_CONFIG[OS_KERNEL_NAME]}" == "SunOS" ]; then
```
karianna: Are we not covering this AIX case elsewhere?  How is it currently working?

sbin/prepareWorkspace.sh
```diff
@@ -47,7 +47,7 @@ checkoutAndCloneOpenJDKGitRepo()
     set +e
     git --git-dir "${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/.git" remote -v
     echo "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}"
-    git --git-dir "${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/.git" remote -v | grep "origin.*fetch" | grep "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" | grep --quiet "${BUILD_CONFIG[REPOSITORY]}"
+    git --git-dir "${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/.git" remote -v | grep "origin.*fetch" | grep "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" | grep "${BUILD_CONFIG[REPOSITORY]}"
```
karianna: I think we still want quiet....

gdams: quiet isn't a flag on solaris grep and this only logs a single line matching the repo so IMO not much noise

sbin/common/config_init.sh
gdams: AIX was hacked on the machines whereas I've now made it universal

PR #817, opened 2019-01-03
Add type=jdk parameter to JDK10 download API call

sxa555:


sxa555:
Causing a build break so pushing through ...

Issue #796, opened 2018-12-06
Missing JDK11 JRE builds


tony--:
Originally I noticed that OpenJ9/macOS JRE builds were missing, but @sxa555 responded that "there haven't been any published JREs anywhere for a while".

[Related thread on Slack](https://adoptopenjdk.slack.com/archives/C09NLQQAV/p1544056768014000).

Looking at [JDK11+OpenJ9 nightlies](https://adoptopenjdk.net/nightly.html?variant=openjdk11&jvmVariant=openj9), the most JRE recent builds for Mac/Linux are from 8 November and the most recent for Windows is from 4 November.

taku0:
Maybe related: https://github.com/AdoptOpenJDK/openjdk-build/issues/697

bennofs:
There used to be OpenJ9/macOS jre builds, as http://blog.io7m.com/2018/11/18/choosing-coffee.xhtml confirms:

```
0901dc5946fdf967f92f7b719ddfffdcdde5bd3fef86a83d7a3f2f39ddbef1f8       | x64      | macos    | 11       | openj9   | jre  | 38.63   MB | urn:net.adoptopenjdk.raw     | production
```

The link for it (https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.1%2B13/OpenJDK11-jre_x64_mac_openj9_11.0.1_13.tar.gz) doesn't work anymore though.

gdams:
JRE builds for JDK11 are now working

PR #837, opened 2019-01-15
Fix java 11 jre build

johnoliver:


johnoliver:
Also includes additional fix of comparing core versions that relates to macos configuration

tony--:
Thanks for the fix!
When should we expect to see new release build JREs?
I'd like to verify that #796 is fixed.

karianna:
I think we'll look to release 11.0.2 some time next week

victornoel:
@karianna will we see docker image with only the jre also?

karianna:
> @karianna will we see docker image with only the jre also?

Yes, @dinogun Is working on it.

victornoel:
excellent, thanks @dinogun :)

PR #856, opened 2019-01-22
Add nasm to Windows path for OpenJ9 builds

sxa555:

### DIFF
--- a/build-farm/platform-specific-configurations/windows.sh
+++ b/build-farm/platform-specific-configurations/windows.sh
@@ -50,7 +50,7 @@ then
 
   if [ "${VARIANT}" == "openj9" ]
   then
-    export PATH="/cygdrive/c/Program Files (x86)/LLVM/bin:$PATH"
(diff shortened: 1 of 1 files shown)


build-farm/platform-specific-configurations/windows.sh
```diff
@@ -64,12 +64,11 @@ then
       # Next line a potentially tactical fix for https://github.com/AdoptOpenJDK/openjdk-build/issues/267
       export PATH="/usr/bin:$PATH"
     fi
-    # This needs to be before cygwin as at least one machine has 64-bit clang in cygwin #813
-    export PATH="/cygdrive/c/Program Files (x86)/LLVM/bin:$PATH"
+    # LLVM needs to be before cygwin as at least one machine has 64-bit clang in cygwin #813
+    export PATH="/cygdrive/c/Program Files (x86)/LLVM/bin:/cygdrive/c/openjdk/nasm-2.13.03:$PATH"
```
karianna: Needs a comment as to why we're adding nasm.  we should possibly extract the version out as a const higher up in the script as well.

sxa555: Nits addressed :-)

PR #857, opened 2019-01-23
Add build metadata file

johnoliver:


johnoliver:
added an example of some of the data

pipelines/build/openjdk_build_pipeline.groovy
```diff
@@ -144,11 +145,56 @@ def sign(config) {
...
+            variant: config.variant,
+            version: config.javaVersion,
+            tag    : config.parameters.TAG
+    ]
+
+    /*
+    example data:
+    {
+        "os": "linux",
+        "arc": "x64",
```
karianna: `arc` -> `arch`

pipelines/build/openjdk_build_pipeline.groovy
```diff
@@ -144,11 +145,56 @@ def sign(config) {
...
+        "variant": "hotspot",
+        "version": "jdk8u",
+        "tag": "jdk8u202-b08",
+        "adopt_build_number": 2,
```
karianna: Should we make this part of the version number?

pipelines/build/openjdk_build_pipeline.groovy
```diff
@@ -144,11 +145,56 @@ def sign(config) {
...
+            os     : config.os,
+            arc    : config.arch,
+            variant: config.variant,
+            version: config.javaVersion,
+            tag    : config.parameters.TAG
+    ]
+
+    /*
+    example data:
+    {
```
karianna: I wonder if this needs to be in a top-level block.  What if for example, we added a `build_configure` section with all of the `bash configure` output or the versions of all of the O/S native libs we build with etc....

pipelines/build/openjdk_build_pipeline.groovy
johnoliver: well, this is a question of how we would bring this into the version that would be compatible with the upstream version number

karianna: Did you see https://bugs.openjdk.java.net/browse/JDK-8207849 - that might help - although not as of today I guess.

karianna:
No bot - it needs discussion.

PR #860, opened 2019-01-24
Add sanity check when releasing

johnoliver:


pipelines/build/build_base_file.groovy
```diff
@@ -233,6 +233,24 @@ def createJob(jobName, jobFolder, config, enableTests, scmVars) {
     return create
 }
 
+def checkSaneConfig(releaseTag, jobConfigurations) {
+
+    if (releaseTag != null && releaseTag.length() > 0) {
```
AdamBrousseau: Does ` if (release tag) {` accomplish the same thing?

johnoliver: just had a test and it appears so yes

PR #870, opened 2019-01-29
Monster PR that combines previous ones

johnoliver:
1. Relocates common pipeline scripts into a different dir
2. Adds metadata file
3. Parses and generates semver 

The adopt semver is of the form `8.0.202+8.2`. The final two metadata numbers are the upstream build number followed by a `.` then our own adopt openjdk build number. We need to agree this format before accepting this PR


README.md
```diff
@@ -305,6 +305,30 @@ to your local platform and then build (Adopt) OpenJDK and place the result in
 These scripts default to using AdoptOpenJDK as the OpenJDK source repo to build 
 from, but you can override this with the `-r` flag.
 
+#Metadata
+Along side the built assets a metadata file will be created with info about the build. This will look of the form:
```
karianna: * "Along side" is one word I think
* Suggestion: _"This will be a JSON document of the form"_

pipelines/build/common/build_base_file.groovy
```diff
@@ -40,7 +42,7 @@ def toBuildParams(enableTests, cleanWorkspace, params) {
     return buildParams
 }
 
-def buildConfiguration(javaToBuild, variant, configuration, releaseTag, branch, additionalConfigureArgs, additionalBuildArgs) {
+def buildConfiguration(javaToBuild, variant, configuration, releaseTag, branch, additionalConfigureArgs, additionalBuildArgs, adoptBuildNumber) {
```
karianna: 8 args - do we need to think about refactoring this?

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -31,16 +33,110 @@ import groovy.json.JsonSlurper
  *
  */
 
-def getJavaVersionNumber(version) {
-    // version should be something like "jdk8u"
-    def matcher = (version =~ /(\d+)/)
-    return Integer.parseInt(matcher[0][1])
+def addOr0(map, name, matched, groupName) {
```
karianna: Purpose of this function?  As a standalone, it's not clear to me.  Perhaps a comment.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -144,11 +240,70 @@ def sign(config) {
...
+    return sh(
+            script: """find workspace/target/ | egrep '.tar.gz|.zip'""",
+            returnStdout: true,
+            returnStatus: false
+    )
+            .trim()
+            .split('\n')
+}
+
+def writeMetadata(config, filesCreated) {
```
karianna: I'm guessing we could split out all of our metadata functionality into its own script?

johnoliver: having separate scripts in jenkins is a pain, if you can manage it let me know

PR #872, opened 2019-01-29
Rewrite of mac installer guess

johnoliver:


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -245,6 +246,44 @@ def sign(config) {
     }
 }
 
+
+def buildInstaller(config) {
```
karianna: @johnoliver - Ideally, this would be a separate pipeline stage no?

PR #912, opened 2019-02-21
add 12 build

johnoliver:
This will not work until https://github.com/AdoptOpenJDK/openjdk-jdk12u has something in it


sophia-guo:
Depends on https://github.com/AdoptOpenJDK/openjdk-build/issues/883 ?

PR #940, opened 2019-03-04
Update node labels, fix typo in script and be more explicit on versions

ali-ince:
JDK12u is supposed to be built with VS2017 toolchain. The PR changes a couple of things across build scripts;
1. It adds `vs2017` label to jdk12u windows build definitions.
2. Fixes a typo that resulted in picking `VS2013` as the toolchain in `windows.sh` file.
3. Re-organize `windows.sh` file to be more explicit on VS toolchains for jdk12u and jdk11u.


gdams:
run tests

PR #959, opened 2019-03-12
Remove /Home from JDK_BOOT_DIR_dir setting

lumpfish:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/958


lumpfish:
Changed to strip top three directories.  @sxa555 , please can you recheck?

PR #960, opened 2019-03-12
add --no-adopt-patches

johnoliver:


johnoliver:
The test failure should be ignored, as the PR builder has not been merged yet, that test is not valid

sbin/build.sh
```diff
@@ -166,7 +166,7 @@ configuringVersionStringParameter()
       addConfigureArg "--with-user-release-suffix=" "${dateSuffix}"
     fi
 
-    if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_HOTSPOT}" ]; then
+    if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_HOTSPOT}" ] && [ ${BUILD_CONFIG[ADOPT_PATCHES]} == true ]; then
       addConfigureArg "--with-company-name=" "AdoptOpenJDK"
```
karianna: This is going to be true for any `dev` branch build effectively yes?

karianna: Will need to think about this with regards to say if we build Coretto.

johnoliver: for any NON-dev branch based builds. Yes for Coretto 8 we will probably need to pass the no patch arg

karianna: OK we can resolve this once we sort out variant.

PR #962, opened 2019-03-12
mac: Only strip 3 dirs on boot JDK11, not JDK10

sxa555:


sxa555:
Forcing this through as it's a build break and towards the end of the day ...

PR #968, opened 2019-03-17
WIP: add version parser

gdams:
Changes the directory of the extracted tarball to be `jdk-8.0.202+08.1` or `jre-8.0.202+08.1`

@smlambert please confirm whether any changes will be required to the test scripts with this new directory name?

closes: https://github.com/AdoptOpenJDK/openjdk-build/issues/617

### DIFF
--- a/version-parser/version-parser.py
+++ b/version-parser/version-parser.py
@@ -1,5 +1,19 @@
 #!/usr/bin/python
 
+################################################################################
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


karianna:
I thought we already had this logic in the API and/or Website repo...

gdams:
@johnoliver wrote this is groovy to write the metadata but this wasn't useful because we needed to know the metadata at build stage in order to name the directory accordingly.

version-parser/version-parser.py
```diff
@@ -0,0 +1,86 @@
...
+try:
+    isinstance(int(split_version[0]), (int, long))
+except ValueError:
+    split_version = version.split('+')
+
+if int(split_version[0]) > 1:
+    # detected openjdk 9 or above
+    major = int(split_version[0]) # 11
+    try:
+        minor = int(split_version[1]) # 0
```
johnoliver: does this work for `11+28`?

gdams: tested here: https://github.com/AdoptOpenJDK/openjdk-build/blob/21cd7cf4162f444fa9ceb95c87f3f71870dd8bca/version-parser/test-version-parser.sh#L68-L70

karianna:
As discussed on the call - this may not be the optimal approach, will see what you and John come up with on take 2 :-)

PR #971, opened 2019-03-19
Extract Jenkins into shared library

johnoliver:
Take some of the classes used in jenkins and extract them to a shared lib so that they can be reused.

Rename dir inside archive to be semver based on the parsed version.

Apply a bunch of the fixes also in #968

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1,5 +1,6 @@
-import NodeHelper
-@Library('openjdk-jenkins-helper@master')
+@Library('local-lib@master')
+import common.VersionInfo
+import common.MetaData
 import groovy.json.JsonOutput
 
 import java.util.regex.Matcher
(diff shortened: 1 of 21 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -1,5 +1,5 @@
-import NodeHelper
-@Library('openjdk-jenkins-helper@master')
+@Library('local-lib@master')
+import common.*
```
karianna: prefer explicit imports

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -375,7 +242,7 @@ class Build {
                         context.string(name: 'PRODUCT_MAJOR_VERSION', value: "${versionData.major}"),
                         context.string(name: 'PRODUCT_MINOR_VERSION', value: "${versionData.minor}"),
                         context.string(name: 'PRODUCT_MAINTENANCE_VERSION', value: "${versionData.security}"),
-                        context.string(name: 'PRODUCT_PATCH_VERSION', value: "${versionData.build}"),
+                        context.string(name: 'PRODUCT_PATCH_VERSION', value: "${String.format("%02d", versionData.build)}"),
```
karianna: Could we ever have 3 decimals?

pipelines/library/src/ParseVersion.groovy
```diff
@@ -0,0 +1,52 @@
...
+
+class ParseVersion {
+
+    static void main(String[] args) {
+        OptionAccessor parsedArgs = parseArgs(args)
+        VersionInfo version
+        if (parsedArgs.s) {
+            version = parseFromStdIn(args)
+        } else {
+            version = new VersionInfo().parse(args[args.length - 2], args[args.length - 1])
```
karianna: Could extract the 2 args into local vars that are named for readability.

pipelines/library/src/ParseVersion.groovy
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        printVersion(parsedArgs, version)
+    }
+
+    private static void printVersion(OptionAccessor parsedArgs, VersionInfo version) {
+        if (parsedArgs.f) {
```
karianna: `f`?

pipelines/library/src/ParseVersion.groovy
```diff
@@ -0,0 +1,52 @@
...
+        def reader = System.in.newReader()
+        def line
+        while ((line = reader.readLine()) != null) {
+            Matcher matcher = (line =~ /.*\(build (?<version>.*)\).*/)
+            if (matcher.matches()) {
+                VersionInfo version = new VersionInfo().parse(matcher.group("version"), args[args.length - 1])
+                return version
+            }
+        }
+        return null
```
karianna: What happens to the caller when they get a null?

pipelines/library/src/ParseVersion.groovy
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private static OptionAccessor parseArgs(String[] args) {
+        CliBuilder cliBuilder = new CliBuilder()
+
+        cliBuilder.s('read input from stdin', args: 0)
```
karianna: `s` and `f`?

pipelines/library/src/common/VersionInfo.groovy
```diff
@@ -0,0 +1,135 @@
...
+    Integer minor // 0
+    Integer security // 181
+    Integer build // 8
+    String opt
+    String version
+    String pre
+    Integer adopt_build_number = 1
+    String semver
+
+    VersionInfo() {
```
karianna: Public constructor needed?

sbin/build.sh
```diff
@@ -368,6 +371,17 @@ executeTemplatedFile() {
...
 }
 
+buildSharedLibs() {
+    cd "${LIB_DIR}"
+    ./gradlew clean uberjar
+}
+
+parseJavaVersionString() {
+  ADOPT_BUILD_NUMBER="${ADOPT_BUILD_NUMBER:-1}"
+  local version=$("$PRODUCT_HOME"/bin/java -version 2>&1 | java -cp "${LIB_DIR}/target/libs/pipelines-all-1.0.0-SNAPSHOT.jar" ParseVersion -s -f semver $ADOPT_BUILD_NUMBER)
```
karianna: 1.0.0-SNAPSHOT seems brittle

sbin/build.sh
```diff
@@ -408,47 +422,49 @@ printJavaVersionString()
...
+  mv "${jdkPath}" "${jdkTargetPath}"
 
   if [ -d "$(ls -d ${BUILD_CONFIG[JRE_PATH]})" ]
   then
-    JRE_TARGET_PATH="${openJdkVersion}-jre"
-    [ "${JRE_TARGET_PATH}" == "${openJdkVersion}" ] && JRE_TARGET_PATH="${openJdkVersion}.jre"
-    echo "moving $(ls -d ${BUILD_CONFIG[JRE_PATH]}) to ${JRE_TARGET_PATH}"
-    rm -rf "${JRE_TARGET_PATH}" || true
-    mv "$(ls -d ${BUILD_CONFIG[JRE_PATH]})" "${JRE_TARGET_PATH}"
+    [ "${jreTargetPath}" == "${openJdkVersion}" ] && jreTargetPath="${openJdkVersion}.jre"
```
karianna: Should that be `jreTargetPath=="${openJdkVersion}.jre"` I'mconfused why an assignment occurs after an `&&`

pipelines/build/common/openjdk_build_pipeline.groovy
johnoliver: So this is a question re what the installer should expect to see here. For padding build numbers seem to get padded by 1 0, i.e `08`. 

However from what I understand of the installers they are using this info to build up what the directory inside the archive is. But under this new scheme the dir inside the archive is now `jdk-<semver>` so not padded at all. I doubt padded or not will fix this. This is a question for @gdams as to what should be provided here

pipelines/library/src/ParseVersion.groovy
johnoliver: the sort of interesting way groovys CliBuilder works. If the user gave a `-f` argument then `parsedArgs.f` will exist

pipelines/library/src/ParseVersion.groovy
johnoliver: same as before, this defines you can give arguments '-s' and '-f'

pipelines/library/src/common/VersionInfo.groovy
johnoliver: groovy defaults to public

karianna:
As an FYI to any watchers - we'll hold off landing this until after we release 12

pipelines/build/common/openjdk_build_pipeline.groovy
gdams: leave it without for now. If your fix is changing the directory name to be the semver version we should be good

karianna:
@johnoliver Rebase.

PR #974, opened 2019-03-20
Correctly find latest tag from repo when there are multiple for same SHA

groeges:
Signed-off-by: Steve Groeger <GROEGES@uk.ibm.com>

The current build.sh doesnt get the correct tag when there are multiple tags with the same date on the same SHA. This change is to correctly get the latest tag from the SHA.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -518,13 +518,17 @@ makeACopyOfLibFreeFontForMacOSX() {
 }
 
 
-# Get the first tag from the git repo
+# Get the tags from the git repo and choose the latest tag when there is more than one for the same SHA.
 # Excluding "openj9" tag names as they have other ones for milestones etc. that get in the way
 getFirstTagFromOpenJDKGitRepo()
 {
     git fetch --tags "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
     revList=$(git rev-list --tags --topo-order --max-count=$GIT_TAGS_TO_SEARCH)
     firstMatchingNameFromRepo=$(git describe --tags $revList | grep jdk | grep -v openj9 | head -1)
+    # this may not find the correct tag if there are multiples on the commit so find commit
+    # that contains this tag and then use `git tag` to find the real tag
+    revList=$(git rev-list -n 1 $firstMatchingNameFromRepo) 
+    firstMatchingNameFromRepo=$(git tag --points-at $revList | tail -1)
     echo "$firstMatchingNameFromRepo"
 }
 


sbin/build.sh
```diff
@@ -518,13 +518,13 @@ makeACopyOfLibFreeFontForMacOSX() {
...
-# Get the first tag from the git repo
+# Get the tags from the git repo and choose the latest tag when there is more than one for the same SHA.
 # Excluding "openj9" tag names as they have other ones for milestones etc. that get in the way
 getFirstTagFromOpenJDKGitRepo()
 {
     git fetch --tags "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
-    revList=$(git rev-list --tags --topo-order --max-count=$GIT_TAGS_TO_SEARCH)
-    firstMatchingNameFromRepo=$(git describe --tags $revList | grep jdk | grep -v openj9 | head -1)
+    revList=$(git rev-list --exclude="*_openj9*" --tags="jdk-*" --topo-order --max-count=$GIT_TAGS_TO_SEARCH)
+    firstMatchingNameFromRepo=$(git tag --points-at $revList | grep '+' | sort -V | tail -1)
```
sxa555: `sort -V` isn't a standard unix option so won't work on AIX

sxa555: What is the `grep '+'` for in here? I don't believe that'll match tags in the JDK8 repo

groeges: OK, have removed the `sort -V` option.

groeges: Seems this was from some jdk12 code. Have removed this so it should work with all jdk versions.

sbin/build.sh
```diff
@@ -518,13 +518,13 @@ makeACopyOfLibFreeFontForMacOSX() {
...
 
-# Get the first tag from the git repo
+# Get the tags from the git repo and choose the latest tag when there is more than one for the same SHA.
 # Excluding "openj9" tag names as they have other ones for milestones etc. that get in the way
 getFirstTagFromOpenJDKGitRepo()
 {
     git fetch --tags "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
     revList=$(git rev-list --tags --topo-order --max-count=$GIT_TAGS_TO_SEARCH)
-    firstMatchingNameFromRepo=$(git describe --tags $revList | grep jdk | grep -v openj9 | head -1)
+    firstMatchingNameFromRepo=$(git tag --points-at $revList | grep jdk | grep -v openj9 | tail -1)
```
keithc-ca: This probably wants to use `sort -V` like [1] to be sure to choose the highest version tag.

[1] https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/closed/autoconf/custom-hook.m4#L320

sxa555: As per my comment on an earlier version of this, `sort -V` is a GNU specific option so I don't really want to be using it (won't work with the default AIX sort)

sxa555:
This is currently failing due to the fact that `revList` has multiple entries (the last 100 due to PR https://github.com/AdoptOpenJDK/openjdk-build/pull/783, issue https://github.com/AdoptOpenJDK/openjdk-build/issues/748) and `git tag --points-at` won't work with multiples.

If we're going to change this we need to understand whether there is still a risk of errantly named tags.

sbin/build.sh
keithc-ca: I hadn't noticed your earlier comment: sorry for that.
On the other hand, it _is_ working on all platforms where openj9 is built, including AIX.

groeges:
Following discussions with @sxa555 we need to have the existing code left as-is in order to find the correct tags, based on updates due to earlier issues (see above) but once we have found the tag, we need to check that is is the latest tag on that commit. This is done by finding the commit the tag is on and then checking that commit to see if there are multiple tags and to return the latest tag.

sbin/build.sh
sxa555: I suspect your openj9 systems have the GNU coreutils installed and in the PATH before the system ones.

sxa555:
Based on the success of testing this at https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/test-build-script-pull-request/187/ I'm happy with this solution :-)

pshipton:
FYI @keithc-ca has re-written the logic to find the proper git tag to fix https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/142
We also found there are problems with using `sort -V` on osx, and stopped using it for jdk8. We also plan to stop using it for jdk11 as per https://github.com/eclipse/openj9/issues/4932

pshipton:
The fix to remove `sort -V` is here https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/144

PR #978, opened 2019-03-21
Check archive dir

johnoliver:


sxa555:
Traps the issue in https://github.com/AdoptOpenJDK/openjdk-build/issues/977

PR #979, opened 2019-03-21
Stop tag detection failing when there's a subdir with name of tag

sxa555:

Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/977

Also will throw a message to stderr if the tag detection fails - I could do an `exit 1` in that case but wanted to keep this critical fix simple for now in case that causes other side effects with some builds

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -522,14 +522,22 @@ makeACopyOfLibFreeFontForMacOSX() {
 # Excluding "openj9" tag names as they have other ones for milestones etc. that get in the way
 getFirstTagFromOpenJDKGitRepo()
 {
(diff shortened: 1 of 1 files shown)


groeges:
@sxa555 you could have just added `--` to the end of the command, ie:
```
revList=$(git rev-list -n 1 $firstMatchingNameFromRepo --)
```

sxa555:
> @sxa555 you could have just added `--` to the end of the command, ie:
> 
> ```
> revList=$(git rev-list -n 1 $firstMatchingNameFromRepo --)
> ```
> 
> this will ensure it is not treated as a path

Yeah that should be ok too I think -  Have pushed another commit for that (will probably squash before merging)

PR #981, opened 2019-03-21
Dont use readlink on mac

johnoliver:


sxa555:
NOTE for historic reference: This was the result of an error on the mac machines: `usage: dirname path` introduced by https://github.com/AdoptOpenJDK/TSC/issues/68 earlier today

PR #985, opened 2019-03-22
enforce compilation of freetype with 32 bit on i686 systems

pinhead84:
In case of cross compilation on x86-64 systems for i686 targets, Freetype is not compiled for 32 bit.
See also: https://github.com/AdoptOpenJDK/TSC/issues/71#issuecomment-475451019

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -224,14 +224,14 @@ checkingAndDownloadingFreeType()
       pngArg="--with-png=no";
     fi
 
-    local freetypeCC="";
+    local freetypeEnv="";
     if [[ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "i686" ]] || [[ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "i386" ]] ; then
-        freetypeCC="gcc -m32";
+      freetypeEnv="export CC=\"gcc -m32\"";
     fi
 
     # We get the files we need at $WORKING_DIR/installedfreetype
     # shellcheck disable=SC2046
-    if ! (CC="${freetypeCC}" bash ./configure --prefix="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}"/installedfreetype "${pngArg}" "${BUILD_CONFIG[FREETYPE_FONT_BUILD_TYPE_PARAM]}" && ${BUILD_CONFIG[MAKE_COMMAND_NAME]} all && ${BUILD_CONFIG[MAKE_COMMAND_NAME]} install); then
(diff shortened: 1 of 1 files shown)


sbin/prepareWorkspace.sh
```diff
@@ -224,9 +224,14 @@ checkingAndDownloadingFreeType()
...
 
+    local freetypeCC="";
+    if [[ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "i686" ]] || [[ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "i386" ]] ; then
+        freetypeCC="gcc -m32";
+    fi
+
     # We get the files we need at $WORKING_DIR/installedfreetype
     # shellcheck disable=SC2046
-    if ! (bash ./configure --prefix="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}"/installedfreetype "${pngArg}" "${BUILD_CONFIG[FREETYPE_FONT_BUILD_TYPE_PARAM]}" && ${BUILD_CONFIG[MAKE_COMMAND_NAME]} all && ${BUILD_CONFIG[MAKE_COMMAND_NAME]} install); then
+    if ! (CC="${freetypeCC}" bash ./configure --prefix="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}"/installedfreetype "${pngArg}" "${BUILD_CONFIG[FREETYPE_FONT_BUILD_TYPE_PARAM]}" && ${BUILD_CONFIG[MAKE_COMMAND_NAME]} all && ${BUILD_CONFIG[MAKE_COMMAND_NAME]} install); then
```
karianna: If it's x64 then `CC=""` - That potentially overrides the default no?

pinhead84: I've checked this on Debian amd64. `CC=""` behaves as expected and produces a 64 bit build of Freetype. It looks like `CC=""` is the same like not using this environment variable at all.

karianna: Still makes me a little nervous :-).  Can you rework the logic so that the CC is only assigned in the case where it is x32?

pinhead84: Okay. :)

pinhead84: I've updated the pull request. Now the `CC` environment variable is only passed on i686 systems.

After Freetype is compiled the `CC` environment variable is **not** defined in the following `else` block and therefore won't interfere with following commands.

I've tested this on Debian amd64 - with a native 64bit build and with a 32bit cross compilation. In both cases Freetype is compiled for the correct architecture. Also a native build on Debian i386 works as expected.

I did not find a better solution than using `eval` without modifying the code too much. This command is also supported on macOS, which uses a pretty old version of Bash.

karianna: great thank you!

PR #989, opened 2019-03-26
Extract configuration passing into json

johnoliver:
Now that we are using a shared lib we can use a shared class to pass data between upstream builds and individual builds. This reduces the amount of code that needs to be kept in sync between the upstream and individual builds.

This will make individual build args go from looking like:

https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-x64-hotspot/154/parameters/

to

https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/build-test/job/jobs/job/jdk8u/job/jdk8u-linux-x64-hotspot/149/parameters/

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -1,7 +1,7 @@
-@Library('local-lib@master')
-import common.VersionInfo
 import common.IndividualBuildConfig
 import common.MetaData
+@Library('local-lib@master')
+import common.VersionInfo
 import groovy.json.JsonOutput
 
 import java.util.regex.Matcher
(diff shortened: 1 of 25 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -291,7 +289,7 @@ class Build {
 
     List<String> listArchives() {
         return context.sh(
-                script: """find workspace/target/ | egrep '.tar.gz|.zip'""",
+                script: """find workspace/target/ | egrep '.tar.gz|.zip|.msi|.pkg'""",
```
karianna: worth pre-adding .deb and.rpm to this?

pipelines/src/test/groovy/TestCompilation.groovy
```diff
@@ -59,6 +59,7 @@ class TestCompilation {
         } catch (Exception e) {
             println("This test checks compilation against Stub implementations that Mimic the jenkins Environment")
             println("Check that any methods you wish to use from the environment are represented in the testDoubles package")
+            e.printStackTrace()
```
karianna: Do we want to print stacktrace?

PR #1000, opened 2019-03-29
move to java 11 for groovy

johnoliver:


sbin/build.sh
```diff
@@ -371,14 +371,36 @@ executeTemplatedFile() {
...
+  echo $gradleJavaHome
+}
+
 buildSharedLibs() {
     cd "${LIB_DIR}"
-    GRADLE_USER_HOME=./gradle-cache ./gradlew clean uberjar
+    local gradleJavaHome=$(getGradleHome)
+    echo "Running gradle with $gradleJavaHome"
+
+    JAVA_HOME="$gradleJavaHome" GRADLE_USER_HOME=./gradle-cache ./gradlew clean uberjar
```
karianna: Split into separate lines please

sbin/build.sh
```diff
@@ -371,14 +371,36 @@ executeTemplatedFile() {
...
+
+    # Test parser
+    "$gradleJavaHome"/bin/java -version 2>&1 | "$gradleJavaHome"/bin/java -cp "${LIB_DIR}/target/libs/adopt-shared-lib.jar" ParseVersion -s -f semver 1
```
karianna: Could be clearer as to what's this is testing in the comment, e.g. What's the expected output

sbin/common/common.sh
```diff
@@ -33,6 +33,24 @@ function setOpenJdkVersion() {
...
+function crossPlatformRealPath() {
+  local target=$1
+
+  local currentDir="$PWD"
+
+  if [[ -d $target ]]; then
+    cd "$target"
+    local name=""
+  elif [[ -f $target ]]; then
+    cd "$(dirname "$target")"
```
karianna: Is 'double' double quoting like this legal?

sbin/build.sh
johnoliver: you cant really

PR #1002, opened 2019-03-29
Remove unsupported versions (jdk9/jdk10)

grzesuav:


karianna:
I'd actually like to discuss this again at the TSC, if we agree then we should consistently remove this and update the documentation accordingly.

karianna:
TSC deferred this back to a build team decision. Build team is actually OK keeping 9 and 10 as we may need to produce bootstrap builds using these in the future.

@grzesuav Apologies for steering you wrong here!

PR #1008, opened 2019-04-02
Stop generating hash json, Use openjdk version for dir name, Fix using readlink on solaris

johnoliver:
- Stop generating hash json. 
- Use openjdk version for dir name. 
- Fix using readlink on solaris

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -289,7 +289,7 @@ class Build {
 
     List<String> listArchives() {
         return context.sh(
-                script: """find workspace/target/ | egrep '.tar.gz|.zip|.msi|.pkg|.deb|.rpm'""",
+                script: '''find workspace/target/ | egrep '(.tar.gz|.zip|.msi|.pkg|.deb|.rpm)$' ''',
                 returnStdout: true,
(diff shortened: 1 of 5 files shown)


dinogun:
@gdams Thanks for the heads up ! The docker scripts create a dir with a standard name (with no versioning) and move all of the content to that, so we should be ok. Will confirm though.

sxa555:
Can we please put more descriptive titles on PRs - "A few small changes" is not appropriate for something that can potentially break users.

johnoliver:
this no longer renames the dir, it will go back to:
```
jdk-11.0.3+9
jdk-11.0.3+9-jre
```

karianna:
> THIS IS A BREAKING CHANGE:
> directory layout used to be:
> 
> ```
> jdk-11.0.3+9
> jdk-11.0.3+9-jre
> ```
> it is now:
> 
> ```
> jdk-11.0.3+9
> jre-11.0.3+9
> ```
> before this can land:
> 
> * [ ]  check with @AdoptOpenJDK/testing that the new directory layout won't break anything
> * [ ]  check with @dinogun if the new directory layout will break anything in the docker scripts.
> * [ ]  update the mac installer scripts
> * [ ]  update the windows installer scripts

No longer a breaking change so landing this.

PR #1015, opened 2019-04-04
Use downloaded freetype instead of relying on a pre-built version

ali-ince:
Freetype dependency on windows requires building for 32-bit / 64-bit targets with the corresponding MSVC version that is used to build the jdk binaries. 

Given that;
1. we're already downloading freetype source code on the fly during build phase,
2. both hotspot and openj9 configure processes support `--with-freetype-src` argument to build the freetype binaries from given source tree with the detected MSVC compiler toolset, 
it seems no longer a requirement for us to handle freetype as a static dependency to configure on each build agent.

This PR removes several freetype configuration flags from windows configuration script, sets freetype version to be downloaded and adds the corresponding `--with-freetype-src` flags to the configure arguments.


karianna:
@ali-ince Have you been able to manually test this change?

ali-ince:
Yes @karianna, it definitely works with hotspot and I couldn't complete an openj9 build but got past the point where configure was able to build the freetype binary.

PR #1023, opened 2019-04-08
OpenJ9: Build against system openssl on Linux and don't reship it

sxa555:
This should supercede https://github.com/AdoptOpenJDK/openjdk-build/pull/1018 but I've had reports that it may cause some build failures so I want to verify it more before integrating.

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -39,7 +39,7 @@ then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


sxa555:
Looks to produce something usable on the platforms I've tested on so I'm now happy for this to be approved and go in.

sxa555:
Particularly to stop us reshipping openssl now I'd like to get this in ASAP if possible.

sxa555:
@lumpfish FYI

Issue #1030, opened 2019-04-11
OpenJ9 builds not showing "based on .." jdk tag level


andrew-m-leonard:
**Platform:**
All

**Architecture:**
All

Latest OpenJ9 builds are missing the based on JDK tag level in java -version.
```
openjdk version "1.8.0_212"
OpenJDK Runtime Environment (build 1.8.0_212-201904101301-b01)
Eclipse OpenJ9 VM (build v0.14.0-release-581f9de22, JRE 1.8.0 Linux ppc64le-64-Bit Compressed References 20190410_259 (JIT enabled, AOT enabled)
OpenJ9   - 581f9de22
OMR      - 4a4278e6
JCL      - fe984ddb37 based on )
```

andrew-m-leonard:
Issue caused by this change: https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/7f76ddcbf1e0db1873c11ea6f2ffc326725a0a8c

karianna:
@andrew-m-leonard Discussed at Adopt - a full clone represents a performance regression for our builds.  Is it possible to fix this upstream?

pshipton:
@keithc-ca is looking at this.

keithc-ca:
See:
* ibmruntimes/openj9-openjdk-jdk8#285
* ibmruntimes/openj9-openjdk-jdk8#286
* ibmruntimes/openj9-openjdk-jdk11#157
* ibmruntimes/openj9-openjdk-jdk11#158
* ibmruntimes/openj9-openjdk-jdk12#40
* ibmruntimes/openj9-openjdk-jdk#94

pshipton:
The fixes are merged, the next set of builds should be working.

karianna:
Thank you all

keithc-ca:
The most recent nightly of JDK8 at adopt predates the change, but JDK11 on Windows looks good:
$ java -version
```
openjdk version "11.0.3" 2019-04-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.3+2-201904131223)
Eclipse OpenJ9 VM AdoptOpenJDK (build master-9da13cbf0, JRE 11 Windows 10 amd64-64-Bit Compressed References 20190413_197 (JIT enabled, AOT enabled)
OpenJ9   - 9da13cbf0
OMR      - f4de44cc
JCL      - 5c13382d6b based on jdk-11.0.3+2)
```

PR #1033, opened 2019-04-12
Fix crash url

johnoliver:


johnoliver:
fixes #1030

sbin/build.sh
```diff
@@ -224,8 +224,12 @@ configuringVersionStringParameter()
     addConfigureArg "--with-vendor-url=" "https://adoptopenjdk.net/"
     addConfigureArg "--with-vendor-name=" "AdoptOpenJDK"
     addConfigureArg "--with-vendor-bug-url=" "https://github.com/AdoptOpenJDK/openjdk-build/issues"
-    addConfigureArg "--with-vendor-vm-bug-url=" "https://github.com/eclipse/openj9/issues"
 
+    if [[ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_OPENJ9}" ]]; then
+      addConfigureArg "--with-vendor-vm-bug-url=" "https://github.com/eclipse/openj9/issues"
+    else
+      addConfigureArg "--with-vendor-vm-bug-url=" "https://github.com/AdoptOpenJDK/openjdk-build/issues"
```
gdams: is this not already the default? https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/dev/jdk/src/share/native/java/lang/System.c#L114

gdams: oh not on jdk9+

PR #1040, opened 2019-04-15
OpenJ9/aarch64: --disable-warnings-as-errors

sxa555:
Required for https://github.com/AdoptOpenJDK/openjdk-build/issues/966

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -51,6 +51,10 @@ then
   else
     export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-openssl=system"
   fi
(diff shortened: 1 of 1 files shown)


karianna:
Won't merge until after April 16th release though

tellison:
From the peanut gallery: seems like a sledgehammer approach to solve the one reported instance of a warning, thereby leaving us open to missing new warnings in the future that may be valuable.

As @knn-k [wrote in the cause](https://github.com/eclipse/openj9/issues/4486#issuecomment-483511330), it would be preferable for OpenJ9 to add a compiler pragma.

knn-k:
I just opened https://github.com/eclipse/openj9/pull/5508 as an alternative solution for AArch64 build.

sxa555:
@tellison Totally agree which is why I really didn't want to do this and delayed it a while. I only ever intended this to be a temporary fix until we got something better. As it happens https://github.com/eclipse/openj9/pull/5508 is now ready to go in, therefore I am going to cancel this PR.

PR #1047, opened 2019-04-18
Fix filename for java 8 releases

johnoliver:


johnoliver:
Actually should hold off on this, just noticed that 11 works correctly by default, in fact I believe it is only required for java 8

pipelines/build/common/build_base_file.groovy
```diff
@@ -2,7 +2,6 @@
 import common.IndividualBuildConfig
 import groovy.json.JsonSlurper
 
-import java.text.SimpleDateFormat
```
karianna: Deliberate?

pipelines/jobs/pipeline_job_template.groovy
```diff
@@ -57,7 +57,7 @@ pipelineJob("${BUILD_FOLDER}/${JOB_NAME}") {
         booleanParam('enableTests', runTests, 'If set to true the test pipeline will be executed')
         stringParam('additionalConfigureArgs', "", "Additional arguments that will be ultimately passed to OpenJDK's <code>./configure</code>")
         stringParam('additionalBuildArgs', "", "Additional arguments to be passed to <code>makejdk-any-platform.sh</code>")
-        stringParam('overrideFileNameVersion', "", "REQUIRED for OpenJ9: e.g <code>8u202b08_openj9-0.12.1</code> or <code>11.0.2_9_openj9-0.12.1</code>")
+        stringParam('overrideFileNameVersion', "", "When forming the filename, ignore the build version and override it. For instance if you set this to 'FOO' the final file name will be of the form: <code>OpenJDK8U-jre_ppc64le_linux_openj9_FOO.tar.gz</code>")
```
karianna: Not sure this comment gives guidance to OpenJ9 folks as to what they should normally set it to for a release?

johnoliver: Assuming this pr works correctly, leaving it blank should be fine from now on

lumpfish: I don't think 'build version' has a context here.  Is 'ignore the part of the filename derived from the publishName or timestamp' more accurate?

karianna: @lumpfish Is that clear enough for you?

karianna: Heh - snap.

pipelines/build/common/build_base_file.groovy
johnoliver: yes

PR #1068, opened 2019-05-02
[WIP] Fix bad directories being formed in nightly builds

johnoliver:


karianna:
@johnoliver merge conflict to resolve

adamfarley:
Is this still a change we want to make?

I ask because @johnoliver and @karianna made significant changes to getFirstTagFromOpenJDKGitRepo() a couple of weeks ago, so perhaps they're not as keen to delete it.

Regardless, here's a branch with all the non-getFirstTagFromOpenJDKGitRepo()-related changes. If this PR is still needed, would this change be sufficient?

https://github.com/AdoptOpenJDK/openjdk-build/compare/master...adamfarley:fix-dir2

PR #1104, opened 2019-05-26
Add client variant to 32-bit hotspot builds 

### LINKED ISSUE #871: Windows x32 8u202-b08 Hotspot JDK/JRE build is missing file jre/bin/client/jvm.dll
**Platform:**

Windows

**Architecture:**

x32

The following build obtained from the link below is missing the file jre/bin/client/jvm.dll.

https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u202-b08/OpenJDK8U-jre_x86-32_windows_hotspot_8u202b08.zip

ali-ince:
This PR makes 32-bit hotspot builds (for windows) to include `client` and `server` vm's.

Fixes #871.

### DIFF
--- a/pipelines/build/openjdk11_pipeline.groovy
+++ b/pipelines/build/openjdk11_pipeline.groovy
@@ -61,6 +61,9 @@ def buildConfigurations = [
                         hotspot: 'win2012',
                         openj9:  'win2012&&mingw-standalone'
                 ],
+                buildArgs : [
+                        hotspot : '--jvm-variant client,server'
+                ],
                 test                : ['openjdktest']
         ],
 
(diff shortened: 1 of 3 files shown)


jt214:
I have a question..?   Why was this only just added for 32-bit...?   Was it for a specific app that is only 32-bit?   It just seems like 64-bit would be first, nowadays?   Are you still using 32-bit windows..?

karianna:
There was a specific user requirement for it, I've merged the 64 bit one as well, some folks will find it useful.

ali-ince:
Actually jdk8 won’t build a client vm for 64-bit builds.

karianna:
> Actually jdk8 won’t build a client vm for 64-bit builds.

Ah oK, I'll revert that then.

karianna:
https://github.com/AdoptOpenJDK/openjdk-build/pull/1114

PR #1113, opened 2019-06-07
Add jvm variant build arguments for 8, 11, 12 64-bit hotspot builds

jt214:
This was merged for 32 bit builds in PR #1104, can this be added to 64-bit builds too?


karianna:
See https://github.com/AdoptOpenJDK/openjdk-build/pull/1114 as well

jt214:
@karianna Thanks for looking into this.   Sorry about my confusion.   You're right, there is no client jvm.dll in the 64-bit jres.   Our product is still only using the 32-bit jre because of this.

PR #1124, opened 2019-06-18
Stop cloning OpenSSL when building OpenJ9 as it's no longer needed

sxa555:
We now use openssl which is preinstalled on the build machine, so we don't build it during the build process, therefore we have no need to clone it, therefore removing the option that does that


sxa555:
Closing due to the fact we still need it for macos

PR #1129, opened 2019-06-25
Add download checks

johnoliver:


sbin/prepareWorkspace.sh
```diff
@@ -164,6 +166,27 @@ checkingAndDownloadingAlsa()
...
 }
 
+checkFingerPrint() {
+  local sigFile="$1"
+  local fileName="$2"
+  local publicKey="$3"
+  local expectedFingerPrint="$4"
+
+  rm /tmp/public_key.gpg || true
+  gpg --output /tmp/public_key.gpg --dearmor "${SCRIPT_DIR}/sig_check/${publicKey}.asc"
```
karianna: Do we know gpg is installed on all hosts?

sbin/prepareWorkspace.sh
```diff
@@ -164,6 +166,27 @@ checkingAndDownloadingAlsa()
   fi
 }
 
+checkFingerPrint() {
```
karianna: `checkFingerprint()` to be pedantic

sbin/prepareWorkspace.sh
karianna: Does `${SCRIPT_DIR}/sig_check/` always exist?  Hmm, I think it does because you're adding .asc files into it below

sbin/prepareWorkspace.sh
```diff
@@ -164,6 +166,27 @@ checkingAndDownloadingAlsa()
...
+  local fileName="$2"
+  local publicKey="$3"
+  local expectedFingerPrint="$4"
+
+  rm /tmp/public_key.gpg || true
+  gpg --output /tmp/public_key.gpg --dearmor "${SCRIPT_DIR}/sig_check/${publicKey}.asc"
+
+  gpg -v --no-default-keyring --keyring "/tmp/public_key.gpg" --verify $sigFile $fileName
+
+  local verify=$(gpg -v --no-default-keyring --keyring "/tmp/public_key.gpg" --verify $sigFile $fileName 2>&1)
```
karianna: Effectively executing the same command as above?

sbin/prepareWorkspace.sh
```diff
@@ -164,6 +166,27 @@ checkingAndDownloadingAlsa()
...
+  local fingerprint=$(echo $verify | grep "Primary key fingerprint" | egrep -o "([0-9A-F]{4} ? ?){10}" | sed -e 's/[[:space:]]*$//')
```
karianna: Comment to show what egrep and sed are doing for us mere mortals ;-)

sbin/prepareWorkspace.sh
```diff
@@ -177,22 +200,40 @@ checkingAndDownloadingFreemarker()
...
   # Temporary fudge as curl on my windows boxes is exiting with RC=127
   if [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]] ; then
     wget -O "${targetFileName}" "${url}"
   else
     curl -L -o "${targetFileName}" "${url}"
   fi
+
+  if [ $# -ge 3 ]; then
+
+    local checksum="$3"
```
karianna: checksum or actually expectedChecksum?

sbin/prepareWorkspace.sh
johnoliver: Yeah, first one is so we can see the output in the log

sbin/prepareWorkspace.sh
```diff
@@ -166,18 +169,36 @@ checkingAndDownloadingAlsa()
   fi
 }
 
-checkFingerPrint() {
+checkFingerprint() {
   local sigFile="$1"
   local fileName="$2"
   local publicKey="$3"
   local expectedFingerPrint="$4"
+  local expectedChecksum="$4"
```
karianna: expectedChecksum == excpectedFingerPrint?  and FingerPrint should be Fingerprint

PR #1153, opened 2019-07-10
Ensure git remotes are added to the jdk8 repo

johnoliver:


git-hg/jdk8u/merge.sh
```diff
@@ -41,6 +40,18 @@ function initRepo() {
   git fetch upstream --tags
 }
 
+function addRemotes() {
+
+  cd "$REPO"
+  if ! git config remote.upstream.url > /dev/null; then
```
AdamBrousseau: I'm not sure if it's faster but you don't actually need to check if you just add the remote via config commands. This also ensures the remote url is correct.
https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/jobs/infrastructure/Update-Reference-Repos.groovy#L121
```
git config remote.upstream.url <URL>
git config remote.upstream.fetch +refs/heads/*:refs/remotes/upstream/*
```

PR #1168, opened 2019-07-15
Add extended.system/special.functional to the set of launched test targets

smlambert:
Additional test targets to launch via build pipelines (did not update jdk9/jdk10 as they are no longer run).


smlambert:
If this is not merged prior to the imminent release (due to code freeze), we can/will manually launch the additional test targets.

smlambert:
Note: this could be improved by having common variables with the nightly & release set of tests.  They used to be more different per platform, but as we get enough machines for each platform, the set of high-level test targets per platform should converge to be identical.

PR #1172, opened 2019-07-16
Allow diffing tags

johnoliver:


git-hg/diff.sh
```diff
@@ -25,7 +25,7 @@
 #
 ################################################################################
 
-set -euo
+set -euxo
```
karianna: Shouldn't `o` have an option name after it?

git-hg/diff.sh
```diff
@@ -53,6 +54,17 @@ function cloneRepos() {
...
+
+  if [ -n "${tag}" ]; then
+    cd openjdk-hg
+    hg update $tag
+    cd ../openjdk-git
+    git fetch --all
+    git checkout $tag
+    cd ..
+  fi
+
```
karianna: Kill extra whitespace

git-hg/diff.sh
```diff
@@ -53,6 +54,17 @@ function cloneRepos() {
 
   git clone -b master "https://github.com/AdoptOpenJDK/openjdk-${git_repo_version}.git" openjdk-git || exit 1
   hg clone "https://hg.openjdk.java.net/${hg_root_forest}/${hg_repo_version}" openjdk-hg || exit 1
+
+  if [ -n "${tag}" ]; then
+    cd openjdk-hg
+    hg update $tag
+    cd ../openjdk-git
+    git fetch --all
+    git checkout $tag
```
karianna: We're happy that we've left this function with git pointing at the tag (do we need to reset later?)

git-hg/diff.sh
```diff
@@ -65,7 +77,10 @@ function updateMercurialClone() {
 }
 
 function runDiff() {
-  diffNum=$(diff -rq openjdk-git openjdk-hg -x '.git' -x '.hg' -x '.hgtags' -x '.hgignore' -x 'get_source.sh' -x 'README.md' | wc -l)
+
+  diff -rq openjdk-git openjdk-hg -x '.git' -x '.hg' -x '.hgtags' -x '.hgignore' -x 'get_source.sh' -x 'README.md' > changes.diff
+
+  diffNum=$(wc -l < changes.diff)
 
   if [ "$diffNum" -gt 0 ]; then
```
karianna: Do we want to log somewhere that the `changes.diff` file can be inspected?

Do we need to delete that file later?

git-hg/diff.sh
johnoliver: the workspace gets wiped out every time

git-hg/diff.sh
johnoliver: can archive it in the job config

git-hg/diff.sh
```diff
@@ -42,6 +42,8 @@ checkArgs $@
 git_repo_version=$1
 hg_root_forest=${2:-${1}}                  # for backwards compatibility
 hg_repo_version=${3:-${hg_root_forest}}    # for backwards compatibility
+tag=${4:-} # tag to check
+expectedDiffLimit=${5:-0} # expected diff limit
```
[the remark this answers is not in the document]
karianna: Ah, number of diffs that we allow before we say "there's a problem" - perhaps update comment with that

git-hg/diff.sh
```diff
@@ -53,53 +55,56 @@ function cloneRepos() {
...
+    cd openjdk-hg
+    hg update $tag
+    cd ../openjdk-git
+    git fetch --all
+    git checkout $tag
+    cd ..
+  fi
 }
 
 function updateMercurialClone() {
```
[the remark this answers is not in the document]
johnoliver: should do yes

PR #1175, opened 2019-07-16
 Compare dev branch by default

johnoliver:


git-hg/diff.sh
```diff
@@ -63,6 +63,12 @@ function cloneRepos() {
     git fetch --all
     git checkout $tag
     cd ..
+  else
+    # By default compare the dev branch
+    cd ./openjdk-git
```
karianna: Should that be `../`?

johnoliver: no, this is correct

PR #1186, opened 2019-07-22
fix arm32 jdk8 build

gdams:


build-farm/platform-specific-configurations/linux.sh
```diff
@@ -61,7 +61,10 @@ fi
 
 if [ "${ARCHITECTURE}" == "arm" ]
 then
-  export CONFIGURE_ARGS_FOR_ANY_PLATFORM="--with-jobs=4 --with-memory-size=2000 --disable-warnings-as-errors"
+  export CONFIGURE_ARGS_FOR_ANY_PLATFORM="--with-jobs=4 --with-memory-size=2000"
+  if [ "$JAVA_FEATURE_VERSION" -ge 11 ]; then
```
karianna: Should that be `gt`? We built 11 successfully.

gdams: nope it's `ge` essentially that flag is not supported in jdk8

PR #1199, opened 2019-07-31
enable hardened runtime support for macos

gdams:
This allows users to bundle AdoptOpenJDK binaries into notarized macOS applications. 

This is confirmed to be fixed on JDK11+

On JDK8 we still hit the same issue that the Corretto folks do which is this error:
```
The binary uses an SDK older than the 10.9 SDK
```

My suggestion would be that we investigate shipping a JDK8 build that is optimized for bundling (i.e built on a later toolchain) This [repo](https://github.com/stooke/jdk8u-xcode10) may be usefil as it contains some patches required to get JDK8 to compile on xcode10.

### DIFF
--- a/entitlements.plist
+++ b/entitlements.plist
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>com.apple.security.cs.allow-jit</key>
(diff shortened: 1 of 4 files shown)


tiberiughioca:
Have you seen the work of Simon Tooke regarding this issue for the openjdk project?
https://github.com/stooke/jdk8u-xcode10/
Maybe you can use something from it for the AdoptOpenJDK...

karianna:
> Have you seen the work of Simon Tooke regarding this issue for the openjdk project?
> https://github.com/stooke/jdk8u-xcode10/
> Maybe you can use something from it for the AdoptOpenJDK...

We're working directly with him :)

PR #1208, opened 2019-08-07
sign builds before running tests

gdams:
This change forces the pipelines to sign the binaries before running tests


smlambert:
Thanks for updating the pipelines!!

PR #1215, opened 2019-08-13
Use gcc7 on all Linux builds (HotSpot-JDK8 specifically!)

sxa:


karianna:
@sxa555 Did you want to take this out of the draft?

jerboaa:
Considering https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1133 we probably should something like this get in.

sxa555:
@jerboaa Slightly curious about this as the existing logic in there does appear to suggest that any version of JDK11+ should already be building with the later GCC. Only JDK8/HotSpot should still be on the earlier GCC.

@karianna Whether this wants to come out of draft depends on whether we want to change all HotSpot JDK releases to run on GCC7. If we decide "yes" then I'm good to take it out of draft

jerboaa:
@sxa555 Right. I was experiencing the GCC version issue on a non-adopt-build-script build. It didn't have the devtoolset-7 magic setting. It does now.

sxa:
We should expedite a decion on this particularly since https://openjdk.java.net/groups/build/doc/building.html says The minimum accepted version of gcc is 5.0 and we're currently still using 4.8 for some of our HotSpot builds

sxa:
Old conflicts resolveld (We've switched to GCC7.5 from 7.3 since I originally made this PR) and running through JDK8 at https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/919 although the automatic PR tests should pick up any issues

sxa:
Note to reviewers - only the conditional at the start has been removed - the other changes are purely indentation of the stuff that was inside the conditional.

karianna (approved):
LGTM

aahlenst (approved):

PR #1221, opened 2019-08-14
Update parameters for create_installer_linux

aahlenst:
This should fix the recent build failures with the linux installers. The Jenkins job `create_installer_linux` expects a parameter `RELEASE_TYPE`. The value should be `Release` if it's a release and `Nightly` otherwise.

It's my first change to the build pipeline, so please double check.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -231,15 +231,20 @@ class Build {
 
         def buildNumber = versionData.build
 
+        String releaseType = "Nightly"
+        if (buildConfig.RELEASE) {
+            releaseType = "Release"
+        }
+
         def installerJob = context.build job: "build-scripts/release/create_installer_linux",
                 propagate: true,
                 parameters: [
                         context.string(name: 'UPSTREAM_JOB_NUMBER', value: "${env.BUILD_NUMBER}"),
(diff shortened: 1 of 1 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -231,15 +231,20 @@ class Build {
...
         def installerJob = context.build job: "build-scripts/release/create_installer_linux",
                 propagate: true,
                 parameters: [
                         context.string(name: 'UPSTREAM_JOB_NUMBER', value: "${env.BUILD_NUMBER}"),
                         context.string(name: 'UPSTREAM_JOB_NAME', value: "${env.JOB_NAME}"),
                         context.string(name: 'FILTER', value: "${filter}"),
+                        context.string(name: 'RELEASE_TYPE', value: "${releaseType}"),
                         context.string(name: 'VERSION', value: "${versionData.version}"),
                         context.string(name: 'MAJOR_VERSION', value: "${versionData.major}"),
-                        context.string(name: 'JVM', value: "${buildConfig.VARIANT}"),
```
karianna: Why is this removed?

aahlenst: The type of JVM (Hotspot, OpenJ9 etc.) is now inferred from the filename of the artifacts copied into the workspace.

PR #1230, opened 2019-08-21
Roll back hardening

johnoliver:


gdams:
Linter issue then good to go

PR #1253, opened 2019-08-29
Fix build scripts breaking when building non-dev branch

johnoliver:


sbin/prepareWorkspace.sh
```diff
@@ -102,7 +102,7 @@ checkoutAndCloneOpenJDKGitRepo()
 
   if [[ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_HOTSPOT}" ]] && [[ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -ge 11 ]] ; then
      # Verify Adopt patches tag is being built, otherwise we may be accidently just building "raw" OpenJDK
-     if [ ! -f "${ADOPTOPENJDK_MD_MARKER_FILE}" ]; then
+     if [ ! -f "${ADOPTOPENJDK_MD_MARKER_FILE}" ] && [ "${BUILD_CONFIG[DISABLE_ADOPT_BRANCH_SAFETY]}" == "false" ]; then
```
karianna: Should we flip this so we don't have a double negative (DISABLE & false)?

karianna: Nevermind - this is a flag without args

karianna:
For those watching this makes at home builds much easier

Issue #1216, opened 2019-08-13
Docker builds failing on MacOS


kittylyst:
**Platform:** MacOS 10.14.6

**Architecture:** x64

**Related:** 1073

The command:

```
./makejdk-any-platform.sh --docker jdk8u
```
fails out of the box.

Adding --debug-docker and --ssh

causes the Docker container to start, but the build command now fails.

Setting:

```
build@cc958f466f61:/openjdk/build$ JAVA_HOME=/usr/lib/jvm/zulu-7-amd64/jre
build@cc958f466f61:/openjdk/build$ export JAVA_HOME
build@cc958f466f61:/openjdk/build$ java -version
openjdk version "1.7.0_232"
OpenJDK Runtime Environment (Zulu 7.31.0.5-CA-linux64) (build 1.7.0_232-b6)
OpenJDK 64-Bit Server VM (Zulu 7.31.0.5-CA-linux64) (build 24.232-b6, mixed mode)
```

enables the build to find a JVM, but this still fails:

```
build@cc958f466f61:/openjdk/build$ /openjdk/sbin/build.sh
/openjdk/sbin/common/common.sh: line 55: name: unbound variable
Running gradle with /usr/lib/jvm/zulu-7-amd64/jre
bash: ./gradlew: No such file or directory
```

johnoliver:
this is probably because the pipelines directory required at https://github.com/AdoptOpenJDK/openjdk-build/blob/fd939c142a0370cda48eebca9f5dee9d22f0dfc1/sbin/build.sh#L46 is not brought into the docker container

sxa555:
@willsparker Perhaps we could include the testing of builds in docker as part of the playbook testing too

Issue #1073, opened 2019-05-07
makejdk-any-platform.sh docker support is broken (Gradle-related)


sxa555:
May be other issues but at the moment gradle is unable to find a suitable JDK to use at the moment, so docker builds cannot be performed.

Willsparker:
The version of Gradle that is being supported only seems to work using Java-8. With the fixes listed in #1216, I'm able to run `buildDocker.sh -v jdk8u` in `/openjdk-build/docker/`. Once I've altered the `../docker/jdk8/../Dockerfile` to install `zulu-8` instead of `zulu-7`, it was able to run Gradle correctly and start the build.

Willsparker:
I was also able to correctly start building JDK10 in a docker container. This was done by installing both `zulu-8` and `zulu-9` into the container, in which the `JDK_BOOT_DIR` was detected to be `zulu-9` as required. From here, `JAVA_HOME` was set to `zulu_8`, and the build on the Docker container started.

Willsparker:
This shouldn't be an issue anymore with #1278 :-)

PR #1278, opened 2019-09-13
Changed scripts to fix broken Docker builds

Willsparker:
Fixes: #1216  
Ref: #1073 

These changes should allow for the Docker Builds to work, and subsequently the `buildDocker.sh` script.
The unbounded variables problem has been fixed by the `if` statements in `docker-build.sh`. It's not a very clean way of fixing it, but it works if the variables haven't been used in the case of not using `--debug-docker` or `--ssh` with `make-any-platform.sh`.
The problem of the `./gradlew` script no being found has been fixed by adding that into the Docker container, in `docker-build.sh` by the addition of `-v "${hostDir}/pipelines":/openjdk/pipelines \ `.

With the current version of Gradle that is put onto the containers, it can only be executed with Java-8. Therefore, all Dockerfiles have `zulu-8`, along with the JDK version needed for the `JDK_BOOT_DIR` var, with `zulu-8` being set to their `JAVA_HOME` variable. 

`build.sh` had to be changed slightly, as in its current form it is putting `JAVA_HOME` to be what the `JDK_BOOT_DIR` is.

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -135,19 +135,54 @@ buildOpenJDKViaDocker()
      dockerEntrypoint=(--entrypoint "/bin/sh" "${BUILD_CONFIG[CONTAINER_NAME]}" -c "echo 'DEBUG DOCKER BUILD\\nTo build jdk run\\n/openjdk/sbin/build.sh'; /bin/bash")
   fi
 
-  # shellcheck disable=SC2140
-  # Pass in the last important variables into the Docker container and call
-  # the /openjdk/sbin/build.sh script inside
-  ${BUILD_CONFIG[DOCKER]} run \
+  #if both dockerMode and gitSshAccess are unbounded
+  if [ ! -n "${dockerMode[@]:-}" ]  && [ ! -n "${gitSshAccess[@]:-}" ]; then
+    ${BUILD_CONFIG[DOCKER]} run \
+         --cpuset-cpus="${cpuSet}" \
+         -v "${BUILD_CONFIG[DOCKER_SOURCE_VOLUME_NAME]}:/openjdk/build" \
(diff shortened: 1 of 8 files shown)


docker-build.sh
```diff
@@ -135,19 +135,54 @@ buildOpenJDKViaDocker()
      dockerEntrypoint=(--entrypoint "/bin/sh" "${BUILD_CONFIG[CONTAINER_NAME]}" -c "echo 'DEBUG DOCKER BUILD\\nTo build jdk run\\n/openjdk/sbin/build.sh'; /bin/bash")
   fi
 
-  # shellcheck disable=SC2140
-  # Pass in the last important variables into the Docker container and call
-  # the /openjdk/sbin/build.sh script inside
-  ${BUILD_CONFIG[DOCKER]} run \
+  #if both dockerMode and gitSshAccess are unbounded
+  if [ ! -n "${dockerMode[@]:-}" ]  && [ ! -n "${gitSshAccess[@]:-}" ]; then
```
karianna: Seems like a fair amount of Don't Repeat Yourself (DRY) here - is it possible to extract common logic?

karianna:
I think I overrode my other comment.  Is it also possible to use an adopt binary as opposed to zulu?  Just for dog fooding purposes.

Willsparker:
@karianna That makes sense, and @sxa555 suggested it too. I mainly put in zulu for ease, as a placeholder :-)

docker-build.sh
Willsparker: Done. I've changed where the debug error message executes in the script, but it looks the same as It did before. This is because I was having a lot of issues with missing speech marks when adding `dockerMode` and `gitSshAccess` onto `commandString`

karianna:
Approved but not landing until Java 13 is out

Issue #1283, opened 2019-09-17
Allow for user-custom make arguments, similar to how --configure-args works


d3r3kk:
When building the jdk8u using patches to allow building on recent macOS/XCode, we discovered that upstream a `make` argument of `COMPILER_WARNINGS_FATAL=false` was being used.

There is no current facility in `makejdk-any-platform` allowing us to set this argument (nor any other).

Specifically in the case of the jdk8u branch, where bug fixes are going to necessarily be kept to a minimum, it may be prudent to allow such custom overrides and make it convenient, analogous to the `--configure-args` switch present today.

d3r3kk:
Closing as this is addressed by #1284

PR #1284, opened 2019-09-17
Add custom  commands accessible from makejdk-any-platform

d3r3kk:
Add the command line --make-args, similar to --config-args but for `make`

Fix for #1283

### DIFF
--- a/makejdk-any-platform.1
+++ b/makejdk-any-platform.1
@@ -119,7 +119,7 @@ the one you are trying to build.
 .BR \-k ", " \-\-keep
 if using docker, keep the container after the build.
 .TP
-.BR \-C ", " \-\-make-args " " \fI<args>\fR
+.BR \-\-make-args " " \fI<args>\fR
 specify any custom user make arguments.
 .TP
 .BR \-\-no\-adopt\-patches
--- a/makejdk-any-platform.1
+++ b/makejdk-any-platform.1
@@ -119,6 +119,9 @@ the one you are trying to build.
 .BR \-k ", " \-\-keep
 if using docker, keep the container after the build.
 .TP
+.BR \-C ", " \-\-make-args " " \fI<args>\fR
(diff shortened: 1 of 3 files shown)


makejdk-any-platform.1
```diff
@@ -119,6 +119,9 @@ the one you are trying to build.
 .BR \-k ", " \-\-keep
 if using docker, keep the container after the build.
 .TP
+.BR \-C ", " \-\-make-args " " \fI<args>\fR
```
karianna: I think there's a confusion clash here as `-C` is used for `--configure-args` in _config_init.sh_?

sbin/common/config_init.sh
```diff
@@ -406,8 +410,9 @@ function configDefaults() {
   # JVM variant, e.g. client or server, defaults to server
   BUILD_CONFIG[JVM_VARIANT]=${BUILD_CONFIG[JVM_VARIANT]:-""}
 
-  # Any extra args provided by the user
+    # Any extra config/make args provided by the user
```
karianna: Whitespace

makejdk-any-platform.1
d3r3kk: copypasta - fixing

sbin/common/config_init.sh
d3r3kk: :+1:

sbin/common/config_init.sh
```diff
@@ -406,8 +410,9 @@ function configDefaults() {
   # JVM variant, e.g. client or server, defaults to server
   BUILD_CONFIG[JVM_VARIANT]=${BUILD_CONFIG[JVM_VARIANT]:-""}
 
-  # Any extra args provided by the user
+    # Any extra config / make args provided by the user
```
karianna: @d3r3kk - I think the whitespace indent is incorrect here.

PR #1292, opened 2019-09-23
Modified buildDocker for option to use Eclipse OpenJ9 dockerfiles

Willsparker:
Requested by @sxa555 

Upon use of the `--eclipse` option, the script will retrieve the correct dockerfile from the `ibmruntimes/openjdk-openj9-jdkXX` repository, build a container using it and build the JDK on it.

### DIFF
--- a/docker/buildDocker.sh
+++ b/docker/buildDocker.sh
@@ -4,6 +4,7 @@ set -u
 jdkVersion=''
 bootDir=''
 openJ9=false
+useEclipseDockerFiles=false
 
 #takes in all arguments to determine script options
 parseCommandLineArgs()
@@ -25,6 +26,8 @@ parseCommandLineArgs()
 					bootDir="$1"; shift;;
 				"--openj9" | "-j9")
 					openJ9=true;;
+				"--use-eclipse-docker-files" | "-e" )
+					useEclipseDockerFiles=true;;
 				"--help" | "-h" )
 					usage; exit 0;;
 				*) echo >&2 "Invalid option: ${opt}"; echo "This option was unrecognised."; usage; exit 1;;
@@ -36,10 +39,11 @@ parseCommandLineArgs()
 usage()
 {
 	echo
-	echo "Usage: ./buildDocker.sh	--all|-a 		Build all support JDK versions"
-	echo "			--version|-v 		Build the specified JDK version"
-	echo "			--jdk-boot-dir|-J	Specify the boot JDK directory"
-	echo "			--openj9|-j9		Builds using OpenJ9 instead of Hotspot"
+	echo "Usage: ./buildDocker.sh	--all|-a 			Build all support JDK versions"
+	echo "			--version|-v 			Build the specified JDK version"
+	echo "			--jdk-boot-dir|-J		Specify the boot JDK directory"
+	echo "			--openj9|-j9			Builds using OpenJ9 instead of Hotspot"
+	echo "			--use-eclipse-docker-files|-e	Builds the specified jdk using the Eclipse Openj9 dockerfiles"
 	echo
 }
 
@@ -58,7 +62,8 @@ checkJDKVersion()
 			jdkVersion="jdk12u";;
 		"jdk13u" | "jdk13" | "13" | "13u" )
 			jdkVersion="jdk13u";;
-		"all" ) ;;
+		"all" )
+			jdkVersion="jdk8u jdk9u jdk10u jdk11u jdk12u jdk13u";;
 		*)
 			echo "Not a valid JDK Version" ; jdkVersionList; exit 1;;
 	esac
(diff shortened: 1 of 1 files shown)


docker/buildDocker.sh
```diff
@@ -4,6 +4,7 @@ set -u
 jdkVersion=''
 bootDir=''
 openJ9=false
+eclipseDockFiles=false
```
karianna: Maybe `useEclipseDockerFiles`

docker/buildDocker.sh
```diff
@@ -25,6 +26,8 @@ parseCommandLineArgs()
 					bootDir="$1"; shift;;
 				"--openj9" | "-j9")
 					openJ9=true;;
+				"--eclipse" | "-e" )
```
karianna: maybe `use-eclipse-docker-files`

docker/buildDocker.sh
```diff
@@ -82,6 +87,21 @@ removeBuild()
 	cd $WORKSPACE/DockerBuildFolder/openjdk-build/workspace && rm -r build
 }
 
+buildDockerEclipse()
```
karianna: maybe `useEclipseDockerFiles`

karianna:
@Willsparker did you see @AdamBrousseau comment above?  I think that needs to be resolved before merging.

Willsparker:
@AdamBrousseau I was following the instructions given on https://www.eclipse.org/openj9/oj9_build.html
@sxa555 suggested I put it into the script as it opens discussions on if its worth maintaining both our own Dockerfiles to build with OpenJ9, or use Eclipse's.

sxa555:
@AdamBrousseau This was purely based on the documentation that an end user was likely to find - I hadn't seen the other set so thanks for pointing at this issue however it looks like that issue hasn't been touched for over two months. I'm happy to put either set in depending on which are the "live" ones. To be honest the job that @Willsparker put together is to test the Dockerfiles on a regular basis, so I'd be totally happy with including both sets for now and removing one later once the upstream issue has been resolved ...

Willsparker:
@AdamBrousseau @sxa555 I've added the option for either set of Dockerfiles :-)

AdamBrousseau:
Yes the issue kind of went stale. There's a related issue @keithc-ca is working on to create a script which will generate a Dockerfile for whatever OS/Arch combination you choose (fyi).
In regards to regular testing, the ones in docker-slaves are tested in the internal nightly build. At some point I'd like to have them tested at the Eclipse farm but we're not there yet.
Also should point out, the docker-slaves ones are available on DockerHub if you'd rather pull them prebuilt, they take a long time to build because of all the pieces we have in there.

karianna:
@AdamBrousseau and @keithc-ca - so it looks like we can merge this and update once more when the upstream work is done?

docker/buildDocker.sh
```diff
@@ -76,12 +86,58 @@ jdkVersionList()
...
+		cp Dockerfile $PWD/Dockerfile.$jdk
+		sharedDockerCommands $jdk
+	done
+}
+
+sharedDockerCommands()
+{
+	local jdk=$1
+	docker build -t $jdk -f Dockerfile.$jdk .
+	docker run -it -u root -d --name=$jdk $jdk
```
AdamBrousseau: Was the root user tested with the `docker-slaves` files? Generally we use the `jenkins` user so I'm not sure the env will work. For example, the freemarker.jar is in `/home/jenkins`. There is also a line about ulimits for the user. I think that's a test req. Also, if you do end up using the Jenkins user, I seem to recall there was env differences depending on if you started the container with `u jenkins` vs starting as `root` and then `su - jenkins`. Possibly related to openssl. Those issues may have been resolved though.

sxa555:
@keith-ca Where are you planning to host this makefile - Not sure it if makes sense to have it in the `openj9` repo requireing you to clone the whole thing to generate the makefile. Or is the plan to publish the generated dockerfiles elsewhere? At the moment the end user experience is a straight "Click here to download to Dockerfile, then run it" and it'd be a shame to overcomplicate that.

sxa555:
To be honest if the changes in here work I'm happy (subject to @AdamBrousseau's comment about the users being used) to go ahead and merge this and update for now and adjust it when the alternate mechanism to generate the `Dockerfile`s is complete.

keithc-ca:
> @keith-ca Where are you planning to host this

Users should need only to fetch a single bash script from github and run it; no need to clone the whole repo.

docker/buildDocker.sh
Willsparker: Running `./buildDocker.sh -v jdk8u --use-eclipse-docker-slave-files` as is, does seem to work, giving this output:

```
/usr/bin/make -s VERBOSE="-s" LOG_LEVEL="warn" -I /root/openj9-openjdk-jdk8/make/common -f /root/openj9-openjdk-jdk8/closed/OpenJ9.gmk SPEC=/root/openj9-openjdk-jdk8/build/linux-x86_64-normal-server-release/spec.gmk openj9_test_image
/root/openj9-openjdk-jdk8/build/linux-x86_64-normal-server-release/images/j2re-image/bin/java -version 2>&1 | /usr/bin/tee /root/openj9-openjdk-jdk8/build/linux-x86_64-normal-server-release/images/test/openj9/java-version.txt
openjdk version "1.8.0_232-internal"
OpenJDK Runtime Environment (build 1.8.0_232-internal-_2019_09_30_07_12-b00)
Eclipse OpenJ9 VM (build master-230f0fa, JRE 1.8.0 Linux amd64-64-Bit Compressed References 20190930_000000 (JIT enabled, AOT enabled)
OpenJ9   - 230f0fa
OMR      - fd3d7a7
JCL      - aaea930 based on jdk8u232-b07)
----- Build times -------
Start 2019-09-30 07:12:25
End   2019-09-30 07:41:21
00:00:28 corba
00:00:33 demos
00:02:03 docs
00:01:58 images
00:00:15 jaxp
00:00:26 jaxws
00:03:57 jdk
00:00:24 langtools
00:00:11 nashorn
00:28:56 TOTAL
-------------------------
```

karianna:
I'm still a little unclear whether this is safe to land :-)

Willsparker:
@AdamBrousseau Are you happy with the output of `./buildDocker.sh -v jdk8u --use-eclipse-docker-slave-files` posted above?

sxa555:
Happy to merge this for now to get the jenkins jobs running green - @AdamBrousseau  if you are concerned about the users etc. let's discuss further (possible in #build) then we can look at that in further comments here, a separate issue, or have a discussion in #build at adopt :-)

PR #1311, opened 2019-10-03
Script to merge from AdoptOpenJDK into FreeBSD child repos

battleblow:
* Add a script that will allow the AdoptOpenJDK bot to automatically merge
  changes from the AdoptOpenJDK repos into the FreeBSD child repos.
* The script itself doesn't contain the automation, but it gives an
  automated facility (e.g. cron, an event queue) a single command to
  execute to do the update.
* This could be used to perform a sync (manually or automatic) for any
  child repository with a few tweaks.  In particular see CHILD_REPO_PATH
  and SYNC_BRANCH.

### DIFF
--- a/git-hg/freebsdRepoSync.sh
+++ b/git-hg/freebsdRepoSync.sh
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+
+# Script to facilitate automatic merging of changes from the AdoptOpenJDK
+# repos into child repos
+
+# The reference AdoptOpenJDK repos
+export ADOPT_REPO_PATH=https://github.com/AdoptOpenJDK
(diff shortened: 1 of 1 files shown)


battleblow:
Script discussed with @gdams to allow automatic merging of the FreeBSD child repos.

I've tested this except for the push part (because doing so makes it hard to test again), however the push is trivial assuming the script runner has the correct github credentials.

gdams:
thanks for this @battleblow! maybe you could include `freebsd` in the filename so it's clear what this script is for?

battleblow:
> thanks for this @battleblow! maybe you could include `freebsd` in the filename so it's clear what this script is for?

Of course :).  Renamed to freebsdRepoSync.sh

PR #1318, opened 2019-10-10
pass --enable-jfr to jfr builds

johnoliver:


pipelines/build/common/build_base_file.groovy
```diff
@@ -153,10 +153,17 @@ class Builder implements Serializable {
         return labels
     }
 
-    static String getConfigureArgs(Map<String, ?> configuration, String additionalConfigureArgs) {
+    static String getConfigureArgs(Map<String, ?> configuration, String additionalConfigureArgs, String variant) {
         def configureArgs = ""
 
-        if (configuration.containsKey('configureArgs')) configureArgs += configuration.configureArgs
+        if (configuration.containsKey('configureArgs')) {
```
karianna: This seems pretty hardcoded to having a variant attribute in the map.  Not sure if this is too tightly coupled?

johnoliver: hmmm, not sure what you mean there, this is similar to how additionalNodeLabels work such as in: https://github.com/AdoptOpenJDK/openjdk-build/blob/580c111be3bf377d8a70a187d3de2e8f323bed71/pipelines/build/openjdk8_pipeline.groovy#L72

karianna: I think I'm thinking about the fact we've added a variable for `variant` and added some processing logic just for that specifically, what happens if we added more nested attributes?  Am I making sense?

PR #1322, opened 2019-10-14
FreeBSD sync: Push changes as well as tags

battleblow:
* Issuing 'git push --tags' will only push the tags, so also do a regular
  'git push' as well to push the changes.

  Interestingly, 'git fetch --tags' fetches both the changes and the tags


battleblow:
CC @gdams

PR #1341, opened 2019-10-21
Pull changes when working with an existing repository

battleblow:
* In the situation where the repository already exists, pull changes
  before attempting to reset ourselves to origin/branch to ensure that
  all of the origin changes are present.  Otherwise there may be
  additional changes on the origin that will cause push to fail.


battleblow:
@gdams , this should fix the job failures.  That said, I suspect they will fix themselves if the script continues to run.

Issue #1293, opened 2019-09-24
BUILD_CONFIG[JDK_PATH] variable is configured wrong when running a build in a Docker container


Willsparker:
**Platform:**
macOS Mojave
**Architecture:**
x86_64

When debugging the Dockerfiles, I ran into the following error. 
```
Should have the source, I'm at /openjdk/build/src
ls: cannot access '/openjdk/build/src/build/*/images/jdk-bundle/jdk-*.jdk': No such file or directory
```
When building JDK8,11,12 and 13 in a Docker container
The actual directory is at `/openjdk/build/src/build/*/images/jdk`, which points to an issue with 
https://github.com/AdoptOpenJDK/openjdk-build/blob/4d1e8bcb9eb1a932901ca526826ed9f8c2e1108f/sbin/build.sh#L447
and
https://github.com/AdoptOpenJDK/openjdk-build/blob/4d1e8bcb9eb1a932901ca526826ed9f8c2e1108f/configureBuild.sh#L127

It seems that `configureBuild.sh` is being ran on the host system and assigning `BUILD_CONFIG[JDK_PATH]` not on the docker container.

sxa555:
@Willsparker Can you confirm that this is no longer an issue now that the PR was landed so this can be closed?

Willsparker:
@sxa555 The PR isn't currently merged. However, I've checked #1347 with jdk8-hotspot and it does fix the issue :-)

sxa555:
It is now.:-D

PR #1347, opened 2019-10-24
Configure path variables accordingly when choosing to build in Docker containers

Haroon-Khel:
This PR is a solution to issue #1293 
The JRE_PATH and JDK_PATH variables are configured according to the OS of the docker container, not the native host.

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -82,6 +82,41 @@ buildOpenJDKViaDocker()
   # shellcheck disable=SC1090
   source "${BUILD_CONFIG[DOCKER_FILE_PATH]}/dockerConfiguration.sh"
 
+    local openjdk_core_version=${BUILD_CONFIG[OPENJDK_CORE_VERSION]}
+    local openjdk_test_image_path=""
+    local jdk_path=""
+    local jre_path=""
+
+    if [ "$openjdk_core_version" == "${JDK8_CORE_VERSION}" ]; then
+      case "${BUILD_CONFIG[OS_KERNEL_NAME]}" in
+      "darwin")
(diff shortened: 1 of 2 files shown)


docker-build.sh
```diff
@@ -82,6 +82,41 @@ buildOpenJDKViaDocker()
   # shellcheck disable=SC1090
   source "${BUILD_CONFIG[DOCKER_FILE_PATH]}/dockerConfiguration.sh"
 
+    local openjdk_core_version=${BUILD_CONFIG[OPENJDK_CORE_VERSION]}
```
johnoliver: Typically we try to keep the configuration building inside `configureBuild.sh` so that the config is set in stone when the build is initiated. This block should probably be in `configureBuild.sh`

johnoliver: actually this block is already in `configureBuild.sh`, is there an issue with it being there?

Haroon-Khel: The block in `configureBuild.sh` currently configures the variables JRE/JDK_PATH for native builds only and does not reconfigure these variables (or create new variables) when building inside a docker container, hence the cause of the bug. Therefore the solution I implemented involved copying the code block from `configureBuild.sh` and adding to `docker-build.sh`.

sxa555: @johnoliver Any further comment after Haroon's statement above?

sxa555:
In the absence of further input from @johnoliver I'm pushing this through on the basis that it has been verified to solve the problem as per https://github.com/AdoptOpenJDK/openjdk-build/issues/1293#issuecomment-561541357

PR #1351, opened 2019-10-28
Skip download of Alsa on macOS.

reitermarkus:
Alsa is not used on macOS.


reitermarkus:
> I think the whitespace is out?

Where exactly?

PR #1358, opened 2019-10-30
Correct the camel case of the function call

karianna:


gdams:
we need to improve our linting on spaces, thanks for tidying things up!

PR #1371, opened 2019-11-06
change windows and linux installer jobs to fail if failed

gdams:


sxa555:
Can we have a better commit message please - I assume "fail high level build pipeline if installer job fails" is the intention here?

Issue #1381, opened 2019-11-13
JDK HEAD does not build testimage file


junyzheng:
**Platform: any**

**Architecture: x64**

The issue is caused by: https://github.com/AdoptOpenJDK/openjdk-build/blob/65e067fafc1a3d786c8ac1b79c3e5af4ee53dbf1/sbin/build.sh#L351

For JDK head, the `OPENJDK_CORE_VERSION` is: `jdk`. Based on logic, it won't be able to add `test-image` configuration.

I will create a PR later to fix this issue.

PR #1382, opened 2019-11-13
[#1381] Fix testimage missing when build JDK head

junyzheng:


junyzheng:
Ready for review

johnoliver:
run tests

junyzheng:
@jerboaa Hi Severin. IMO, to build JDK head, the configuration will always be `jdk`. So we don't need to update this every 6 months.

PR #1383, opened 2019-11-14
Dockerfiles: Refactor buildDocker script

Willsparker:
The main thing done here was to more appropriately split up the various versions that `buildDocker.sh` can run. As it currently stands, a person couldn't run 2 docker builds in parallel from the same `openjdk-build` repo.
I've also fixed a bug in which if the first time `buildDocker.sh` is ran is using the `-use-eclipse-docker-files` option, the script won't work as the `$WORKSPACE/DockerBuildFolder/` directory hasn't been made yet.

### DIFF
--- a/docker/buildDocker.sh
+++ b/docker/buildDocker.sh
@@ -3,7 +3,7 @@ set -u
 
 jdkVersion=''
 bootDir=''
-openJ9=false
+buildVariant="hotspot"
 useEclipseDockerFiles=false
 useEclipseDockerSlavesFiles=false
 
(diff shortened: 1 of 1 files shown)


docker/buildDocker.sh
```diff
@@ -138,34 +144,40 @@ buildDocker()
...
 	fi
 	for jdk in $jdkVersion
 	do
 		echo "$commandString $jdk being executed"
-		cd $WORKSPACE/DockerBuildFolder/openjdk-build && $commandString $jdk
+		cd $WORKSPACE/DockerBuildFolder/$jdk-$buildVariant && $commandString $jdk
 	done
 }
 
+setupFolders()
```
sxa555: This whole section can just be replaced by
`mkdir -p $WORKSPACE/DockerBuildFolder/$jdkVersion-$buildVariant`
which will create the full path and also not complain if the directory is already there

Willsparker:
@sxa555 This is to fix any issues when running multiple instances of `buildDocker.sh` locally - Its to clone a copy of openjdk-build for each jdk-version and each variant, as not to have multiple containers sharing the same directory

docker/buildDocker.sh
Willsparker: Sorted :)

docker/buildDocker.sh
```diff
@@ -26,11 +26,11 @@ parseCommandLineArgs()
 				"--jdk-boot-dir" | "-J")
 					bootDir="$1"; shift;;
 				"--openj9" | "-j9")
-					openJ9=true;;
+					buildVariant="openj9";;
 				"--use-eclipse-docker-files" | "-e" )
```
karianna: I wonder if these vars should be more explicitly called `use-eclipse-openj9...`

Willsparker: We could do, I just ran out of ideas to differentiate using the adopt openj9 dockerfiles, and the two sets that are in the Eclipse OpenJ9 repo.

PR #1385, opened 2019-11-14
DockerBuild: Made sure variables are named appropriately

Willsparker:
This change will more appropriately name the volume sources for the docker containers, and the containers themselves. This change will make it so when running `./makejdk-any-platform.sh --docker` and `./buildDocker.sh` , there will be unique volumes for the J9 and Hotspot variants. Without this, Jenkins jobs that run 2 instances of `./buildDocker.sh` on the same OS but different build variants at the same time, result in one of the two containers outputting: 
```
Incorrect Source Code for <jdk_version>.  This is an error, please check what is in /openjdk/build and manually remove, exiting...
```
as it's detecting the other build variant's source code.

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -157,11 +157,11 @@ buildOpenJDKViaDocker()
   fi
 
   # Run the command string in Docker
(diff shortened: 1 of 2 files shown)


sxa555:
LGTM but I'm going to request a review from @johno just in case there are likely to be any side-effects which I may have missed as I don't know that code in as much detail

johnoliver:
Think this should be good

Issue #1391, opened 2019-11-19
JDK11/tip mirror script is failing.


d3r3kk:
The mirror script for both jdk and jdk11 have been failing since late October it seems. Both are failing with this output:

```bash
+ mergeTag=true
./mercurialToGit.sh: line 227: releaseTagExcludeList: unbound variable
Build step 'Execute shell' marked build as failure
```

Looks as if a new variable was added `$releaseTagExcludeList` that isn't always going to be set prior to being used?

Proposed fix might be:
```bash
226:        # and do not create an _adopt tag
227: -      if [ ! -z "$releaseTagExcludeList" ] ; then
227: +      if [ -n "${releaseTagExcludeList-}" ] ; then
228:          for skipTag in $releaseTagExcludeList; do
```

Such that `releaseTagExcludeList` would get a default value when left otherwise unset.

PR #1392, opened 2019-11-19
Give optional variable releaseTagExcludeList a default value.

d3r3kk:
JDK11 and JDK mirror scripts are failing.

Unbound variable `$releaseTagExcludeList` is the cause.

Fix for #1391

CC: @johnoliver

### DIFF
--- a/git-hg/mercurialToGit.sh
+++ b/git-hg/mercurialToGit.sh
@@ -224,7 +224,7 @@ function performMergeIntoReleaseFromMaster() {
       mergeTag=true
       # Check if tag is in the releaseTagExcludeList, if so do not bring it into the release branch
       # and do not create an _adopt tag
-      if [ ! -z "$releaseTagExcludeList" ] ; then
+      if [ -n "${releaseTagExcludeList-}" ] ; then
         for skipTag in $releaseTagExcludeList; do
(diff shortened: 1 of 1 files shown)


karianna:
@johnoliver - I'm not convinced this is the right fix, or if it is, does the existing logic actually do what we intended?  I suspect not :-)

d3r3kk:
I see - without the background info of why the `$releaseTagExcludeList` exists I may have provided this fix too hastily.

If I can get a bit of guidance I would be happy to modify the solution in a more appropriate manner.

karianna:
> I see - without the background info of why the `$releaseTagExcludeList` exists I may have provided this fix too hastily.
> 
> If I can get a bit of guidance I would be happy to modify the solution in a more appropriate manner.

If only I knew the background better as well! :-D

d3r3kk:
The error is exactly this:

```bash
./mercurialToGit.sh: line 227: releaseTagExcludeList: unbound variable
```

Which is correct when using `set -euxo pipefail` as we are at the top of the file.

Another fix would be to give that variable a value... was the intention to give that variable a value in all cases?

johnoliver:
Yeah this logic is good, thanks

PR #1396, opened 2019-11-24
Reset before pull

battleblow:
* Reset to the upstream state before doing a pull instead of after.  If
  there were conflicts during the previous merge then the pull will fail
  otherwise.  Since we've done a reset prior to the pull, there is no need
  to do so after, since local changes were wiped out by the initial reset.

Sponsored by:	The FreeBSD Foundation


battleblow:
@gdams This should fix the current job failures

PR #1404, opened 2019-11-27
Enable large heap OpenJ9 builds for s390xLinux and ppc64leLinux

sxa555:
Shelley - asking you to check over this in case you want a different set of tests configured for the new builds

Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/1400 and the aarch64/J9 failing builds (support for JITted builds is only in  them for large heap variants just now)


johnoliver:
run tests

sxa555:
@johnoliver has indicated he thinks this will be ok, therefore merging for now

PR #1410, opened 2019-11-28
Stop pulling cacerts from jdk11u for jdk8u builds

sxa555:
Causing a build break now. Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/1086 (assuming it works and hasn't got any other side effects ...)

Reviews welcome but please do not merge until I've had a chance to test it [EDIT: I'm comfortable that this is putting an identical `cacerts` in the same place as before, so this is good to go]


sxa555:
No comments, so merging this on  the basis that tonight's builds won't be any more broken with this fix.

jerboaa:
> No comments, so merging this on the basis that tonight's builds won't be any more broken with this fix.

:)

PR #1419, opened 2019-12-02
Enable bootJDK to overwrite to 11

adam-thorpe:
Overwrites boot jdk to 11 when building jdk 11 since a lot of machines don't seem to have bootjdk11.
Issue: https://github.com/AdoptOpenJDK/openjdk-build/issues/1409
Thoughts? @sxa555 

### DIFF
--- a/build-farm/platform-specific-configurations/aix.sh
+++ b/build-farm/platform-specific-configurations/aix.sh
@@ -37,8 +37,12 @@ fi
 echo LDR_CNTRL=$LDR_CNTRL
 
 # Any version above 8 (11 for now due to openjdk-build#1409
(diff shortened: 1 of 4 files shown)


adam-thorpe:
My understanding is that these sections of the scripts will fetch a boot jdk if none is present on the machine? So these lines enable it to fetch a jdk 11 when it is building 11, instead of just skipping it.

sxa555:
Understood ... To be honest I want to strip out that whole section entirely since I want the bootstrap JVMs to be on the machines anyway - that was always more of a temporary hack / backstop in case that hadn't been done, but it now really just masks underlying issues :-)

PR #1420, opened 2019-12-02
Add special.openjdk to the set of jdk8's launched test targets

sophia-guo:
Currently special.openjdk target only exists in jdk8 and correspondingly only jdk8 Test_openjdk_special builds were created. https://ci.adoptopenjdk.net/view/Test_openjdk_special/ 
Will expand to other versions when needed.


sophia-guo:
Related https://github.com/AdoptOpenJDK/openjdk-tests/pull/1470

Issue #1415, opened 2019-11-29
Add a requirements section to README


galderz:
A requirements section should be added to the README.

First, the first I run this I got:

```
Searching for JDK_BOOT_DIR
Unable to find any JVMs matching version "(null)".
No Java runtime present, try --request to install.
```

I had `JAVA_HOME` set and `java` was present in path, what is `/usr/libexec/java_home` looking for exactly? Is that only covering java versions installed via Apple or something along those lines? I worked around it by giving it `--jdk-boot-dir`

Then, I'm seeing:

```
Autoconf is not found on the PATH, and AUTOCONF is not set.
You need autoconf to be able to generate a runnable configure script.
You might be able to fix this by running 'brew install autoconf'.
Error: Cannot find autoconf
```

So `autoconf` should be one of those requirements?

I can help with this

sxa555:
Out machines are set up using ansible playbooks, so they provide most of the information you want want as prereqs. For HotSpot they are as per the upstream OpenJDK project, and for OpenJ9 they are at https://github.com/eclipse/openj9/blob/master/doc/build-instructions/Build_Instructions_V8.md (which I believe is a supserset of that required for the HotSpot builds.

I'm not personally a macos expert but it sounds like there's a definite opportunity for autodetecting the right thing on there that we're currently missing.

galderz:
I think the problem with `/usr/libexec/java_home` is that it only detects those Java versions that have been installed via Apple. I had installed mine directly from AdoptOpenJDK and that wasn't being found, even though my `JAVA_HOME` and `javac` where there... I wonder if the code should fallback on looking for `javac`, like it does for other envs, if `/usr/libexec/java_home` fails?

sxa555:
@galderz I presume this can be now closed with your PR being in - If you disagree feel free to repoen this :-)

PR #1422, opened 2019-12-02
Handle macOS installations with non-Apple Java setups #1415

galderz:


sbin/common/common.sh
```diff
@@ -129,7 +129,14 @@ function setBootJdk() {
 
     # shellcheck disable=SC2046,SC2230
     if [[ "${BUILD_CONFIG[OS_KERNEL_NAME]}" == "darwin" ]]; then
+      set +e
       BUILD_CONFIG[JDK_BOOT_DIR]="$(/usr/libexec/java_home)"
+      set -e
+      local returnCode=$?
```
sxa555: Isn't this going to use the return code of `set -e` which is always zero?

galderz: Good point, let me fix that

galderz:
@sxa555 Pushed an update

PR #1426, opened 2019-12-03
Revert "Temporarily use JDK11 to boot JDK11 builds (#1411)"

sxa555:
This reverts commit 572d554169a10c1b02a5f1440acd1657cbcb7648.

May cause issues if some of our build machines don't have a JDK10 boot JDK on them as such issues will no longer be masked after https://github.com/AdoptOpenJDK/openjdk-build/pull/1416/files was merged ... I'd prefer to sit and watch this after integrating tomorrow but I'll leave it to any reviewer as to whether to just approve or approve & merge before the automated runs tonight ;-)

Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/1409


sxa555:
Forcing this through despite the absence of any reviews as - from a sysadmin perspective - I need to be able to debug this during the day if/when it goes wrong due to machines not having a suitable JDK10 boot JDK available.

PR #1438, opened 2019-12-10
Ensure Debian is built with openssl-1.0.2

Willsparker:
With https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1021 , I changed the OpenSSL102 role to include Debian8 - therefore the build script needs to be updated to change the build configuration when building on Debian 8.
I checked this wouldn't break CentOS6 too.

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -43,7 +43,7 @@ fi
 if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]
 then
   # CentOS 6 has openssl 1.0.1 so we use a self-installed 1.0.2 from the playbooks
-  if grep 'release 6' /etc/redhat-release >/dev/null; then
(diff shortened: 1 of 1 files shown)


build-farm/platform-specific-configurations/linux.sh
```diff
@@ -43,7 +43,7 @@ fi
 if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]
 then
   # CentOS 6 has openssl 1.0.1 so we use a self-installed 1.0.2 from the playbooks
-  if grep 'release 6' /etc/redhat-release >/dev/null; then
+  if grep 'release 6' /etc/redhat-release >/dev/null || grep 'jessie' /etc/os-release >/dev/null; then
```
sxa555: Add 2>&1 after each `>/dev/null` otherwise it'll chuck out errors when theat doesn't exist (I suspect it is throwing such an error on non-RHEL6 systems already)

Willsparker: Done 👍

PR #1448, opened 2019-12-16
suppress 'which pigz' output as it looks like an error

sxa555:
This is frequently seen in the build log and has a tendancy to confuse people, when it's just checking if pigz is in place to see if it can be used in preference to gzip, which it will fall back to.

`which: no pigz in (/usr/local/gcc/bin:/usr/local/bin:/usr/bin)`

It's not an error, so suppress the output to avoid confusing people :-)

### DIFF
--- a/sbin/common/common.sh
+++ b/sbin/common/common.sh
@@ -94,8 +94,8 @@ createOpenJDKArchive()
   fi
 
   COMPRESS=gzip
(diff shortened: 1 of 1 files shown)


sxa555:
Merging in the absence of any official review but two Hooray reactions :-P

PR #1459, opened 2019-12-31
Fix datestamps on last week of the year

sxa555:
Somewhat belated since it's the last day of 2019, but this fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/1456

Ref https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html for the datestamp formats


sxa555:
Colleague has eyeballed it, so merging

PR #1461, opened 2020-01-08
Use --with-vendor-name rather than --with-company-name on jdk8

lumpfish:
Signed-off-by: Simon Rushton <rushton@uk.ibm.com>


johnoliver:
yeah will do when the release is out

sxa555:
I would rather this hadn't been landed (hence my rejection of it) since we have not done the final  quarterly OpenJ9 JDK8 builds.

karianna:
> I would rather this hadn't been landed (hence my rejection of it) since we have not done the final quarterly OpenJ9 JDK8 builds.

Apologies - I hadn't caught the extra delay there.

PR #1473, opened 2020-01-13
Fix formatting issues in RELEASING.md

sxa555:


sxa555:
Pushing this through since the formatting isn't right in the current version and this is required this week

PR #1475, opened 2020-01-14
Fix bash syntax error for builds on jdk/jdk

jerboaa:
When invoking the build via './makejdk-any-platform.sh [...] jdk' the
BUILD_CONFIG[OPENJDK_CORE_VERSION] ends up being set to the same value
as JDK_PREFIX, namely 'jdk'. This results in
${BUILD_CONFIG[OPENJDK_CORE_VERSION]#$JDK_PREFIX} evaluating to the empty
string. Thus, the if statement:

if [ "$JDK_VERSION_NUMBER" -gt 8 ] ...

ends up being evaluated as 'if [ "" -gt 8 ] ... resulting in:

bash: [...] integer expression expected

Use BUILD_CONFIG[OPENJDK_FEATURE_NUMBER] instead which should always
be a numeric value.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -346,7 +346,7 @@ buildTemplatedFile() {
 
   # If it's Java 9+ then we also make test-image to build the native test libraries
   JDK_PREFIX="jdk"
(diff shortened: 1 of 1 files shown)


jerboaa:
@johnoliver Could you take a look at this, please? Thanks!

PR #1480, opened 2020-01-17
Disable JDK13 nightly builds

sxa555:
JDK13 now out of support, and the OpenJ9 repositories no longer can no longer build it.

### DIFF
--- a/pipelines/jobs/configurations/jdk13u.groovy
+++ b/pipelines/jobs/configurations/jdk13u.groovy
@@ -49,4 +49,6 @@ targetConfigurations = [
         ]
 ]
 
+disableJob = true
+
 return this


karianna:
IIRC there's more than one place these need to be disabled - also should we disable the git-hg job?

sxa555:
If there are multiple places we need that documented in the FAQ ;-)

sxa555:
> IIRC there's more than one place these need to be disabled - also should we disable the git-hg job?

@gdams @karianna What else is needed?

johnoliver:
You can cancel the hg job too, but that is a separate issue, I dont believe we automate that so that is just a manual disable

PR #1483, opened 2020-01-21
add script to mirror skara repos into github

gdams:
Used to setup https://github.com/AdoptOpenJDK/openjdk-jdk14


karianna:
@gdams - Have we confirmed with Skara folks that their repositories are going to be stable (not get removed)?  Doesn't stop this PR from going through of course :-)

gdams:
> Have we confirmed with Skara folks that their repositories are going to be stable (not get removed)?

@karianna I am in the process of double-checking :)

PR #1487, opened 2020-01-22
build: Allow freetype to be detected on OpenSUSE

sxa555:
Not entirely sure why but on OpenSUSE freetype appears to put intself in a `lib64` directory under the install location instead of `lib`. Perhaps this should be fixed at source, but for now this is a workaround that will symlink `lib` to `lib64` if it's put it in the "wrong" place.


sxa555:
Succeeded at https://ci.adoptopenjdk.net/view/work%20in%20progress/job/VagrantPlaybookCheck/OS=SUSE12,label=vagrant/344/console

@Willsparker 😍

Willsparker:
HUZZAH!! 🎉

PR #1489, opened 2020-01-23
Switch to API v3 and provide better diagnostics

jerboaa:
Currently there is logic where when there is no directory
present pointed to by JDKxx_BOOT_DIR a fallback is in
place to download the JDK N-1 boot JDK version via the
API. If this fails for some reason, no diagnostics are
available.

This patch tries to simplify the logic a bit and adds an
additional 'java -version' output for the - potentially -
changed boot JDK the build will be using.


jerboaa:
@gdams @johnoliver Would you be able to review/push this as it's blocking JDK 14 pipelines? Thanks very much.

PR #1493, opened 2020-01-24
Fix java -version diagnostic output

jerboaa:
With later JDK versions 'java -version' output
goes to the error stream (fd 2). Account for that
in diagnostics so that "BOOT JDK: " is properly
pre-fixed.

### DIFF
--- a/build-farm/platform-specific-configurations/aix.sh
+++ b/build-farm/platform-specific-configurations/aix.sh
@@ -51,7 +51,7 @@ if [ "$JAVA_FEATURE_VERSION" -gt 11 ]; then
       fi
     fi
     export JDK_BOOT_DIR="$(eval echo "\$$BOOT_JDK_VARIABLE")"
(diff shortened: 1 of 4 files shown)


build-farm/platform-specific-configurations/aix.sh
```diff
@@ -51,7 +51,7 @@ if [ "$JAVA_FEATURE_VERSION" -gt 11 ]; then
       fi
     fi
     export JDK_BOOT_DIR="$(eval echo "\$$BOOT_JDK_VARIABLE")"
-    $JDK_BOOT_DIR/bin/java -version | sed 's/^/BOOT JDK: /'
+    $JDK_BOOT_DIR/bin/java -version 2>&1 | sed 's/^/BOOT JDK: /'
```
gdams: probably a good idea to quote all of the `$JDK_BOOT_DIR/bin/java` like I did in Windows

jerboaa: Done.

PR #1494, opened 2020-01-24
Disable JDK 10u pipeline job

jerboaa:
This would be a step towards auto-detecting new pipelines and
generating them.

JDK 10u is deprecated anyway.


jerboaa:
@johnoliver similar to #1480 Thoughts?

PR #1497, opened 2020-01-27
Fix compilation of pipelines

jerboaa:
Caused by b8ffae5cf892a7efb35714783def2fb65796079b.
ff2f113a1ee68a2fd4b3b5b70d0c7a73ea68da00 seems to have happened in
between and the revert was not clean?

### DIFF
--- a/pipelines/build/openjdk8_pipeline.groovy
+++ b/pipelines/build/openjdk8_pipeline.groovy
@@ -135,7 +135,7 @@ def buildConfigurations = [
                 additionalNodeLabels : 'centos6',
                 arch                 : 'x64',
                 additionalFileNameTag: "linuxXL",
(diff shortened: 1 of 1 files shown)


jerboaa:
Nightly pipelines currently fail with:

```
/home/jenkins/workspace/build-scripts/jobs/jdk11u/jdk11u-linux-aarch64-hotspot/pipelines/build/openjdk8_pipeline.groovy: 139: expecting ']', found 'test' @ line 139, column 17.
                   test                 : ['sanity.openjdk', 'sanity.system', 'extended.system', 'special.openjdk'],
                   ^

1 error
```

The reason seems to be a syntax error. The comma needed to be there if the configure args come after. It wasn't. Seems a bad revert.

PR #1501, opened 2020-01-27
Add JDK 15 (HEAD) build pipeline

jerboaa:
- Add fallback for ea boot JDKs. For JDK head builds we might
  not have JDK N-1 present on the build machine. Download it
  via the API, preferring GA releases and falling back to EA
  releases.
- Enable it for hotspot only (at this point)
- Use JAVA_FEATURE_VERSION also for BOOT_JDK_VERSION logic
  so as to avoid empty strings when working with JAVA_TO_BUILD
  See #1475 for a similar issue.
- Remove old pipelines/build/openjdk_pipeline.groovy and
  rename it to pipelines/build/openjdk15_pipeline.groovy
- Add configuration for JDK 15 (currently HEAD).
- Add logic so as to accept 'jdk' as javaToBuild value which
  doesn't contain a number (previous expectation).


jerboaa:
Not sure what to do about those `${releaseType}` unused var warnings. This is intentional - used via `eval` -  but I don't know the syntax to ignore it.

sxa555:
I'm ok with this (other than the fact I don't like all the logic for downloading boot JDKs in there - I wanted to strip out what's there and ensure they're in the playbooks, but I'm not going to block  this for now).

I'd like to get a formal review from @johnoliver before merging though.

johnoliver:
Think my only thing here would be to try to get those download blocks that look like they are duplicated into one place if possible

karianna:
@jerboaa  - if you can fix : https://app.codacy.com/manual/AdoptOpenJDK/openjdk-build/pullRequest?prid=4894746 then this is good to go.

jerboaa:
> @jerboaa - if you can fix : https://app.codacy.com/manual/AdoptOpenJDK/openjdk-build/pullRequest?prid=4894746 then this is good to go.

This is a false positive. Not sure why it mentions any curly braces. They're gone a while ago: https://github.com/AdoptOpenJDK/openjdk-build/pull/1501/files#diff-89e86b4e9e8e519824f70635359fbf5cR53

PR #1509, opened 2020-01-29
Work-around 'set -e' for the JDK N-1 fallback download

jerboaa:
This would be a (temporary) fix to actually get the JDK HEAD
pipelines building. Currently they fail with:

```
Downloading GA release of boot JDK version 14...
gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

That is, 'set -e', prevents for the fallback download
using the EA version of JDK 14 to actually happen.


jerboaa:
@gdams @johnoliver Could you please take a look? Thanks!

jerboaa:
Thanks, I'll merge this and see how it goes. It should affect stable builds as they'll have relevant boot JDKs present.

PR #1512, opened 2020-01-30
Fix Java version output

adam-thorpe:
Closes: https://github.com/AdoptOpenJDK/openjdk-build/issues/1511
FYI @jerboaa 

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -57,12 +57,12 @@ class Build {
     Integer getJavaVersionNumber() {
         // version should be something like "jdk8u" or "jdk" for HEAD
         def matcher = (buildConfig.JAVA_TO_BUILD =~ /(\d+)/)
-        if (matcher.matches()) {
+        if (matcher.find()) {
             List<String> list = matcher[0] as List
             return Integer.parseInt(list[1] as String)
         } else if ("jdk".equalsIgnoreCase(buildConfig.JAVA_TO_BUILD.trim())) {
             // This needs to get updated when JDK HEAD version updates
-            return Integer.valueOf("15")
+            return buildConfig.JAVA_FEATURE_VERSION
         } else {
             return Integer.valueOf("-1")
         }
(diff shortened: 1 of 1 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -57,12 +57,12 @@ class Build {
     Integer getJavaVersionNumber() {
         // version should be something like "jdk8u" or "jdk" for HEAD
         def matcher = (buildConfig.JAVA_TO_BUILD =~ /(\d+)/)
-        if (matcher.matches()) {
+        if (matcher.find()) {
             List<String> list = matcher[0] as List
             return Integer.parseInt(list[1] as String)
         } else if ("jdk".equalsIgnoreCase(buildConfig.JAVA_TO_BUILD.trim())) {
             // This needs to get updated when JDK HEAD version updates
```
jerboaa: With your update to use `buildConfig.JAVA_FEATURE_VERSION` we should remove that comment as it's no longer a hard-coded value.

jerboaa:
> Nice work! Thanks for fixing this and sorry for the breakage.
> 
> There is a second hard-coded value here:
> https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L65
> 
> Could you use `buildConfig.JAVA_FEATURE_VERSION` there too, please? Thank you!

Err, I meant:
`pipelines/build/common/build_base_file.groovy:211`

It's fine to not do that, in this PR though. Removing the comment in `openjdk_build_pipeline.groovy` would be sufficient.

pipelines/build/common/openjdk_build_pipeline.groovy
jerboaa: On second thought. Will `JAVA_FEATURE_VERSION` be available at this point. Looks to me that there are only these parameters going to be generated:
https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/common/create_job_from_template.groovy#L64

jerboaa:
Perhaps @johnoliver should look this over.

pipelines/build/common/openjdk_build_pipeline.groovy
adam-thorpe: I'm not too sure, I didn't realise only this list of values we're passed into the groovy scripts. Surely just adding JAVA_FEATURE_VERSION to this list would be a good idea as we don't want too many hardcoded values that we'd need to change in 6 months time

adam-thorpe: Or adding it as a constant somehow?

adam-thorpe:
@jerboaa I've reverted the change so that it is just the regex fix as I would like this to get fixed asap so we can get some testing done in the nightlies tonight. I'll create an issue for the magic numbers

PR #1517, opened 2020-01-31
move jdk14 and 15 to macos machine that can support notarization

gdams:
@sxa555 this will now move all jdk14 and 15 builds to our Mojave build machine too. I suggest that we consider deprecating the macos10.12 machine and spin up an additional 10.14 machine.

We should be able to take the 10.12 machine offline as soon as we stop building jdk13 (if we haven't already)


gdams:
This fixes the macOS failures that we're seeing on jdk13+ right now

PR #1527, opened 2020-02-06
env: tolerate gcc 7.5 in /usr/local/gcc bin

### LINKED ISSUE #1508: OpenJ9 linux builds on gcc 7.5
OpenJ9 is ready to make the switch on Linux (x, p, z) to switch to gcc 7.5. This version contains a fix which improves the speed of the OpenJ9 bytecode interpreter (on x, z), which may benefit customers in some situations.

OpenJ9 should coordinate the change with Adopt, which will need to have gcc 7.5 installed on the build machines.
https://github.com/eclipse/openj9/pull/8443

sxa555:
Fixes #1508 once updates have been deployed onto the machines. New provisional tarballs are in https://ci.adoptopenjdk.net/userContent/gcc/


sxa555:
Forcing this through as I need it to be tolerated as I've started deploying it to the build machines. Once deployed universally we can look at removing the support for other gcc 7 versions which will stop us accidentally picking up an incorrect one.

PR #1530, opened 2020-02-07
Removed erroneous tags

adam-thorpe:
Was preventing these statements from returning 1 if the command failed, therefore not fetching correct bootjdk on jdk next

### DIFF
--- a/build-farm/platform-specific-configurations/aix.sh
+++ b/build-farm/platform-specific-configurations/aix.sh
@@ -55,7 +55,7 @@ if [ "$JAVA_FEATURE_VERSION" -gt 11 ]; then
         # for the fallback mechanism, as downloading of the GA binary might
         # fail.
         set +e
-        wget -q -O - "${apiURL}" | tar xpzf - --strip-components=1 -C "$bootDir"
+        wget -q -O "${apiURL}" | tar xpzf --strip-components=1 -C "$bootDir"
         retVal=$?
         set -e
         if [ $retVal -ne 0 ]; then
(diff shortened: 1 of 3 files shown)


build-farm/platform-specific-configurations/aix.sh
```diff
@@ -55,7 +55,7 @@ if [ "$JAVA_FEATURE_VERSION" -gt 11 ]; then
         # for the fallback mechanism, as downloading of the GA binary might
         # fail.
         set +e
-        wget -q -O - "${apiURL}" | tar xpzf - --strip-components=1 -C "$bootDir"
```
jerboaa: Are you sure about removing `-` from the `wget` command? From the man page:

```
-O file
       --output-document=file
           The documents will not be written to the appropriate files, but all will be concatenated together and written to file.  If - is used as file, documents will be printed to standard output,
           disabling link conversion.  (Use ./- to print to a file literally named -.)
```

adam-thorpe: Oh right, I thought it was a typo like the tar command, didn't realise it was doing something I'll add it back

sxa555:
@adam-thorpe @karianna Why were those dashes considered erroneous - the `-` after a  tar `f` modifier has to be a filename, and `-` indicates the tarfile is coming from stdin - in this case from the `wget` command. I want to be clear on the problem description before putting in any alternate fix.

Reverting under https://github.com/AdoptOpenJDK/openjdk-build/pull/1536

sxa555:
@adam-thorpe @karianna Why were those dashes considered erroneous - the `-` after a  tar `f` modifier has to be a filename, and `-` indicates the tarfile is coming from stdin - in this case from the `wget` command. I want to be clear on the problem description before putting in any alternate fix.

Reverted under https://github.com/AdoptOpenJDK/openjdk-build/pull/1536

adam-thorpe:
@sxa555 The wget command is expected to fail on jdk15, since there is no ga version of jdk14. Because there is no input file to the tar command, that should fail with an exit code != 0. Except on mac, no exit code is set (so default is 0). When I was testing this, removing the - (which I thought was just a typo) fixed the problem and returned 1, so I thought this was the solution to the problem.

karianna:
Yeah my bad for merging, I saw that `-` on its own and went "I've not seen that before, makes sense to remove them"

PR #1534, opened 2020-02-10
Skip solaris sparcv9

adam-thorpe:
Boxes are offline for the time being


sxa555:
This should be removed from the 11 pipeline too

sxa555:
Doesn't appear to be showing up in [the build job list](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/) despite it being listed in https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/openjdk11_pipeline.groovy#L102 so I'll let this go as-is

PR #1548, opened 2020-02-17
Implement fixed trigger schedule

johnoliver:


karianna:
Fixes #1549

karianna:
@johnoliver fails with:

`/home/jenkins/.jenkins/workspace/build-scripts-pr-tester/openjdk-build-pr-tester/pipelines/jobs/configurations/jdk14u.groovy`

jerboaa:
@johnoliver @karianna: It's been failing like that for a while. Note that JDK 14 doesn't use the jdk14u tree as of yet. it's `openjdk-jdk14`. So my suspicion would be that there is some hard-coding of jdkXXu somewhere.

I might be wrong entirely ;-)

johnoliver:
@karianna yeah that failure will be unrelated to this PR

PR #1552, opened 2020-02-19
Docker: Add JDK15 Dockerfiles & buildDocker.sh support them

Willsparker:
fixes: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1150

Although the script won't currently work for JDK15, this future proofs it for when it does come out. For now, the changes to `buildDocker.sh` will at least allow the `DockerfileCheck` job to run/test JDK14. 

JDK15 Dockerfiles were made by cp'ing the JDK13 repo and `sed -i -e 's/jdk13/jdk15/g' * && sed -i -e 's/JDK13/JDK15/g' *` in it.

### DIFF
--- a/docker/buildDocker.sh
+++ b/docker/buildDocker.sh
@@ -67,8 +67,12 @@ checkJDKVersion()
 			jdkVersion="jdk12u";;
 		"jdk13u" | "jdk13" | "13" | "13u" )
 			jdkVersion="jdk13u";;
+                "jdk14u" | "jdk14" | "14" | "14u" )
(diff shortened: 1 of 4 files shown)


sxa555:
Our [JDK15 pipeline](https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk15-pipeline/) is building from https://github.com/adoptopenjdk/openjdk-jdk using the jobs [here](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk/) so let's go with that and have a "build from HEAD" Dockerfile

PR #1560, opened 2020-02-21
Set TMPDIR to ramdisk on AIX if /ramdisk0/build/tmp exists

sxa555:


sxa555:
Pushing this through so it can be tested over the weekend in an attempt to alleviate apparent I/O slowness on the AIX build machines

PR #1566, opened 2020-02-26
docker: Change jdkVersion to jdk14, not jdk14u

Willsparker:
Was causing the `DockerfileCheck` jenkins job to fail on JDK14, as it was attempting to find the jdk14u repository which isn't there (yet). Also thought I'd remove this for JDK15 so it doesn't trip up the job when it comes to running it.

### DIFF
--- a/docker/buildDocker.sh
+++ b/docker/buildDocker.sh
@@ -70,7 +70,7 @@ checkJDKVersion()
                 "jdk14u" | "jdk14" | "14" | "14u" )
                         jdkVersion="jdk14";;
                 "jdk15u" | "jdk15" | "15" | "15u" )
-                        jdkVersion="jdk15";;
+                        jdkVersion="jdk";;
 		"all" )
 			jdkVersion="jdk8u jdk9u jdk10u jdk11u jdk12u jdk13u jdk14u jdk15u";;
 		*)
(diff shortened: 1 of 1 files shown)


docker/buildDocker.sh
```diff
@@ -68,9 +68,9 @@ checkJDKVersion()
 		"jdk13u" | "jdk13" | "13" | "13u" )
 			jdkVersion="jdk13u";;
                 "jdk14u" | "jdk14" | "14" | "14u" )
-                        jdkVersion="jdk14u";;
+                        jdkVersion="jdk14";;
                 "jdk15u" | "jdk15" | "15" | "15u" )
-                        jdkVersion="jdk15u";;
+                        jdkVersion="jdk15";;
```
sxa555: Are you sure this is correct? Given the jdk14u problem and the fact we don't have an `openjdk-jdk15` repo I suspect this might have to be `jdk`

Willsparker: That's a good point - I put that in more as a future-proof, as I didn't think it'd be used in DF Check job for a little bit. Forgot that `jdk` could be put in to build the next one.

Willsparker: Commit in.

PR #1573, opened 2020-02-28
Add corretto test build names. Fixes #1572

### LINKED ISSUE #1572: Corretto builds should fire off test pipelines named by variant
In order to easily differentiate test results against Corretto builds versus AdoptOpenJDK hotspot builds, we should auto-gen test pipelines with a different variant name.

https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L77 

if (buildConfig.VARIANT == "corretto") {
            variant = "corretto" 
} ...

johnoliver:


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -100,11 +100,11 @@ class Build {
     def runTests() {
         def testStages = [:]
         List testList = []
-        
+
         if (buildConfig.VARIANT == "hotspot-jfr") {
```
smlambert: Can you also add corretto to this if so we do not run sanity.external tests (which pull hs/j9 docker images from docker hub repo to test with by default).  If we want to include this set for corretto, we will have to add special test targets that mount the SDK using EXTRA_DOCKER_ARGS="-v $TEST_JDK_HOME:/opt/java/openjdk".  Since we don't have those targets at the moment, let's not call this set of tests.

johnoliver: done

smlambert:
Learned in TSC mtg the failure is a pre-existing problem so am merging this.

jerboaa:
This PR check failure is pre-existing. I think it happens since the PR checker also included JDK 14, which is openjdk-jdk14 not openjdk-jdk14u. So there might be some hard coded value of XXu somewhere in the groovy script.

Issue #1576, opened 2020-03-02
Add RISC-V support in the build scripts / nightly pipelines


Willsparker:
Requested by @sxa555 

ref: https://github.com/eclipse/openj9/issues/7421 https://github.com/AdoptOpenJDK/TKG/pull/42 
https://ci.adoptopenjdk.net/job/jdk11-linux-riscv-openj9-build/

PR #1577, opened 2020-03-02
Add riscv to jdk11 build pipeline

Willsparker:
fixes: #1576 

Adding riscv64 jdk11-openj9 to the build pipelines. As this build is required to use a cross compiler, I've configured it specifically to use the compiler specifically at `/opt/riscv-toolchain/linux/bin`, with the thought process being that we'd put that file specifically on machines we want to run this on ( shown using the `riscvcross` tag on the machine such as https://ci.adoptopenjdk.net/computer/build-digitalocean-centos69-x64-2/ )


Willsparker:
The most recent commit sets the environment variables based on what I had the `jdk11-linux-riscv-openj9-build` Jenkins job configured to do.

adam-thorpe:
You'll need to add an entry to https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/jobs/configurations/jdk11u.groovy as well, just look at the other examples pretty self explanatory. Other than that I'm not sure, we'll know if it works and whats missing when this gets merged

pipelines/build/openjdk11_pipeline.groovy
```diff
@@ -166,6 +166,13 @@ def buildConfigurations = [
                 additionalFileNameTag: "linuxXL",
                 configureArgs        : '--with-noncompressedrefs --disable-ccache'
         ],
+        riscv64Linux      :  [
+                os                   : 'linux',
+                additionalNodeLabels : 'riscvcross'
+                arch                 : 'riscv'
+                test                 : ['sanity.openjdk', 'sanity.system', 'extended.system'],
```
adam-thorpe: Pretty sure we didn't want testing so setting this to `test : []` will not run tests

pipelines/build/openjdk11_pipeline.groovy
```diff
@@ -166,6 +166,13 @@ def buildConfigurations = [
                 additionalFileNameTag: "linuxXL",
                 configureArgs        : '--with-noncompressedrefs --disable-ccache'
         ],
+        riscv64Linux      :  [
+                os                   : 'linux',
+                additionalNodeLabels : 'riscvcross'
+                arch                 : 'riscv'
```
adam-thorpe: Comma at the end of this line and the line above ;)

pipelines/build/openjdk11_pipeline.groovy
Willsparker: For now, I agree - We'll want to test it _eventually_ - but probably when the JDK itself is more stable :)

pipelines/build/openjdk11_pipeline.groovy
Willsparker: 👍

pipelines/jobs/configurations/jdk11u.groovy
```diff
@@ -51,6 +51,9 @@ targetConfigurations = [
         ],
         "aarch64LinuxXL": [
                 "openj9"
+        ],
+	"riscv64Linux": [
```
adam-thorpe: Real picky but indent?

Willsparker: Fixed 👍

Willsparker:
I've managed to get riscv building on the machine by running the following script, using this branch.

```
#!/bin/bash
export JAVA_TO_BUILD=jdk11u
export TARGET_OS=linux
export VARIANT=openj9
export ARCHITECTURE=riscv
export JAVA_HOME=/usr/lib/jvm/java-1.8.0
export JDK_BOOT_DIR=/usr/lib/jvm/jdk-10.0.2+13/
openjdk-build/build-farm/make-adopt-build-farm.sh
```

 Additional changes were made to ensure the correct versions of the `openj9-openjdk-jdk11`, `omr` and `openj9` were cloned, however this won't be required when riscv support has officially been put into openj9

The only issue I ran into was when building `Freetype` in the build script - as `CC` & `CXX` are set to the cross compiler, when `freetype` is building, this is picked up and errors. I got round this by skipping the freetype build with `-F`, however I'm not sure this is ideal

Willsparker:
With latest commit, I believe all has been set to run the riscv build using the build scripts 👍 Just waiting on the additions into `Eclipse/OpenJ9` for this to become useful.

Willsparker:
With ChengJin's changes being merged, this is ready for review

sxa:
@M-Davies Can you check over this and see if it's adequate or if any other changes will be required?

pipelines/jobs/configurations/jdk11_pipeline_config.groovy
```diff
@@ -166,9 +166,15 @@ class Config11 {
                 additionalFileNameTag: "linuxXL",
                 configureArgs        : '--with-noncompressedrefs --disable-ccache --enable-dtrace=auto'
         ],
+        riscv64Linux      :  [
```
M-Davies, line 169: We should have a field in here specifying whether we want to run tests or not. I.e. https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/jobs/configurations/jdk11_pipeline_config.groovy#L7 OR https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/jobs/configurations/jdk11_pipeline_config.groovy#L99 

If we ARE going to run tests against this, we should have the relevant test jobs created in Jenkins, otherwise there will be nowhere for the tests to run

pipelines/jobs/configurations/jdk11u.groovy
```diff
@@ -51,6 +51,9 @@ targetConfigurations = [
         ],
         "aarch64LinuxXL": [
                 "openj9"
+        ],
```
M-Davies, line 54: Just to confirm, we DO want to have riscv run in the nightlies yes? Otherwise, we shouldn't need this

sxa: It's quick enough and reliable enough (based on https://ci.adoptopenjdk.net/job/jdk11-linux-riscv-openj9-build/) that I'm happy with it being run regularly now

pipelines/jobs/configurations/jdk11_pipeline_config.groovy
sxa: At the moment we do not have a system functional enough to run tests properly, so tests should be disabled - we can add that later

sxa:
I'd like to give this another day for anyone to voice their objections in order to ensure we do not risk destabilising the nightlies - assuming there are no extra concerns we can look at merging and running a new pipeline tomorrow to test this.

Issue #1596, opened 2020-03-09
Builds should archive debugimage archives


keithc-ca:
OpenJ9 builds also produce `build/.../images/debug-image` containing debug information relating to native code. That folder contains platform-specific (`.debuginfo`, `.pdb`, `.dSYM` or `.diz`) files in locations that relate to the corresponding executable or shared library in a jdk.

Copying the `debug-image` folder over top of the jdk makes symbols available to greatly ease the task of debugging an issue with native code.

Archives of this information are already made available in nightly and pull-request builds at ci.eclipse.org/openj9 (e.g. [1]); it would be helpful to make them available via adoptopenjdk.net as well.

[1] https://ci.eclipse.org/openj9/view/Build_Nightly/job/Build_JDK11_x86-64_linux_Nightly/lastSuccessfulBuild/

pshipton:
JDK8 should also be configured with the `--disable-zip-debug-info` option so the debug image contains .debuginfo and .pdb files instead of .diz files which the user (developer) would have to unzip before they can be used with a debugger.
https://github.com/eclipse/openj9/pull/8808/

pshipton:
FYI https://github.com/eclipse/openj9/issues/8874. The OpenJ9 project is going to include the debug-image into the testimage archive, to reduce the number of different files that need to be specified and expanded when running the tests. We are looking at modifying the test scripting to recognize the presence of the debug-image and copy/move the files to the correct place in the JVM build so they are used if a crash occurs when running the tests.
https://github.com/AdoptOpenJDK/openjdk-tests/blob/6d4670a158607cb0e0fd30e76b5c6a0f03a7a74d/get.sh#L273

pshipton:
@keithc-ca what would it take to have a Hotspot build create a debug-image as well?

keithc-ca:
> @keithc-ca what would it take to have a Hotspot build create a debug-image as well?

I expect the additions for OpenJ9 would also work for hotspot. We introduced a new main target ([1]) which invokes a new makefile ([2]). The changes are similar, but not identical, in other versions.

[1] https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/closed/custom/Main.gmk#L72-L75
[2] https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/closed/DebugImage.gmk

sxa555:
@johnoliver Can you take a look at this please - the OpenJ9 team want it available for the April release

[...]

PR #1597, opened 2020-03-09
Capture debug-image if present

keithc-ca:
This is meant to address #1596.

Also captures test-image for jdk8.


configureBuild.sh
```diff
@@ -119,7 +119,8 @@ setVariablesForConfigure() {
   local openjdk_core_version=${BUILD_CONFIG[OPENJDK_CORE_VERSION]}
   # test-image target is JDK 11+ only. Set to the empty string
   # as build scripts check whether the variable is non-empty.
-  local openjdk_test_image_path=""
+  local openjdk_test_image_path="test"
```
karianna: I think this change from the non-empty string will cause issues in later scripts (see the comment above).

docker-build.sh
```diff
@@ -83,7 +83,8 @@ buildOpenJDKViaDocker()
   source "${BUILD_CONFIG[DOCKER_FILE_PATH]}/dockerConfiguration.sh"
 
     local openjdk_core_version=${BUILD_CONFIG[OPENJDK_CORE_VERSION]}
-    local openjdk_test_image_path=""
+    local openjdk_test_image_path="test"
```
karianna: Same comment as above.

configureBuild.sh
keithc-ca: As I understand this, the only difference here is to enable the test-image for jdk8 like other versions. Likewise for docker-build.sh.

keithc-ca:
> One question and does the README also need updating?

I didn't see an obvious place to add information there about this change.

sxa555:
Have queued up https://ci.adoptopenjdk.net/view/work%20in%20progress/job/VagrantPlaybookCheck/434/ to verify if it is likely to break anything in the pipelines (FYI @Willsparker )

gdams:
run tests

configureBuild.sh
karianna: OK I updated the comment inline

johnoliver: in the java 8 repo there is a dir called test:

https://github.com/AdoptOpenJDK/openjdk-jdk8u/tree/master/test

I think this change will have the effect of packaging and archiving that dir

keithc-ca:
> On reflection, default of test will break Java 8 builds.

I'm not sure why this claim is being made. In any case, perhaps I should revert the test-related changes so we can focus on the debug image for now?

karianna:
@johnoliver can explain better but I think that because the Java 8 builds have a pre-existing test folder there's a clash...

johnoliver:
@keithc-ca that value is written into `BUILD_CONFIG[DEBUG_IMAGE_PATH]`
```
BUILD_CONFIG[DEBUG_IMAGE_PATH]=$openjdk_debug_image_path
```
Now that DEBUG_IMAGE_PATH is set the following will now be true when it was previously not:

https://github.com/AdoptOpenJDK/openjdk-build/blob/298421668973dc66c14cf220dbc04fea03f0d1c9/sbin/build.sh#L533

Is that what you intended to do with that change, and is that behaviour is correct? If so, should be fine but the previous comments on this indicate that was only meant to be for java 11+

karianna:
> @keithc-ca that value is written into `BUILD_CONFIG[DEBUG_IMAGE_PATH]`
> 
> ```
> BUILD_CONFIG[DEBUG_IMAGE_PATH]=$openjdk_debug_image_path
> ```
> 
> Now that DEBUG_IMAGE_PATH is set the following will now be true when it was previously not:
> 
> https://github.com/AdoptOpenJDK/openjdk-build/blob/298421668973dc66c14cf220dbc04fea03f0d1c9/sbin/build.sh#L533
> 
> Is that what you intended to do with that change, and is that behaviour is correct? If so, should be fine but the previous comments on this indicate that the logic of that if statement was only meant to be for java 11+ only

@johnoliver did you mean TEST_IMAGE_PATH in your above comment?

johnoliver:
@karianna yes I did

keithc-ca:
Both `${BUILD_CONFIG[TEST_IMAGE_PATH]}` and `${BUILD_CONFIG[DEBUG_IMAGE_PATH]}` are relative paths and will be interpreted based on the current directory set at [1] (`build/*/images`).
All jdk versions may have `images/test`, so I'm still not sure why this would be a 'clash'.
The primary goal of this PR is to also recognize `images/debug-image`.

[1] https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh#L509

keithc-ca:
Updated and rebased.
* updated comments
* use local variables for better clarity
* don't use `$(ls -d $dir)` when `$dir` will not contain wildcards; just use `$dir`

johnoliver:
looks good

Issue #1580, opened 2020-03-03
build scripts "leaking" files in /tmp (public_key.gpg)


sxa555:
These are being left around on build machines which prevents other users from running build jobs on the machine as it can't write to that file. We should either ensure they are removed immediately afterwards, or give it a unique name for each run.

Seen while validating the RISC-V builds as part of #1576 as we were validating the work as a different user ID on the build machines

Willsparker:
Could an alternative of this be that we store this within the workspace?

PR #1599, opened 2020-03-10
Move GPG key to be generated/stored in `WORKSPACE_DIR/WORKING_DIR`

Willsparker:
fixes: #1580 

Ensures that the `public_key.gpg` is specific to each build.

Testing at : https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/392/


Willsparker:
The job succeeded, except for Windows which has a prior issue.

PR #1601, opened 2020-03-11
Add jre to build for jdk14

lumpfish:
Signed-off-by: Simon Rushton <rushton@uk.ibm.com>


lumpfish:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/1600

lumpfish:
Changed list of "or" checks to single "not jdk8" check.
Note that this would change the behaviour for jdk9 or jdk10 should anyone try to build them using this script.
Also spotted somewhere else which needed "or jdk14" adding.

PR #1615, opened 2020-03-19
build: Enable OpenSSL1.1.1e on Windows

Willsparker:
Ref: 
https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1209 
https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1205

Update to the build scripts to allow them to pick up the new version of OpenSSL.

### DIFF
--- a/build-farm/platform-specific-configurations/windows.sh
+++ b/build-farm/platform-specific-configurations/windows.sh
@@ -72,7 +72,12 @@ then
 
   if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]
   then
-    export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-openssl=/cygdrive/c/openjdk/OpenSSL-1.1.1e-x86_32 --enable-openssl-bundling"
+    # If statements to ensure build machines don't break if they're yet to have openssl-1.1.1e put on them
+    if [ -d /cygdrive/c/openjdk/OpenSSL-1.1.1e-x86_32 ]; then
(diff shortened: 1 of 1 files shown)


build-farm/platform-specific-configurations/windows.sh
```diff
@@ -133,11 +133,7 @@ then
     elif [ "$JAVA_FEATURE_VERSION" -ge 11 ]
     then
       TOOLCHAIN_VERSION="2017"
-      if [ -d /cygdrive/c/openjdk/OpenSSL-1.1.1d-x86_64.VS2017 ]; then
-        export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-freemarker-jar=/cygdrive/c/openjdk/freemarker.jar --with-openssl=/cygdrive/c/openjdk/OpenSSL-1.1.1d-x86_64.VS2017 --enable-openssl-bundling"
-      else
-        export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-freemarker-jar=/cygdrive/c/openjdk/freemarker.jar --with-openssl=/cygdrive/c/openjdk/OpenSSL-1.1.1d-x86_64 --enable-openssl-bundling"
-      fi
+      export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-freemarker-jar=/cygdrive/c/openjdk/freemarker.jar --with-openssl=/cygdrive/c/openjdk/OpenSSL-1.1.1e-x86_64-VS2017 --enable-openssl-bundling"
```
Willsparker: Would it be preferable to keep this `if` statement, as it would allow already provisioned build machines to still pick up the old version- I know the plan is to get 1.1.1e on them eventually, but in the meantime?

karianna: Yes keep it but perhaps comment why it's there and could be removed later.

PR #1620, opened 2020-03-19
add missing fi

gdams:
fixes the build scripts


sxa555:
Introduced by https://github.com/AdoptOpenJDK/openjdk-build/pull/1615

PR #1627, opened 2020-03-23
Because the Diabolical Developer was bad, need legacy jre target

karianna:


gdams:
the diabolical developer strikes again....

PR #1638, opened 2020-03-30
bump jdk8u hotspot builds to win2012

gdams:
FYI @sxa555 this will allow us to destroy the 2008 machines and provision more 2012 build machines

### DIFF
--- a/pipelines/jobs/configurations/jdk8_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk8_pipeline_config.groovy
@@ -36,8 +36,8 @@ class Config8 {
                 os                  : 'windows',
                 arch                : 'x64',
                 additionalNodeLabels: [
-                        hotspot : 'win2008',
-                        corretto: 'win2008',
+                        hotspot : 'win2012',
+                        corretto: 'win2012',
                         openj9  : 'win2012&&mingw-cygwin'
                 ],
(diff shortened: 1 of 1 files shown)


gdams:
test builds:

jdk8u-windows-x86-32 (https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x86-32-hotspot/529/)

aahlenst:
If this gets merged, do we need to change the support statement on the website to "Windows Server 2012 or newer" instead of "Windows 2008r2 or later"?

gdams:
@aahlenst we probably should even though we won't be breaking backwards compatibility because it's built on the same toolchain version

gdams:
run tests

karianna:
> @aahlenst we probably should even though we won't be breaking backwards compatibility because it's built on the same toolchain version

I'd actually state the opposite.  We should probably make it clear what we build on vs what we can support on.

karianna:
`Execution error: No such property: versionInfo for class: Build`

gdams:
@karianna this PR checker has been broken by the recent Pipeline changes that allow the jobs to run in parallel, that particular failure has actually come from my other PR here: https://github.com/AdoptOpenJDK/openjdk-build/pull/1639

This PR is good to go (see the two test builds at the top)

Issue #1644, opened 2020-03-31
OpenJ9 builds have incomplete debug symbols


keithc-ca:
Debug information is present for OpenJ9 shared libraries, but not for those derived from openjdk code.

PR #1645, opened 2020-03-31
Don't disable native debug symbols for OpenJ9

keithc-ca:
Addresses #1644.


keithc-ca:
FYI @pshipton

karianna:
If we decide to ship debug builds for HotSpot we can revisit this.

Issue #1531, opened 2020-02-07
AdoptOpenJDK 8u242 JIT enabled aarch32 build missing (Raspberry Pi)


justinyunke:
Hi --

The current AdoptOpenJDK arm (aarch32) release for 8u232 is unusable on the Raspberry Pi, causing a CPU 100% issue similar to that documented here: https://github.com/AdoptOpenJDK/openjdk-build/issues/1287

The daily build (8u252 series) does not experience this issue.

Can we please see an 8u24x or fixed 8u23x release in your repositories soon?

-- Justin

karianna:
@gdams - did we miss this in the latest release?

sxa555:
@karianna We are still missing JDK8/HS/arm32 (subject of this issue), JDK8/HS/aarch64 (I've been managing the resolution to a broken mirror for that one) and JDK13/HS/AIX

sxa555:
We enabled the JIT on arm32/JDK8/HS and need to properly go through the contents of https://github.com/AdoptOpenJDK/openjdk-tests/issues/1548 to ensure it's safe before looking to release that one.

justinyunke:
Thanks for answering so quickly.  Please post to this issue thread if there's something I can do to help.  I have Pi Zero/Pi 2/Pi 3/Pi 4 platforms on which I can build/run tests.

karianna:
> We enabled the JIT on arm32/JDK8/HS and need to properly go through the contents of [AdoptOpenJDK/openjdk-tests#1548](https://github.com/AdoptOpenJDK/openjdk-tests/issues/1548) to ensure it's safe before looking to release that one.

Perhaps we could give @justinyunke an RC binary for him to test on his hardware?

karianna:
I'm following up on AdoptOpenJDK/openjdk-tests#1548 - I'm personally OK with releasing this since it seems to be more about intermittent hardware failures than anything else.

sxa555:
Currently awaiting advice from @johnoliver on the mirroring since I can't build the JIT-enabled version at the correct level from our mirror of the source repository. If we can't get this resolved soon I might just release a non-JIT version again :'(

sxa555:
In the absence of an alternative solution being  forthcoming I have reverted the build jobs back to using the main JDK8u repository and will attempt to push out an 8u242 without a JIT (Consistent with previous releases) ASAP

sxa555:
[...]

PR #1655, opened 2020-04-06
Build aarch32 (arm32/armv7l) builds from main repository instead of the aarch32-specific one for April GA

sxa555:
Since we do not have a resolution to the aarch32 repository issues covered in #1531 I'm going to assume that we will not have it ready for next weeks' quarterly release so it will not be possible ot have a GA build from the JIT-enabled repository. In order to prevent any delay (such as those covered in https://github.com/AdoptOpenJDK/openjdk-docker/issues/303) I am switching back to the zero VM builds from the main jdk8u repository for the April GA :-(


karianna:
Holding off on this to see if I can unlock the git-hg issue

PR #1658, opened 2020-04-08
Inhibit nightly testing during April 2020 release cycle

sxa:
Note to reviewers: DO NOT MERGE THIS YET (I'll do so over the weekend)

As per standard process, this will keep the nightlies running but will not run testing against them, thus preventing executor tie-up for release builds.

### DIFF
--- a/pipelines/jobs/configurations/jdk11_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk11_pipeline_config.groovy
@@ -4,7 +4,7 @@ class Config11 {
                 os                  : 'mac',
                 arch                : 'x64',
                 additionalNodeLabels : 'macos10.14',
-                test                : ['sanity.openjdk', 'sanity.system', 'extended.system', 'sanity.perf'],
+                test                : false,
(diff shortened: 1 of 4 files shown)


pipelines/jobs/configurations/jdk11_pipeline_config.groovy
```diff
@@ -45,14 +42,14 @@ class Config11 {
...
-                test                : ['sanity.openjdk', 'sanity.perf', 'sanity.system', 'extended.system']
+                test                : false
         ],
 
         x64WindowsXL    : [
                 os                   : 'windows',
                 arch                 : 'x64',
                 additionalNodeLabels : 'win2012&&vs2017',
-                test                 : ['sanity.openjdk', 'sanity.system'],
+                test                 : false
```
karianna, line 52: Nees a ','

sxa: Added - thanks :-) If anything was likely to be wrong with this it was going to be that - I even tried to double check them

PR #1668, opened 2020-04-15
Fix checksums on platforms without sha256sum

johnoliver:


gdams:
do you not need to update https://github.com/AdoptOpenJDK/openjdk-build/blob/bd07a4bda17cd4d32c7ba0d36d2fe45b72daa3ba/pipelines/build/common/build_base_file.groovy#L345 too?

PR #1675, opened 2020-04-17
add propagate: true to all installer jobs

gdams:
this means that the builds will now fail if the installers fail to build


karianna:
@gdams Is this true for nightlies as well?

gdams:
@karianna correct

karianna:
OK I see that it could slow down nightlies but it will raise our installers to first-class citizens, so all good.

PR #1682, opened 2020-04-20
remove zulu-7 from the jdk8 build

tmancill:
zulu-7 no longer appears to be needed or used during the build.
(Perhaps it was used previously as part of a bootstrap?)

The PR also installs ca-certificates, which is still needed and was
being pulled in by software-properties-common.

I realize that this doesn't change much.  This saves one `apt-get update` during the build process and avoids unnecessary traffic to Azul's repo, so perhaps speeds up the build just a tad.  But I was curious about "why we can't have nice things..."

### DIFF
--- a/docker/jdk8/x86_64/ubuntu/Dockerfile
+++ b/docker/jdk8/x86_64/ubuntu/Dockerfile
@@ -16,17 +16,11 @@ FROM ubuntu:16.04
 
 MAINTAINER AdoptOpenJDK <adoption-discuss@openjdk.java.net>
 
-# Install required OS tools.  Yes we have to apt-get update first in order to
-# get to software-properties-common, which includes the apt-add-repository
-# package, so we can add the Azul repo and then apt-get update that so we cam
-# get zulu-7.  This is why we can't have nice things.
+# Install required OS tools.
 RUN apt-get update \
-  && apt-get install -y software-properties-common \
-  && apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0x219BD9C9 \
-  && apt-add-repository 'deb http://repos.azulsystems.com/ubuntu stable main' \
-  && apt-get update \
   && apt-get -y upgrade \
   && apt-get install -qq -y --no-install-recommends \
+    ca-certificates \
     ccache \
     cpio \
     curl \
@@ -49,7 +43,6 @@ RUN apt-get update \
     unzip \
     wget \
     zip \
-    zulu-7 \
 && rm -rf /var/lib/apt/lists/*
 
 # Pick up build instructions


karianna:
@tmancill - how does the docker build bootstrap itself without the Java 7?

tmancill:
> @tmancill - how does the docker build bootstrap itself without the Java 7?

@karianna I was puzzling over this as well when I started looking into it.  Ostensibly the assumption is that we're already successfully bootstrapped and can retrieve jdk8 from api.adoptopenjdk.net [in that Dockerfile here](https://github.com/tmancill/openjdk-build/blob/no-zulu7/docker/jdk8/x86_64/ubuntu/Dockerfile#L52-L55); these lines changed in https://github.com/AdoptOpenJDK/openjdk-build/pull/1278.

To test, I cleaned my environment with `docker system prune -a` to ensure that a container with a JDK wasn't being reused from cache, and then built with `./makejdk-any-platform.sh --docker jdk8u`:

```
...
Step 4/24 : RUN mkdir -p /openjdk/target
 ---> Running in 4b5869d461f2
Removing intermediate container 4b5869d461f2
 ---> a06a1507ac08
Step 5/24 : RUN wget 'https://api.adoptopenjdk.net/v2/binary/releases/openjdk8?openjdk_impl=hotspot&os=linux&arch=x64&release=latest&type=jdk' -O jdk8.tar.gz
 ---> Running in 44d266989564
--2020-04-20 18:36:59--  https://api.adoptopenjdk.net/v2/binary/releases/openjdk8?openjdk_impl=hotspot&os=linux&arch=x64&release=latest&type=jdk
Resolving api.adoptopenjdk.net (api.adoptopenjdk.net)... 104.17.158.60, 104.17.159.60
Connecting to api.adoptopenjdk.net (api.adoptopenjdk.net)|104.17.158.60|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u252-b09/OpenJDK8U-jdk_x64_linux_hotspot_8u252b09.tar.gz [following]
...
```

And then later on in `openjdk_container:latest`:
```
...
Searching for JDK_BOOT_DIR
Guessing JDK_BOOT_DIR: /usr/lib/jvm/jdk8
If this is incorrect explicitly configure JDK_BOOT_DIR
Boot dir set to /usr/lib/jvm/jdk8
Running gradle with /usr/lib/jvm/jdk8
...
```

So we're currently building jdk8 with itself.  Should we break the recursion?

sxa:
While building with JDK8 works, we generally use the previous version, but if we've already got it pulling the adoptopenjdk8 then this seems like a reasonable change ... thoughts @karianna ?

karianna:
I'm actually going to be a pian here and state that we should stay with the Zulu 7 as the bootstrap JDK.  We should ideally always build with version -1 (as recommended by the upstream guides).

tmancill:
Makes sense to me.  Closing this PR.  I'll have another look to see what it will take to build with 7 have Java 8 to run Gradle.

PR #1685, opened 2020-04-21
Revert "Fix checksums on platforms without sha256sum"

gdams:
Reverts AdoptOpenJDK/openjdk-build#1668


sxa:
Approving since George has assered that it fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/1683

PR #1689, opened 2020-04-21
Allow building the --with-add-source-root to work with full paths

johnoliver:
https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1263


sxa:
> lgtm but are we holding off until after Mac Os X builds go out?

Yes - we can fully test in independently in the environment that the problem showed up in without it being merged

Willsparker:
https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/506/
Indeed, it worked 👍 Thanks

PR #1697, opened 2020-04-22
Upgrade openj9 builds on Windows and MacOS to use OpenSSL 1.1.1g

sxa:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/1693 but will  https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1269 is also related to oavoid the playbooks getting out of sync. @Willsparker is installing OpenSSL1.1.1g on the Windows machines for me now as the become freed up from running the HotSpot builds to fix the signing issues


sxa:
I don't believe it needs any changes to mac.sh as the version number isn't in there, only the options to enable download + bundling

Issue #1701, opened 2020-04-23
Create "Updating Concurrent Pipeline Scripts" Documentation


M-Davies:
Although we have concurrent pipelines enabled, there is currently no documentation in place to instruct people on how to update the pipelines when the JDK HEAD changes.

- Create a doc with the following specs 
  - [x] How the job regenerator works
  - [x] Updating the [config_regeneration](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/common/config_regeneration.groovy) file. I.e:
    - [folder_name](https://github.com/AdoptOpenJDK/openjdk-build/blob/2041d51db738b741f2a630bc803ca1f038ef62b2/pipelines/build/common/config_regeneration.groovy#L359) will need to be changed when HEAD updates. 
    - [freebsd](https://github.com/AdoptOpenJDK/openjdk-build/blob/2041d51db738b741f2a630bc803ca1f038ef62b2/pipelines/build/common/config_regeneration.groovy#L234). We currently exclude `freebsd` from regeneration since there is no config in the `pipeline_config` files (needs to be removed if we want to build `freebsd` again in the future).
  - [x] Creating/Updating the [pipeline_config files](https://github.com/AdoptOpenJDK/openjdk-build/tree/master/pipelines/jobs/configurations) for each pipeline.
  - [x] Creating/Updating the [regeneration_scripts](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/regeneration) for each JDK version
  - [x] Updating the JDK-HEAD version in the [pipeline](https://github.com/AdoptOpenJDK/openjdk-build/tree/master/pipelines/build) files
  - [x] Creating/Updating the names and configs of the `pipeline_jobs_generator_jdkxx` files in the [utils](https://ci.adoptopenjdk.net/job/build-scripts/job/utils/) folder in jenkins.

M-Davies:
Ideally the place to put this would be the wiki or the [docs](https://github.com/AdoptOpenJDK/openjdk-build/tree/master/docs) folder although there doesn't seem to be any docs in there. 

@karianna could I get editing access to the wiki please?

sxa:
Part of this should be a minimal set of instructions to add to RELEASING.MD to tell people creating new releases what needs to be updated for new releases I.e. which new jobs to add. There's another issue specifically for updating that but it would be good to have it in there even if the full details you are describing in the checklist above go elsewhere.

[...]

PR #1702, opened 2020-04-23
Updating Concurrent Pipeline Scripts

M-Davies:
Part 1 of Issue: #1701 

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -103,7 +103,25 @@ Don't be scared off by this document! If you already understand the stuff in th
 Here are the steps:
 
 1. If desired, add a banner to the website to indicate that the releases are coming in the near future ([Sample PR](https://github.com/AdoptOpenJDK/openjdk-website/pull/702/files))
-2. Build and Test the OpenJDK for "release" at AdoptOpenJDK using a build pipeline job as follows:
+2. Update the job generators to match the JDK HEAD version as follows:
+   * Job Folder - https://ci.adoptopenjdk.net/job/build-scripts/job/utils/: NOTE, As of writing this, the `utils` folder in jenkins is restricted. Ask for access in Slack:#build if you cannot see this folder. The jobs themselves you want are called `pipeline_jobs_generator_jdkxx` (`pipeline_jobs_generator_jdk` for HEAD).
+    * If you are ADDING a JDK version: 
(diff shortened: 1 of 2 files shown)


M-Davies:
Ready for review

sxa:
This shouldn't be in the main section as it's something that needs to be done asynchronously from a release and certainly not something that releasers should be worrying about during the quarterly cycle.

sxa:
Also I would suggest that the line about updating the `JDK_HEAD` job isn't necessary in the section about removing an (old) JDK version as it's unlikely we'd be removing something that would affect the head (and people can hopefully use common sense if such an odd situation did occur.

M-Davies:
Hmm. In that case, would that section be better suited in https://github.com/AdoptOpenJDK/openjdk-build/blob/master/RELEASING.md#openjdk-new-release in that case?

M-Davies:
Or I could add the details on how to update the pipeline job generators to the [README](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/regeneration/README.md) in `/regeneration` and add a link in https://github.com/AdoptOpenJDK/openjdk-build/blob/master/RELEASING.md#openjdk-new-release to it? 

What is your opinion @sxa ?

sxa:
I've been pondering this for a couple of hours (while doing other work I hasten to add!) and I'm still not quite sure ...

Note that there are other enhancements related to creating new releases that need to be added (probably not by you, but anyone who feels they have the information can do it!) under https://github.com/AdoptOpenJDK/openjdk-build/issues/1581

Given that the new sections is (fairly) long with quite a lot of steps, I'm inclined to put it in the `OpenJDK "New" Release` section, but move that section to the end of the document - purely because I want the important information for people doing a quarterly release to be at the top, and that's a separate thing dont every six months asynchronous to the release ... That also means there's an argument for having it as a separate document but I'd prefer to keep it in there just now.

So ... Let's move that section to the bottom of the document, add your bits into it (ideally change the bullet points to numbered steps for consistency and so that people can talk about things like "step 5" when discussing it online) By all means add a "For more information on the pipeline generation scripts see <link to your other doc that you've just added>" into that section too.

The info itself looks great though, so thanks for writing it :-D

M-Davies:
Follow up PR to address the above https://github.com/AdoptOpenJDK/openjdk-build/pull/1709

PR #1703, opened 2020-04-23
Remove debug files from JDK and JRE

### LINKED ISSUE #1699: macos and AIX OpenJ9 April releases are larger than expected
HotSpot does not suffer the same problem - here are the macos download sizes:

jdk8u252 - 140Mb (JDK) 74MB (JRE)
jdk8u242 - 114MB (JDK) 49MB (JRE)
jdk-11.0.7 - 248MB (JDK) 95MB (JRE)
jdk-11.0.6 - 196MB (JDK) 39MB (JRE)

Initial investigation by @pshipton suggests this is related to extra debug symbols being added and various fixes for it such as https://github.com/AdoptOpenJDK/openjdk-build/issues/1636

AIX builds also seems to have got a lot bigger too ... A quick look suggests that other platforms are ok.

keithc-ca:
This should fix #1699.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -574,7 +574,24 @@ removingUnnecessaryFiles() {
   esac
   rm -rf "${dirToRemove}"/demo || true
 
-  find "${jdkTargetPath}" "${jreTargetPath}" -type f "(" -name "*.debuginfo" -o -name "*.diz" -o -name "*.pdb" -o -name "*.map" ")" -delete || true
+  # .diz files may be present on any platform
+  find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.diz" -delete || true
+
+  case "${BUILD_CONFIG[OS_KERNEL_NAME]}" in
+    *cygwin*)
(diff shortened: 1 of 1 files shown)


keithc-ca:
> I think this should be wrapped in a darwin detection block

Should we do the same for other platforms, e.g. only remove `*.map` and `*.pdb` files on Windows?

karianna:
> > I think this should be wrapped in a darwin detection block
> 
> Should we do the same for other platforms, e.g. only remove `*.map` and `*.pdb` files on Windows?

I think that makes it much clearer for the next maintainer yeah

keithc-ca:
Updated as suggested.

My theory is that on AIX, the `find` command wasn't handling the predicate
```
    "(" -name "*.debuginfo" -o -name "*.diz" -o -name "*.pdb" -o -name "*.map" ")"
```
properly, leaving the `.debuginfo` files alone. Now that the commands are platform-specific, there's no need to use `-o` on AIX (only Windows might benefit).

PR #1711, opened 2020-04-27
Docker: Update Dockerfiles to use APIv3

Willsparker:
ping @sxa 

Update to the more stable APIv3 as to stop 404 errors.
(ref: https://ci.adoptopenjdk.net/job/DockerfileCheck/label=docker-aws-ubuntu1604-x64-1,variant=hotspot,version=jdk14u/116/ )

### DIFF
--- a/docker/jdk/x86_64/ubuntu/Dockerfile
+++ b/docker/jdk/x86_64/ubuntu/Dockerfile
@@ -59,12 +59,10 @@ RUN apt-get update \
 RUN mkdir -p /openjdk/target
 
 # Extract AdoptOpenJDK8 to run Gradle.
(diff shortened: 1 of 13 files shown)


Willsparker:
Indeed we can - taking JDK14 as an example:
```
RUN sh -c "mkdir -p /usr/lib/jvm/jdk14 && wget 'https://api.adoptopenjdk.net/v2/binary/releases/openjdk14?openjdk_impl=hotspot&os=linux&arch=x64&release=latest&type=jdk' -O - | tar xzf - -C /usr/lib/jvm/jdk14 --strip-components=1"
```
(The URL will use the v3 API, I was just testing on a standard openjdk-build repo).

PR #1715, opened 2020-04-27
Fix deletion of .debuginfo and .diz files

keithc-ca:
Looking closely at the build logs on AIX, you'll notice two instances of this as it's trying to remove `.debuginfo` and `.diz` files:
```
find: 0652-017 -delete is not a valid option.
```
This removes use of the `-delete` option in favour of using `xargs`.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -575,7 +575,8 @@ removingUnnecessaryFiles() {
   rm -rf "${dirToRemove}"/demo || true
 
   # .diz files may be present on any platform
-  find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.diz" -delete || true
+  # Note that on AIX, find does not support the '-delete' option.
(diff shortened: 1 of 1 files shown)


sxa:
@keithc-ca What is the reason for all of these operations having `|| true` afterwards. Are you expecting them to have not produced the files sometime but want to ignore that situation? Regardless, the `-f` on the `rm` commands you use will never issue a failure error code, so the `|| true` after an `rm -f` serves no purpose

keithc-ca:
> @keithc-ca What is the reason for all of these operations having `|| true` afterwards. Are you expecting them to have not produced the files sometime but want to ignore that situation? Regardless, the `-f` on the `rm` commands you use will never issue a failure error code, so the `|| true` after an `rm -f` serves no purpose

Initially, I thought the same, but I left those there because `find` might fail and someday the script might need `pipefail` to be set. On the other hand, I wouldn't object to removing the `|| true` clauses if `pipefail` is not set.

PR #1717, opened 2020-04-28
Provide notification that signing or installer failed

M-Davies:

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -189,6 +189,11 @@ class Build {
                     def signJob = context.build job: "build-scripts/release/sign_build",
                             propagate: true,
                             parameters: params
+                    
+                    // Output notification of downstream failure (the build will fail automatically)
+		    if (signJob.getResult() == 'FAILURE') {
+                        context.println "ERROR: downstream sign_build FAILED. See ${signJob.getAbsoluteUrl()}"
+                    } 
 
                     //Copy signed artifact back and rearchive
                     context.sh "rm workspace/target/* || true"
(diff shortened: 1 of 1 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -189,6 +189,10 @@ class Build {
                     def signJob = context.build job: "build-scripts/release/sign_build",
                             propagate: true,
                             parameters: params
+                    
+                    if (signJob.getResult() == 'FAILURE') {
+                        context.println "ERROR: downstream sign_build FAILED. See ${signJob.getAbsoluteUrl()}"
+                    }
```
karianna, line 195: Should the script not exit at this point?

M-Davies: It should and the build fails because of it. However, in the logs, it gives no indication that the downstream job, `sign_build`, was the reason the build failed. 

Ref the thread in #build https://adoptopenjdk.slack.com/archives/C09NW3L2J/p1588065913158000, when the sign_build job fails, it doesn't give any feedback to build job that it failed, just a rather unhelpful [Execution error: null](https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk/job/jdk-windows-x64-hotspot/88/console). This change will actually give some feedback in the console log  that it was the sign_build that failed.

karianna: OK, but after this, if statement the script continues to do work, do we not need an exit; here?

M-Davies: The script will exit automatically when the downstream job fails since there is no `catchError` block on the installers or sign build job. I didn't think it was necessary to add an additional exit

karianna: OK, worth adding a comment to that effect though, it's hidden implicit logic to teh next maintainer :-)

PR #1725, opened 2020-05-04
Revert: "Ensure testType isn't false"

M-Davies:
* https://github.com/AdoptOpenJDK/openjdk-build/pull/1723 is no longer needed. It was not the test job's that were at fault.


M-Davies:
Pretty sure I reverted this correctly but could use a review just to ensure I didn't accidently delete a curly bracket somewhere

PR #1734, opened 2020-05-07
Move AIX OpenJ9 builds to AIX 7.1.5 machines

sxa:
Required for fixing https://github.com/AdoptOpenJDK/openjdk-build/issues/1733

### DIFF
--- a/pipelines/jobs/configurations/jdk11_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk11_pipeline_config.groovy
@@ -73,7 +73,10 @@ class Config11 {
         ppc64Aix    : [
                 os                  : 'aix',
                 arch                : 'ppc64',
-                additionalNodeLabels: 'xlc13',
+                additionalNodeLabels: [
+                        hotspot: 'xlc13&&aix710',
+                        openj9:  'xlc13&&aix715'
+                ],
                 test                : [
                         nightly: ['sanity.openjdk'],
                         release: ['sanity.openjdk', 'sanity.system', 'extended.system']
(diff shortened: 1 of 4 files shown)


karianna:
This seems to tie to openj9 machines only - what about hotspot builds for that platform?

sxa:
> This seems to tie to openj9 machines only - what about hotspot builds for that platform?

What about them? Do you want them changed too? There is no issue with the HotSpot builds so if we're moving them it would be for consistency rather than any requirement to do so, but it may adjust the minimum OS level required to run those builds going forward

karianna:
I'm probably confused but for example, I look at the change in pipelines/jobs/configurations/jdk11_pipeline_config.groovy and the change adds the buildj9 label which I think means that this build must choose an (xlc13 and buildj9) labeled machine.  Are pp64Aix only OpenJ9 builds?  a label of buildj9 wouldn't make sense for a hotspot build.

ppc64Aix    : [
                os                  : 'aix',	                os                  : 'aix',
                arch                : 'ppc64',	                arch                : 'ppc64',
                additionalNodeLabels: 'xlc13',	                additionalNodeLabels: 'xlc13&&buildj9',
                test                : [	                test                : [
                        nightly: ['sanity.openjdk'],	                        nightly: ['sanity.openjdk'],
                        release: ['sanity.openjdk', 'sanity.system', 'extended.system']	                        release: ['sanity.openjdk', 'sanity.system', 'extended.system']
                ]	                ]
        ],	        ],

sxa:
I've now adjusted the way I'm doing the labels as I didn't like the use of `buildj9` anyway, even though we have used it elsewhere previously. Have now used `aix710` and `aix715` since that's the capability - the difference that causes it to pass/break is in the `dump` command that shipeed with AIX 7.1SP5.

At the moment all of the `aix710` systems are offline as I was artificially forcing everything through to the new boxes to ensure the OpenJ9 builds went there, but once this is in we can re-enable the old systems again :-)

PR #1747, opened 2020-05-18
Strip java directories from the AIX LIBPATH

sxa:
Not pretty (and if anyone wants to implement another solution then go for it) but this will exclude the Jenkins Agent's PATH from ending up in the java executables, since that has the potential to be confusing.


sxa:
I need to push this through as a fix for https://github.com/AdoptOpenJDK/openjdk-build/issues/1521 - @karianna I've set you as an approver for this so if you object please shout and I we casn revert / do something else.

PR #1762, opened 2020-05-22
Do not attempt to run java -version on riscv cross-compile

sxa:
Error from [latest bulid](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-riscv-openj9/2/console)
`
 Error executing 'java' does not exist in '/home/jenkins/workspace/build-scripts/jobs/jdk11u/jdk11u-linux-riscv-openj9/workspace/build/src/build/linux-riscv64-normal-server-release/images/jdk'.
`
I'd prefer a bit more generic support for cross-compiled builds but this is the only one we have at present so I'm happy with this "quick fix" on one platform for now.

@johnoliver Can you comment on the line in here please:
```
       # do not modify these strings as jenkins looks for them
```
When does jenkins look for these, and what are the implications if that portion is skipped on these builds?

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -473,17 +473,18 @@ printJavaVersionString()
   esac
   if [[ -d "$PRODUCT_HOME" ]]; then
      echo "'$PRODUCT_HOME' found"
-     if ! "$PRODUCT_HOME"/bin/java -version; then
+     if [ ! -r "$PRODUCT_HOME/bin/java" ]; then
        echo "===$PRODUCT_HOME===="
        ls -alh "$PRODUCT_HOME"
 
        echo "===$PRODUCT_HOME/bin/===="
        ls -alh "$PRODUCT_HOME/bin/"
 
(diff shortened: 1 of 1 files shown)


lumpfish:
The code comments say 'Jenkins looks for the strings around the java -version output'.  What do the build pipelines do if they can't find them (which they now won't)?

sxa:
That was the question I targetted to @johnoliver in the initial post, but I'm not sure he's around so I'm tempted in the interests of cleaning up this build (worst case it won't be any worse than it is now I believe!) I'm keen to get this in today

lumpfish:
OK, let's see what it does, approving.

sxa:
Thanks - merged and verifying at https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk11-pipeline/944

Hopefully @johnoliver will be able to comment as to what it's used for.

Issue #1016, opened 2019-04-05
Clear up directory naming scheme inside archives


johnoliver:
We need to decide what our dir naming scheme is inside our archives.

Currently we output dirs of the form:

```
jdk-8.0.202
jdk-8.0.202-jre
jdk-12.0.0
jdk-12.0.0-jre
```
I.e the semver version without any build numbers. In the past dirs were of the form:

```
jdk8u202-b08
jdk8u202-b08-jre
jdk-12+33
jdk-12+33-jre
```

I.e the adopt version number including build.

Currently there is a mismatch between what the installers are accepting as a dir and what we produce. Recently the build number was removed in #1012 to help the installers but this does not seem to have worked. Can we agree on a format and make sure the installers are in sync. My preference would be:

```
jdk-8.0.202+8
jdk-8.0.202+8-jre
jdk-12.0.0+33
jdk-12.0.0+33-jre
```

I.e semver including build

johnoliver:
also possible option of adding `aojdk` to make it clear that it is us

sxa555:
I quite like the current (nightly build) format, although as per John's comment I think it would be good to have `aojdk` at the start if there were no objections as it's useful to identify what's what when doing an `ls` in `/usr/lib/jvm` on my machine :-)

It might be worth collating what all other providers are doing with the directory name in here to help get a consensus.

johnoliver:
I am happy to keep it as:

```
jdk-8.0.202
jdk-8.0.202-jre
jdk-12.0.0
jdk-12.0.0-jre
```

Just need to fix the windows installer to make sure they understand the format

sxa555:
I've got a feeling we have discussed this somewhere previously but could we not avoid jdk in the jre case? i.e.
```
aojdk-8.0.202
aojre-8.0.202
aojdk-12.0.0
aojre-12.0.0
```

lumpfish:
With this convention all variants (openjdk-hotspot and openjdk-openj9) have the same top level directory, so they cannot be unzipped side by side (and it is not obvious from the directory name which variant it is).

aahlenst:
What about compiling a list of requirements first? Like:

* Hotspot and OpenJ9 can be in the same directory
* OpenJDK 8..12 can be in the same directory
* Semver (e.g. 8.0.202) instead of upstream naming scheme (8u202)
* …

This should make it easier to evaluate each proposal for its (dis-)advantages.

sxa555:
[...]

PR #1764, opened 2020-05-22
Use JDK-Head version number instead of 'jdk'

M-Davies:
* Renames JDK-Next binaries to include the major version number
* I.e.`OpenJDK-jdk_x64_windows_hotspot_2020-03-27-03-31.zip` to `OpenJDK15U-jdk_x64_windows_hotspot_2020-03-27-03-31.zip`
* Includes various documentation and script updates to allow this to function

Ref: #1016 (Doesn't close it)

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -417,6 +417,12 @@ class Build {
 
         javaToBuild = javaToBuild.toUpperCase()
 
+	// Use version number for HEAD binaries
+	if (javaToBuild == "JDK") {
+		// This needs to be updated each time JDK-Head updates
+		javaToBuild = "15U"
+	}
+
         def fileName = "Open${javaToBuild}-jdk_${architecture}_${os}_${variant}"
 
         if (additionalFileNameTag) {


M-Davies:
# OFF SIGHT REQUIREMENTS FOR THIS
- Update names of downstream jdk-next jobs to be `jdk15u-blahblahblah` https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk/
- Update job regenerator to `pipeline_jobs_generator_jdk15u` https://ci.adoptopenjdk.net/job/build-scripts/job/utils/job/pipeline_jobs_generator_jdk/

pipelines/build/openjdk15_pipeline.groovy
```diff
@@ -12,7 +12,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-def javaToBuild = "jdk"
```
karianna, line 15: does jdk15u exist?

M-Davies: I'm not sure. I mostly just went off all of the other builds that use `jdk<version>u`. How can I find out?

karianna: https://github.com/AdoptOpenJDK/openjdk-jdk14u exists but https://github.com/AdoptOpenJDK/openjdk-jdk15u doesn't yet (the upstream mirror doesn't exist yet either)

M-Davies: Oh ok I see what you mean. The correct repo to use would be https://github.com/AdoptOpenJDK/openjdk-jdk. In that case, due to the tags, it's probably easier to just rename the binary instead of renaming the jdk version at start. I'll reset this branch and make the changes

M-Davies:
At this point, all I'm doing is renaming the binary to `OpenJDK15u blah blah`. Im not touching the actual `javaVersion` set by the pipeline

douph1:
as discussed here :
https://github.com/AdoptOpenJDK/openjdk-build/pull/1574

I don't think hard coded JDK-Latest to "15" like already here `getJavaVersionNumber()` or "15U" a second time is a good idea (maintainability point of view )

If you intended only to fix https://github.com/AdoptOpenJDK/openjdk-installer/issues/211 here I think I can better handle it in installer script because upstream job always pass the major version number.
So I can search for OpenJDK (without major number) and use it as JDK-Latest = PRODUCT_MAJOR_VERSION and not hardcoded any number either on my side or your side.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -417,6 +417,12 @@ class Build {
 
         javaToBuild = javaToBuild.toUpperCase()
 
+	// Use version number for HEAD binaries
+	if (javaToBuild == "JDK") {
+		// This needs to be updated each time JDK-Head updates
+		javaToBuild = "15U"
```
karianna, line 423: Should that be a lower case u?

M-Davies: Line 418 sets `javaToBuild` as uppercase before this `if`

M-Davies:
@douph1 Yes this was more to fix recent installer failures but if you think you can handle it on the installer side then I would prefer to do that. In addition, 1574 seems to be a better solution to this in the end than this since it avoids extra documentation when JDK-Head gets bumped.

PR #1766, opened 2020-05-22
Fake version.txt on cross compiles on RISC-V

sxa:
This isn't the best idea, but if it works to get green pipelines I'll take it for now until someone implementes a better solution. I'm not sure what the implications are of faking the metadata like this, but we're a little short of options. I suspect the best option would be to modify the code that reads it and handle an empty case.

[EDIT: The more i think about this the more I dislike it event as a temporary solution]

[@M-Davies @Willsparker If you can figure out how it's used and implement a better alternative that might be useful]


sxa:
@karianna Done

PR #1767, opened 2020-05-22
Use OpenJ9 x64 nightly as RISC-V build jdk

sxa:
Use an Openj9 nightly as the "build" (not boot) JDK for the RISC-V cross compilation process.


sxa:
Because the Build JDK is supposed to match the same source as the code you are compiling apparently. I figured potentially one day behind would be ok :-)

PR #1768, opened 2020-05-25
Pass timestamps and timeouts as a context

M-Davies:


sxa:
This will no longer apply cleanly as I [reverted the timestamp change](https://github.com/AdoptOpenJDK/openjdk-build/pull/1771).

M-Davies:
@sxa Resolved conflicts. To avoid the same problems as yesterday, is there anyway to test changes in the build-scripts before a merge? (I could look into implementing a system if it doesn't currently exist 🙂)

karianna:
> @sxa Resolved conflicts. To avoid the same problems as yesterday, is there any way to test changes in the build-scripts before a merge? (I could look into implementing a system if it doesn't currently exist 🙂)

We need to enable the pull request builder.  I'm going to try to resolve that with John tomorrow.

sxa:
@karianna Can you have a look at https://github.com/AdoptOpenJDK/openjdk-build/issues/1528 as part of that please as it has had a tendancy to chew up a lot of space on the build systems and leave stuff around

sxa:
(@M-Davies We can also test through [VagrantPlaybookCheck](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/) if required although that runs through a full ansible playbook run so is a bit slow!

sxa:
Running at https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk14-pipeline/183/ on three platforms

PR #1797, opened 2020-05-29
Rename riscv in pipelines to riscv64

sxa:
This should really have been riscv64 from the start as that's what `uname -m` is on the boxes, and it's a 64-bit build. Hopefully these are the only places it needs to be changed :-)


M-Davies:
This PR is fine but the changes might not be brought into the pipelines until https://github.com/AdoptOpenJDK/openjdk-build/pull/1792#issuecomment-636004346 is resolved

Issue #1820, opened 2020-06-04
jdknext windows build break: 05:36:06  Failed test: hudson.AbortException: Environment variable JAVA_FEATURE_VERSION not set


andrew-m-leonard:
**Platform:** jdk-windows-x64-openj9

https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk/job/jdk-windows-x64-openj9/3/console

M-Davies:
Caused by https://github.com/AdoptOpenJDK/openjdk-build/pull/1574. 

Seems like it's something wrong with [this line](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/make-adopt-build-farm.sh#L27)

M-Davies:
That line is ok after local testing and it [this line can be seen in the output](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/make-adopt-build-farm.sh#L52) so that means the groovy script can't see variables created in `make-adopt-build-farm.sh`. I'll have a look to see if there any variables set in jenkins that we can compare against instead

FYI @ShishirH

M-Davies:
Reverting the problem commit for now since it's a build break https://github.com/AdoptOpenJDK/openjdk-build/pull/1821

PR #1822, opened 2020-06-04
Query the adopt api to get the JDK-Head version

M-Davies:
Uses the info/available_releases/ endpoint to obtain the JDK-HEAD version number

Closes: #1820 

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -210,10 +210,10 @@ class Builder implements Serializable {
         } else if ("jdk".equalsIgnoreCase(javaToBuild.trim())) {
             // Query the Adopt api to get the "most_recent_feature_release" (currently 15)
             def JobHelper = context.library(identifier: 'openjdk-jenkins-helper@master').JobHelper
-            context.println "Querying Adopt Api for the JDK-Head number (most_recent_feature_version)..."
+            context.println "Querying Adopt Api for the JDK-Head number (most_recent_feature_release)..."
 
             response = JobHelper.getAvailableReleases()
-            Integer headVersion = Integer.valueOf(response.most_recent_feature_version)
+            Integer headVersion = Integer.valueOf(response.most_recent_feature_release)
             if (headVersion.equalsIgnoreCase(null)) {
                 context.println "Failure on api connection or parsing."
                 throw new Exception()
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -1,6 +1,6 @@
 @Library('local-lib@master')
 import common.IndividualBuildConfig
-import groovy.json.JsonSlurper
+import groovy.json.*
 
 import java.util.regex.Matcher
 
@@ -208,13 +208,18 @@ class Builder implements Serializable {
         if (matcher.matches()) {
             return Integer.parseInt(matcher.group('version'))
         } else if ("jdk".equalsIgnoreCase(javaToBuild.trim())) {
-            String javaFeatureVersion = System.getenv("JAVA_FEATURE_VERSION")
-            if (javaFeatureVersion) {
-                return Integer.valueOf(javaFeatureVersion)
-            } else {
-                context.error("Environment variable JAVA_FEATURE_VERSION not set")
+            // Query the Adopt api to get the "most_recent_feature_release" (currently 15)
+            def JobHelper = context.library(identifier: 'openjdk-jenkins-helper@master').JobHelper
+            context.println "Querying Adopt Api for the JDK-Head number (most_recent_feature_version)..."
+
+            response = JobHelper.getAvailableReleases()
+            Integer headVersion = Integer.valueOf(response.most_recent_feature_version)
+            if (headVersion.equalsIgnoreCase(null)) {
+                context.println "Failure on api connection or parsing."
                 throw new Exception()
             }
+            context.println "Found Java Version Number: ${headVersion}"
+            return headVersion
         } else {
             context.error("Failed to read java version '${javaToBuild}'")
             throw new Exception()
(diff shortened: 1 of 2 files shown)


M-Davies:
This is currently **BLOCKED** by https://github.com/AdoptOpenJDK/openjdk-api-v3/pull/218.

M-Davies:
@sxa In the api query https://api.adoptopenjdk.net/v3/info/available_releases, we could alternatively use `most_recent_feature_release` instead of the in-progress `most_recent_feature_version` as they currently produce the same version number. 

You can see the schema for both at https://github.com/AdoptOpenJDK/openjdk-api-v3/pull/218/files. I doubt we will need a version number if we don't produce a build with it but I'd like to get your opinion on this before changing this PR

M-Davies:
Code works as it should in a test instance I created on my machine:
```groovy
import groovy.json.*
import java.util.regex.Matcher

Integer getJavaVersionNumber() {
        // version should be something like "jdk8u" or "jdk" for HEAD
        Matcher matcher = javaToBuild =~ /.*?(?<version>\d+).*?/
        if (matcher.matches()) {
            return Integer.parseInt(matcher.group('version'))
        } else if ("jdk".equalsIgnoreCase(javaToBuild.trim())) {
            // Query the Adopt api to get the "most_recent_feature_release" (currently 15)
            String releases = "https://api.adoptopenjdk.net/v3/info/available_releases"
            Integer headVersion

            println "Querying Adopt Api for the JDK-Head number: ${releases}..."
            try {
                def get = new URL(releases).openConnection()
                def parser = new JsonSlurper()
                def response = parser.parseText(get.getInputStream().getText())
                headVersion = Integer.valueOf(response.most_recent_feature_release)
            } catch (Exception e) {
                println "Failure on api connection or parsing.\n${e}"
                throw new Exception()
            }
            println "Found Java Version Number: ${headVersion}"
            return headVersion
        } else {
            context.error("Failed to read java version '${javaToBuild}'")
            throw new Exception()
        }
}

javaToBuild = "jdk"
print(getJavaVersionNumber())
```
Output: 
Screenshot 2020-06-04 at 14 11 43
NOTE: This will require sudo (Script approval) to run the GET request otherwise we get a 403 error similar to this 
```java
java.io.IOException: Server returned HTTP response code: 403 for URL: https://api.adoptopenjdk.net/v3/info/available_releases
Caught: java.lang.Exception
java.lang.Exception
        at test.getJavaVersionNumber(test.groovy:23)
        at test.run(test.groovy:34)
```

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -62,13 +62,22 @@ class Build {
         if (matcher.matches()) {
             return Integer.parseInt(matcher.group('version'))
         } else if ("jdk".equalsIgnoreCase(javaToBuild.trim())) {
-            String javaFeatureVersion = System.getenv("JAVA_FEATURE_VERSION")
-            if (javaFeatureVersion) {
-                return Integer.valueOf(javaFeatureVersion)
-            } else {
-                context.error("Environment variable JAVA_FEATURE_VERSION not set")
+            // Query the Adopt api to get the "most_recent_feature_release" (currently 15)
```
johnoliver, line 65: This seems wrong, would this not mean that all builds would run tests on 15/the latest version?

johnoliver:
If you are having issues with permissions, try creating a method to do what you need in https://github.com/AdoptOpenJDK/openjdk-jenkins-helper

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies: Yes if the tests were enabled. We don't run tests for JDK-Next currently so there isn't any history for what would happen if we did run tests with this but it should be fine as they will expect JDK-Next binaries to be whatever the latest jdk-head number is (unless we only want to work with GA releases as described here https://github.com/AdoptOpenJDK/openjdk-build/pull/1822#issuecomment-638767984)

M-Davies:
> If you are having issues with permissions, try creating a method to do what you need in https://github.com/AdoptOpenJDK/openjdk-jenkins-helper

Do we use the job helper much anymore? I brought this up with @karianna before but I was curious if it was actually used outside of [this case](https://github.com/AdoptOpenJDK/openjdk-build/issues/1013#issuecomment-630689667). It seems like most of the functionality in it is replicated with the Jenkins API so we would just duplicating effort by using it?

M-Davies:
Looks like this will require higher perms https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk-pipeline-morgan-test/3/console
@karianna Can you do the honours when/if this is merged?

karianna:
> Looks like this will require higher perms https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk-pipeline-morgan-test/3/console
> @karianna Can you do the honours when/if this is merged?

Yep happy to do so when this lands - ping me on slack tomorrow.

M-Davies:
Note this PR is now dependant on https://github.com/AdoptOpenJDK/openjdk-jenkins-helper/pull/31

M-Davies:
In the absence of a reply, I've altered the PR to use the existing `most_recent_feature_release` since I doubt we will need the version number of something we don't produce a release for. This is ready for re-review

M-Davies:
Running a quick test job before the nightlies kick in https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk15-pipeline/151/console

pipelines/build/common/build_base_file.groovy
```diff
@@ -208,13 +208,18 @@ class Builder implements Serializable {
         if (matcher.matches()) {
             return Integer.parseInt(matcher.group('version'))
         } else if ("jdk".equalsIgnoreCase(javaToBuild.trim())) {
-            String javaFeatureVersion = System.getenv("JAVA_FEATURE_VERSION")
-            if (javaFeatureVersion) {
-                return Integer.valueOf(javaFeatureVersion)
-            } else {
-                context.error("Environment variable JAVA_FEATURE_VERSION not set")
+            // Query the Adopt api to get the "most_recent_feature_release" (currently 15)
+            def JobHelper = context.library(identifier: 'openjdk-jenkins-helper@master').JobHelper
```
johnoliver, line 212: for future reference can simplify this with 

```
@Library('openjdk-jenkins-helper@master')
import JobHelper
```

At the top of the file

PR #1827, opened 2020-06-05
Fail build if valid boot dir isn't found

M-Davies:
Closes: https://github.com/AdoptOpenJDK/openjdk-build/issues/1826


M-Davies:
> Hmm, this is the first time we've exit 1'd out of the script. I'm not sure this would interact cleanly with the pipeline status etc. Is there a way we can test this?

I could just exit with a different code or use a different function? I'm open to ideas as my bash scripting skills are a little rudimentary

karianna:
I trust Stewart ;-)

Issue #1846, opened 2020-06-10
Use API to get the JDK-Head number


M-Davies:
Now that we can obtain the JDK-Head number (via [most_recent_feature_version](https://api.adoptopenjdk.net/v3/info/available_releases) using the API), we can look into replacing places in the code where jdk next numbers are hardcoded with an api call.
E.g.
-  https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/make-adopt-build-farm.sh#L32
- https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/common/common.sh#L38

PR #1849, opened 2020-06-10
Remove Hard coded JDK-Next Values

M-Davies:
* Queries the adopt api to obtain the most_recent_feature_version

Closes: #1846 

### DIFF
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -28,8 +28,18 @@ JAVA_FEATURE_VERSION=$(echo "${JAVA_TO_BUILD}" | tr -d "[:alpha:]")
 
 if [ -z "${JAVA_FEATURE_VERSION}" ]
 then
-    # THIS NEEDS TO BE UPDATED WHEN HEAD UPDATES (the latest tag that jdk/jdk contains)
-    JAVA_FEATURE_VERSION=15
+  # Use Adopt API to get the JDK Head number
+  echo "This appears to be JDK Head. Querying the Adopt API to get the JDK HEAD Number (https://api.adoptopenjdk.net/v3/info/available_releases)..."
+  JAVA_FEATURE_VERSION=$(curl https://api.adoptopenjdk.net/v3/info/available_releases | jq '.most_recent_feature_version')
+    
+  # Checks the api request was successfull and the return value is a number
+  isNum='^[0-9]+$'
+  if [ -z "${JAVA_FEATURE_VERSION}" ] || ! [[ $JAVA_FEATURE_VERSION =~ $isNum ]]
+  then
+    echo "Failed to query or parse the adopt api"
+    exit 1
+  fi
+  echo "JAVA_FEATURE_VERSION is ${JAVA_FEATURE_VERSION}"
 fi
 
 echo "BUILD TYPE: "
--- a/sbin/common/common.sh
+++ b/sbin/common/common.sh
@@ -34,8 +34,24 @@ function setOpenJdkVersion() {
 
   local featureNumber=$(echo "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" | tr -d "[:alpha:]")
 
+  if [ -z "${featureNumber}" ]
+  then
+    # Use Adopt API to get the JDK Head number
+    echo "This appears to be JDK Head. Querying the Adopt API to get the JDK HEAD Number (https://api.adoptopenjdk.net/v3/info/available_releases)..."
+    local featureNumber=$(curl https://api.adoptopenjdk.net/v3/info/available_releases | jq '.most_recent_feature_version')
+    
+    # Checks the api request was successfull and the return value is a number
+    isNum='^[0-9]+$'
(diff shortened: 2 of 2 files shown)


build-farm/make-adopt-build-farm.sh
```diff
@@ -28,8 +28,18 @@ JAVA_FEATURE_VERSION=$(echo "${JAVA_TO_BUILD}" | tr -d "[:alpha:]")
 
 if [ -z "${JAVA_FEATURE_VERSION}" ]
 then
-    # THIS NEEDS TO BE UPDATED WHEN HEAD UPDATES (the latest tag that jdk/jdk contains)
-    JAVA_FEATURE_VERSION=15
+  # Use Adopt API to get the JDK Head number
+  echo "This appears to be JDK Head. Querying the Adopt API to get the JDK HEAD Number (https://api.adoptopenjdk.net/v3/info/available_releases)..."
+  JAVA_FEATURE_VERSION=$(curl https://api.adoptopenjdk.net/v3/info/available_releases | jq '.most_recent_feature_version')
```
karianna, line 33: Is that call correct?  Is it returning 15?  There's also the tip_version API call.

sbin/common/common.sh
```diff
@@ -34,8 +34,24 @@ function setOpenJdkVersion() {
 
   local featureNumber=$(echo "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" | tr -d "[:alpha:]")
 
+  if [ -z "${featureNumber}" ]
+  then
+    # Use Adopt API to get the JDK Head number
+    echo "This appears to be JDK Head. Querying the Adopt API to get the JDK HEAD Number (https://api.adoptopenjdk.net/v3/info/available_releases)..."
+    local featureNumber=$(curl https://api.adoptopenjdk.net/v3/info/available_releases | jq '.most_recent_feature_version')
```
karianna, line 41: Same as above

build-farm/make-adopt-build-farm.sh
```diff
@@ -28,8 +28,18 @@ JAVA_FEATURE_VERSION=$(echo "${JAVA_TO_BUILD}" | tr -d "[:alpha:]")
...
+    
+  # Checks the api request was successfull and the return value is a number
+  isNum='^[0-9]+$'
+  if [ -z "${JAVA_FEATURE_VERSION}" ] || ! [[ $JAVA_FEATURE_VERSION =~ $isNum ]]
```
karianna, line 37: Is that if bracketing correct for bash?

sbin/common/common.sh
M-Davies: Reply above

build-farm/make-adopt-build-farm.sh
M-Davies: The call returns 15 as it's the latest JDK head number. What do you mean by tip_version API call?

build-farm/make-adopt-build-farm.sh
M-Davies: It has to be double bracketed as regex's don't seem to work in singular ones (tested locally on mac)

build-farm/make-adopt-build-farm.sh
gdams: @M-Davies Martijn is referring to the tip_version call in the API response.

```json
{
"available_lts_releases": [
8,
11
],
"available_releases": [
8,
9,
10,
11,
12,
13,
14
],
"most_recent_feature_release": 14,
"most_recent_feature_version": 15,
"most_recent_lts": 11,
"tip_version": 15
}
```

`most_recent_feature_version` is the most recent version that we have built whereas tip_version is the definitive version of tip. In principle they will nearly always be the same but when we switch to 16 `most_recent_feature_version` will remain at 15 until we've created a nightly/release

M-Davies: Ah I see. I thought they were the other way around. In that case, I'll go back and amend the other places where I'm using `most_recent_feature_version`

gdams:
is there a reason that the code is duplicated in both `make-adopt-build-farm.sh` and `common.sh`? could you not create a function in common.sh and use it in both places?

M-Davies:
> is there a reason that the code is duplicated in both `make-adopt-build-farm.sh` and `common.sh`? could you not create a function in common.sh and use it in both places?
> 
I'm not sure if `make-adopt-build-farm.sh` can access the functions declared in `common.sh`

M-Davies:
> but we should also check that it's not blank after this
> 
@sxa Can you clarify this? We have an if check after the command to ensure the value isn't null and is a number

build-farm/make-adopt-build-farm.sh
sxa: You cold possibly get away with

`if [ "${JAVA_FEATURE_VERSION}" -gt 0 ]`

Which would fail one way or another on all the situations we're trying to trap

sxa:
Yep that logic is probably ok (although I'm not an expert on regex in bash as I tend not to use it!) but I've added an alternative that's potentially simpler code-wise.

PR #1865, opened 2020-06-11
Switch JDK15 Linux/x64 builds to CentOS7/RHEL7

sxa:
Part of https://github.com/AdoptOpenJDK/openjdk-build/issues/1819 - seems reasonable to switch the latest version over first... This will obviously affect minimum OS levels (although it will bring them in line with Linux on other architectures).

I have a GoDaddy RHEL7 box earmarked for build that's been used for some other jobs that can be used for this.

### DIFF
--- a/pipelines/jobs/configurations/jdk15_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk15_pipeline_config.groovy
@@ -14,7 +14,7 @@ class Config15 {
         x64Linux  : [
                 os                  : 'linux',
                 arch                : 'x64',
-                additionalNodeLabels: 'centos6',
+                additionalNodeLabels: 'rhel7',
                 test                : [
                         nightly: false,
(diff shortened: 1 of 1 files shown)


pipelines/jobs/configurations/jdk15_pipeline_config.groovy
```diff
@@ -14,7 +14,7 @@ class Config15 {
         x64Linux  : [
                 os                  : 'linux',
                 arch                : 'x64',
-                additionalNodeLabels: 'centos6',
+                additionalNodeLabels: 'rhel',
```
gdams, line 17: probably needs to be `rhel7` to ensure that it's locked to 7.

sxa: Cut & paster error - thanks :-) (`centos7||rhel7` would e best if it would work I guess, but the ones I have are rhel7 so that'll do for now)

sxa:
Approved in principle by two TSC members in slack so we can give this a go and see what happens - removing from draft;-)

PR #1871, opened 2020-06-12
Trial copyArtifactPermission

M-Davies:
* copyArtifactPermissionProperty might be deprecated in prod mode

Closes: https://github.com/AdoptOpenJDK/openjdk-build/issues/1879

### DIFF
--- a/pipelines/build/common/create_job_from_template.groovy
+++ b/pipelines/build/common/create_job_from_template.groovy
@@ -50,7 +50,7 @@ pipelineJob("$buildFolder/$JOB_NAME") {
     }
     properties {
 	disableConcurrentBuilds()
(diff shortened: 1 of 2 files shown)


M-Davies:
Draft for now while we migrate the JDK-Head https://github.com/AdoptOpenJDK/openjdk-build/pull/1870

karianna:
It's worth a try although we won't know if it's the right call until it gets deployed.

sxa:
I'm going to let this through and see what happens then back it out if it breaks things.
JDK11 pipeline executing with restricted platforms (though including all windows) at https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk11-pipeline/1006/

PR #1901, opened 2020-06-17
Use OpenJ9 as boot JDK if downloading one during OpenJ9 build

sxa:
This will resolve the symbol issue affecting latest OpenJ9 builds until OpenJ9 have merged their latest changes. Note that this won't take effect if:
1. There is a JDKxx_BOOT_DIR pointing to a valid location in the machine definition
2. There is already a suitable one in the workspace (So they will need to be cleared for this to work as they currently have cached HotSpot ones)


sxa:
Alternate fix already in as https://github.com/AdoptOpenJDK/openjdk-build/pull/1899/files (although mine is slicker ... meh)

gdams:
@sxa I actually prefer your solution... A) It's less code, B) It's more consistent.

Reopening to get other people's opinions

karianna:
won't this pick OpenJ9 for OpenJ9 builds regardless of jdk 15+?

gdams:
> won't this pick OpenJ9 for OpenJ9 builds regardless of jdk 15+?

@karianna it will but this shouldn't break anything. I prefer consistency where possible and I would argue that building OpenJ9 with an OpenJ9 boot JDK makes more sense IMO

karianna:
We'd need to make that change everywhere though no?  Probably should get
the OpenJ9 folks to weigh in here.

Cheers,
Martijn

sxa:
The OpenJ9 folks have generally had a preference for using an OpenJ9 boot VM ...
It's also a temporary solution in order to get past the immediate problem where HotSpot can't be used as the OpenJ9 build JDK at the moment and it will be resolved (We could reverse this later, but the auto-download is a backstop anyway...)

We normally have the boot JDKs installed on the machine (and normally it's only Hotspot) and I don't really want to change that going forward.

PR #1911, opened 2020-06-18
docker: Update buildDocker.sh to use new eclipse dockerfile script

Willsparker:
This PR is to allow for Eclipse's `mkDocker.sh` script ( https://github.com/eclipse/openj9/pull/9371 ) to be tested under the `DockerfileCheck` [Jenkins Job](https://ci.adoptopenjdk.net/job/DockerfileCheck/)

Currently there are some concerns that the `git gc --agressive --prune=all` command will fail when building the image due to the command taking up to 8GB of Memory (https://adoptopenjdk.slack.com/archives/C09NW3L2J/p1592317252094400 ). However, this will depend on a change in the Eclipse/OpenJ9 repository. Once that's fixed, the `DockerfileCheck` job should work for the Eclipse Dockerfiles. 

The `cleanWorkspace` option has been put in so the `DockerfileCheck` job won't use old dockerfiles or create multiple copies of `mkDocker.sh` when running with the Eclipse option.

### DIFF
--- a/docker/buildDocker.sh
+++ b/docker/buildDocker.sh
@@ -5,7 +5,7 @@ jdkVersion=''
 bootDir=''
 buildVariant="hotspot"
 useEclipseDockerFiles=false
-useEclipseDockerSlavesFiles=false
+cleanWorkspace=false
 
 #takes in all arguments to determine script options
 parseCommandLineArgs()
@@ -21,6 +21,8 @@ parseCommandLineArgs()
 			case "$opt" in
 				"--all" | "-a" )
 					jdkVersion="all";;
+				"--clean" | "-c" )
+					cleanWorkspace=true;;
 				"--version" | "-v" )
 					jdkVersion="$1"; shift;;
 				"--jdk-boot-dir" | "-J")
@@ -29,8 +31,6 @@ parseCommandLineArgs()
 					buildVariant="openj9";;
 				"--use-eclipse-docker-files" | "-e" )
 					useEclipseDockerFiles=true; buildVariant="eclipse";;
-				"--use-eclipse-docker-slave-files" | "-es" )
-					useEclipseDockerSlavesFiles=true; buildVariant="eclipse_slave";;
 				"--help" | "-h" )
 					usage; exit 0;;
 				*) echo >&2 "Invalid option: ${opt}"; echo "This option was unrecognised."; usage; exit 1;;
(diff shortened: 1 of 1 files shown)


docker/buildDocker.sh
```diff
@@ -99,51 +99,46 @@ checkArgs()
...
-		echo "Unable to use both kinds of dockerfiles at once, Select a single option."
-		exit 1
-        fi
+	if [ "$cleanWorkspace" == true ]; then
+		echo "Cleaning workspace"
+		rm -rf $WORKSPACE/DockerBuildFolder/$jdkVersion-$buildVariant
+	fi
 }
 
 useEclipseDockerFiles()
```
karianna, line 108: I wonder if that needs to be called eclipseOpenJ9DockerFiles throughout the script.  I know it seems overkill, but I thought initially that Eclipse had some special docker base-files for all Java  variants (as opposed to OpenJ9)

docker/buildDocker.sh
```diff
@@ -99,51 +99,46 @@ checkArgs()
...
-	cd $WORKSPACE/DockerBuildFolder/$jdkVersion-$buildVariant/buildenv/jenkins/docker-slaves/x86/centos6.9
-	if [ -f "known_hosts" ]; then 
-		rm known_hosts
-	fi
+	mkdir -p $WORKSPACE/DockerBuildFolder/$jdkVersion-$buildVariant/EclipseDockerfiles
+	cd $WORKSPACE/DockerBuildFolder/$jdkVersion-$buildVariant/EclipseDockerfiles
+	wget https://raw.githubusercontent.com/eclipse/openj9/master/buildenv/docker/mkdocker.sh
+	chmod +x mkdocker.sh
+	# Generate an Ubuntu1804 Dockerfile using mkdocker.sh
+	$WORKSPACE/DockerBuildFolder/$jdkVersion-$buildVariant/EclipseDockerfiles/mkdocker.sh --dist=ubuntu --version=18 --print >> $WORKSPACE/DockerBuildFolder/$jdkVersion-$buildVariant/EclipseDockerfiles/Dockerfile
```
karianna, line 115: Do we need to parameterize that 18?

docker/buildDocker.sh
```diff
@@ -99,51 +99,46 @@ checkArgs()
...
-		sharedEclipseDockerCommands $jdk
+		if [ ${jdk} != "jdk" ]; then
+			jdk=${jdk%?}
+		fi
+		manualDockerCommands ${jdk}
 	done
 }
 
-sharedEclipseDockerCommands()
+manualDockerCommands()
```
karianna, line 132: Are they manual commands?  Looks like we're running them automated...

docker/buildDocker.sh
Willsparker: Happy to make the change, if it can be clearer, it should be :-)

docker/buildDocker.sh
Willsparker: For now, I'd rather not. While we _can_ make dockerfiles with different distributions and versions with OpenJ9's `mkDocker.sh` script, to keep it consistent with our own `dockerfile-generator.sh` script and for the purposes of `DockerfileCheck`, I think they should be all on Ubuntu1804. Extended `dockerfile-generator.sh` would certainly be a good future enhancement however

docker/buildDocker.sh
Willsparker: haha, good point. I'll rename it- by 'manual', I meant 'not using makejdk-any-platform to create the docker container'.

Issue #1745, opened 2020-05-15
When executing a docker build I'm not guided to where I get the resulting binary


karianna:
On Mac OS X 10.15.4 with the latest Docker client for Mac.

I run:

`./makejdk-any-platform.sh -c --docker --sudo --jdk-boot-dir /Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home/ jdk8u`

It builds successfully and toward the end it states:

```
Archiving the build OpenJDK image and compressing with gzip
OpenJDK JDK path will be jdk8u262-b02. JRE path will be jdk8u262-b02-jre
Archiving the build OpenJDK image and compressing with gzip
Your final archive was created at /openjdk/build/src/build/linux-x86_64-normal-server-release/images/OpenJDK.tar.gz
Moving the artifact to /openjdk//target/
Archiving the build OpenJDK image and compressing with gzip
Your final archive was created at /openjdk/build/src/build/linux-x86_64-normal-server-release/images/OpenJDK.tar.gz
Moving the artifact to /openjdk//target/
All done!
Removing container jdk8-hotspot
```

It's not clear if `/openjdk//target/` is inside the container or on the host.  Mac OS X 10.15.4 doesn't allow writing of directories to `/`, so if it attempted to write it there then it should fail.  If this is inside the docker container, I'd like some guidance on how to get it out.

M-Davies:
I get a File not found when trying to run that on my machine (Mac 10.15.5)
```
➜  openjdk-build git:(master) ✗ ./makejdk-any-platform.sh -c --docker --sudo --jdk-boot-dir /Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home/ jdk8u
Starting ./makejdk-any-platform.sh to configure, build (Adopt)OpenJDK binary
Parsing opt: -c
Possible opt arg: --docker
Parsing opt: --docker
Possible opt arg: --sudo
Parsing opt: --sudo
Possible opt arg: --jdk-boot-dir
Parsing opt: --jdk-boot-dir
Possible opt arg: /Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home/
Working dir is ./build/
[debug] COPY_MACOSX_FREE_FONT_LIB_FOR_JDK_FLAG=true
[debug] COPY_MACOSX_FREE_FONT_LIB_FOR_JRE_FLAG=true
[...]

PR #1912, opened 2020-06-18
docker: Fixes to get `dockerfile-generator.sh` working with `makejdk-any-platform.sh --docker`

Willsparker:
Ref: https://ci.adoptopenjdk.net/job/DockerfileCheck/121/

Signed-off-by: William Parker william.parker1@ibm.com

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -83,7 +83,7 @@ buildOpenJDKViaDocker()
     build_variant_flag="--openj9"
   fi
 
-  docker/dockerfile-generator.sh --jdk "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" --path "${BUILD_CONFIG[DOCKER_FILE_PATH]}" "$build_variant_flag"
+  docker/dockerfile-generator.sh --version "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" --path "${BUILD_CONFIG[DOCKER_FILE_PATH]}" "$build_variant_flag"
   # shellcheck disable=SC1090
   source "${BUILD_CONFIG[DOCKER_FILE_PATH]}/dockerConfiguration.sh"
 
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -82,8 +82,8 @@ buildOpenJDKViaDocker()
   if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "openj9" ]; then
     build_variant_flag="--openj9"
   fi
-
   docker/dockerfile-generator.sh --version "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" --path "${BUILD_CONFIG[DOCKER_FILE_PATH]}" "$build_variant_flag"
+
   # shellcheck disable=SC1090
   source "${BUILD_CONFIG[DOCKER_FILE_PATH]}/dockerConfiguration.sh"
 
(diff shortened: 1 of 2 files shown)


Willsparker:
With latest commit : 
- Added a message to tell where the built JDK is on the host system. Fixes: #1745 
- Added some packages to the apt list
- Made OpenJ9 use GCC-7.5, not 7.3
- Make sure that when building `jdk`, use the latest general availability release. (Noticed this was an issue as 'jdk' refers to (what will be) jdk16 now, however jdk15 is not yet out and therefore can't be retrieved from the api)
- Stopped a blank argument being passed to `dockerfile-generator.sh` from `docker-build.sh`, when building a hotspot JDK. Side benefit of it being more clear to read.
- Rearranged the `user` section of the Dockerfile so the `WORKDIR` is created before the `chown` command is ran, as that was causing permission issues when executing a build.

I'm having a bit of an issue with specifically JDK11-J9. For some reason, there are cases where JDK11 is unable to pickup that `libdwarf-dev` and `libZ` is on the system, however JDK14/J9 and JDK8/J9 don't have the same issues. e.g: https://ci.adoptopenjdk.net/job/DockerfileCheck/130/

Related: #1911

docker-build.sh
```diff
@@ -76,14 +76,14 @@ buildOpenJDKViaDocker()
...
   # architectures going forwards
   local container_architecture="x86_64/ubuntu"
-  local build_variant_flag=""
   BUILD_CONFIG[DOCKER_FILE_PATH]="docker/${BUILD_CONFIG[OPENJDK_CORE_VERSION]}/$container_architecture"
   
   if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "openj9" ]; then
-    build_variant_flag="--openj9"
+    docker/dockerfile-generator.sh --version "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" --path "${BUILD_CONFIG[DOCKER_FILE_PATH]}" --openj9
+  else
+    docker/dockerfile-generator.sh --version "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" --path "${BUILD_CONFIG[DOCKER_FILE_PATH]}"
```
karianna, line 84: Looks like you could just append the --openj9 in an if else block as opposed to repeating the whole line?

Willsparker: Hmm, I think I may have a way of doing it as it was before (i.e. using `build_variant_flag` ), which would stop the repetition. I did change it as the `dockerfile-generation.sh` script was complaining about having a blank argument passed to it (see the hotspot builds here: https://ci.adoptopenjdk.net/job/DockerfileCheck/125/ ), however I think I've found a solution to this :-)

Willsparker:
@karianna I did see that, I'm a bit unsure what it means by 'useless cat' - I can't remove the `cat` command otherwise it complains that `available_release` isn't a command. I suppose it could mean do something like `wget -q $URL -O - | grep ..etc`, but as I'm accessing the file twice for 2 different variables, it'd be inefficient to `wget` it twice.

karianna:
Yeah, I think your approach is actually cleaner - is there a command we can give to tell Codacy to ignore this next time?

Willsparker:
Hmm, I think I can put this in to the function
`# shellcheck disable=SC2002`
https://github.com/koalaman/shellcheck/wiki/Ignore
https://github.com/koalaman/shellcheck/wiki/SC2002

Willsparker:
Hmm, it appears Codacy doesn't like that. I think you can add the file to an `ignore` list, but I think the owner of the Codacy account has to do that?

karianna:
Hmm, do you know if we let this through whether it gets pulled up every time?

Willsparker:
I don't think so - If you click on the `files` tab in the Codacy run thing, it shows a list of files in the master branch, and some other scripts have issues it seems (i.e. apparently `mercurial-tags/java-tool/MercurialTracker.java` has 37 codacy issues )

Issue #1914, opened 2020-06-19
MissingContextVariableException: Required context class hudson.FilePath is missing


archanaNogriya:
**Platform:**
jdk-mac-x64-hotspot
jdk14u-windows-x64-hotspot
jdk14u-windows-x64-openj9
jdk8u-aix-ppc64-hotspot

**Architecture:**

Please describe the issue that you are having with our binaries and if possible how to recreate it.

https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk/job/jdk-mac-x64-hotspot/157/console

There is not much error except this.

`org.jenkinsci.plugins.workflow.steps.MissingContextVariableException: Required context class hudson.FilePath is missing
Perhaps you forgot to surround the code with a step that provides this, such as: node`

archanaNogriya:
#1913 Print out file perms on build fail

PR #1915, opened 2020-06-19
Revert "Print out file perms on build fail (#1913)"

M-Davies:
This reverts #1913 
Closes: #1914


M-Davies:
Reverting until we can think of some other way to print out the perms

Issue #1878, opened 2020-06-11
file permissions Issue - windows


archanaNogriya:
**jdk11u-windows-x64-openj9**
https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-windows-x64-openj9/630/console

`Execution error: Unable to get file permissions`

sxa:
Potential duplicate of https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1360 so marking as duplicate but leaving open until verified (https://github.com/AdoptOpenJDK/openjdk-build/pull/1871 is a potential fix)

M-Davies:
https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x86-32-openj9/596/console
```
19:43:49  Starting building: build-scripts » release » create_installer_windows #4760
[Pipeline] copyArtifacts
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // node
[Pipeline] echo
19:48:08  Execution error: Unable to get file permissions
```

M-Davies:
https://github.com/AdoptOpenJDK/openjdk-build/pull/1896/files might print out the stacktrace for us

M-Davies:
https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x64-hotspot/648/console
```
19:02:41  Scheduling project: build-scripts » release » create_installer_windows
19:02:47  Starting building: build-scripts » release » create_installer_windows #4841
[Pipeline] copyArtifacts
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // node
[Pipeline] echo
19:07:20  Execution error: hudson.os.PosixException: Unable to get file permissions null
```

archanaNogriya:
https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x64-openj9-windowsXL/330/console

```
22:01:43  Scheduling project: build-scripts » release » create_installer_windows
22:12:53  Starting building: build-scripts » release » create_installer_windows #4847
[Pipeline] copyArtifacts
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[...]

PR #1916, opened 2020-06-19
Try printing file perms outside of script tag

M-Davies:
Issue: #1878 


M-Davies:
Draft until monday so we don't screw up the weekend nightlies

M-Davies:
I realise that this actually won't work so as we're running on the jenkins master at this point which doesn't allow printing of directories for security reasons

PR #1935, opened 2020-06-26
switch linux x64 builds to use docker

gdams:
Swaps all <15 hotspot builds to use the centos6 build image (https://hub.docker.com/r/adoptopenjdk/centos6_build_image)

Also adds codebuild machines to handle the docker images (saves us maintaining any docker only machines.

This PR technically allows us to use a docker build environment for any architecture that supports docker.

I'm currently not doing this on OpenJ9 because it requires CUDA which bloats the docker image substantially, I'm working on a solution for that.

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -75,6 +75,7 @@ class Builder implements Serializable {
                 SCM_REF: scmReference,
                 BUILD_ARGS: buildArgs,
                 NODE_LABEL: "${additionalNodeLabels}&&${platformConfig.os}&&${platformConfig.arch}",
+                DOCKER_IMAGE: platformConfig.dockerImage as String,
                 CONFIGURE_ARGS: getConfigureArgs(platformConfig, additionalConfigureArgs, variant),
                 OVERRIDE_FILE_NAME_VERSION: overrideFileNameVersion,
                 ADDITIONAL_FILE_NAME_TAG: platformConfig.additionalFileNameTag as String,
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -58,6 +58,8 @@ class Builder implements Serializable {
 
         def additionalNodeLabels = formAdditionalBuildNodeLabels(platformConfig, variant)
 
+        def dockerImage = getDockerImage(platformConfig, variant)
+
         def buildArgs = getBuildArgs(platformConfig, variant)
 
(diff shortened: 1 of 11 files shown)


gdams:
blocked by https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1418

pipelines/build/common/build_base_file.groovy
```diff
@@ -75,6 +77,8 @@ class Builder implements Serializable {
                 SCM_REF: scmReference,
                 BUILD_ARGS: buildArgs,
                 NODE_LABEL: "${additionalNodeLabels}&&${platformConfig.os}&&${platformConfig.arch}",
+                CODEBUILD: platformConfig.codebuild as Boolean,
```
johnoliver, line 80: I take it codebuild == true and no docker image provided is an invalid config? Worth doing a sanity check on that

gdams: if codebuild == true and no docker image is provided it will already skip and continue as usual. e.g it won't break anything

gdams:
> https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/openjdk-build-test-pr/375/console

Those failures are not related to my PR, they came in a recent PR which @douph1 is fixing I think

keithc-ca:
> I'm currently not doing this on OpenJ9 because it requires CUDA which bloats the docker image substantially, I'm working on a solution for that.

See the `--cuda` option of [mkdocker.sh](https://github.com/eclipse/openj9/blob/master/buildenv/docker/mkdocker.sh#L586) which adds about 20MB to the images instead of more than 1GB if built on top of `nvidia/cuda:9.0-devel-ubuntu16.04`, for example.

gdams:
> See the --cuda option of mkdocker.sh which adds about 20MB to the images instead of more than 1GB if built on top of nvidia/cuda:9.0-devel-ubuntu16.04, for example.

@keithc-ca is there a way of modifying our ansible cuda role to simply do that?

Alternatively, I can add that directly to our dockerfile here (https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/master/ansible/Dockerfile.CentOS6) but I'mn trying to avoid adding dependencies directly there.

douph1:
> > https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/openjdk-build-test-pr/375/console
> 
> Those failures are not related to my PR, they came in a recent PR which @douph1 is fixing I think

@gdams I really don't know what it is about ..

PR #1941, opened 2020-06-29
add config arguments for adopt build repo uri and branch

jennli:
Introducing 2 additional config arguments to `sbin/prepareWorkspace.sh`, for other vendors to have the option to build openjdk with different openjdk-build source code.
`OPENJDK_BUILD_REPO_BRANCH`
and
`OPENJDK_BUILD_REPO_URI`

### DIFF
--- a/sbin/common/config_init.sh
+++ b/sbin/common/config_init.sh
@@ -32,6 +32,8 @@
 # map. This is why we can't have nice things.
 CONFIG_PARAMS=(
 ADOPT_PATCHES
+ADOPTOPENJDK_BUILD_REPO_BRANCH
+ADOPTOPENJDK_BUILD_REPO_URI
 BRANCH
 BUILD_FULL_NAME
 BUILD_VARIANT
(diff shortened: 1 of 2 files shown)


sbin/prepareWorkspace.sh
```diff
@@ -469,10 +469,10 @@ checkingAndDownloadCaCerts() {
     downloadCerts "$caLink"
   elif [ "${BUILD_CONFIG[USE_JEP319_CERTS]}" != "true" ]; then
     git init
-    git remote add origin -f https://github.com/AdoptOpenJDK/openjdk-build.git
+    git remote add origin -f "${BUILD_CONFIG[ADOPTOPENJDK_BUILD_REPO_URI]}"
```
jennli, line 472: so we can build openjdk with our own mirrored openjdk-build source code

sbin/prepareWorkspace.sh
```diff
@@ -469,10 +469,10 @@ checkingAndDownloadCaCerts() {
...
     git config core.sparsecheckout true
     echo "security/*" >>.git/info/sparse-checkout
-    git pull origin master
+    git pull origin "${BUILD_CONFIG[ADOPTOPENJDK_BUILD_REPO_BRANCH]}"
```
jennli, line 475: since we are now naming our default branch `main` instead of `master`

johnoliver:
run tests

johnoliver:
add to whitelist

sbin/common/config_init.sh
```diff
@@ -32,6 +32,8 @@
 # map. This is why we can't have nice things.
 CONFIG_PARAMS=(
 ADOPT_PATCHES
+ADOPTOPENJDK_BUILD_REPO_BRANCH
```
karianna, line 35: I'm wondering, could/should we abstract this further so that it's not just Adopt, but any valid OpenJDK source repo?  So maybe it's 

`OPENJDK_BUILD_REPO_BRANCH`

jennli: sure, I think your suggestion is better, will update.

karianna:
run tests

Issue #474, opened 2018-09-10
Commit reference for releases should be available via the /releases GitHub API


karianna:
_From @Joe-Brady on March 17, 2017 14:25_

Should be part of the automated process.

E.g. added as a tag, or by any method that makes it appear in this JSON:

api.github.com/repos/AdoptOpenJDK/openjdk-releases/releases

_Copied from original issue: AdoptOpenJDK/openjdk8-releases#2_

M-Davies:
@karianna I'm guessing this is still relevant and a `help wanted` issue?

karianna:
> @karianna I'm guessing this is still relevant and a `help wanted` issue?

Correct. the `scmRef` JSON key in our metadata still needs to be hooked up.

M-Davies:
For some context on this, our current meta data is created without setting a scm reference
https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L401

@karianna What exactly would the scm reference be? Would it be the latest commit hash that the binary was built with?

karianna:
> For some context on this, our current meta data is created without setting a scm reference
> https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/common/openjdk_build_pipeline.groovy#L401
> 
> @karianna What exactly would the scm reference be? Would it be the latest commit hash that the binary was built with?

Yes, in our openjdk-jdk<version> repos whatever the tagged commit in the `dev` branch was.

PR #1949, opened 2020-07-02
Populate metadata scmRef

M-Davies:
Hooks up the scmRef in the metadata to whatever the tag has been specified as OR whatever the output of git describe is. On a failed git describe, the tag is used whether it is an empty string or not

Closes: #474 

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -3,6 +3,7 @@ import common.MetaData
 @Library('local-lib@master')
 import common.VersionInfo
 import groovy.json.*
+import java.nio.file.NoSuchFileException
 
 import java.util.regex.Matcher
 
@@ -46,6 +47,7 @@ class Build {
     final def env
     final def currentBuild
     VersionInfo versionInfo = null
+    String scmRef = ""
 
     Build(IndividualBuildConfig buildConfig, def context, def env, def currentBuild) {
         this.buildConfig = buildConfig
@@ -217,7 +219,7 @@ class Build {
 
                     context.sh 'for file in $(ls workspace/target/*.tar.gz workspace/target/*.zip); do sha256sum "$file" > $file.sha256.txt ; done'
 
-                    writeMetadata(versionInfo)
+                    writeMetadata(versionInfo, false)
                     context.archiveArtifacts artifacts: "workspace/target/*"
                 }
             }
@@ -367,7 +369,7 @@ class Build {
                 if (buildConfig.TARGET_OS == "mac" || buildConfig.TARGET_OS == "windows") {
                     try {
                         context.sh 'for file in $(ls workspace/target/*.tar.gz workspace/target/*.pkg workspace/target/*.msi); do sha256sum "$file" > $file.sha256.txt ; done'
-                        writeMetadata(versionData)
+                        writeMetadata(versionData, false)
                         context.archiveArtifacts artifacts: "workspace/target/*"
                     } catch (e) {
                         context.println("Failed to build ${buildConfig.TARGET_OS} installer ${e}")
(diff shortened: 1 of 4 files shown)


M-Davies:
Draft for now while testing is underway at https://ci.adoptopenjdk.net/view/work%20in%20progress/job/jdk8u-windows-x64-hotspot/

M-Davies:
run tests

M-Davies:
run tests

karianna:
Are we not just writing Metadata once?  What's the use case where we edit an existing one?

M-Davies:
> Are we not just writing Metadata once? What's the use case where we edit an existing one?

We write the metadata several times during the process (once after the initial build, once after signing and once after the installers). The problem is that the later write outs overwrite the initial scmref value so they all end up being blank strings if `SCM_REF` is not set. We need to set the scmref in every write out otherwise we won't have a scmref at all in the metadata

M-Davies:
This is ready to be merged but I would like to get some reviews on it first. @johnoliver @karianna @sxa @gdams

M-Davies:
run tests

M-Davies:
run tests

M-Davies:
Ready for review

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -390,11 +392,91 @@ class Build {
...
+                        )
+                    }
+
+                } catch (NoSuchFileException e) {
+                    // In rare cases, we will fail to create the scmref.txt file
+                    context.println "WARNING: $scmRefPath was not found. Using SCM_REF instead..."
+
+                    return new MetaData(
+                        buildConfig.TARGET_OS,
+                        overrideScmref,
```
johnoliver, line 454: does this variable exist at this point?

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -390,11 +392,91 @@ class Build {
                 .toList()
     }
 
-    MetaData formMetadata(VersionInfo version) {
-        return new MetaData(buildConfig.TARGET_OS, buildConfig.SCM_REF, version, buildConfig.JAVA_TO_BUILD, buildConfig.VARIANT, buildConfig.ARCHITECTURE)
+    MetaData formMetadata(VersionInfo version, Boolean initialWrite) {
+
+        if (initialWrite) {
```
johnoliver, line 397: from what I can see of this section it could probably be simplified to something like:

```
scmRef=buildConfig.SCM_REF
        
        if (initialWrite) {

            String scmRefPath = "workspace/target/scmref.txt"

            context.println "FIRST METADATA WRITE OUT! Checking if we have a scm reference in the build config..."

            // Use the buildConfig scmref if it exists
            if (buildConfig.SCM_REF != "") {
                context.println "SCM_REF has been set (${buildConfig.SCM_REF})! Using it to build the inital metadata over ${scmRefPath}..."
                scmRef=buildConfig.SCM_REF
            } else {
                // If we don't have a scmref set in config, check if we have a scmref from the build
                context.println "SCM_REF is not set. Attempting to read ${scmRefPath}..."
                try {
                    scmRef = context.readFile(scmRefPath).trim()
                } catch (NoSuchFileException e) {
                    // In rare cases, we will fail to create the scmref.txt file
                    context.println "WARNING: $scmRefPath was not found."
                }
            }
        }
        
        return new MetaData(
            buildConfig.TARGET_OS,
            scmRef,
            version,
            buildConfig.JAVA_TO_BUILD,
            buildConfig.VARIANT,
            buildConfig.ARCHITECTURE
        )
```

M-Davies: I'm not sure I agree with setting `scmRef` to never be an empty string. In later writeouts, when `initialWrite` is set to false, `scmRef` will never be an empty string. So if the user has specified a tag to build, it will never be put into the metadata

karianna:
TestCompilation > openjdk_build_pipelineTest() STANDARD_ERROR
    org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
    common/openjdk_build_pipeline.groovy: 454: [Static type checking] - The variable [overrideScmref] is undeclared.
     @ line 454, column 25.
                               overrideScmref,
                               ^

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies: Copy paste error. Will fix

pipelines/build/common/openjdk_build_pipeline.groovy
johnoliver: from what I can tell this would mirror the functionality you already have, I believe in all paths in your code scmref is set to either `buildConfig.SCM_REF` or the contents of the scmRefPath file

M-Davies: Looking at it again, I've found a way to make it work (I think)! @johnoliver Can you check the latest commit?

johnoliver: yup looks good, although you can probably go further as both the if and else return the same thing on 420 and 430

M-Davies: Agreed

PR #1950, opened 2020-07-02
If API retries exceeeded, show RC of curl command

sxa:
Display return code from the final debug `curl -v` since it doesn't appear to be showing any output when it fails (Ref: https://adoptopenjdk.slack.com/archives/C09NW3L2J/p1593680142251300)

Also modifies the main curl commands to use `-q` to curl instead of `-v` as the `-v` can only cause confusion to the subsequent grep operation which will always remove all the verbose stuff anyway.


karianna:
run tests

johnoliver:
add to whitelist

PR #1960, opened 2020-07-06
set GRADLE_USER_HOME only when it's unset

jennli:
Proposing a small improvement to make GRADLE_USER_HOME configurable through `BUILD_CONFIG`.

In Azure Pipelines, we need to specify the location of `GRADLE_USER_HOME` in order to perform gradle caching between builds. This improvement grants us control for that location and prevents future issues when the location of GRADLE_USER_HOME gets modified.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -458,12 +458,12 @@ buildSharedLibs() {
     cd "${LIB_DIR}"
 
     local gradleJavaHome=$(getGradleJavaHome)
-    export GRADLE_USER_HOME=$(getGradleUserHome)
+    local gradleUserHome=$(getGradleUserHome)
 
-    echo "Running gradle with $gradleJavaHome at ${GRADLE_USER_HOME}"
+    echo "Running gradle with $gradleJavaHome at ${gradleUserHome}"
 
     gradlecount=1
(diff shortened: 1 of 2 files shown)


sbin/build.sh
```diff
@@ -442,16 +442,28 @@ getGradleHome() {
   echo $gradleJavaHome
 }
 
+getGradleUserHome() {
+  local gradleUserHome=""
+
+  if [ -z "${BUILD_CONFIG[GRADLE_USER_HOME]}" ]; then
+    gradleUserHome="${BUILD_CONFIG[WORKSPACE_DIR]}/.gradle"
```
karianna, line 449: Maybe flip this logic around so we're checking for the positive as opposed to the negative?  Not sure how easy that would be.

sbin/build.sh
```diff
@@ -442,16 +442,28 @@ getGradleHome() {
...
+  else
+    gradleUserHome="${BUILD_CONFIG[GRADLE_USER_HOME]}"
+  fi
+
+  echo $gradleUserHome
```
karianna, line 454: make the echo more verbose so there's context

jennli: I thought this serves as a return statement -- is this not how you return values in bash?

karianna: Ah, I see what you're doing now, I'm being dense - carry on :-)

M-Davies:
run tests

PR #1968, opened 2020-07-13
arm32: Build with gcc 7.5 on JDK8/arm32/HotSpot

sxa:
I am proposing pushing this into the July quarterly release this week since we cannot currently make https requests

Ref: https://github.com/AdoptOpenJDK/openjdk-support/issues/33#issuecomment-657340402


sxa:
Based on the two approvals and the fact this is a reasonably important fix, I'm merging this and will kick off a test build so we can verify this before tomorrow evening. :-)

PR #1981, opened 2020-07-21
Remove master node from pipeline files

M-Davies:
* Stops the pipelines from holding up a master executor for the whole build


M-Davies:
@AdamBrousseau Can I get a review from you too please?

AdamBrousseau:
I was just looking through the PR tester and it appears that job does the same thing as we're solving here. The config for that job is in the UI config (which is a bit scary).
https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/openjdk-build-pr-tester/configure

PR #1982, opened 2020-07-21
docker: Refactor of buildDocker.sh

Willsparker:
Ref: 
https://ci.adoptopenjdk.net/view/Tooling/job/DockerfileCheck/159/
https://ci.adoptopenjdk.net/view/Tooling/job/DockerfileCheck/162/
(The job for both of these runs had the config changed to run from this branch to test it)

A big refactor of `buildDocker.sh` changing the following:
- Rename variables, replace duplicated file paths with variables, add whitespace where it makes the code more readable
- Use the same logic in `dockerfile-generator.sh` ( using the API ) to set `JDK_MAX` and `JDK_GA`, so the script doesn't have to be changed in the future to add in later JDKs.
- The script runs within the openjdk-build directory it's executed, instead of re-cloning the openjdk-build repository again (requested by @sxa )
- Removed the `all` option as it seemed superfluous.

### DIFF
--- a/docker/buildDocker.sh
+++ b/docker/buildDocker.sh
@@ -1,13 +1,30 @@
 #!/bin/bash
 set -u
 
-jdkVersion=''
-bootDir=''
-buildVariant="hotspot"
+BOOTDIR=''
+VARIANT="hotspot"
 useEclipseOpenJ9DockerFiles=false
-cleanWorkspace=false
+CLEAN=false
+WORKSPACE=${PWD}
+JDK_VERSION=
+JDK_MAX=
+JDK_GA=
 
-#takes in all arguments to determine script options
+# shellcheck disable=SC2002 # Disable UUOC error
+setJDKVars() {
+	wget -q https://api.adoptopenjdk.net/v3/info/available_releases
+	JDK_MAX=$(cat available_releases \
+		| grep 'tip_version' \
+		| cut -d':' -f 2 \
+		| sed 's/,//g; s/ //g')
+	JDK_GA=$(cat available_releases \
+		| grep 'most_recent_feature_release' \
+		| cut -d':' -f 2 \
+		| sed 's/,//g; s/ //g')
+	rm available_releases
+}
+
+# Takes in all arguments to determine script options
 parseCommandLineArgs()
 {
 	if [ $# -lt 1 ]; then
@@ -19,18 +36,22 @@ parseCommandLineArgs()
 			local opt="$1";
 			shift;
 			case "$opt" in
-				"--all" | "-a" )
-					jdkVersion="all";;
 				"--clean" | "-c" )
-					cleanWorkspace=true;;
+					CLEAN=true;;
 				"--version" | "-v" )
-					jdkVersion="$1"; shift;;
+					if [ $1 == "jdk" ]; then
+	  					JDK_VERSION=$JDK_MAX
+					else
+	  					JDK_VERSION=$(echo $1 | tr -d [:alpha:])
+        				fi
+					checkJDK
+					shift;;
 				"--jdk-boot-dir" | "-J")
-					bootDir="$1"; shift;;
+					BOOTDIR="$1"; shift;;
 				"--openj9" | "-j9")
-					buildVariant="openj9";;
+					VARIANT="openj9";;
 				"--use-eclipse-docker-files" | "-e" )
-					useEclipseOpenJ9DockerFiles=true; buildVariant="eclipse";;
+					useEclipseOpenJ9DockerFiles=true; VARIANT="eclipse";;
 				"--help" | "-h" )
 					usage; exit 0;;
 				*) echo >&2 "Invalid option: ${opt}"; echo "This option was unrecognised."; usage; exit 1;;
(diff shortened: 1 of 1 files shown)


docker/buildDocker.sh
```diff
@@ -43,135 +64,102 @@ parseCommandLineArgs()
...
-		- jdk9u
-		- jdk10u
-		- jdk11u
-		- jdk12u
-		- jdk13u
-		- jdk14u
-		- jdk"
+checkJDK() {
+  if ! ((JDK_VERSION >= 8 && JDK_VERSION <= JDK_MAX)); then
+    echo "Please input a JDK between 8 & ${JDK_MAX}, or 'jdk'"
```
karianna, line 77: a concrete example will help here

docker/buildDocker.sh
```diff
@@ -1,13 +1,30 @@
 #!/bin/bash
 set -u
 
-jdkVersion=''
-bootDir=''
-buildVariant="hotspot"
+BOOTDIR=''
```
karianna, line 4: single quotes used here and double quotes on the next line - should we be consistent here?

docker/buildDocker.sh
```diff
@@ -19,18 +36,22 @@ parseCommandLineArgs()
...
+					shift;;
 				"--jdk-boot-dir" | "-J")
-					bootDir="$1"; shift;;
+					BOOTDIR="$1"; shift;;
 				"--openj9" | "-j9")
-					buildVariant="openj9";;
+					VARIANT="openj9";;
 				"--use-eclipse-docker-files" | "-e" )
-					useEclipseOpenJ9DockerFiles=true; buildVariant="eclipse";;
+					useEclipseOpenJ9DockerFiles=true; VARIANT="eclipse";;
```
karianna, line 54: is variant actually openj9?

docker/buildDocker.sh
```diff
@@ -43,135 +64,102 @@ parseCommandLineArgs()
...
+	local dockerfileDir="$WORKSPACE/EclipseDockerfiles"
+	local jdk="jdk"
+
+	mkdir -p $dockerfileDir
+	cd $dockerfileDir
 	wget https://raw.githubusercontent.com/eclipse/openj9/master/buildenv/docker/mkdocker.sh
 	chmod +x mkdocker.sh
 	# Generate an Ubuntu1804 Dockerfile using mkdocker.sh
-	$WORKSPACE/DockerBuildFolder/$jdkVersion-$buildVariant/EclipseDockerfiles/mkdocker.sh --dist=ubuntu --version=18 --print >> $WORKSPACE/DockerBuildFolder/$jdkVersion-$buildVariant/EclipseDockerfiles/Dockerfile
+	$dockerfileDir/mkdocker.sh --dist=ubuntu --version=18 --print >> $dockerfileDir/Dockerfile
```
karianna, line 110: do we want to pass in 18 as a parameter?

Willsparker: https://github.com/AdoptOpenJDK/openjdk-build/pull/1911#discussion_r444010951 
For now, it's hardcoded to stay `18` as all of the Dockerfiles generated using `dockerfile-generator.sh` can only be Ubuntu1804, so I think the `mkDocker.sh` generated dockerfiles should be consistent with those :-)

docker/buildDocker.sh
Willsparker: Fixed with latest commit

docker/buildDocker.sh
Willsparker: Fixed with latest commit

docker/buildDocker.sh
Willsparker: Fixed with latest commit

Willsparker:
@karianna Yep, it's been through `dockerfileCheck` a few times to check it's good

docker/buildDocker.sh
```diff
@@ -1,13 +1,30 @@
...
+WORKSPACE=${PWD}
+JDK_VERSION=
+JDK_MAX=
+JDK_GA=
 
-#takes in all arguments to determine script options
+# shellcheck disable=SC2002 # Disable UUOC error
+setJDKVars() {
+	wget -q https://api.adoptopenjdk.net/v3/info/available_releases
+	JDK_MAX=$(cat available_releases \
```
sxa, line 16: You could simplify this with awk which would reduce the number of processes by 3:
JDK_MAX=$(awk -F: '/tip_version/{gsub("[, ]","",$2); print$2}' < available_releases)

docker/buildDocker.sh
```diff
@@ -43,135 +64,103 @@ parseCommandLineArgs()
...
-	ssh-keyscan github.com >> $PWD/known_hosts
+	ssh-keygen -q -f $dockerfileDir/id_rsa -t rsa -N ''
+	cat id_rsa.pub >> $dockerfileDir/authorized_keys
+	ssh-keyscan github.com >> $dockerfileDir/known_hosts
 
-	# ${jdk%?} will remove the 'u' from 'jdkXXu'.
-	for jdk in $jdkVersion
-	do
-		if [ ${jdk} != "jdk" ]; then
-			jdk=${jdk%?}
```
sxa, line 126: Any reason not to use `${jdk%u}` instead of just stripping anything?

docker/buildDocker.sh
sxa, line 16: You could simplify this with awk which would reduce the number of processes by 3:
JDK_MAX=$(awk -F: '/tip_version/{gsub("[, ]","",$2); print$2}' < available_releases)

docker/buildDocker.sh
Willsparker: That's a bit that's been deleted

docker/buildDocker.sh
Willsparker: Tested it and committed it :-)

docker/buildDocker.sh
sxa: @karianna Are you ok with that explanation? Would be good to merge this and test the merge overnight if possible

karianna: I'll live with it :-)

sxa:
Testing at https://ci.adoptopenjdk.net/view/Tooling/job/DockerfileCheck/164/

sxa:
@Willsparker The above run seems to have broken on JDK11u/J9 with the cmake libdwarf error :-(

Willsparker:
@sxa `DFC` is still running on this branch - which I guess I made this PR before #1971 . Changing the config to test the `adoptopenjdk/master` branch :-) https://ci.adoptopenjdk.net/view/Tooling/job/DockerfileCheck/165/

Issue #1980, opened 2020-07-21
Intermittent unknown mac failure in checkoutRequiredCodeToBuild


M-Davies:
**Platform:**
All
**Architecture:**
All

See https://github.com/AdoptOpenJDK/openjdk-build/issues/1969 for a report of this (the title is misleading)
The actual error is unknown as no output appears in the log but it appears to be an issue in the [checkoutRequiredCodeToBuild](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/prepareWorkspace.sh#L111-L238) function
```
04:37:50  Pulling latest tags and getting the latest update version using git fetch -q --tags --depth=1
04:37:50  NOTE: This can take quite some time!  Please be patient
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // node
[Pipeline] }
[Pipeline] // stage
[Pipeline] echo
04:37:53  Execution error: hudson.AbortException: script returned exit code 1
```

M-Davies:
Seen during testing on `build-macstadium-macos1014-x64-2`https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk15/job/jdk15-mac-x64-openj9/49/consoleFull
```
11:21:26  checkoutRequiredCodeToBuild:
11:21:26    workspace = /Users/jenkins/workspace/build-scripts/jobs/jdk15/jdk15-mac-x64-openj9/workspace/./build//src
11:21:26    BUILD_VARIANT = openj9
11:21:26    TAG = 
11:21:26    BRANCH = openj9
11:21:26  Fetching origin
11:21:29  HEAD is now at 3604912533 Update OPENJDK_TAG to merged level jdk-15+32
11:21:29  Checked out origin/openj9
11:21:29  INFO: Extracting latest tag for the scmref using git describe since no tag is set in the build config...
11:21:29  fatal: No annotated tags can describe '3604912533cd23d5c77e28617e44d74546904f00'.
11:21:29  However, there were unannotated tags: try --tags.
11:21:29  WARNING: git describe FAILED. There is likely additional error output above (exit code was 128). Trying again using git describe --tags to match a lightweight (non-annotated) tag
11:21:29  jdk-15+32-71-g3604912533
[...]

PR #1984, opened 2020-07-22
Always pre-clean workspace on mac

M-Davies:
Fixes: #1980 

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -70,6 +70,12 @@ class Builder implements Serializable {
 
         def testList = getTestList(platformConfig)
 
+        // Always clean on mac due to https://github.com/AdoptOpenJDK/openjdk-build/issues/1980
+        def cleanWorkspace = cleanWorkspaceBeforeBuild
+        if (platformConfig.os == "mac") {
+            cleanWorkspace = true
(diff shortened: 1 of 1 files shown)


M-Davies:
Groovy compile error should not be due to these changes. Moving into review readiness

M-Davies:
https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/openjdk-build-pr-tester/503/console tests this on mac

M-Davies:
> is there something in the code that's forcing AIX to do the same?

I haven't seen AIX produce the error seen in #1980 . @adamfarley @andrew-m-leonard have you?

karianna:
> Seems a bit of a shame to do this as it means we are unable to update from an existing clone on the repo, requiring us to clone each time :-(
> 
> Related - is there something in the code that's forcing AIX to do the same? We were using a ramdisk previously but are not any more so it would be good to disable cleaning (probably post-cleaning thinking about it) on AIX if possible. I couldn't find where it was being done despite playing with a few options :-)

It would be good to know why it fails...

karianna:
I think you need to rebase to get that Groovy compile bug gone.

M-Davies:
> I think you need to rebase to get that Groovy compile bug gone.

Done

PR #1988, opened 2020-07-27
Disable jdk14 build+test, and enable jdk15 build+test

adamfarley:
This is being done because we've passed the last release for jdk14,
and we need start testing 15 in preparation for the September release.

Most JDK15 builds are already active, but there are a few platforms
(windows 32bit and XL variants) that needed to be added, based on the
fact that jdk14 was creating builds for them.

### DIFF
--- a/pipelines/jobs/configurations/jdk15.groovy
+++ b/pipelines/jobs/configurations/jdk15.groovy
@@ -3,12 +3,24 @@ targetConfigurations = [
                 "hotspot",
                 "openj9"
         ],
+        "x64MacXL"    : [
+                "openj9"
+        ],
         "x64Linux"    : [
                 "hotspot",
                 "openj9"
         ],
+        "x64LinuxXL"  : [
+                "openj9"
+        ],
+        "x32Windows"  : [
+                "hotspot"
+        ],
         "x64Windows"  : [
                 "hotspot",
+                "openj9"
+        ],
+        "x64WindowsXL": [
                 "openj9"
         ],
         "ppc64Aix"    : [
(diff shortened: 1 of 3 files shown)


pipelines/jobs/configurations/jdk15.groovy
```diff
@@ -23,16 +35,22 @@ targetConfigurations = [
...
+        ],
         "arm32Linux"  : [
                 "hotspot"
         ]
 ]
 
-// 03:30
-triggerSchedule="TZ=UTC\n30 03 * * *"
+// 20:26
+triggerSchedule="TZ=UTC\n26 20 * * *"
```
M-Davies, line 54: Why the time change here?

adamfarley: Because the jdk15 nightlies originally included a smaller number of builds (no xl builds), and included no testing.

My assumption was that since I'm increasing the machine requirements of this timeline, that I should give it the jdk14 time slot (which is disabled in this PR, and is already that big).

adamfarley: So instead of making the jdk15 pipeline bigger in the same time slot, and potentially causing problems, I moved it into the old jdk14 timeslot which is around that same size.

Does that make sense?

gdams:
run tests

gdams:
run tests

adamfarley:
Ah, George's test shows I'm missing a vital comma.

Searching for more commas now.

adamfarley:
run tests

gdams:
run tests

M-Davies:
add to whitelist

gdams:
one last observation, we should probably remove 14 from the PR tester too (https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/prTester/pr_test_pipeline.groovy#L133)

adamfarley:
Thanks @sxa 

@gdams - Sure. See here: https://github.com/AdoptOpenJDK/openjdk-build/pull/1993

PR #1992, opened 2020-07-28
Disable nightly builds while we release the 8u265 update

sxa:


sxa:
> We enabling ARM 32 C2 builds for this?

Not sure - we'll need to verify that we have the tags for it, and I'd prefer to do a bit more testing ... Gut feel is to err on the side of saying no at the moment since we haven't even looked at the recent builds with it enabled and then look again later (I didn't get the time I'd hoped to to test against 262) It's a risk, but then we do minimal testing against it anyway ;-)

karianna:
@sxa I think this PR should remove that change then?

sxa:
> @sxa I think this PR should remove that change then?

Ah I see - that commit wasn't supposed to be in here :-) I'll rebase properly

karianna:
@gdams do we still need to block nightlies to get a release out?

gdams:
nope this should be good to be closed

PR #1996, opened 2020-07-29
Correct dtrace argument for jdk15

adamfarley:
Dtrace does not accept the "auto" option on jdk15, and configure
fails if it is set. Changing this so it matches the other usages
in the same groovy file.


sxa:
@karianna Adding you as a reviewer - this is causing a build break on https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk15/job/jdk15-linux-aarch64-openj9-linuxXL/ as a result of https://github.com/AdoptOpenJDK/openjdk-build/pull/1988 and makes it consistent with the changes originally made in https://github.com/AdoptOpenJDK/openjdk-build/pull/1731

adamfarley:
Rebased.

Issue #1997, opened 2020-07-29
Add param to exclude docker building on specific jobs


M-Davies:
By entering a `dockerImage` and/or `dockerFile` field into the [pipeline_config files](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy), you are telling jenkins to build with docker. If those fields are not present, it won’t build using docker. What we need is an extra parameter that overrides this logic to tell jenkins to never build with docker (similar to how the current `enableTests` parameter works).

PR #1998, opened 2020-07-29
Add a dockerExcludes override parameter

M-Davies:
Adds a parameter that overrides our current system of always building docker if there is dockerImage and dockerFile parameter.

Closes: #1997 


M-Davies:
run tests

M-Davies:
run tests

M-Davies:
run tests

M-Davies:
run tests

M-Davies:
Had to use a `stringParam` instead of a `textParam` for this because `textParam` doesn't allow you to specify a `defaultValue` as far as I can see

pipelines/build/common/build_base_file.groovy
```diff
@@ -132,27 +133,56 @@ class Builder implements Serializable {
...
+        Boolean overrideDocker = false
+        if (dockerExcludes == {}) {
+            return overrideDocker 
+        }
+
+        String stringArch = configuration.arch as String
+        String stringOs = configuration.os as String
+        String estimatedKey = stringArch + stringOs.capitalize()
+
+        if (configuration.containsKey("additionalFileNameTag")) {
```
karianna, line 146: Possibly brittle?  Are there other suffixes that might apply when additionalFileNameTag is detected?

M-Davies: Not that I know of. This code is also used to [form node labels](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/common/build_base_file.groovy#L176) that helps jenkins pick which machines should run which builds so if this is brittle, that makes the linked code brittle too

Issue #2003, opened 2020-07-30
If JDKxx_BOOT_DIR not found, invalid params passed to makejdk_any_platform.sh


sxa:
If the `JDKxx_BOOT_DIR` parameters are not defined on the build machine, the build process does not add a parameter to `--jdk-boot-dir` on the `makejdk_any_platform.sh` command line. This situation needs to be trapped. Sample output:

```
12:24:27  Filename will be: OpenJDK11U-jdk_x64_linux_hotspot_2020-07-30-11-24.tar.gz
12:24:27  /var/lib/jenkins/workspace/build-scripts/jobs/jdk11u/jdk11u-linux-x64-hotspot/build-farm/../makejdk-any-platform.sh --clean-git-repo --jdk-boot-dir  --configure-args --enable-dtrace=auto --disable-warnings-as-errors --enable-ccache --target-file-name OpenJDK11U-jdk_x64_linux_hotspot_2020-07-30-11-24.tar.gz --use-jep319-certs --build-variant hotspot jdk11u
```

PR #2004, opened 2020-07-30
Terminate build if we dont have valid JDK_BOOT_DIR

M-Davies:
Fixes: #2003 


M-Davies:
> Assuming this passes the tests and doesn't have any side effects I think this is the right thing to do, but can you add a comment saying `Without this, a blank value can be passed into makejdk-any-platform.sh which causes a parsing failure` or something along those lines

@sxa Tests pass. Can you please re-review?

Issue #1989, opened 2020-07-27
gradle: "Unable to find java to run gradle with"


sxa:
If a set of variables are unset you get a message such as this:

```
Unable to find java to run gradle with, set JAVA_HOME, JDK8_BOOT_DIR or JDK11_BOOT_DIR: 
Running gradle with  at /home/jenkins/...
```
Regardless of this it appears to run so this is not a "hard" failure We should clarify the wording (I assume it's using a `java` from the PATH) and make sure the users are not confused by this message

M-Davies:
The code that spits out that line is here: https://github.com/AdoptOpenJDK/openjdk-build/blob/6de791416bb7f18af4de8aa4692061427d5e96db/sbin/build.sh#L438
I've noticed that sometimes it can be related to hard failures to do with gradle. [Example:](https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/build-test/job/jobs/job/jdk8u/job/jdk8u-linux-x64-hotspot/499/console)
```gradle
17:26:52  Unable to find java to run gradle with, set JAVA_HOME, JDK8_BOOT_DIR or JDK11_BOOT_DIR: 
17:26:52  Running gradle with  at /home/adoptopenjdk/workspace/build-scripts-pr-tester/build-test/jobs/jdk8u/jdk8u-linux-x64-hotspot/workspace/.gradle
17:26:52  Downloading https://services.gradle.org/distributions/gradle-6.5-bin.zip
17:26:51   > git rev-parse 3faf2ce437e05d9442df742bb36c4aba52ca17c8^{commit} # timeout=10
17:26:51   > git config core.sparsecheckout # timeout=10
17:26:51   > git checkout -f 3faf2ce437e05d9442df742bb36c4aba52ca17c8 # timeout=10
17:26:53  .................................................................................................
17:26:54  
17:26:54  Welcome to Gradle 6.5!
17:26:54  
17:26:54  Here are the highlights of this release:
17:26:54   - Experimental file-system watching
17:26:54   - Improved version ordering
17:26:54   - New samples
17:26:54  
17:26:54  For more details see https://docs.gradle.org/6.5/release-notes.html
17:26:54  
17:26:54  To honour the JVM settings for this build a new JVM will be forked. Please consider using the daemon: https://docs.gradle.org/6.5/userguide/gradle_daemon.html.
17:26:56  Daemon will be stopped at the end of the build stopping after processing
[...]

PR #2007, opened 2020-07-30
Re-jig Gradle not found as warning

M-Davies:
This is not necessarily a hard failure as first thought since gradle will still run without gradleJavaHome set

Fixes: #1989 

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -435,8 +435,7 @@ getGradleJavaHome() {
   fi
 
   if [ ! -d "$gradleJavaHome" ]; then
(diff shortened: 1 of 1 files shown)


brunoborges:
If you want a second (third?) pair of eyes, then yeah, looks good!

karianna:
> If you want a second (third?) pair of eyes, then yeah, looks good!

Hey Bruno, thanks, FYI you need to be a reviewer in the build team for it to count as an official +1 (to pass the GitHub check).  We can load you up with some tickets if you like :-)

Issue #1972, opened 2020-07-15
debug symbols are deleted during build despite of setting --with-native-debug-symbols=external


jennli:
impacting all `hotspot` builds.

when `--with-native-debug-symbols=external` flag is passed to configure, I expect the subsequent build will output all the debug symbol files. However, I am not getting any debug symbols due to the following lines explicitly removing them:

```bash
 # .diz files may be present on any platform
  # Note that on AIX, find does not support the '-delete' option.
  find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.diz" | xargs rm -f || true

  case "${BUILD_CONFIG[OS_KERNEL_NAME]}" in
    *cygwin*)
      # on Windows, we want to remove .map and .pdb files
      find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.map" -delete || true
      find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.pdb" -delete || true
      ;;
    darwin)
      # on MacOSX, we want to remove .dSYM folders
      find "${jdkTargetPath}" "${jreTargetPath}" -type d -name "*.dSYM" | xargs -I "{}" rm -rf "{}"
      ;;
    *)
      # on other platforms, we want to remove .debuginfo files
      find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.debuginfo" | xargs rm -f || true
      ;;
  esac
```
https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh#L623-L641

I feel this can be replaced with setting `--with-native-debug-symbols=none`. But the following line seems to indicate that there was a use case for OpenJ9 to delete all the debug symbol files while skipping the `--with-native-debug-symbols=none` flag.
https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh#L291

 I'd like to bring up this question and figure out if we can have a better approach here, since setting that flag to `none` should just remove the debug symbols from the output.

sxa:
@keithc-ca From memory the debug symbol changes were yours - can you take a look?

keithc-ca:
Yes, I recently made changes to the code that removes debug files, but the practice of deleting them precedes my involvement in that area. In fact, I remember being surprised it was happening (at first) because it was removing the files that OpenJ9 wanted to be in debug images.

[...]

PR #2008, opened 2020-07-30
Only remove debug symbols from JDK for OpenJ9 builds

keithc-ca:
OpenJ9 builds capture debug symbols in the debug image; this avoids the duplication while retaining them when present in other (e.g. hotspot) builds.

Fixes: #1972.


karianna:
@jennli - JIC you wanted to see this change

jennli:
Thanks!!!

M-Davies:
run tests

PR #2033, opened 2020-08-13
Build cacerts during build from Mozilla's list

aahlenst:
Most of the motivation/thinking behind the change is in the updated README. I manually checked the resulting cacerts of an 8u build on Linux and it has 138 entries as it should. I hope it works on Windows 🤞.

I removed `checkingAndDownloadCaCerts` in `prepareWorkspace.sh` because that has been unused for a while.

For context: https://github.com/AdoptOpenJDK/openjdk-support/issues/13


gdams:
@aahlenst is it not better to dynamically pull down the cacerts at build time? I'm worried that we might end up with an out of date list? It's also a rather large file to add to our repo.

aahlenst:
Arguments against downloading:

* Another possible cause for intermittent failures. 
* We have no idea what's in a given build except we lock and verify it with checksums.
* It is harder to review changes.
* We make it significantly harder for us to track changes over time.

gdams:
run tests

aahlenst:
```
/home/adoptopenjdk/workspace/build-scripts-pr-tester/build-test/jobs/jdk8u/jdk8u-linux-x64-hotspot/sbin/prepareWorkspace.sh: line 495: cd: /home/adoptopenjdk/workspace/build-scripts-pr-tester/build-test/jobs/jdk8u/jdk8u-linux-x64-hotspot/workspace/security: No such file or directory
```

Local docker build was okay, so something must be different.

gdams:
run tests

gdams:
add to whitelist

aahlenst:
Pipeline check has passed.

gdams:
run tests

karianna:
@aahlenst - merge conflicts.

I'm also calling on this PR to be gated by a @johnoliver review as his tinfoil hat is far better than my paper one ;-)

aahlenst:
Conflicts have been resolved.

karianna (approved):

aahlenst:
Blocked on (at least) https://github.com/AdoptOpenJDK/openjdk-infrastructure/pull/1525. I won't merge before the weekend. Monday at the earliest.

johnoliver (approved):

karianna:
@aahlenst - dependent PR was merged so I think you're good to go...

aahlenst:
Still needs playbook runs on the non-dockerized platforms and I don't know whether the container images have already been regenerated. So next step would be another test run here after the container images are ready.

aahlenst:
Prompted by a comment of @sxa on Slack, I changed `mk-cacerts.sh` to accept the path to keytool as an argument. This allows us to use keytool from the boot JDK to generate cacerts which makes infra problems less likely and should be more in line with idea of the boot JDK.

@karianna @gdams Please take another look. Changes: https://github.com/AdoptOpenJDK/openjdk-build/pull/2033/commits/a29ea75988ab0bba33c797522987d2a4e9bf6f9f

sxa:
FYI for OpenJ9 team awareness: @DanHeidinga @pshipton @shellberg @AdamBrousseau

sbin/prepareWorkspace.sh
```diff
@@ -492,44 +492,13 @@ checkingAndDownloadingFreeType() {
...
-
-  mkdir -p "security"
-  # Temporary fudge as curl on my windows boxes is exiting with RC=127
-  if [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]]; then
-    wget -O "./security/cacerts" "${caLink}"
-  else
-    curl -L -o "./security/cacerts" "${caLink}"
-  fi
+# Generates cacerts file
+prepareCacerts() {
```
sxa, line 496: Sorry for being a pain with another change but can we add something into this function that puts a message including "cacerts" to the log to make it easy to identify the correct section (previous messages about this all had it to make it easy to jump to the log) ... Just something like:

`echo "Generating cacerts from Mozilla's bundle"`

Then I'll be happy with it

sxa (changes requested):

sbin/prepareWorkspace.sh
aahlenst: Done

sxa (approved):

aahlenst:
The failing pipelines had trouble after the build. The builds themselves were successful.

PR #2035, opened 2020-08-19
Add support for Azure Devops testing

llxia:
Signed-off-by: lanxia <lan_xia@ca.ibm.com>

### DIFF
--- a/.azure-devops/test/jobs/steps/parallel.yml
+++ b/.azure-devops/test/jobs/steps/parallel.yml
@@ -1,0 +1,66 @@
+steps:
+    - script: |
+        brew update
+      displayName: 'Update brew'
+      condition: and(succeeded(), eq(variables['Agent.OS'], 'Darwin'))
+
+    # Install macOS required dependencies
+    - script: |
+        set -e
+        brew install xmlstarlet ant ant-contrib tree gnu-sed
+      displayName: "macOS: install dependencies"
+      condition: and(succeeded(), eq(variables['Agent.OS'], 'Darwin'))
+
+    - script: |
+        echo "##vso[task.setvariable variable=OPENJDK_BINARY]$PIPELINE_WORKSPACE/openjdkbinary";
+        echo "##vso[task.setvariable variable=TEST_JDK_HOME]$PIPELINE_WORKSPACE/openjdkbinary/j2sdk-image";
+        echo "##vso[task.setvariable variable=TEST_DIR]$COMMON_TESTRESULTSDIRECTORY/openjdk-tests"
+        echo "##vso[task.setvariable variable=TKG_DIR]$COMMON_TESTRESULTSDIRECTORY/openjdk-tests/TKG"
+      displayName: "Set environment varaibles"
+    - script: |
+        echo "TEST_JDK_HOME ${TEST_JDK_HOME}";
+        echo "TEST_DIR ${TEST_DIR}";
+        echo "TEST_DIR ${TEST_DIR}";
+        echo "OPENJDK_BINARY ${OPENJDK_BINARY}";
+        echo "ADOPT_TEST_REPO_BRANCH ${ADOPT_TEST_REPO_BRANCH}";
+        echo "ADOPT_TEST_REPO_URI ${ADOPT_TEST_REPO_URI}";
+      displayName: "Print environment varaibles"
+
+    # clone AdoptOpenJDK/openjdk-tests repo
+    - script: |
+        cd $COMMON_TESTRESULTSDIRECTORY
+        git clone -b master --depth 1 https://github.com/AdoptOpenJDK/openjdk-tests
+      displayName: "Clone openjdk test repo"
+
+    - task: DownloadPipelineArtifact@2
+      inputs:
+        artifact: $(Agent.OS)_$(ARCHITECTURE)
+        patterns: '**/*$(ARCHITECTURE)_$(TARGET_OS)*.tar.gz'
+        path: $(OPENJDK_BINARY)
+      displayName: "Download *$(ARCHITECTURE)_$(TARGET_OS)*.tar.gz"
+
+    # Get test framework and SDK
+    - script: |
+        cd $TEST_DIR
+        echo "./get.sh -s $PIPELINE_WORKSPACE -t $TEST_DIR -r upstream";
+        ./get.sh -s $PIPELINE_WORKSPACE -t $TEST_DIR -r upstream
+        if [ -f "job.properties" ]; then
+          TEST_JDK_HOME=`grep "TEST_JDK_HOME" job.properties | cut -d'=' -f2`
+          echo "TEST_JDK_HOME is set to $TEST_JDK_HOME"
+          echo "##vso[task.setvariable variable=TEST_JDK_HOME]$TEST_JDK_HOME";
+        fi
+      displayName: "Get test framework"
+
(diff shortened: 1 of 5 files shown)


karianna:
@junyzheng - can you review this?

.azure-devops/test/jobs/sanity.openjdk.yml
```diff
@@ -0,0 +1,44 @@
+parameters:
```
junyzheng, line 1: i feel like there is now significant difference between this yaml file and sanity.functional.yml file, maybe we can merge those 2?

.azure-devops/test/jobs/steps/parallel.yml
```diff
@@ -0,0 +1,66 @@
...
+        echo "ADOPT_TEST_REPO_URI ${ADOPT_TEST_REPO_URI}";
+      displayName: "Print environment varaibles"
+
+    # clone openjdk-test repo
+    - script: |
+        cd $COMMON_TESTRESULTSDIRECTORY
+        git clone -b master --depth 1 https://github.com/AdoptOpenJDK/openjdk-tests
+      displayName: "Clone openjdk test repo"
+
+    - task: DownloadPipelineArtifact@2
```
junyzheng, line 35: do we still need this step if you are going to run ` ./get.sh -s $PIPELINE_WORKSPACE -t $TEST_DIR -p x86-64_mac -r upstream`?

.azure-devops/test/jobs/steps/test.yml
```diff
@@ -0,0 +1,108 @@
...
+        echo "TEST_TARGET ${TEST_TARGET}";
+        echo "Agent.OS_ARCHITECTURE $(Agent.OS)_$(ARCHITECTURE)";
+      displayName: "Print environment varaibles"
+
+    - script: |
+        cd $COMMON_TESTRESULTSDIRECTORY;
+        git clone -b master --depth 1 https://github.com/AdoptOpenJDK/openjdk-tests
+      displayName: "Clone openjdk test repo"
+
+    - task: DownloadPipelineArtifact@2
```
junyzheng, line 45: same here do we still need to download jdk artifacts?

.azure-devops/test/jobs/sanity.openjdk.yml
junyzheng: I think you can take a look at our juniper-tests, we will pass a configurations object into the YAML templates and it will dynamically generate create those template.

.azure-devops/test/jobs/steps/parallel.yml
```diff
@@ -0,0 +1,66 @@
+steps:
+    - script: |
```
junyzheng, line 2: and for those dependencies, you can check our python script, we moved all those install dependencies into a python program so it is easier to manage in the future.

.azure-devops/test/jobs/steps/parallel.yml
llxia: yes, we do. This `DownloadPipelineArtifact ` downloads archived SDK, whereas  `./get.sh -s $PIPELINE_WORKSPACE -t $TEST_DIR -p x86-64_mac -r upstream` downloads test framework (TKG) and other test materials. It does not download SDK in this case.

.azure-devops/test/jobs/sanity.functional.yml
```diff
@@ -0,0 +1,44 @@
...
+    dependsOn: generate_parallelList_${{ parameters.buildList }}
+    variables:
+      BUILD_LIST: ${{ parameters.buildList }}
+      FAIL_TASK_ON_FAILED_TESTS: ${{ parameters.failTaskOnTestFailures }}
+      TEST_RESULT_SEARCH_PATH: ${{ parameters.testResultSearchPath }}
+      ORIGINAL_TEST_TARGET: ${{ parameters.testTarget }}
+    pool: ${{ parameters.pool }}
+    ${{ insert }}: ${{ parameters.useContainer }}
+    strategy:
+      matrix:
```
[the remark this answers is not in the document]
junyzheng:
```
{{ each i in counter(0, NUM_MACHINES )}}:
  testList_${{ i }}:
          TEST_TARGET: "-f parallelList.mk testList_${{ i }}"
          TEST_NAME: "testList_${{ i }}"
```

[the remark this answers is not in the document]
junyzheng: In the first job you can save a variable by using 
`echo "##vso[task.setvariable variable={you name it};isOutput=true]{some json data}"` in a step `stepabc`

In the second job you can use that variable like this:

```
dependsOn: job1 # reference of the first job
matrix: $[ dependencies.job1.outputs['stepabc.{you name it}'] ] 
```

Actually check this one: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#set-a-multi-job-output-variable

karianna:
@llxia - Are there further comments to address or should we re-review?

llxia:
I updated the PR and merged two files into one. Thanks so much for your suggestions! @junyzheng  
As for creating a JSON file and passing it as a matrix data (instead of a loop), I feel it makes the code harder to read and maintain. Since there is [not a good way to support loops](https://stackoverflow.com/questions/62409478/need-to-use-for-loop-not-for-each-loop-in-azure-pipeline-template-yaml) in Azure DevOps atm, I would like to open a feature request to support it. IMO, a simple solution for supporting a loop without inserting variables would be to add a range, something like `range(start, end, step)`. Any idea where I should submit my request?

Also, this PR is the initial version of the test pipelines for running tests in parallel on mac. And we will enhance it to add support for more platforms and tests. I think we should look into the optimization once we have the full picture.

karianna:
@junyzheng - Can you give this another review?

junyzheng (approved):
looks good me!

junyzheng:
@karianna done!

karianna:
@d3r3kk - Can you have a final review as well please?

d3r3kk:
Hey @llxia,

You had commented before:

> As for creating a JSON file and passing it as a matrix data (instead of a loop), I feel it makes the code harder to read and maintain. Since there is [not a good way to support loops](https://stackoverflow.com/questions/62409478/need-to-use-for-loop-not-for-each-loop-in-azure-pipeline-template-yaml) in Azure DevOps atm, I would like to open a feature request to support it. IMO, a simple solution for supporting a loop without inserting variables would be to add a range, something like `range(start, end, step)`. Any idea where I should submit my request?

[You can raise feature requests here.](https://developercommunity.visualstudio.com/content/idea/post.html?space=21)
[There is a request already in place that might be relevant here.](https://developercommunity.visualstudio.com/content/idea/1080699/implement-range-loop-construct-warning.html)

d3r3kk:
@llxia is there a run I can look at the end result of here? I don't see a pipeline that runs this in your AdoptOpenJDK org today?

llxia:
I tested this using my personal account ([previous run](https://dev.azure.com/helenxbc/AdoptOpenJDK/_build/results?buildId=94&view=results)). I just launched a new one: https://dev.azure.com/helenxbc/AdoptOpenJDK/_build/results?buildId=95&view=results

d3r3kk (approved):
LGTM, and I look forward to seeing how you leverage TRSS with AzDO! I left a couple of FYI comments, please don't see them as blockers.

PR #2039, opened 2020-08-26
add exploded image build to scripts

gdams:
**By default, this PR changes nothing. It just adds the ability to pause a build before the jmods are created and then continue.**

This PR allows us to create an exploded image `--make-exploded-image` (an image before the jmods have been compressed)

This allows us to then codesign the jmods before the get packaged up and then continue the build job using the `--assemble-exploded-image` flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,9 @@ the one you are trying to build.
 -k, --keep
 if using docker, keep the container after the build.
 
+--make-exploded
+creates an exploded image (useful for codesigning jmods). Use --assemble-exploded-image once you have signed the jmods to complete the packaging steps.
+
 -n, --no-colour
 disable colour output.
 
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ the one you are trying to build.
 -k, --keep
 if using docker, keep the container after the build.
 
---make-exploded
+--make-exploded-image
 creates an exploded image (useful for codesigning jmods). Use --assemble-exploded-image once you have signed the jmods to complete the packaging steps.
 
 -n, --no-colour
--- a/configureBuild.sh
+++ b/configureBuild.sh
@@ -281,6 +281,9 @@ function setMakeArgs() {
     "darwin") BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]=${BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]:-"product-images mac-legacy-jre-bundle"} ;;
     *) BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]=${BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]:-"product-images legacy-jre-image"} ;;
     esac
+    if [ "${BUILD_CONFIG[MAKE_EXPLODED]}" == "true" ]; then
+      BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]=""
+    fi
   else
     BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]=${BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]:-"images"}
   fi
--- a/configureBuild.sh
+++ b/configureBuild.sh
@@ -281,6 +281,7 @@ function setMakeArgs() {
     "darwin") BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]=${BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]:-"product-images mac-legacy-jre-bundle"} ;;
     *) BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]=${BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]:-"product-images legacy-jre-image"} ;;
     esac
+    # In order to build an exploded image, no other make targets can be used
     if [ "${BUILD_CONFIG[MAKE_EXPLODED]}" == "true" ]; then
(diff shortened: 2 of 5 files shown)


aahlenst:
Could you give some context why this is needed? Might help in case we later have to figure out why this change was applied. Except it's obvious for everyone but me.

gdams:
@aahlenst in Microsoft (like most companies) we don't have direct access to a codesign cert. We, therefore, cannot currently produce a JDK11 build using the Adopt build scripts because, by the time the image is created, it's too late to sign the JMOD binaries. This PR is a workaround to make this work.

aahlenst:
I that case, wouldn't it make sense to document how to use it? The arguments are not surfaced in the README and others might need a little help to successfully apply them.

gdams:
@aahlenst updated PTAL

configureBuild.sh
```diff
@@ -281,6 +281,9 @@ function setMakeArgs() {
     "darwin") BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]=${BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]:-"product-images mac-legacy-jre-bundle"} ;;
     *) BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]=${BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]:-"product-images legacy-jre-image"} ;;
     esac
+    if [ "${BUILD_CONFIG[MAKE_EXPLODED]}" == "true" ]; then
+      BUILD_CONFIG[MAKE_ARGS_FOR_ANY_PLATFORM]=""
```
karianna, line 285: Could this be subsequently overridden?  We might need to do this later or stick warning label on it.

sbin/build.sh
```diff
@@ -387,6 +387,11 @@ buildTemplatedFile() {
     ADDITIONAL_MAKE_TARGETS=" test-image"
   fi
 
+  if [[ "${BUILD_CONFIG[MAKE_EXPLODED]}" == "true" ]]; then
+    # In order to make an exploded image we cannot have any additional targets
+    ADDITIONAL_MAKE_TARGETS=""
```
karianna, line 392: Same goes for the previous comment, have we bound this as late as possible?

sbin/build.sh
```diff
@@ -846,11 +861,13 @@ configureCommandParameters
 buildTemplatedFile
 executeTemplatedFile
 
-printJavaVersionString
+if [[ "${BUILD_CONFIG[MAKE_EXPLODED]}" != "true" ]]; then
```
karianna, line 864: Isn't this true by default as we would have exited previously in the other if case?

sbin/common/config_init.sh
```diff
@@ -65,6 +66,7 @@ TEST_IMAGE_PATH
 JVM_VARIANT
 MACOSX_CODESIGN_IDENTITY
 MAKE_ARGS_FOR_ANY_PLATFORM
+MAKE_EXPLODED
```
karianna, line 69: prefer MAKE_EXPLODED_IMAGE

sbin/common/config_init.sh
```diff
@@ -198,6 +200,12 @@ function parseConfigurationArguments() {
         "--make-args" )
         BUILD_CONFIG[USER_SUPPLIED_MAKE_ARGS]="$1"; shift;;
 
+        "--make-exploded" )
```
karianna, line 203: prefer make-exploded-image

README.md
```diff
@@ -166,6 +166,9 @@ the one you are trying to build.
 -k, --keep
 if using docker, keep the container after the build.
 
+--make-exploded
+creates an exploded image (useful for codesigning jmods). Use --assemble-exploded-image once you have signed the jmods to complete the packaging steps.
```
karianna, line 170: So it's 'compulsory' to use both together?  I'd make that clear

karianna (changes requested):

configureBuild.sh
gdams: this is the latest place it can be set

sbin/build.sh
gdams: this is the latest place it can be set

sbin/build.sh
gdams: no, in a normal build (without `--make-exploded` or `--assemble-exploded-image`) the steps would need to be run

README.md
gdams: it's not compulsory, most people would want to run the latter for packing

sbin/common/config_init.sh
gdams: updated

gdams:
@karianna updated PTAL

gdams:
Build failure is unrelated

README.md
```diff
@@ -166,6 +166,9 @@ the one you are trying to build.
 -k, --keep
 if using docker, keep the container after the build.
 
+--make-exploded-image
+creates an exploded image (useful for codesigning jmods). Use --assemble-exploded-image once you have signed the jmods to complete the packaging steps.
```
aahlenst, line 170: It's fine for the README. Ideally, there should be some sort of a recipe, too. I could not find info how to sign the jmods and our tooling isn't that approachable in general. Separate PR would be okay.

configureBuild.sh
aahlenst: A comment would be useful why `MAKE_ARGS_FOR_ANY_PLATFORM` is set to `""` and there might need some sort of guidance for future editors that they need to ensure that `MAKE_ARGS_FOR_ANY_PLATFORM` stays empty.

gdams:
@aahlenst @karianna updated PTAL

aahlenst (approved):

karianna (approved):
LGTM

PR #2043, opened 2020-08-28
Add option to create debug-symbols in a separate file

milderhc:
Currently, we either have the option to make the images alone or with the debug symbols included in the same tar.gz/zip file.

This change is to have the option of including the debug symbols in a separate package, outside the JDK archive. If this option is called, we basically take the debug symbols of each platform, copy them to a new folder and then delete them from the images folder, and that folder just created is zipped as an artifact.

Issue: https://github.com/AdoptOpenJDK/openjdk-build/issues/2042

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ branch which may contain a very small diff set to the master branch
 (which is a clone from the OpenJDK mercurial forest).
 
 -B, --build-number <build_number>
-specify the OpenJDK build number to build from, e.g. b12.
+specify the OpenJDK build number to build hfrom, e.g. b12.
 For reference, OpenJDK version numbers look like 1.8.0_162-b12 (for Java 8) or
 9.0.4+11 (for Java 9+) with the build number being the suffix at the end.
 
@@ -138,7 +138,7 @@ specify any custom user configuration arguments.
 clean out any 'bad' local git repo you already have.
 
 --create-debug-symbols-package
-create a debug-symbols only archive if debug symbols are generated.
+create a debug-symbols only archive (if debug symbols were generated).
 
 -d, --destination <path>
 specify the location for the built binary, e.g. /path/.
--- a/README.md
+++ b/README.md
@@ -137,6 +137,9 @@ specify any custom user configuration arguments.
 --clean-git-repo
 clean out any 'bad' local git repo you already have.
 
+--create-debug-symbols-package
+create a debug-symbols only archive if debug symbols are generated.
+
 -d, --destination <path>
 specify the location for the built binary, e.g. /path/.
 This is typically used in conjunction with -T to create a custom path
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -297,14 +297,19 @@ configureDebugParameters() {
   # other options include fastdebug and slowdebug.
   addConfigureArg "--with-debug-level=" "release"
 
-  if [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK8_CORE_VERSION}" ]; then
-    addConfigureArg "--disable-zip-debug-info" ""
-    if [[ "${BUILD_CONFIG[BUILD_VARIANT]}" != "${BUILD_VARIANT_OPENJ9}" ]]; then
-      addConfigureArg "--disable-debug-symbols" ""
-    fi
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -123,6 +123,9 @@ specify any custom user configuration arguments.
 --clean-git-repo
 clean out any 'bad' local git repo you already have.
 
+--create-debug-symbols-package
+create a debug-symbols only archive if debug symbols are generated.
```
karianna, line 127: I think this needs a slight change in grammar.  Perhaps:

`create a debug-symbols only archive (if debug symbols were generated).`

sbin/build.sh
```diff
@@ -631,30 +636,58 @@ removingUnnecessaryFiles() {
...
         ;;
       *)
-        # on other platforms, we want to remove .debuginfo files
-        find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.debuginfo" | xargs rm -f || true
+        # on other platforms, we want to take .debuginfo files
+        debugSymbols=$(find "${jdkTargetPath}" -type f -name "*.debuginfo")
         ;;
     esac
+
+    if [ -n "${debugSymbols}" ] ; then
```
karianna, line 658: Worth adding a comment on what this effectively does

karianna (changes requested):
A couple of small nit-picks.  What I *think* I'm missing though is the creation of the debug symbols themselves for Hotspot builds.  That would need to occur in _build.sh_ in the `configureDebugParameters` function.  Do you think you could add that to this PR or add it as a separate PR?

milderhc:
> A couple of small nit-picks. What I _think_ I'm missing though is the creation of the debug symbols themselves for Hotspot builds. That would need to occur in _build.sh_ in the `configureDebugParameters` and `buildTemplatedFile` functions. Do you think you could add that to this PR or add it as a separate PR?

@karianna I originally thought this as if you want to get a debug symbols package you also need to provide something like `--with-native-debug-symbols=external` to generate them yourself. But if I understand correctly, what you are saying is if we call out this option we would like to ensure the debug symbols are going to be generated, right?

karianna:
> > A couple of small nit-picks. What I _think_ I'm missing though is the creation of the debug symbols themselves for Hotspot builds. That would need to occur in _build.sh_ in the `configureDebugParameters` and `buildTemplatedFile` functions. Do you think you could add that to this PR or add it as a separate PR?
> 
> @karianna I originally thought this as if you want to get a debug symbols package you also need to provide something like `--with-native-debug-symbols=external` to generate them yourself. But if I understand correctly, what you are saying is if we call out this option we would like to ensure the debug symbols are going to be generated, right?

That looks correct to me.  We can adjust if folks discover they're not getting what they expect.  I'll approve this but not yet merge as we're blocking updates until after next Tuesday's release of Java 15

karianna (approved):
LGTM

karianna:
run tests

karianna:
@milderhc - this will need rebasing I'm afraid.

sbin/build.sh
```diff
@@ -647,30 +657,59 @@ removingUnnecessaryFiles() {
...
-        find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.map" -delete || true
-        find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.pdb" -delete || true
+        # on Windows, we want to take .pdb files
+        debugSymbols=$(find "${jdkTargetPath}" -type f -name "*.pdb")
         ;;
       darwin)
-        # on MacOSX, we want to remove .dSYM folders
-        find "${jdkTargetPath}" "${jreTargetPath}" -type d -name "*.dSYM" | xargs -I "{}" rm -rf "{}"
+        # on MacOSX, we want to take .dSYM folders
+        debugSymbols=$(find "${jdkTargetPath}" -print -type d -name "*.dSYM")
```
tellison, line 671: Is the `-print` necessary?

tellison (commented):
One minor question

sbin/build.sh
milderhc: Yes, to get the actual files inside the `.dSYM` folders and then be copied by `cpio`. If not, only folders are copied.

gdams (approved):

PR #2051, opened 2020-09-04
add dragonwell support

gdams:


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -135,8 +138,10 @@ class Build {
             suffix="corretto/corretto-${javaNumber}"
         } else if (buildConfig.VARIANT == "openj9") {
```
karianna, line 139: Hmm we need to refactor this overloaded term VARIANT should probably refer to the underlying VM and we should introduce PROVIDER for 'who is behind this thing'

pipelines/build/dockerFiles/dragonwell.dockerfile
```diff
@@ -0,0 +1,12 @@
+ARG image
+
+FROM $image
+
+RUN \
+    # Dragonewell 8 requires a dragonwell 8 BootJDK
```
karianna, line 6: So we're only doing 8 for now?

pipelines/jobs/configurations/jdk8u.groovy
```diff
@@ -6,7 +6,8 @@ targetConfigurations = [
         "x64Linux"      : [
                 "hotspot",
                 "openj9",
-                "corretto"
+                "corretto",
+                "alibiba"
```
karianna, line 10: should be called dragonwell?

karianna (changes requested):

pipelines/jobs/configurations/jdk8u.groovy
smlambert: I think so, also alibaba is spelled wrong here

pipelines/build/dockerFiles/dragonwell.dockerfile
gdams: Dragonwell 11 can be built with a regular bootJDK. This is specific to 8

pipelines/build/common/openjdk_build_pipeline.groovy
gdams: agreed, we can do some logic at the publish phase to control this for now.

karianna (approved):

karianna:
run tests

pipelines/build/common/openjdk_build_pipeline.groovy
gdams: PR to push these binaries to a separate repo for the API: https://github.com/AdoptOpenJDK/openjdk-website-backend/pull/72/files

smlambert (approved):
Thanks @gdams !

PR #2062, opened 2020-09-08
Stop building client VM on Windows for 15, 16

### LINKED ISSUE #2047: EA15 fails compiling jlink image under windows: module not found: jdk.internal.vm.compiler
EA15
**Platform:**
Windows
**Architecture:**
x64

When compiling our project we noticed that building a jlink image fails under Windows
module not found: jdk.internal.vm.compiler

You can found a log of our run here:
https://github.com/JabRef/jabref/pull/6829/checks?check_run_id=1055196773

Please describe the issue that you are having with our binaries and if possible how to recreate it.

aahlenst:
Starting with 15, building the client VM together with the server VM on Windows causes the exclusion of modules/functionality related to Graal, namely:

```
jdk.internal.vm.compiler@15
jdk.internal.vm.compiler.management@15
```

Windows is the last x64 platform that we build the client VM on together with the server VM. Of the other vendors I checked (OpenJDK, Oracle, Azul, Bellsoft), only Bellsoft Liberica still has a client VM bundled with their Windows x64 build of OpenJDK **14**. The state of 15 for Bellsoft is unknown to me. From a technical standpoint, the server VM is superior nowadays (higher memory limits, ...) and the slower start-up compared to the client VM shouldn't really be an issue anymore.

I therefore propose dropping the client/server combination VM on Windows x64 starting with 15.

Fixes #2047.

### DIFF
--- a/pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
@@ -51,7 +51,7 @@ class Config15 {
                         hotspot: 'win2012&&vs2017'
                 ],
                 buildArgs : [
-                        hotspot : '--jvm-variant client,server'
+                        hotspot : '--jvm-variant server'
                 ],
                 test                : ['sanity.openjdk', 'sanity.perf', 'sanity.system', 'extended.system']
         ],
--- a/pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
@@ -50,9 +50,6 @@ class Config15 {
                 additionalNodeLabels: [
                         hotspot: 'win2012&&vs2017'
                 ],
-                buildArgs : [
-                        hotspot : '--jvm-variant server'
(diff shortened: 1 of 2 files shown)


pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
```diff
@@ -51,7 +51,7 @@ class Config15 {
                         hotspot: 'win2012&&vs2017'
                 ],
                 buildArgs : [
-                        hotspot : '--jvm-variant client,server'
```
gdams, line 54: I believe that the default is server, Therefore you should be able to completely remove the buildArgs block.

aahlenst: Removed.

jerboaa (approved):
This seem OK to me, but I don't know who relied on -client JVM.

karianna:
Ask discussed in TSC meeting this is OK (I scanned historical usage, it doesn't seem to be required).

karianna (approved):
LGTM

PR #2089, opened 2020-09-17
Add checks to prevent Github Actions error

austin0:
Adds a check to prevent Github Actions from failing due to a git error inside the release file. the error occurs when it attempts to `git rev-parse` in a directory with no tree as GHA doesn't seem to create one for the build repository.

Also adds comments to provide status reports in the console during release file generation.

Signed-off-by: Austin Bailey <Austin.Bailey@ibm.com>

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -854,22 +854,35 @@ fixJavaHomeUnderDocker() {
 
 addInfoToReleaseFile(){
   # Extra information is added to the release file here
+  echo "===GENERATING RELEASE FILE==="
   cd $PRODUCT_HOME
   JAVA_LOC="$PRODUCT_HOME/bin/java"
+  echo "ADDING IMPLEMENTOR"
   addImplementor
+  echo "ADDING BUILD SHA"
   addBuildSHA
+  echo "ADDING FULL VER"
   addFullVersion
+  echo "ADDING SEM VER"
   addSemVer
+  echo "ADDING BUILD OS"
   addBuildOS
+  echo "ADDING VARIANT"
   addJVMVariant
+  echo "ADDING JVM VERSION"
   addJVMVersion
   # OpenJ9 specific options
   if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_OPENJ9}" ]; then
+    echo "ADDING HEAP SIZE"
     addHeapSize
+    echo "ADDING J9 TAG"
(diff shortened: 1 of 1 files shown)


austin0:
Seems to be all good, error is gone from GHA logs and jenkins artifacts appear the same.
Example of new console logs:
```
=JAVA VERSION OUTPUT=
openjdk version "1.8.0_272"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_272-202009171930-b07)
OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.272-b07, mixed mode)
=/JAVA VERSION OUTPUT=
===GENERATING RELEASE FILE===
ADDING IMPLEMENTOR
ADDING BUILD SHA
ADDING FULL VER
ADDING SEM VER
ADDING BUILD OS
ADDING VARIANT
ADDING JVM VERSION
MIRRORING TO JRE
ADDING IMAGE TYPE
===RELEASE FILE GENERATED===
```

gdams (approved):

sbin/build.sh
```diff
@@ -903,8 +916,12 @@ addJVMVariant(){
 }
 
 addBuildSHA(){ # git SHA of the build repository i.e. openjdk-build
-  local buildSHA=$(git -C ${BUILD_CONFIG[WORKSPACE_DIR]} rev-parse --short HEAD)
-  echo -e BUILD_SOURCE=\"git:$buildSHA\" >> release
+  local buildSHA=$(git -C ${BUILD_CONFIG[WORKSPACE_DIR]} rev-parse --short HEAD 2>/dev/null)
+  if [[ $buildSHA ]]; then
+    echo -e BUILD_SOURCE=\"git:$buildSHA\" >> release
+  else
+    echo "Unable to fetch build SHA, does a work tree exist?..."
```
karianna, line 923: Do you need to crash out at this point or can you continue?

karianna (changes requested):

sbin/build.sh
austin0: I don't believe so as it would mean that GHA is never be able to build. For jenkins builds it should crash before it ever reaches this point as it would mean that something went wrong when fetching the `openjdk-build` repo.

karianna: OK, thanks

karianna (approved):

M-Davies (approved):

PR #2124, opened 2020-10-07
Update OpenSSL to version 1.1.1h for all non-Windows platforms

### LINKED ISSUE #2123: Update OpenJ9 builds to openssl 1.1.1h
There is a new version of openssl, 1.1.1h, which contains bug fixes. I don't think any of the fixes affect areas that OpenJ9 uses, but we should update to use the latest version anyway. This can occur after the Oct release.

AdamBrousseau:
This will fetch the current latest version (1.1.1h)
for MacOS as well as all linuxs and AIX

Fixes #2123


sxa (approved):
@Willsparker FYI - we need to try and do this change along with the windows one during each update

Willsparker (approved):

sxa:
(Verifying via https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk11-pipeline/1198/console so I can catch this before the weekend if anything goes wrong with it)

PR #2125, opened 2020-10-07
Fixes #2114 - Add Shenandoah for Java 11 builds.

### LINKED ISSUE #2114: Build JDK11u 11.0.9 (or a variant) with Shenandoah
Hello,

Do you plan to release a jdk11u build with Shenandoah ? As you are certainly aware Shenandoah was back-ported this summer, it just need to be activated in the build.

While the modification in this [file](https://github.com/AdoptOpenJDK/openjdk-build/blob/b5941737145dfbe5c10030737f6c69bce6132749/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy#L32-L36) is likely straightforward, it has certainly implications which I do not grasp. Wether to include Shanandoah in the _default_ build may be something up to debate, but in the case it's a no I think that shipping a JDK11u variant with Shenandoah could be appreciated, I certainly would.

Cheers,
Thank you for this project !!!

karianna:
As Shenandoah is ready for production in 16, I'm assuming the flag is not required there.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -75,9 +75,10 @@ configureBootJDKConfigureParameter() {
   addConfigureArgIfValueIsNotEmpty "--with-boot-jdk=" "${BUILD_CONFIG[JDK_BOOT_DIR]}"
 }
 
-# Configure the Shenandoah GC build parameter for JDK 11-15
+# Shenandaoh was backported to Java 11 as of 11.0.9 but requires this build
+# parameter to ensure its inclusion. For Java 12+ this is automatically set
 configureShenandoahBuildParameter() {
-  if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_HOTSPOT}" ] && [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" -ge "${JDK11_CORE_VERSION}" ] && [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" -le "${JDK15_CORE_VERSION}" ]; then
+  if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_HOTSPOT}" ] && [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" -eq "${JDK11_CORE_VERSION}" ]; then
       addConfigureArg "--with-jvm-features=" "shenandoahgc"
   fi
 }
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -28,7 +28,7 @@
 
 set -eu
 
-SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 # shellcheck source=sbin/prepareWorkspace.sh
 source "$SCRIPT_DIR/prepareWorkspace.sh"
@@ -41,7 +41,6 @@ source "$SCRIPT_DIR/common/constants.sh"
 
 # shellcheck source=sbin/common/common.sh
 source "$SCRIPT_DIR/common/common.sh"
-
 
 export LIB_DIR=$(crossPlatformRealPath "${SCRIPT_DIR}/../pipelines/")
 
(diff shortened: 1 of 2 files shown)


gdams (changes requested):
can you make these changes in build.sh? E.g https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh#L211

sxa (commented):
The location for changes such as this is something we need to discuss really. I came across it when putting [this section of the FAQ](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/FAQ.md#how-do-i-change-the-parameters-such-as-configure-flags-for-a-jenkins-build) together as things are currently in one of three places. Which to use depends on who we want to be affected and to me knowledge we've avoided being clear on where changes should be made which has already lead to confusion in the past. Quick summary:

| Location | Impact |
| --- | --- |
| groovy files (as per this PR) | Only when run through our jenkins pipelines |
| [platform-specific-configurations](https://github.com/AdoptOpenJDK/openjdk-build/tree/master/build-farm/platform-specific-configurations) scripts | Those using [build-farm/make-adopt-build-farm.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/make-adopt-build-farm.sh) (inc. our pipelines) - should be stuff specific to our machines |
| [build.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh) | Anyone (including end users) who are running [makejdk-any-platform.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/makejdk-any-platform.sh) |

So it depends what we want the last line of those to be. If it's to allow users to replicate the adopt builds with the same configuration options as us as far as possible, then I think it ought to be in build.sh but if we want it to be optional for users building it themselves then the jenkins pipelines aren't a bad choice. But we should really make it clear for new people to the project where changes should be made e.g. via an update to the FAQ.

To be clear, I'm ok with this going in as-is for now, or in `build.sh` (which is why this is a comment rather than an explicit approve/reject for now) but let's make sure we clarify for the future (I'll spin off another issue for it to ensure we get consensus - as any outcome from that should include tidying up where things are currently defined)

sxa:
(Side note: Is this really a bug as opposed to an enhancement?)

karianna:
@gdams I've given that a go, I'm not sure if the whitespace changes my IDE enforced will mean I need to redo this PR, let me know.

sbin/build.sh
```diff
@@ -56,50 +55,53 @@ function parseArguments() {
...
-{
+configureBootJDKConfigureParameter() {
   addConfigureArgIfValueIsNotEmpty "--with-boot-jdk=" "${BUILD_CONFIG[JDK_BOOT_DIR]}"
 }
 
+# Configure the Shenandoah GC build parameter for JDK 11-15
+configureShenandoahBuildParameter() {
+  if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_HOTSPOT}" ]; then
+    if [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" -ge "${JDK11_CORE_VERSION}" ] && [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" -le "${JDK15_CORE_VERSION}" ]; then
+      addConfigureArgIfValueIsNotEmpty "--with-jvm-features=" "shenandoahgc"
```
sxa, line 82: Two comments:
- Why use the `IfValueIsNotEmpty` verison of `addConfigureArg` when it's explicitly not-empty?
```suggestion
      addConfigureArg "--with-jvm-features=" "shenandoahgc"
```
Also the nested if as opposed to just `&&`'ing them together on the same line seems to me to make it less readable, unless the intention is to adjust it to change it to an "or" of other variants going forward (but if that's the intention I would suggest splitting it out at that time)

karianna:
@sxa fixes in place.

sxa (approved):
Unless I've missed something in amongst all the whitespace adjustments this looks ok to me now :-)

karianna:
run tests

karianna:
run tests

sbin/build.sh
```diff
@@ -56,50 +55,51 @@ function parseArguments() {
...
 }
 
 # Configure the boot JDK
-configuringBootJDKConfigureParameter()
-{
+configureBootJDKConfigureParameter() {
   addConfigureArgIfValueIsNotEmpty "--with-boot-jdk=" "${BUILD_CONFIG[JDK_BOOT_DIR]}"
 }
 
+# Configure the Shenandoah GC build parameter for JDK 11-15
```
aahlenst, line 78: Is it enabled by default on 16? Would be great if that was documented. Otherwise I've to google it every time. We should also record why <11 doesn't have it.

aahlenst (changes requested):

aahlenst:
Can we do something to prevent accidental whitespace changes in the future? Terrible to review.

bric3 (approved):

aahlenst (approved):

PR #2130, opened 2020-10-08
WIP: Active Node Timeout for Build Pipelines

adamfarley:
Currently, if there are no active nodes that match a build job's
labels, we fail immediately.

This change is intended to add a modifiable timeout value, so if we
don't find an active node that matches the build job's labels, we
wait for x minutes, periodically checking to see if a node matching
the labels has come online.

### DIFF
--- a/pipelines/build/common/config_regeneration.groovy
+++ b/pipelines/build/common/config_regeneration.groovy
@@ -266,6 +266,7 @@ class Regeneration implements Serializable {
                 SCM_REF: "",
                 BUILD_ARGS: buildArgs,
                 NODE_LABEL: "${additionalNodeLabels}&&${platformConfig.os}&&${platformConfig.arch}",
+                ACTIVE_NODE_TIMEOUT: "",
                 CODEBUILD: platformConfig.codebuild as Boolean,
                 DOCKER_IMAGE: dockerImage,
                 DOCKER_FILE: dockerFile,
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -659,6 +659,40 @@ class Build {
         }
     }
 
+    def waitForANodeToBecomeActive(def label) {
+        def NodeHelper = context.library(identifier: 'openjdk-jenkins-helper@master').NodeHelper
+
+        if (NodeHelper.nodeIsOnline(label)) {
+            return
+        }
+
+        context.println("No active node matches this label: " + label)
+
+        int activeNodeTimeout = 0
+        if (buildConfig.ACTIVE_NODE_TIMEOUT.isInteger()) {
+            activeNodeTimeout = buildConfig.ACTIVE_NODE_TIMEOUT as Integer
+        }
+
+
+        if (activeNodeTimeout > 0) {
+            context.println("Will check again periodically until a node labelled " + label + " comes online, or " + buildConfig.ACTIVE_NODE_TIMEOUT + " minutes (ACTIVE_NODE_TIMEOUT) has passed.")
+            int x = 0
+            while (x < activeNodeTimeout) {
+                context.sleep(60 * 1000)  // 1 minute sleep
+                if (NodeHelper.nodeIsOnline(label)) {
+                    context.println("A node which matches this label is now active: " + label)
+                    return
+                }
+                x++
+            }
+            context.error("No node matching this label became active prior to the timeout: " + label)
+            throw new Exception()
+        } else {
+            context.error("As the timeout value is set to 0, we will not wait for a node to become active.")
+            throw new Exception()
+        }
+    }
+
     def build() {
         context.timestamps {
             context.timeout(time: 18, unit: "HOURS") {
@@ -678,8 +712,6 @@ class Build {
                     def cleanWorkspace = Boolean.valueOf(buildConfig.CLEAN_WORKSPACE)
 
                     context.stage("queue") {
-                        def NodeHelper = context.library(identifier: 'openjdk-jenkins-helper@master').NodeHelper
-
                         if (buildConfig.DOCKER_IMAGE) {
                             // Docker build environment
                             def label = buildConfig.NODE_LABEL + "&&dockerBuild"
@@ -690,6 +722,8 @@ class Build {
                             if (buildConfig.CODEBUILD) {
                                 label = "codebuild"
                             }
+
+                            waitForANodeToBecomeActive(label)
                             context.node(label) {
                                 // Cannot clean workspace from inside docker container
                                 if (cleanWorkspace) {
@@ -714,27 +748,23 @@ class Build {
                             }
 
                         } else {
-                            if (NodeHelper.nodeIsOnline(buildConfig.NODE_LABEL)) {
-                                context.node(buildConfig.NODE_LABEL) {
-                                    // This is to avoid windows path length issues.
-                                    context.echo("checking ${buildConfig.TARGET_OS}")
-                                    if (buildConfig.TARGET_OS == "windows") {
-                                        // See https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1284#issuecomment-621909378 for justification of the below path
-                                        def workspace = "C:/workspace/openjdk-build/"
-                                        if (env.CYGWIN_WORKSPACE) {
-                                            workspace = env.CYGWIN_WORKSPACE
-                                        }
-                                        context.echo("changing ${workspace}")
-                                        context.ws(workspace) {
-                                            buildScripts(cleanWorkspace, filename)
-                                        }
-                                    } else {
+                            waitForANodeToBecomeActive(buildConfig.NODE_LABEL)
+                            context.node(buildConfig.NODE_LABEL) {
+                                // This is to avoid windows path length issues.
+                                context.echo("checking ${buildConfig.TARGET_OS}")
+                                if (buildConfig.TARGET_OS == "windows") {
+                                    // See https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1284#issuecomment-621909378 for justification of the below path
+                                    def workspace = "C:/workspace/openjdk-build/"
+                                    if (env.CYGWIN_WORKSPACE) {
+                                        workspace = env.CYGWIN_WORKSPACE
+                                    }
+                                    context.echo("changing ${workspace}")
+                                    context.ws(workspace) {
                                         buildScripts(cleanWorkspace, filename)
                                     }
-                                }   
-                            } else {
-                                context.error("No node of this type exists: ${buildConfig.NODE_LABEL}")
-                                return
+                                } else {
+                                    buildScripts(cleanWorkspace, filename)
+                                }
                             }
                         }
                     }
--- a/pipelines/jobs/pipeline_job_template.groovy
+++ b/pipelines/jobs/pipeline_job_template.groovy
@@ -58,6 +58,7 @@ pipelineJob("${BUILD_FOLDER}/${JOB_NAME}") {
 
     parameters {
         textParam('targetConfigurations', JsonOutput.prettyPrint(JsonOutput.toJson(targetConfigurations)))
+        stringParam('activeNodeTimeout', "0", 'Number of minutes we will wait for a label-matching node to become active.')
         stringParam('dockerExcludes', "", 'Map of targetConfigurations to exclude from docker building. If a targetConfiguration (i.e. { "x64LinuxXL": [ "openj9" ], "aarch64Linux": [ "hotspot", "openj9" ] }) has been entered into this field, jenkins will build the jdk without using docker. This param overrides the dockerImage and dockerFile downstream job parameters.')
         choiceParam('releaseType', ['Nightly', 'Nightly Without Publish', 'Release'], 'Nightly - release a standard nightly build.<br/>Nightly Without Publish - run a nightly but do not publish.<br/>Release - this is a release, this will need to be manually promoted.')
         stringParam('overridePublishName', "", '<strong>REQUIRED for OpenJ9</strong>: Name that determines the publish name (and is used by the meta-data file), defaults to scmReference(minus _adopt if present).<br/>Nightly builds: Leave blank (defaults to a date_time stamp).<br/>OpenJ9 Release build Java 8 example <code>jdk8u192-b12_openj9-0.12.1</code> and for OpenJ9 Java 11 example <code>jdk-11.0.2+9_openj9-0.12.1</code>.')
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -33,6 +33,7 @@ limitations under the License.
 //@CompileStatic(extensions = "JenkinsTypeCheckHelperExtension")
 class Builder implements Serializable {
     String javaToBuild
+    String activeNodeTimeout
     String adoptBuildNumber
     String overrideFileNameVersion
     String additionalBuildArgs
@@ -94,6 +95,7 @@ class Builder implements Serializable {
                 SCM_REF: scmReference,
                 BUILD_ARGS: buildArgs,
                 NODE_LABEL: "${additionalNodeLabels}&&${platformConfig.os}&&${platformConfig.arch}",
+                ACTIVE_NODE_TIMEOUT: activeNodeTimeout,
                 CODEBUILD: platformConfig.codebuild as Boolean,
                 DOCKER_IMAGE: dockerImage,
                 DOCKER_FILE: dockerFile,
@@ -490,6 +492,7 @@ return {
     String javaToBuild,
     Map<String, Map<String, ?>> buildConfigurations,
     String targetConfigurations,
+    String activeNodeTimeout,
     String dockerExcludes,
     String enableTests,
     String enableInstallers,
@@ -536,6 +539,7 @@ return {
             javaToBuild: javaToBuild,
             buildConfigurations: buildConfigurations,
             targetConfigurations: new JsonSlurper().parseText(targetConfigurations) as Map,
+            activeNodeTimeout: activeNodeTimeout,
             dockerExcludes: buildsExcludeDocker,
             enableTests: Boolean.parseBoolean(enableTests),
             enableInstallers: Boolean.parseBoolean(enableInstallers),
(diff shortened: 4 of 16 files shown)


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -659,6 +659,40 @@ class Build {
...
+
+        if (NodeHelper.nodeIsOnline(label)) {
+            return
+        }
+
+        context.println("No active node matches this label: " + label)
+
+        int activeNodeTimeout = 0
+        if (${buildConfig.ACTIVE_NODE_TIMEOUT}.isInteger()) {
+            activeNodeTimeout = ${buildConfig.ACTIVE_NODE_TIMEOUT} as Integer
```
M-Davies, line 673: You shouldn't need the `${}` here

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -659,6 +659,40 @@ class Build {
...
+        }
+
+
+        if (activeNodeTimeout > 0) {
+        context.println("Will check again periodically until either one comes online, or " + ${buildConfig.ACTIVE_NODE_TIMEOUT} + " minutes (ACTIVE_NODE_TIMEOUT) has passed.")
+            int x = 0
+            while (x < activeNodeTimeout) {
+                sleep(60 * 1000)
```
M-Davies, line 681: You could use groovy's advanced sleep for better brevity. See https://github.com/AdoptOpenJDK/openjdk-build/pull/2108

pipelines/jobs/pipeline_job_template.groovy
```diff
@@ -58,6 +58,7 @@ pipelineJob("${BUILD_FOLDER}/${JOB_NAME}") {
 
     parameters {
         textParam('targetConfigurations', JsonOutput.prettyPrint(JsonOutput.toJson(targetConfigurations)))
+        stringParam('activeNodeTimeout', "0", 'Number of minutes we will wait for a label-matching node to become active.')
```
M-Davies, line 61: Are we happy having the default timeout value as 0? This would mean that the nightlies will fail automatically if a node isn't found straight away (and with most of them executing at the same time, it's very possible some platforms which we don't have a lot of machines for will crash out instantly)

pipelines/build/regeneration/jdk9_regeneration_pipeline.groovy
```diff
@@ -61,11 +61,13 @@ node ("master") {
     def jobRoot = "$JOB_ROOT" != "" ? JOB_ROOT : "build-scripts"
     def jenkinsBuildRoot = "$JENKINS_BUILD_ROOT" != "" ? JENKINS_BUILD_ROOT : "https://ci.adoptopenjdk.net/job/build-scripts/"
     def excludes = "$EXCLUDES_LIST" != "" ? EXCLUDES_LIST : ""
+    def activeNodeTimeout = "$ACTIVE_NODE_TIMEOUT" != "" ? ACTIVE_NODE_TIMEOUT : ""
```
M-Davies, line 64: No need for having a user inputted timeout on the `jdkxx_regeneration_pipeline.groovy` files as they don't use nodes to do their jobs. Putting an empty string for `ACTIVE_NODE_TIMEOUT ` in config_regeneration.groovy (what you've done) or putting 0 as that's the current default timeout should be fine.

pipelines/build/common/config_regeneration.groovy
```diff
@@ -513,6 +517,7 @@ return {
     String javaVersion,
     Map<String, Map<String, ?>> buildConfigurations,
     Map<String, ?> targetConfigurations,
+    String activeNodeTimeout,
```
M-Davies, line 520: Can be removed if we're removing if we're not specifying a node timeout in the regen jobs

pipelines/build/common/config_regeneration.groovy
```diff
@@ -539,6 +544,7 @@ return {
             javaVersion,
             buildConfigurations,
             targetConfigurations,
+            activeNodeTimeout,
```
M-Davies, line 547: Can be removed if we're removing if we're not specifying a node timeout in the regen jobs

pipelines/build/common/config_regeneration.groovy
```diff
@@ -60,6 +62,7 @@ class Regeneration implements Serializable {
         this.javaVersion = javaVersion
         this.buildConfigurations = buildConfigurations
         this.targetConfigurations = targetConfigurations
+        this.activeNodeTimeout = activeNodeTimeout
```
M-Davies, line 65: Can be removed if we're removing if we're not specifying a node timeout in the regen jobs

pipelines/build/common/config_regeneration.groovy
```diff
@@ -47,6 +48,7 @@ class Regeneration implements Serializable {
         String javaVersion,
         Map<String, Map<String, ?>> buildConfigurations,
         Map<String, ?> targetConfigurations,
+        String activeNodeTimeout,
```
M-Davies, line 51: Can be removed if we're removing if we're not specifying a node timeout in the regen jobs

pipelines/build/common/config_regeneration.groovy
```diff
@@ -26,6 +26,7 @@ class Regeneration implements Serializable {
     private final String javaVersion
     private final Map<String, Map<String, ?>> buildConfigurations
     private final Map<String, ?> targetConfigurations
+    private final String activeNodeTimeout
```
M-Davies, line 29: Can be removed if we're removing if we're not specifying a node timeout in the regen jobs

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -659,6 +659,40 @@ class Build {
...
+        context.println("No active node matches this label: " + label)
+
+        int activeNodeTimeout = 0
+        if (${buildConfig.ACTIVE_NODE_TIMEOUT}.isInteger()) {
+            activeNodeTimeout = ${buildConfig.ACTIVE_NODE_TIMEOUT} as Integer
+        }
+
+
+        if (activeNodeTimeout > 0) {
+        context.println("Will check again periodically until either one comes online, or " + ${buildConfig.ACTIVE_NODE_TIMEOUT} + " minutes (ACTIVE_NODE_TIMEOUT) has passed.")
```
M-Davies, line 678: Inline if statement

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -659,6 +659,40 @@ class Build {
...
+            int x = 0
+            while (x < activeNodeTimeout) {
+                sleep(60 * 1000)
+                if (NodeHelper.nodeIsOnline(label)) {
+                    context.println("A node which matches this label is now active: " + label)
+                    return
```
M-Davies, line 684: I'm not 100% sure that return statements are allowed in groovy while loops. Has this been tested to verify it works? If it doesn't, [break](https://www.tutorialspoint.com/groovy/groovy_break_statement.htm) should do the same thing

pipelines/build/prTester/pr_test_pipeline.groovy
```diff
@@ -87,12 +87,14 @@ class PullRequestTestPipeline implements Serializable {
             
             String actualJavaVersion = updateRepo ? "jdk${javaVersion}u" : "jdk${javaVersion}"
             def excludedBuilds = ""
+            def activeNodeTimeout = ""
```
M-Davies, line 90: Can be removed if we're removing if we're not specifying a node timeout in the regen jobs (along with passing it into the `regenerationScript` below)

M-Davies (changes requested):
Sorry for the long review, some of my comments are just repeated points to make it easier to change if they need changing 😬

pipelines/jobs/pipeline_job_template.groovy
smlambert: A month ago test pipelines were changed (see https://github.com/AdoptOpenJDK/openjdk-tests/pull/1950) to abort immediately (no timeout) and to date we have seen no issues.  Given this, I think default of 0 is fine.  Perhaps we should have done this in stages, where we first just exit with no activeNodeTimeout (and see if it was actually even needed).

Note, the check is if nothing matches the LABEL being searched for, not if a machine is busy.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -690,6 +722,8 @@ class Build {
                             if (buildConfig.CODEBUILD) {
                                 label = "codebuild"
                             }
+
+                            waitForANodeToBecomeActive(label)
```
karianna, line 726: Can this hang the pipeline forever if no nodes come online?

adamfarley: If it all works, no.

It should only be capable of causing us to pause until ACTIVE_NODE_TIMEOUT number of minutes has passed (default 0), or until a node that matches the labels becomes active.

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: Yes, but would it break out of the IF statement just outside that while loop? Because if not, we'd get an exception.

adamfarley: And no, this hasn't been tested.

pipelines/build/prTester/pr_test_pipeline.groovy
adamfarley: Yes.

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: Sorted.

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: Sorted, though I don't see how this makes it more brief. Thoughts?

pipelines/build/regeneration/jdk9_regeneration_pipeline.groovy
adamfarley: Sorted.

pipelines/build/common/config_regeneration.groovy
adamfarley: Sorted.

pipelines/build/common/config_regeneration.groovy
adamfarley: Sorted.

pipelines/build/common/config_regeneration.groovy
adamfarley: Sorted.

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: I've always been taught that inline IF statements are bad practice, though I don't recall why. Thoughts?

adamfarley:
All issues handled or replied to. Squashing and pushing.

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies: Mainly for brevity sake. Its a lot easier when debugging to see `time: activeNodeTimeout, unit: "MINUTES"` rather than a multiplication sum in the code

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies: Inline as in the line underneath the if statement is not indented so that it's clearly inside the if statement

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies: It shouldn't do as `break` and `continue` statements usually only affect the loop that they have been declared inside. It shouldn't cascade and break the `if (activeNodeTimeout > 0) {` statement

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -659,6 +659,40 @@ class Build {
...
+        if (activeNodeTimeout > 0) {
+        context.println("Will check again periodically until either one comes online, or " + buildConfig.ACTIVE_NODE_TIMEOUT + " minutes (ACTIVE_NODE_TIMEOUT) has passed.")
+            int x = 0
+            while (x < activeNodeTimeout) {
+                context.sleep(time: activeNodeTimeout, unit: "MINUTES")
+                if (NodeHelper.nodeIsOnline(label)) {
+                    context.println("A node which matches this label is now active: " + label)
+                    return
+                }
+                x++
```
andrew-m-leonard, line 686: This loop doesn't look right, the context.sleep() will sleep for activeNodeTimeout "MINUTES" each time through the loop until x == activeNodeTimeout. So if activeNodeTimeout is 5 minutes it is actually going to sleep for 5x5=25 mins ?

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: Actually it is. Github diff display bug, I think.

See that same line here: https://github.com/adamfarley/openjdk-build/blob/timeout_when_no_nodes_found/pipelines/build/common/openjdk_build_pipeline.groovy

adamfarley: Ah, k. Will fix.

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: In which case replacing the return with a break would cause the exception I mentioned.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -659,6 +659,40 @@ class Build {
         }
     }
 
+    def waitForANodeToBecomeActive(def label) {
+        def NodeHelper = context.library(identifier: 'openjdk-jenkins-helper@master').NodeHelper
+
+        if (NodeHelper.nodeIsOnline(label)) {
```
karianna, line 665: I'm wondering if we need to change nodeIsOnline a little here.  What it's really doing is saying that there is 'a' node (perhaps out of several) that is _available_.  It's a subtle distinction but I think an important one.

Even aNodeIsOnline(label) may be more descriptive.

I could also see us enhancing NodeHelper to add a numberOfOnlineNodes(label) or some such.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -659,6 +659,40 @@ class Build {
...
+        context.println("No active node matches this label: " + label)
+
+        int activeNodeTimeout = 0
+        if (buildConfig.ACTIVE_NODE_TIMEOUT.isInteger()) {
+            activeNodeTimeout = buildConfig.ACTIVE_NODE_TIMEOUT as Integer
+        }
+
+
+        if (activeNodeTimeout > 0) {
+        context.println("Will check again periodically until either one comes online, or " + buildConfig.ACTIVE_NODE_TIMEOUT + " minutes (ACTIVE_NODE_TIMEOUT) has passed.")
```
karianna, line 678: Maybe:

"Will check again periodically until a node labeled " + label + " comes online, or " + buildConfig.ACTIVE_NODE_TIMEOUT + " minutes (ACTIVE_NODE_TIMEOUT) has passed."

karianna (commented):
Some more suggestions (not a hill I'm going to die on)

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: Ah, good spot. Fixed.

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: Like the function here?

https://github.com/AdoptOpenJDK/openjdk-tests/pull/1950/files

andrew-m-leonard (approved):

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: Sure. Will change.

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies: I wrote some code to test this
```groovy
x = 15

def waitForANodeToBecomeActive() {
        def returnValue = ""
        while (x < 20) {
                println "X is currently " + x
                sleep(1000) // Sleep for 1 second

                if (x == 17) {
                    println("A node which matches this label is now active: " + "labelname")
                    returnValue = "RETURN BREAKS OUT OF THE FUNCTION"
                    return returnValue
                }

                x++
        }

        returnValue = "RETURN BREAKS OUT OF THE WHILE LOOP"
        return returnValue
}

def retValue = waitForANodeToBecomeActive()
println "RET VALUE = " + retValue
```
and it turns out I was wrong. A return statement inside a function will actually just break out of the function as you originally thought, it doesn't affect the while loop:
```bash
➜  ansible git:(master) ✗ groovy test.groovy
X is currently 15
X is currently 16
X is currently 17
A node which matches this label is now active: labelname
RET VALUE = RETURN BREAKS OUT OF THE FUNCTION
```
As such, I think this is a non issue assuming you want the code to return back to the function caller

adamfarley: That is the intended behaviour, yes. Thanks for checking. :)

M-Davies (approved):
Thanks for implementing those changes! 👍

smlambert:
Removed myself as a reviewer, as 3 reviews seems like plenty already ;)

karianna:
```
TestCompilation > openjdk_build_pipelineTest() FAILED
    org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
    common/openjdk_build_pipeline.groovy: 681: [Static type checking] - Cannot find matching method testDoubles.ContextStub#sleep(java.util.LinkedHashMap <java.lang.String, java.io.Serializable>). Please check if the declared type is correct and if the method exists.
     @ line 681, column 17.
                       context.sleep(time: 1, unit: "MINUTES")
```

adamfarley:
Hmm, looks like the ContextStub class we use for test compilation lacks a stub for the advanced sleep method. Looking to see if it's definitely there in the context class the actual pipeline uses.

PR #2136, opened 2020-10-12
Ensure any old openjdk src repository is removed before re-clone

andrew-m-leonard:
The logic to reclone the openjdk src checks for the absence of .git, but does not clean any old src directory before doing so.

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -77,8 +77,8 @@ checkoutAndCloneOpenJDKGitRepo() {
       exit 1
     fi
   elif [ ! -d "${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/.git" ]; then
-    # If it doesn't exist, clone it
-    echo "Didn't find any existing openjdk repository at $(pwd)/${BUILD_CONFIG[WORKING_DIR]} so cloning the source to openjdk"
+    echo "Didn't find a valid openjdk git repository at $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} so re-cloning the source to openjdk"
+    rm -rf "${BUILD_CONFIG[WORKSPACE_DIR]:?}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
     cloneOpenJDKGitRepo
   fi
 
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
run tests

sxa:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/2135

sbin/prepareWorkspace.sh
```diff
@@ -77,8 +77,8 @@ checkoutAndCloneOpenJDKGitRepo() {
       exit 1
     fi
   elif [ ! -d "${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/.git" ]; then
```
sxa, line 79: I know it's not part of this PR but why is this an `elif` instead of just an `else` given that the condition is the same other than the `!`?

sxa (approved):
One little wording nit but I'm ok with this. We could add a guard to check if the directory exists before doing the `rm` but `-f` means it won't cause a problem either way.

sbin/prepareWorkspace.sh
andrew-m-leonard: I suspect there's some history, possibly this PR https://github.com/AdoptOpenJDK/openjdk-build/commit/b60282a390d0fc0685fef376603716ab84a4ee63
whereby the first if was different before

adamfarley (approved):
LGTM

sbin/prepareWorkspace.sh
```diff
@@ -77,8 +77,8 @@ checkoutAndCloneOpenJDKGitRepo() {
...
-    # If it doesn't exist, clone it
-    echo "Didn't find any existing openjdk repository at $(pwd)/${BUILD_CONFIG[WORKING_DIR]} so cloning the source to openjdk"
+    echo "Didn't find a valid openjdk git repository at $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} so re-cloning the source to openjdk"
+    rm -rf "${BUILD_CONFIG[WORKSPACE_DIR]:?}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
```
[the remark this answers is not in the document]
andrew-m-leonard: yes, I made the same for consistency

andrew-m-leonard:
> # 🟠 PR TESTER RESULT 🟠
> 
> ❎ Some pipelines failed or the job was aborted! ❎
> _See the `pipeline-build-check` below for more information..._

Windows signing failure, unrelated

PR #2154, opened 2020-10-16
Build JDK 15, 16 (x64) on CentOS 6

### LINKED ISSUE #2097: jdk-15+36 linux version can not run with glibc version 2.12
**Platform:**
CentOS 6.10

**Architecture:**
X64

Download page say linux version openjdk 15 work with "Linux glibc version 2.12 or higher"
BUT

```
jdk-15+36]# bin/java -version
bin/java: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /root/jdk-15+36/bin/../lib/libjli.so)
```

OpenJDK 15 build by Oracle works in Centos 6.

aahlenst:
Restores glibc 2.12 support and aligns us with upstream (https://wiki.openjdk.java.net/display/Build/Supported+Build+Platforms).

Fixes #2097.

### DIFF
--- a/pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
@@ -20,7 +20,7 @@ class Config15 {
         x64Linux  : [
                 os                  : 'linux',
                 arch                : 'x64',
-                dockerImage         : 'adoptopenjdk/centos7_build_image',
+                dockerImage         : 'adoptopenjdk/centos6_build_image',
                 dockerFile: [
                         openj9  : 'pipelines/build/dockerFiles/cuda.dockerfile'
                 ],
@@ -34,7 +34,7 @@ class Config15 {
         x64LinuxXL  : [
                 os                  : 'linux',
                 arch                : 'x64',
-                dockerImage         : 'adoptopenjdk/centos7_build_image',
+                dockerImage         : 'adoptopenjdk/centos6_build_image',
                 dockerFile: [
                         openj9  : 'pipelines/build/dockerFiles/cuda.dockerfile'
                 ],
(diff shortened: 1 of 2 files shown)


gdams (approved):

karianna (approved):
LGTM

pshipton:
The OpenJ9 project doesn't support CentOS 6 for Java 15 and later, only glibc level 2.17 and later i.e. CentOS 7.
CentOS 6 is out of support in Nov 2020.
https://www.eclipse.org/openj9/docs/openj9_support/

sxa:
We have had people moving away from adoptopenjdk builds due to not continuing support for the older levels as other distributions have. Are there specific issues with OpenJ9 on glibc 2.12?
I can prepare a PR for switching it back for OpenJ9 but it would be good to understand why we have to diverge on a technical level for OpenJ9

pshipton:
The issue here is customer's not realizing that CentOS 6 isn't supported by OpenJ9. In particular for Java 17 which is expected to be a LTS release. The builds not running on CentOS 6 made that very clear. Even if Java 15+ is built today on CentOS 6, I don't see this continuing for OpenJ9 builds once the move to Adoptium occurs.

It's related to the following statement copied from the OpenJ9 Supported environments document https://www.eclipse.org/openj9/docs/openj9_support/
which is related to the RHEL 6 "Maintenance Support or Maintenance Support 2 ends" date of Nov 30, 2020. Since the last release of Java 15 occurs after this date, OpenJ9 dropped support from the Java 15 release, and for future versions as well.

"When public support for an operating system version ends, OpenJ9 can no longer be supported on that level."

pshipton:
>  Are there specific issues with OpenJ9 on glibc 2.12?

There are no issues today afaik, however the project is free to introduce issues at any time without notice since it's not supported.

PR #2155, opened 2020-10-16
Disable concurrency on pr tester

M-Davies:
Point being that we have pr's stuck in the check status stage for a long time, where their generated job dsl's can be overriden by new jobs. We need to go back to the old system of sequential builds and a queue system as described in the README of the pr-tester...


M-Davies:
Tester is failing due to aarch builds suffering from a laggy filesystem issue (unrelated to the changes in this pr)
```
12:23:37  Updating support/src.zip
12:35:28  wrapper script does not seem to be touching the log file in /home/ubuntu/workspace/build-scripts-pr-tester/build-test/jobs/jdk15u/jdk15u-linux-aarch64-hotspot@tmp/durable-fc09f713
12:35:28  (JENKINS-48300: if on an extremely laggy filesystem, consider -Dorg.jenkinsci.plugins.durabletask.BourneShellScript.HEARTBEAT_CHECK_INTERVAL=86400)
```

karianna:
So are we saying we can't have 2 PR testers in parallel because they override pipelines?

Could we not add a UUID to the temp pipeline in each case?

M-Davies:
> So are we saying we can't have 2 PR testers in parallel because they override pipelines?

They override each other. Say one tester has generated a jdk8 pipeline successfully but it has to sleep when it tries to do the same for a jdk11 one (since jdk11 is building at that moment). Then, another tester comes along and also generates the jdk8 job, overwriting the 1st tester with the 2nd's configuration. It then also sleeps while the jdk11 one is in progress. The 1st tester would then run using the 2nd tester's code and you would get a false positive or negative result.

> Could we not add a UUID to the temp pipeline in each case?

We don't currently have a system for doing that and I think the result would be the same anyway. The tester runs, stops when it realises that a pipeline is in progress and carries on when it's not. So it's still sequential, it just means that you won't have a link to follow in the pr-checks until your job is first in line (it also tells you when your job is in the queue)

karianna:
@gdams Can you see a way around this from your original impl

karianna (approved):
Logic looks OK, I've asked @gdams to comment as well.

gdams (approved):

gdams:
Approved, I also think that we should disable the autorunning to avoid running the PR tester unnecessarily

M-Davies:
> Approved, I also think that we should disable the autorunning to avoid running the PR tester unnecessarily

@gdams We currently have a `no-test` label for PR's that will prevent the tester from running. However, I think that we should have it disabled unless someone comments `run tests`. I could enable `Only use trigger phrase for build triggering` so that only the trigger phrase will execute PR builds?

PR #2156, opened 2020-10-16
Nightly build and test stats job

andrew-m-leonard:

### DIFF
--- a/tools/nightly_build_and_test_stats.groovy
+++ b/tools/nightly_build_and_test_stats.groovy
@@ -21,9 +21,9 @@ node ("master") {
   def buildFailures = 0
   def testStats = []
 
-  // Get the number of "Failed Builds"
+  // Get the number of "Failing Builds"
   stage("getBuildFailures") {
-    def builds = sh(returnStdout: true, script: "wget -q -O - ${jenkinsUrl}/view/Failed%20Builds/api/json")
+    def builds = sh(returnStdout: true, script: "wget -q -O - ${jenkinsUrl}/view/Failing%20Builds/api/json")
     def json = new JsonSlurper().parseText(builds)
     buildFailures = json.jobs.size()
   }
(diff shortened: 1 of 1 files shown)


karianna:
Is this duplicating work that TRSS does?

smlambert:
Much like the [trss-parser](https://www.npmjs.com/package/trss-parser) that George built to do the task of summarizing a release for the TSC issue, this is actually purposefully leveraging the TRSS API to get a specific set of metrics that we are going to track for a while (all in this one spot), to show whether the triage and issue fixing efforts are improving the overall project or not.  

Ideally the trend will be that the following metrics:
- number of build failures
- number of test failures 
- number of tests that are disabled 
tend towards zero where we no longer need to track these metrics to this extent.

andrew-m-leonard:
job: https://ci.adoptopenjdk.net/view/Tooling/job/nightlyBuildAndTestStats/

tools/nightly_build_and_test_stats.groovy
```diff
@@ -0,0 +1,107 @@
...
+        def pipelineJson = new JsonSlurper().parseText(pipeline)
+        if (pipelineJson.size() > 0) {
+          // First in list is last Nightly job
+          def pipeline_id = pipelineJson[0]._id
+          pipelineUrl = pipelineJson[0].buildUrl 
+
+          // Get all child Test jobs for this pipeline job
+          def pipelineTestJobs = sh(returnStdout: true, script: "wget -q -O - ${trssUrl}/api/getAllChildBuilds?parentId=${pipeline_id}\\&buildNameRegex=^Test_.*")
+          def pipelineTestJobsJson = new JsonSlurper().parseText(pipelineTestJobs)
+          if (pipelineTestJobsJson.size() > 0) {
```
smlambert, line 60: Presumably the number of child jobs is not changing often.  It changes when we add more test targets into the build pipelines (which we will do soon, plan to add extended.openjdk).  When this happens, there will be a spike in failures (and disabled jobs).  

It may be worthwhile noting how many children there are, to explain one case where numbers might suddenly change.

smlambert (approved):
LGTM, added one minor comment, in the event you wanted to print out number of children.

tools/nightly_build_and_test_stats.groovy
andrew-m-leonard: Added test job number

karianna (approved):
LGTM

PR #2162, opened 2020-10-20
revert https://github.com/AdoptOpenJDK/openjdk-build/pull/2130

gdams:
fixes: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1630


sxa (changes requested):
Based on https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/build-test/job/jobs/job/jdk/job/jdk-linux-aarch64-openj9/317/consoleFull which still hits the timeout issue that caused this PR to be created I think this can be closed without merging

PR #2170, opened 2020-10-22
Initiate NodeHelper before running build

adamfarley:
There a chance something downstream is preventing test runs on
Docker platforms because the NodeHelper isn't initiated prior to
the build (which was the case prior to an earlier commit).

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -803,6 +803,11 @@ class Build {
                     def cleanWorkspace = Boolean.valueOf(buildConfig.CLEAN_WORKSPACE)
 
                     context.stage("queue") {
+                        // This loads the library containing two Helper classes, and causes them to be
+                        // imported/updated from their repo. Without the library being imported here, runTests 
+                        // method will fail to execute the post-build test jobs for reasons unknown.
+                        context.library(identifier: 'openjdk-jenkins-helper@master')
+
                         if (buildConfig.DOCKER_IMAGE) {
                             // Docker build environment
                             def label = buildConfig.NODE_LABEL + "&&dockerBuild"


lumpfish (approved):
As discussed in Slack, approving this PR because it is not possible to test the pipeline PR change and it is needed for the Oct 2020 release, Will need to back out if necessary.

sxa:
If someone else wants to take the risk on this approach based on "There a chance something downstream" then they are welcome to, but I'm not personally willing to sign this one off for inclusion during a GA cycle. I'd prefer a more conrete understanding that this will resolve things.

"Needed for the Oct 2020 release" isn't quite true as an alternative would be initiating the test jobs manually for the two platforms (Linux/x64 and Linux/aarch64) affected by this.

lumpfish:
Just pointing out that "initiating the test jobs manually for the two platforms (Linux/x64 and Linux/aarch64) affected by this" = manually submitting 216 jobs.

johnoliver:
add to whitelist

johnoliver:
run tests

adamfarley:
Hmm, PR builder error is:

```
org.jenkinsci.plugins.workflow.cps.CpsCompilationErrorsException: startup failed:
Script2.groovy: 753: unable to resolve class NodeHelper 
 @ line 753, column 47.
   NodeToBecomeActive(def label, NodeHelper
                                 ^
Script2.groovy: 804: unable to resolve class NodeHelper 
 @ line 804, column 36.
                           NodeHelper nodeHelper = context.library(identifier: 'openjdk-jenkins-helper@master').NodeHelper
```

adamfarley:
Maybe it'd work if we:
- call "context.library(identifier: 'openjdk-jenkins-helper@master').NodeHelper" in the build method, while ignoring the return object.
- remove NodeHelper from the wait method's params
- and call "context.library(identifier: 'openjdk-jenkins-helper@master').NodeHelper" in the wait method while **not** ignoring the return object.

adamfarley:
Done, with a helpful comment.

Proved to fix the build-test flow on docker, as evidenced here: https://ci.adoptopenjdk.net/job/build-scripts-pr-tester/job/build-test/job/jobs/job/jdk8u/job/jdk8u-linux-x64-openj9/444/console

The only change needed, in the end, was a simple library load. Added a helpful comment so the user knows what's going on.

karianna (approved):
This doesn't look like it can cause any harm and @johnoliver has also reviewed so I'm good with this.

PR #2182, opened 2020-10-26
solaris: only specify freetype location on sparcv9

gdams:
the ansible setup machines seem to have freetype installed in the correct location


sxa (approved):
Let's give it a shot :-)

smlambert (approved):

PR #2193, opened 2020-10-29
Restructure config args to give user args override priority

adamfarley:
Currently, any platform-specific config args are added after the
user's config args, so the user's args get overridden by them.

This is suboptimal, as the user should be in control, regardless of
whether the scripts think a config arg is a good idea.


M-Davies (approved):

karianna (approved):
LGTM

PR #2195, opened 2020-10-29
Ensure CONFIGURE_ARGS isn't replaced on ARM

adamfarley:
This change will ensure arm-specific config options are appended to
CONFIGURE_ARGS_FOR_ANY_PLATFORM, instead of the arm options erasing
and replacing CONFIGURE_ARGS_FOR_ANY_PLATFORM's contents.

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -74,12 +74,12 @@ fi
 
 if [ "${ARCHITECTURE}" == "arm" ]
 then
-  export CONFIGURE_ARGS_FOR_ANY_PLATFORM="--with-jobs=4 --with-memory-size=2000"
+  export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-jobs=4 --with-memory-size=2000"
   if [ "$JAVA_FEATURE_VERSION" -eq 8 ]; then
-    export CONFIGURE_ARGS_FOR_ANY_PLATFORM="$CONFIGURE_ARGS_FOR_ANY_PLATFORM --with-extra-ldflags=-latomic"
+    export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-extra-ldflags=-latomic"
   fi
   if [ "$JAVA_FEATURE_VERSION" -ge 11 ]; then
-    export CONFIGURE_ARGS_FOR_ANY_PLATFORM="$CONFIGURE_ARGS_FOR_ANY_PLATFORM --disable-warnings-as-errors"
+    export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --disable-warnings-as-errors"
   fi
   if [ ! -z "${NUM_PROCESSORS}" ]
   then


sxa (commented):
Is `CONFIGURE_ARGS` a variable that's documented as a variable an end-user might try to use?

We appears to deliberately blank it in [build.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/82d3beb0678f5622e429b2f8f2442a318bfa240c/sbin/build.sh#L48) - so if we really wanted to allow people to use this variable another fix would be not to do that as this proposed solution involves processing the variable and then subsequently blanking it which starts to look like a very odd way to do this.

adamfarley:
> 
> 
> Is `CONFIGURE_ARGS` a variable that's documented as a variable an end-user might try to use?
> 
> We appears to deliberately blank it in [build.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/82d3beb0678f5622e429b2f8f2442a318bfa240c/sbin/build.sh#L48) - so if we really wanted to allow people to use this variable another fix would be not to do that as this proposed solution involves processing the variable and then subsequently blanking it which starts to look like a very odd way to do this.

It is documented (see jenkins pipeline args definitions, screenshot below), and it does seem odd, I agree.

image

I figured the variable was blanked in build.sh to ensure we're starting with a clean variable internally. Maybe it was a mistake. Either way, adding the user's configure args into a parameter-set variable called USER_SUPPLIED_CONFIGURE_ARGS seemed the right place for the user's supplied configure arguments.

Another thing that caught my eye is that we run set-platform-specific-configurations.sh before calling makejdk-any-platform.sh in order to set CONFIGURE_ARGS_FOR_ANY_PLATFORM, we pass that variable in via --configure-args (so it ends up in USER_SUPPLIED_CONFIGURE_ARGS), and then we add that variable to CONFIGURE_ARGS *again* inside build.sh.

If we're set on doing the latter, the former makes little sense to me. Thoughts?

adamfarley:
Turns out this issue is arm-specific. On other platforms, the user's arguments get through fine, though they can be overridden by platform-specific stuff.

Fix has been modified to resolve the underlying arm-specific issue, while also tweaking the existing flow to ensure the user's arguments have priority over the platform arguments.

sxa (approved):
That whole arm section in there likely needs an overhaul to see what's actually necessary and have documentation as to why. The 4 CPU/2000Mb overrides were hard coded for the current scaleway build machines and should probably be defined elsewhere anyway ... Plus we're on borrowed time with those boxes and we expect to be building elsewhere soon.

But since this fixes the issue I'm happy for this to go in :-)

M-Davies (approved):

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -74,7 +74,7 @@ fi
 
 if [ "${ARCHITECTURE}" == "arm" ]
 then
-  export CONFIGURE_ARGS_FOR_ANY_PLATFORM="--with-jobs=4 --with-memory-size=2000"
+  export CONFIGURE_ARGS_FOR_ANY_PLATFORM="$CONFIGURE_ARGS_FOR_ANY_PLATFORM --with-jobs=4 --with-memory-size=2000"
```
karianna, line 77: probably should have braces {}

karianna (changes requested):

build-farm/platform-specific-configurations/linux.sh
adamfarley: Done, and I changed the others in the "arm" section as well, to makie it consistent.

sxa (approved):

adamfarley:
This is ready to go in. Can a committer please merge this?

tellison:
Out of respect to @karianna, who requested changes, I'll leave this for him to update his review status to approve.

karianna (approved):

adamfarley:
Thanks @karianna ! 👍

PR #2197, opened 2020-10-30
Select Xcode 11.7 for building the JDK

aahlenst:
GitHub switched to 12 on Oct 20:
https://github.com/actions/virtual-environments/issues/1712

Should fix build problems that manifest as 
```
configure: error: Xcode 4, 9, 10 or 11 is required to build JDK 8, the version found was 12.0.1. Use --with-xcode-path to specify the location of Xcode or make Xcode active by using xcode-select. configure exiting with result code 1 Makefile:55: *** Cannot continue. Stop.
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,6 +53,22 @@ jobs:
     - name: Build macOS
       run: |
         export JAVA_HOME=$JAVA_HOME_11_X64
+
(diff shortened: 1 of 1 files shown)


gdams (approved):

smlambert (approved):
thanks @aahlenst !

PR #2198, opened 2020-10-30
Give build pipeline user priority control over config options

adamfarley:
The user should always have the ability to specify options on the
commandline. It's useful for debugging, releasing, prototyping fixes,
etc.

However, giving the user a lower priority than the options provided
by the platform-specific scripts means that the user's options can
be overridden, and the user then has to go digging around in
various files to figure out why the scripts aren't doing why they're
told.

I think this is effort wasted, and the user's commands should always
have priority, even if it breaks things.

This command ensures that the user (or at least the surface params
that the user can trivially see and change in jenkins) always has
priority.

### DIFF
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -67,7 +67,8 @@ OPTIONS=""
 
 EXTENSION=""
 # shellcheck disable=SC2034
-CONFIGURE_ARGS_FOR_ANY_PLATFORM=${CONFIGURE_ARGS:-""}
+CONFIGURE_ARGS_FOR_ANY_PLATFORM=""
+CONFIGURE_ARGS=${CONFIGURE_ARGS:-""}
 BUILD_ARGS=${BUILD_ARGS:-""}
 VARIANT_ARG=""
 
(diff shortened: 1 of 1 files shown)


adamfarley:
Test: https://ci.adoptopenjdk.net/view/work-in-progress/job/adamf_aarch64_j9/4/console

See how with-jobs=0 is the last config arg, as it is given persistent priority over all other config args.

Though this is a flawed test, as all user config args on ARM platforms get "lost" prior to the config stage as a result of the issue fixed by https://github.com/AdoptOpenJDK/openjdk-build/pull/2195

karianna (approved):
LGTM

gdams (approved):

adamfarley:
Thank you to the two reviewers.
Merge requested.

tellison:
Kicked off a re-run of the checks

karianna:
Weird - `configure: error: Xcode 4, 9, 10 or 11 is required to build JDK 8, the version found was 12.0.1. Use --with-xcode-path to specify the location of Xcode or make Xcode active by using xcode-select.`

But we landed that fix...

karianna:
Anyhow, shouldn't block this.

adamfarley:
Thanks @karianna :)

PR #2200, opened 2020-10-30
Remove "CONFIGURE_ARGS_FOR_ANY_PLATFORM" variable to avoid confusion

adamfarley:
There's a CONFIGURE_ARGS_FOR_ANY_PLATFORM variable used before we
enter makejdk-all-platforms.sh, and a pair of
CONFIGURE_ARGS_FOR_ANY_PLATFORM variables (one upper case, one lower)
we use sparingly afterwards, but the inner and outer variables do not
seem connected.

This change removes the inner CONFIGURE_ARGS_FOR_ANY_PLATFORM
variables because they are rarely used, can confuse developers by
having multiple inner variables with the same name as the outer
variable, and also because they seem to be vestigial duplicates of
USER_SUPPLIED_CONFIGURE_ARGS. This clarifies the scripts, and also
avoids an extra with-jobs configure arg on ARM when it isn't needed.

Note that the outer variable is untouched, and the user can still set
config variables via makejdk-all-platforms.sh's -C or
--configure-args parameters.


adamfarley:
Test running: https://ci.adoptopenjdk.net/view/work-in-progress/job/adamf-linux-hotspot-build/3/console

karianna (changes requested):
I suspect the README or some other doc will need updating here

adamfarley:
I don't see any reference to CONFIGURE_ARGS_FOR_ANY_PLATFORM inside the build repo, including documentation, besides the references to it that are exclusively in .sh files used prior to makejdk_any_platform.sh.

> 
> 
> I suspect the README or some other doc will need updating here

I don't see any reference to CONFIGURE_ARGS_FOR_ANY_PLATFORM inside the build repo, including documentation, besides the references to it that are exclusively in .sh files used prior to makejdk_any_platform.sh.

Unless there's documentation outside the build repo, I think we're good.

karianna (approved):

configureBuild.sh
```diff
@@ -237,15 +236,17 @@ processArgumentsforSpecificArchitectures() {
       jvm_variant=server,client
       make_args_for_any_platform="DEBUG_BINARIES=true images legacy-jre-image"
     fi
-    configure_args_for_any_platform="--with-jobs=${BUILD_CONFIG[NUM_PROCESSORS]}"
+    if [[ ${BUILD_CONFIG[USER_SUPPLIED_CONFIGURE_ARGS]:-""} != *"--with-jobs"* ]]; then
```
sxa, line 239: Is this change supposed to be in this PR?

sxa (approved):
OK I'll approve this, but let's make sure than after the next build run we do some random checks on the build output just to ensure that nothing that is included in the configure arguments has resulted in a functional change to the executables as a result of this

configureBuild.sh
adamfarley: Yes. It removes the need for the lowercase configure_args_for_any_platform variable, and has the additional benefit of ensuring we have a --with-jobs parameter set here even if one was not set by the user (which seems to be the point of this code, as per the removed line below).

```
BUILD_CONFIG[CONFIGURE_ARGS_FOR_ANY_PLATFORM]=${BUILD_CONFIG[CONFIGURE_ARGS_FOR_ANY_PLATFORM]:-$configure_args_for_any_platform}
```

Currently it looks like the only reason this works is because we don't set CONFIGURE_ARGS_FOR_ANY_PLATFORM anywhere else, which means the user always has at least one --with-jobs arg. But now we're removing CONFIGURE_ARGS_FOR_ANY_PLATFORM, the presence of any text in USER_SUPPLIED_CONFIGURE_ARGS will prevent the required --with-jobs argument being added even if you don't have one already, which I believe to be bad news on armv71.

karianna:
@adamfarley I'm going to land this on Monday so we're around in case of bad-elf factor :-).

PR #2207, opened 2020-11-03
enable Aarch64 jdk8 sanity and extended.functional nightly builds

LongyuZhang:
- enable sanity and extended.functional testing to Aarch64 jdk8 nightly builds
- Issue:AdoptOpenJDK/openjdk-build#2085

Signed-off-by: Longyu Zhang <longyu.zhang@ibm.com>

### DIFF
--- a/pipelines/jobs/configurations/jdk8u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk8u_pipeline_config.groovy
@@ -121,7 +121,10 @@ class Config8 {
                 os                  : 'linux',
                 arch                : 'aarch64',
                 dockerImage         : 'adoptopenjdk/centos7_build_image',
-                test                 : 'default'
+                test                : [
+                        nightly: ['sanity.functional', 'extended.functional', 'sanity.openjdk', 'sanity.system', 'extended.system', 'sanity.perf', 'sanity.external'],
+                        weekly : ['extended.openjdk', 'extended.perf', 'extended.external', 'special.openjdk','special.functional', 'special.system', 'special.perf']
+                ]
         ],
 
         x64LinuxXL       : [


karianna (approved):
LGTM

aahlenst (commented):
I miss the reasoning behind the change, especially in light of https://github.com/AdoptOpenJDK/openjdk-build/pull/2203 which does more or less the opposite for JDK 11. We need better explanation for changes like that (Why have these tests been selected? Why only for that version and this platform? What's the expected impact on test runtimes and machine capacity?). Otherwise this results in another head scratcher when things go awry like during the last release.

cc @smlambert

smlambert:
Yes, given we want to standardize most test lists across all platforms, the special request (https://github.com/AdoptOpenJDK/openjdk-build/issues/2085) to run aarch64 testing on OpenJ9 was handled by just updating/overriding for that platform.  As there is no 'nice way' to do this, and I am still completing my assessment of test duration across platforms, @LongyuZhang was being cautious here to not enable these 2 functional targets on all platforms.  

This is my fault as Longyu actually asked me should he do it in the same way we added jdk11 (and I said yes... without thinking about the fact that I just 'undid' that approach).  

@LongyuZhang - I can actually move these 2 targets up to the nightly/weekly lists in my currently open PR https://github.com/AdoptOpenJDK/openjdk-build/pull/2203 (and verify that we can handle the extra execution time across all platforms... there are very few functional targets that will currently be triggered against hotspot).  Sorry for the confusion.

LongyuZhang:
@smlambert Thanks for clarifying this. Sorry I did not notice your PR before. Sure it would be better if you can add these 2 targets in your list to test. Thanks.

smlambert:
Note: test schedule & duration gathering being done in https://github.com/AdoptOpenJDK/openjdk-tests/issues/2037

PR #2212, opened 2020-11-05
Add VersionInfo support for <fix>

andrew-m-leonard:


andrew-m-leonard:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/2209

karianna (approved):
LGTM

M-Davies:
run tests

pipelines/library/src/common/VersionInfo.groovy
```diff
@@ -148,53 +150,66 @@ class VersionInfo {
...
+        // jdk-11.0.9+11-202011050024
+        // jdk-11.0.9+11-adhoc.username-myfolder
+        //
         //Regexes based on those in http://openjdk.java.net/jeps/223
-        // Technically the standard supports an arbitrary number of numbers, we will support 3 for now
-        final vnumRegex = "(?<major>[0-9]+)(\\.(?<minor>[0-9]+))?(\\.(?<security>[0-9]+))?"
-        final preRegex = "(?<pre>[a-zA-Z0-9]+)"
+        // Technically the standard supports an arbitrary number of numbers, we will support 4 for now
+        final vnumRegex = "(?<major>[0-9]+)(\\.(?<minor>[0-9]+))?(\\.(?<security>[0-9]+))?(\\.(?<fix>[0-9]+))?"
+        final preRegex = "(?<pre>[a-zA-Z]+)"
```
johnoliver, line 166: according to the 223 standard, numbers are allowed in the pre

andrew-m-leonard: from the jdk11 code:
```
      # Only [a-zA-Z] is allowed in the VERSION_PRE. Outer [ ] to quote m4.
      [ VERSION_PRE=`$ECHO "$with_version_pre" | $TR -c -d '[a-z][A-Z]'` ]
      if test "x$VERSION_PRE" != "x$with_version_pre"; then
        AC_MSG_WARN([--with-version-pre value has been sanitized from '$with_version_pre' to '$VERSION_PRE'])
      fi

```

andrew-m-leonard: also the version string is always built as:
```
VERSION_STRING=$VERSION_NUMBER${VERSION_PRE:+-$VERSION_PRE}+$VERSION_BUILD${VERSION_OPT:+-$VERSION_OPT}
```
So wondering if things have moved on somewhat since 223?

pipelines/library/src/common/VersionInfo.groovy
```diff
@@ -148,53 +150,66 @@ class VersionInfo {
...
+        context.println "[WARNING] Failed to match 223 regex."
 
         return false
     }
 
     String formSemver() {
         if (major != null) {
             def semver = major + "." + minor + "." + security
+ 
+            if (fix) {
```
johnoliver, line 210: not sure about this, according to the semver standard you can only have 3 numbers, not sure we can do this and still call it a semver

pipelines/library/src/common/VersionInfo.groovy
```diff
@@ -225,6 +240,10 @@ class VersionInfo {
         if (major != null) {
             def semver = major + "." + minor + "." + security
 
+            if (fix) {
+                semver += "." + fix
```
johnoliver, line 244: not sure about this, according to the semver standard you can only have 3 numbers, not sure we can do this and still call it a semver

pipelines/library/src/common/VersionInfo.groovy
johnoliver: It was updated in http://openjdk.java.net/jeps/322 but that does not really define what PRE means, and implies it is the same as 223

pipelines/library/src/common/VersionInfo.groovy
andrew-m-leonard: See what you mean, I didn't realize is was a standard! Which means OpenJDK are not semver compliant now...
Shall we just keep semver the same?

andrew-m-leonard: same as in just major.minor.security

M-Davies (approved):
Approved so long as the concerns raised by @johnoliver are resolved

andrew-m-leonard:
Thanks @M-Davies , we're still in discussion on what to do...!

pipelines/library/src/common/VersionInfo.groovy
```diff
@@ -148,61 +150,76 @@ class VersionInfo {
...
-
-                major = or0(matched223, 'major')
-                minor = or0(matched223, 'minor')
-                security = or0(matched223, 'security')
-
-                if (matched223.group('pre') != null) {
-                    pre = matched223.group('pre')
-                    context.println "[SUCCESS] regex group (pre) of last 223 regex matched to ${pre}"
-                }
+        String regex = "(?:jdk\\-)?(?<version>${vnumRegex}(\\-${preRegex})?\\+${buildRegex}(\\-${optRegex})?)".toString()
```
johnoliver, line 171: I think this may make build tags required, where as under the previous regex, the second one did not require a build

johnoliver (approved):

PR #2213, opened 2020-11-05
actions: disable macOS PR tester

gdams:
to help free up builds

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,55 +29,55 @@ jobs:
         TARGET_OS: linux
         FILENAME: OpenJDK.tar.gz
 
-  build_macos:
-    name: macOS
-    runs-on: macos-latest
-    strategy:
-      matrix:
-        version: [jdk8u, jdk11u, jdk15u, jdk]
-        vm: [hotspot, openj9]
+  # build_macos:
+  #   name: macOS
+  #   runs-on: macos-latest
+  #   strategy:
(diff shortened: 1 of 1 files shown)


aahlenst (approved):
I just close my eyes and weep inside :see_no_evil:

sxa (approved):
Approved, assuming george will verify it once merged ;-)

karianna:
@gdams Are we enabling this later?

aahlenst:
@karianna There aren't enough macOS workers to keep up with the work. Having to wait >12 hours isn't tenable. So either we add more workers (tough as long as GitHub Actions does not support stateless custom runners) or we come up with a chatbot or something like that that allows us to run the tests only when needed and hereby decrease the load on the available workers.

karianna:
How do we add more workers?

aahlenst:
We cannot at the moment. The problem is https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories and no support for self-hosted elastic agent pools. So either you make one of your magic phone calls (or emails, mind control, or whatever you do in such situations 😁) to increase our allowance (GitHub supposedly uses Orka) or we have to use another CI system. Haven't heard yet from CircleCI. I don't know whether we could use our Jenkins. AppVeyor Server can do all those things (I use it at work), but that would be a separate infra and a lot of work.

PR #2214, opened 2020-11-05
actions: only run builds on pull request

gdams:
currently, we run builds on code that has been merged into master. This is a waste of executor time

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,8 +1,6 @@
 name: Build
 
 on:
-  push:
-    branches: [ master ]
   pull_request:
     branches: [ master ]
 
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -1,8 +1,6 @@
 name: Compile
 
 on:
-  push:
-    branches: [ master ]
   pull_request:
     branches: [ master ]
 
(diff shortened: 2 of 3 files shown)


aahlenst:
I don't think that's a good idea. We don't require PRs to be updated before we merge them. Can you enable all branches instead? This would allow to run the tests on forks and that do not count towards our limits. So people could develop in their accounts and do not have to push WIP PRs just to get the PR tester to run.

gdams:
@aahlenst by the time the PR has been merged it's too late anyway. It's not like anybody actually monitors the branch builds. I'd rather require branches be updated with master before being merged

M-Davies (changes requested):
Regardless whether this is a good idea or not, we probably need to be documenting the pr tester better (all of our code actually 😅). Can you hold off on this until https://github.com/AdoptOpenJDK/openjdk-build/pull/2216 is merged and then add a note in there of these changes?

aahlenst:
> It's not like anybody actually monitors the branch builds.

That can be automated away, for example by notifying the build channel if master breaks.

> I'd rather require branches be updated with master before being merged.

I haven't found an option to enforce this. Furthermore, this would induce and even bigger backlog because mergeable PRs have to be rebased and retested after ever merge to the master branch.

karianna (approved):
LGTM - I agree with George's premise here.

M-Davies (approved):
LGTM now. The documentation I was referencing was designed and merged with this PR in mind

PR #2222, opened 2020-11-10
Revert "Skip solaris sparcv9"

sxa:
Sparc builds are rather slow, but they do complete in an acceptable time for the overnight runs, so we should re-enable them
This reverts commit ec72cf53e326b6b77377c63719c9ba423a7310de.


karianna (approved):
LGTM

gdams (approved):

PR #2224, opened 2020-11-12
add comment trigger for macOS builds

gdams:


gdams:
build-macos

aahlenst:
Can we have something that is more distinctive like `/build-macos`?

PR #2226, opened 2020-11-12
Skip freetype build and bundling on Linux

adamfarley:
Freetype itself has various dependencies on external libraries,
some of which are not present on certain Linux distros.

Better to use the system's freetype library, so the user can handle
their own affairs, reducing the risk of issues (e.g. segmentation
faults on Fedora 28 while starting Eclipse).

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -18,6 +18,9 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=sbin/common/constants.sh
 source "$SCRIPT_DIR/../../sbin/common/constants.sh"
 
+# Bundling our own freetype can cause problems, so we skip that on linux.
+export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
+
 if [ "${ARCHITECTURE}" == "x64" ]
 then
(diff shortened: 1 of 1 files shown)


aahlenst:
Thanks.

Can I get a functional test in https://github.com/AdoptOpenJDK/openjdk-tests or (wherever @smlambert deems it appropriate) that ensures that the Freetype library isn't part of the JDK we ship on Linux but present on Windows probably macOS? I'm unsure about AIX.

aahlenst (commented):
As it's written, it's a config that is only applied on our build farm, not even locally. I don't think that's great.

adamfarley:
Here's some basic testing for Linux x86-64 to ensure we haven't broken anything vital.

https://ci.adoptopenjdk.net/view/work-in-progress/job/adamf-linux-hotspot-build/8/

@aahlenst - To address your points:

- Test? Sure thing. Will work that once Shelley has determined where such a test should go.
- AIX? No, we don't bundle freetype there. I know that because we use the --skip-freetype option in the aix platform script much in the same way I did. 
- Build farm but not locally? Correct, As mentioned above,m the aix platform script does it this way. It's why I assumed that was the correct way to prevent freetype bundling.

karianna (approved):
lgtm

M-Davies:
run tests

M-Davies (approved):

PR #2228, opened 2020-11-13
Stop building shared libs as it hangs on Solaris

sxa:
Since this is currently causing [significant delays](https://github.com/AdoptOpenJDK/openjdk-build/issues/2206) in the Solaris/SPARC pipelines and [this Slack discussion](https://adoptopenjdk.slack.com/archives/C09NW3L2J/p1605110513432600) did not provide a full conclusion as to whether it was needed (I've seen people hit issues with it on ARM builds as well in the past) we should remove this stage unless we can determine whether it is really required for something (metadata is all I can see that might utilise it but if the only reason for the shared lib is to use parseVersion during the build then we should perhaps come up with an alternate method)

[VPC run to check across build environments](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/924/)

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -1110,7 +1110,11 @@ if [[ "${BUILD_CONFIG[ASSEMBLE_EXPLODED_IMAGE]}" == "true" ]]; then
   exit 0
 fi
 
-buildSharedLibs
+# Our Solaris build environment has performance issues so disabling this for now
+# Refhttps://github.com/AdoptOpenJDK/openjdk-build/issues/2206
+if [ "${ARCHITECTURE}" != "sparcv9" ]; then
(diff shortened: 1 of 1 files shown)


sxa:
Adjusted following comment from @karianna saying we should only do this for the platform siginificantly affected by it.
Verifying at https://ci.adoptopenjdk.net/view/work-in-progress/job/SXA-sparcJDK8/23

Also following comments I'm running https://ci.adoptopenjdk.net/view/work-in-progress/job/SXA-sparcJDK8/24 on another agent running on a new, clean user ID on the same build machine

karianna (approved):
LGTM

gdams (approved):

PR #2234, opened 2020-11-16
Allow for null versionData.patch for installer creation and tag sorting

andrew-m-leonard:

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -430,6 +430,9 @@ class Build {
             buildNumber = String.format("%02d", versionData.build)
         }
 
+        // Get version patch number if one is present
+        def patch_version = versionData.patch ?: 0
+
         // Execute installer job
         def installerJob = context.build job: "build-scripts/release/create_installer_windows",
                 propagate: true,
@@ -440,7 +443,7 @@ class Build {
                         context.string(name: 'PRODUCT_MAJOR_VERSION', value: "${versionData.major}"),
                         context.string(name: 'PRODUCT_MINOR_VERSION', value: "${versionData.minor}"),
                         context.string(name: 'PRODUCT_MAINTENANCE_VERSION', value: "${versionData.security}"),
-                        context.string(name: 'PRODUCT_PATCH_VERSION', value: "${versionData.patch}"),
+                        context.string(name: 'PRODUCT_PATCH_VERSION', value: "${patch_version}"),
                         context.string(name: 'PRODUCT_BUILD_NUMBER', value: "${buildNumber}"),
                         context.string(name: 'MSI_PRODUCT_VERSION', value: "${versionData.msi_product_version}"),
                         context.string(name: 'PRODUCT_CATEGORY', value: "jdk"),
@@ -472,7 +475,7 @@ class Build {
                             context.string(name: 'PRODUCT_MAJOR_VERSION', value: "${versionData.major}"),
                             context.string(name: 'PRODUCT_MINOR_VERSION', value: "${versionData.minor}"),
                             context.string(name: 'PRODUCT_MAINTENANCE_VERSION', value: "${versionData.security}"),
-                            context.string(name: 'PRODUCT_PATCH_VERSION', value: "${versionData.patch}"),
+                            context.string(name: 'PRODUCT_PATCH_VERSION', value: "${patch_version}"),
                             context.string(name: 'PRODUCT_BUILD_NUMBER', value: "${buildNumber}"),
                             context.string(name: 'MSI_PRODUCT_VERSION', value: "${versionData.msi_product_version}"),
                             context.string(name: 'PRODUCT_CATEGORY', value: "jre"),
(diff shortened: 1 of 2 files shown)


douph1 (approved):
good

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -430,6 +430,12 @@ class Build {
             buildNumber = String.format("%02d", versionData.build)
         }
 
+        // Get version patch number if one is present
+        def patch_version = 0
+        if (versionData.patch != null) {
+            patch_version = versionData.patch
+        }
```
M-Davies, line 437: Could probably abbreviate to this
```suggestion
        def patch_version = versionData.patch != null ? versionData.patch : 0
```

M-Davies (changes requested):

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: Done, even shorter using ?:

andrew-m-leonard:
Installer now gets a patch of 0 if metadata does not contain a patch.

The ".patch" sorting I added wasn't quite right and hit a sort difference on Windows.
Changed the tag sort algo to add a .0 during the sort if no .patch was present, and also add a tailing "." so the reset of teh line is not used in the sort.

sbin/build.sh
```diff
@@ -792,19 +792,21 @@ getFirstTagFromOpenJDKGitRepo() {
...
+  # that don't have one, and the trailing "." to terminate the 5th field from the +
   # First, sort on build number (B):
-  local jdk11plus_tag_sort1="sort -t+ -k2n"
-  # Second, (stable) sort on (V), (W), (X):
-  local jdk11plus_tag_sort2="sort -t. -k2n -k3n -k4n -k5n"
-  jdk11plus_get_tag_cmd="grep -v _openj9 | grep -v _adopt | sed 's/jdk-/jdk./g' | $jdk11plus_tag_sort1 | nl | $jdk11plus_tag_sort2 | cut -f2- | sed 's/jdk./jdk-/g' | tail -1"
+  local jdk11plus_tag_sort1="sort -t+ -k2,2n"
+  # Second, (stable) sort on (V), (W), (X), (P): P(Patch) is optional and defaulted to "0"
+  local jdk11plus_tag_sort2="sort -t. -k2,2n -k3,3n -k4,4n -k5,5n"
+  jdk11plus_get_tag_cmd="grep -v _openj9 | grep -v _adopt | sed 's/jdk-/jdk./g' | sed 's/+/.0.+/g' | $jdk11plus_tag_sort1 | nl | $jdk11plus_tag_sort2 | sed 's/.0.+/+/g' | cut -f2- | sed 's/jdk./jdk-/g' | tail -1"
```
sxa, line 809: As mentioned I cannot see where the (existing, not added by this PR) `nl` comes into helping produce a stable sort, but this needs to be changed as `.` is a wildcard character. It works by luck, but best to make it right:
```suggestion
  jdk11plus_get_tag_cmd="grep -v _openj9 | grep -v _adopt | sed 's/jdk-/jdk./g' | sed 's/+/.0.+/g' | $jdk11plus_tag_sort1 | nl | $jdk11plus_tag_sort2 | sed 's/\.0\.+/+/g' | cut -f2- | sed 's/jdk./jdk-/g' | tail -1"
```
The extra `.` after the `0` in these translations is superfluous but doesn't do any harm.

sxa (changes requested):

sbin/build.sh
andrew-m-leonard: i worked it out again (I should have written it down the first time!), and added explanation to the original PR here: https://github.com/AdoptOpenJDK/openjdk-build/issues/2101#issuecomment-728779484

andrew-m-leonard: the extra "." is needed so that the -t. -k5,5n ensures the 5th "numeric" sort field does not include the "+",eg.
jdk.11.0.9.1+1
the 5th field using "." field separator is "1+1"
using
jdk.11.0.9.1.0.+1
the 5th field is "1"
also note the extra ".0." is superflous in this example, but not when there is no patch eg.
jdk.11.0.8+1 => 4th field is "8+1", 5th is "blank" 
jdk.11.0.8.0.+1 => 4th field is "8", 5th field is "0"

andrew-m-leonard:
Test build on Windows: https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-jdk11u-windows-x64-hotspot/12/

andrew-m-leonard:
@M-Davies @sxa i've updated the PR and tested on test build.
The problem was related to incorrectly optional sorting of the ".patch" that may or may not be present and the numeric collation ordering differences on Windows cygwin when the "+" was included in the 5th numeric sort field.

sbin/build.sh
sxa: )To be clear, my "superfluous" comment related to the final `.` (i.e. the use of `.0.` instead of just `.0`

sxa (approved):

M-Davies (approved):

sbin/build.sh
andrew-m-leonard: yes, in my example if + was only changed to .0:
jdk.11.0.8+1 => jdk.11.0.8.0+1 => 4th field is "8", 5th field is **"0+"**
**"0+"** won't numerically sort correctly

andrew-m-leonard:
Will merge once final test build completes successfully: https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-jdk11u-windows-x64-hotspot/13/

Issue #1773, opened 2020-05-26
Handle cross-compilation builds where java -version cannot be run


sxa:
RISC-V bulids are currently failing because the end of the pipeline tries to pick up a `version.txt` file which is the output from `java -version` on the final build. In the cross-compile case, you cannot run `java -version` on the build machine. As a temporary workaround we have implemented a [fake version.txt](https://github.com/AdoptOpenJDK/openjdk-build/pull/1766) but this will likely cause complications on the API so we should look to implement something a little less "hacked together" to do something valid in the absence of being able to run the resulting executable.

M-Davies:
Is there no chance of getting a riscv machine to run the builds on and print out the java version?

sxa:
Yes but it would require connecting into a remote machine, copying the build over somewhere, running it and clearing up and the build scripts currently don't do that.

M-Davies:
@sxa Do we have a riscv machine we can use to print out the java version? I could create new job that is tied to said machine which takes in a built jdk, prints out the java version to a file and returns it back to the main pipeline?

sxa:
At the moment we only have one and it tends to be tied up for quite a long time due to the slowness of executing tests etc. so I'd be concerned that it would block the pipelines for long periods of time until a system became available. @Willsparker does `stage4` run `java -version` reliably?

Willsparker:
I believe so, yes. Just checked and it seems okay!

M-Davies:
@sxa Are there any chances of obtaining more riscv systems? Admittedly, my solution in https://github.com/AdoptOpenJDK/openjdk-build/issues/1773#issuecomment-645395878 is not ideal but would running a `java -version` on the machine slow down the pipelines a lot? If all the job would do is take in a artifact, run java -version on it, archive to a txt and exit, would it really take a long time?

sxa:
Happy for you to give it a shot if you think it's easily doable. My only concern would be how much complexity it would add to the pipelines to support that. Otherwise I'm happy with giving it a shot. We're only building one release (JDK11/J9) at the moment so the RISC-V machines will generally be completely free during the builds (i.e. not running any tests)

sxa:
Another option for this would be to use qemu to run it on the local machine, although building a suitable version for CentOS6 is rather non-trivial

Willsparker:
Would it have to be CentOS6 ...?

sxa:
[...]

PR #2238, opened 2020-11-17
Retrieve RISC-V Version String job

M-Davies:
* Instead of retrieving the faked version string, we run a downstream job on a riscv machine after the build that runs java version and passes it back to the pipeline in time for parseVersionOutput()
* Also clears up white spaces in openjdk_build_pipeline.groovy

Fixes: #1773 

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -177,16 +177,16 @@ class Build {
 
         return jdkBranch
     }
-    
+
     /*
     Retrieve the corresponding OpenJDK source code repository. This is used the downstream tests to determine what source code the tests should run against.
     */
     private def getJDKRepo() {
-        
+
         def jdkRepo
         def suffix
         def javaNumber = getJavaVersionNumber()
-        
+
         if (buildConfig.VARIANT == "corretto") {
             suffix="corretto/corretto-${javaNumber}"
         } else if (buildConfig.VARIANT == "openj9") {
@@ -199,15 +199,15 @@ class Build {
             context.error("Unrecognised build variant '${buildConfig.VARIANT}' ")
             throw new Exception()
         }
-        
+
         jdkRepo = "https://github.com/${suffix}"
         if (buildConfig.BUILD_ARGS.count("--ssh") > 0) {
             jdkRepo = "git@github.com:${suffix}"
         }
-        
+
         return jdkRepo
     }
-    
+
     /*
     Run the downstream test jobs based off the configuration passed down from the top level pipeline jobs.
     If we try to call a test job that doesn't exist, the pipeline will not fail but it will print out a warning.
@@ -227,7 +227,7 @@ class Build {
         }
 
         testList.each { testType ->
-			
+
 			// For each requested test, i.e 'sanity.openjdk', 'sanity.system', 'sanity.perf', 'sanity.external', call test job
 			try {
 				context.println "Running test: ${testType}"
@@ -326,12 +326,12 @@ class Build {
                 def signJob = context.build job: "build-scripts/release/sign_build",
                         propagate: true,
                         parameters: params
-                
+
                 // Output notification of downstream failure (the build will fail automatically)
                 def jobResult = signJob.getResult()
                 if (jobResult != 'SUCCESS') {
                     context.println "ERROR: downstream sign_build ${jobResult}.\nSee ${signJob.getAbsoluteUrl()} for details"
-                } 
+                }
 
                 context.node('master') {
                     //Copy signed artifact back and archive again
@@ -376,7 +376,7 @@ class Build {
                         context.string(name: 'CERTIFICATE', value: "${certificate}"),
                         ['$class': 'LabelParameterValue', name: 'NODE_LABEL', label: "${nodeFilter}"]
                 ]
-        
+
         context.copyArtifacts(
                 projectName: "build-scripts/release/create_installer_mac",
                 selector: context.specific("${installerJob.getNumber()}"),
@@ -413,7 +413,7 @@ class Build {
                         context.string(name: 'ARCHITECTURE', value: "${buildConfig.ARCHITECTURE}"),
                         ['$class': 'LabelParameterValue', name: 'NODE_LABEL', label: "${nodeFilter}"]
                 ]
-        
+
     }
 
     /*
@@ -461,7 +461,7 @@ class Build {
         listArchives().each({ file ->
 
             if (file.contains("-jre")) {
-                
+
                 context.println("We have a JRE. Running another installer for it...")
                 def jreinstallerJob = context.build job: "build-scripts/release/create_installer_windows",
                         propagate: true,
@@ -490,8 +490,8 @@ class Build {
                     target: "workspace/target/",
                     flatten: true
                 )
-            } 
-            
+            }
+
         })
     }
 
@@ -532,7 +532,7 @@ class Build {
 
 
     /*
-    Lists and returns any compressed archived contents of the top directory of the build node 
+    Lists and returns any compressed archived contents of the top directory of the build node
     */
     List<String> listArchives() {
         return context.sh(
@@ -693,7 +693,7 @@ class Build {
     /*
     Calculates and writes out the metadata to a file.
     The metadata defines and summarises a build and the jdk it creates.
-    The adopt v3 api makes use of it in its endpoints to quickly display information about the jdk binaries that are stored on github. 
+    The adopt v3 api makes use of it in its endpoints to quickly display information about the jdk binaries that are stored on github.
     */
     def writeMetadata(VersionInfo version, Boolean initialWrite) {
         /*
@@ -820,6 +820,46 @@ class Build {
     }
 
     /*
+    Run the Riscv version reader downstream job.
+    In short, we archive the build artifacts to expose them to the job and run ./java version, copying the output back to here.
+    See riscv_version_out.groovy.
+    */
+    def readRiscvVersionString() {
+        // Archive the artifacts early so we can copy them over to the downstream job
+        try {
+            context.timeout(time: buildTimeouts.BUILD_ARCHIVE_TIMEOUT, unit: "HOURS") {
+                context.archiveArtifacts artifacts: "workspace/target/*"
+            }
+        } catch (FlowInterruptedException e) {
+            context.println "[ERROR] Build archive timeout (${buildTimeouts.BUILD_ARCHIVE_TIMEOUT} HOURS) has been reached. Exiting..."
+            throw new Exception()
+        }
+
+        // Setup params for downstream job & execute
+        String shortJobName = env.JOB_NAME.split('/').last()
+        String copyFileFilter = "${shortJobName}_${env.BUILD_NUMBER}_version.txt"
+
+        def riscvJob = context.build job: "build-scripts/utils/riscv-version-out",
+            propagate: true,
+            parameters: [
+                context.string(name: 'UPSTREAM_JOB_NAME', value: "${env.JOB_NAME}"),
+                context.string(name: 'UPSTREAM_JOB_NUMBER', value: "${env.BUILD_NUMBER}"),
+                context.string(name: 'JDK_FILE_FILTER', value: "OpenJDK*-jdk*_linux_*.tar.gz"),
+                context.string(name: 'FILENAME', value: "${copyFileFilter}")
+            ]
+
+        context.copyArtifacts(
+            projectName: "build-scripts/utils/riscv-version-out",
+            selector: context.specific("${riscvJob.getNumber()}"),
+            filter: "RiscvVersionOuts/${copyFileFilter}",
+            target: "workspace/target/",
+            flatten: true
+        )
+
+        return context.readFile("workspace/target/${copyFileFilter}")
+    }
+
+    /*
     Executed on a build node, the function checks out the repository and executes the build via ./make-adopt-build-farm.sh
     Once the build completes, it will calculate its version output, commit the first metadata writeout, and archive the build results.
     */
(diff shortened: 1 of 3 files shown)


sxa (commented):
Just because I'm picky ... Can you change the title of the PR and the commit message to write the architecture as `RISC-V` please?

M-Davies:
> Just because I'm picky ... Can you change the title of the PR and the commit message to write the architecture as `RISC-V` please?

Sure but it'll still appear as `riscv` in the code as that's what the architecture is listed as in the config file

karianna:
Can this come out of draft now?

M-Davies:
Hitting a node issue at the moment. Java version will not run in the azure docker container, for ref see https://ci.adoptopenjdk.net/job/build-scripts/job/utils/job/riscv-version-out (builds 12-14):
```
13:45:16  ./java: 1: ./java: ELF��: not found
13:45:16  ./java: 2: ./java: @h@8: not found
13:45:16  ./java: 3: ./java: : not found
13:45:16  ./java: 3: ./java: 0: not found
13:45:16  ./java: 20: ./java: Syntax error: Bad function name
```
I'm pretty sure it's running on the binary due to all the debug I'm spitting out to verify it but I get this error no matter where I execute the binary from. It's likely the environment (since it's actually running on the azure build agent, not the riscv machine itself) which begs the question, which riscv machine am I supposed to use from [this list](https://ci.adoptopenjdk.net/label/riscv/)?

sxa:
I would literally go with "anything that's online" which at the moment is just one from that list - so just have it select the `riscv` label and it'll pick whatever we happen to have online at the time. We don't have enough RISC-V systems to get picky about it ;-)

M-Davies:
> I would literally go with "anything that's online" which at the moment is just one from that list - so just have it select the `riscv` label and it'll pick whatever we happen to have online at the time. We don't have enough RISC-V systems to get picky about it ;-)

I am already picking the `riscv` label and it picks up the docker container every time which is why it's failing. I'll try and pick a node from that list and go with that then

sxa:
> > I would literally go with "anything that's online" which at the moment is just one from that list - so just have it select the `riscv` label and it'll pick whatever we happen to have online at the time. We don't have enough RISC-V systems to get picky about it ;-)
> 
> I am already picking the `riscv` label and it picks up the docker container every time which is why it's failing. I'll try and pick a node from that list and go with that then

Hmmm OK we need to fix that - the x64 machines used to build should NOT have `riscv` on them - I set up a tag of `riscvcross` when we were buiding natively.

But if that's been broken use `riscv&&ci.role.test`

M-Davies:
Note when using the jenkins `sh` tag and `java -version`, make sure you specify `2>&1` to redirect to stdout, even if `returnStdout` is set 🤦

M-Davies:
run tests

M-Davies:
> # 🟠 PR TESTER RESULT 🟠
> ❎ Some pipelines failed or the job was aborted! ❎
> _See the `pipeline-build-check` below for more information..._

Failed due to existing OpenJ9 problems https://github.com/eclipse/openj9/issues/11236

M-Davies:
> 1) JSON files aren't being produced (due to early archive)
> * So we try archiving the binaries first, and then the metadata

Fixed by archiving the binaries, and then the metadata 
> 2) JSON data shows it's using the faked version.txt file and not new one
> * Change name of faked version to ensure it's not picked up
> * Add additional check in formMetadata() to look for riscv info

Fixed by description above

See https://ci.adoptopenjdk.net/view/work-in-progress/job/build-scripts-testing/job/jobs/job/jdk11u/job/jdk11u-linux-riscv64-openj9/26/ for evidence this works. I'll run tests one more time to ensure that I haven't broken anything new and open this up for review

sxa (commented):
Since it's Friday evening I'll try and do a review of this next week (We won't want to merge it tonight regardless!) but thanks for working on this and great to see it seemingly working as it is quite possible that we'll need this for other platforms in the future :-D

M-Davies:
> # 🟠 PR TESTER RESULT 🟠
> ❎ Some pipelines failed or the job was aborted! ❎
> _See the `pipeline-build-check` below for more information..._

Failed due to issues with connecting to the docker agent. Existing issue unrelated to this PR https://github.com/AdoptOpenJDK/openjdk-build/issues/2244
```
17:35:46  Compiling /home/adoptopenjdk/workspace/build-scripts-pr-tester/build-test/jobs/jdk8u/jdk8u-linux-x64-hotspot/workspace/build/src/hotspot/src/share/vm/code/dependencies.cpp
17:35:47  Cannot contact azurebuildagentd92090: hudson.remoting.RequestAbortedException: java.io.IOException: Unexpected termination of the channel
01:28:39  Cancelling nested steps due to timeout
01:28:39  Could not connect to azurebuildagentd92090 to send interrupt signal to process
[Pipeline] }
[Pipeline] // timeout
[Pipeline] echo
01:28:39  [ERROR] Build JDK timeout (8 HOURS) has been reached. Exiting...
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
01:28:39  ERROR: Issue with creating launcher for agent azurebuildagentd92090. The agent has not been fully initialized yet
```

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -583,9 +584,10 @@ class Build {
             try {
                 fullVersionOutput = context.readFile(versionPath)
                 context.println "SUCCESS: version.txt found"
-            } catch (NoSuchFileException e) {
-                context.println "ERROR: ${versionPath} was not found. Exiting..."
-                throw new Exception()
+            } catch (Exception e) {
+                context.println "WARNING: ${versionPath} was not found. Likelihood is this is a riscv build which creates a different version file. Attempting to read ${riscvVersionPath}"
```
karianna, line 588: This is a little bit of a red herring - can we detect if we are meant to be looking for riscv and predicate some of this debug accordingly?

pipelines/build/common/riscv_version_out.groovy
```diff
@@ -0,0 +1,116 @@
...
+ *  3. Run java -version and export to file
+ *  4. Expose to the upstream job by archiving file
+ */
+
+// TODO: ADD THE ACTIVE NODE TIMEOUT LOGIC HERE OR GET IT MERGED INTO JOB HELPER (https://github.com/AdoptOpenJDK/openjdk-build/issues/2235)
+
+node ("riscv&&ci.role.test") {
+    timestamps {
+        try {
+            Integer JOB_TIMEOUT = 1
```
karianna, line 21: I suspect some of this can be extracted to local functions/methods or does the DSL prevent this?

karianna (changes requested):

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies: I figured asking for forgiveness was better than asking for permission. Nevertheless, it should be pretty easy to adapt to look for riscv

pipelines/build/common/riscv_version_out.groovy
M-Davies: I stuck it in a separate file mostly to avoid clogging up `openjdk_build_pipeline.groovy`, as that one is quite big already. It doesn't have to be and can be moved across if people think it's better keeping it all together?

M-Davies:
run tests

karianna (approved):

M-Davies:
Running one final test at https://ci.adoptopenjdk.net/view/work-in-progress/job/build-scripts-testing/job/jobs/job/jdk11u/job/jdk11u-linux-riscv64-openj9/27/console but assuming that passes and no more changes are requested, this should be good to merge at an admin discretions

gdams:
One observation (and it doesn't need to be done in this PR). Are you able to generalise this job to be for any cross platform builds. I'm thinking of windows arm64 and macOS arm64 which will likely need the same logic

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -578,11 +579,14 @@ class Build {
 
             // Get Full Version Output
             String versionPath = "workspace/target/metadata/version.txt"
+            if (buildConfig.ARCHITECTURE.contains("riscv")) {
```
sxa, line 582: I wonder if we can modify this conditional to be more of a generic "If we are cross-compiling" (true for any RIS-V specific references in method names etc. here) but that's 'somethign we can change later as we don't have an obvious way to detect it at the moment so would be quite a bit more work :-)

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -819,6 +823,48 @@ class Build {
         return fileName
     }
 
+    /*
+    Run the Riscv version reader downstream job.
```
sxa, line 827:
```suggestion
    Run the RISC-V version reader downstream job.
```

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -819,6 +823,48 @@ class Build {
...
+    In short, we archive the build artifacts to expose them to the job and run ./java version, copying the output back to here.
+    See riscv_version_out.groovy.
+    */
+    def readRiscvVersionString() {
+        // Archive the artifacts early so we can copy them over to the downstream job
+        try {
+            context.timeout(time: buildTimeouts.BUILD_ARCHIVE_TIMEOUT, unit: "HOURS") {
+                context.archiveArtifacts artifacts: "workspace/target/*"
```
sxa, line 835: I'm assuming this doesn't stop the normal archive at the end  which will still be needed to pick up the metadata

pipelines/build/common/riscv_version_out.groovy
```diff
@@ -0,0 +1,116 @@
...
+ * The job looks like:
+ *  1. Switch into a riscv node
+ *  2. Retrieve the artifacts built in the upstream job
+ *  3. Run java -version and export to file
+ *  4. Expose to the upstream job by archiving file
+ */
+
+// TODO: ADD THE ACTIVE NODE TIMEOUT LOGIC HERE OR GET IT MERGED INTO JOB HELPER (https://github.com/AdoptOpenJDK/openjdk-build/issues/2235)
+
+node ("riscv&&ci.role.test") {
```
sxa, line 18: The build machines used for cross compiling shouldn't have the `riscv` tag on them (but as discused elsewhere they seem to at the moment) but we can roll with this for now and update it later

pipelines/build/common/riscv_version_out.groovy
```diff
@@ -0,0 +1,116 @@
...
+                        filter: "workspace/target/${jdkFileFilter}",
+                        fingerprintArtifacts: true,
+                        flatten: true
+                    )
+
+                    println "[INFO] Unzipping..."
+                    sh "tar -zxvf ${jdkFileFilter} && rm ${jdkFileFilter}"
+
+                    String jdkDir = sh(
+                        script: "ls | grep jdk*",
```
sxa, line 64: `*` is superfluous here - suggest removing it to avoid potential confusion (Could also use `ls -d *jdk*`
```suggestion
                        script: "ls | grep jdk",
```

sbin/build.sh
```diff
@@ -538,11 +538,12 @@ printJavaVersionString() {
...
      elif [ "${ARCHITECTURE}" == "riscv64" ]; then
        # riscv is cross compiled, so we cannot run it on the build system
-       # This is a temporary plausible solution in the absence of another fix
+       # So we fake it for now and retrive the version out from a downstream job on a riscv machine after the build
+       echo "Warning: java version can't be run on RISC-V build system. Faking version for now..."
        local jdkversion=$(getOpenJdkVersion)
        local jdkversionNoPrefix=${jdkversion#jdk-}
        local jdkShortVersion=${jdkversionNoPrefix%%+*}
-       cat << EOT > "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/version.txt"
+       cat << EOT > "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/fakedVersion.txt"
```
sxa, line 546: I guess this can be removed now since it should never be required? Again, not critical and we can leave it in for now if it won't cause problems but perhaps a comment indicating it is no longer required as a minimum

sxa (approved):
Approved. Would be good to merge this today and kick off a pipeline just to validate it in production. Comments should be considered suggestions only so I've approved regardless - if you want to include any changes based on my comments feel free but I won't be hurt if you don't :-)

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies: Yes that happens regardless

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies: I think that (and @gdams [suggestion above](https://github.com/AdoptOpenJDK/openjdk-build/pull/2238#issuecomment-733633743)) should probably be done under a separate PR once we have this job running for linux and we're happy with it

M-Davies:
run tests

gdams (approved):

pipelines/build/common/riscv_version_out.groovy
M-Davies: That's probably worth opening an issue for, as it's worrying they have been labelled as such when they actually aren't

sbin/build.sh
M-Davies: Done

pipelines/build/common/riscv_version_out.groovy
sxa: @M-Davies I can see machines with `riscv64` but not `riscv` - did you hit some problem using `riscv`? I've removed `riscv64` from one of the static machines (which is offline anyway) but I'm surprised that it caused a problem with the current machine sets and labels

M-Davies: I don't think I tried `riscv64` as I didn't realise it existed. The issue I had with `riscv` on it's own was that it always `node` into an azure build agent instead of an actual riscv machine. Adding `ci.role.test` ensured it picked up one of George's machines

PR #2240, opened 2020-11-18
Convert false values to empty arrays to prevent boolean error

adamfarley:
Passing "false" in as the value of these variables appears to cause
a Groovy error where it says "No signature of method:
java.lang.Boolean.clear() is applicable for argument types: ()
values: []"

Switching the values to empty arrays, as on the other platforms, to
fix this.


sxa (approved):
Seems reasonable, especially since this has been tested, but @gdams should be the second review of this before we merge.

gdams (approved):

andrew-m-leonard (approved):

adamfarley:
Thanks everyone. :)

PR #2252, opened 2020-11-23
Ensure createOpenJDKTarArchive is in the working directory

andrew-m-leonard:

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -851,6 +851,8 @@ createArchive() {
 
 # Create a Tar ball
 createOpenJDKTarArchive() {
+  stepIntoTheWorkingDirectory
+
   local jdkTargetPath=$(getJdkArchivePath)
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
The build.sh createOpenJDKTarArchive() method was not running in the build source dir, and thus was not always able to get the git tags. The previous method left the current directory in a sub-directory and I think in some circumstances it was still able to find the parent folder tags.

Test arm32 build: https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-jdk8u-linux-arm-hotspot/5/

sxa (approved):
Despite us not being sure why this has started happening (potentially coincident with the new JDK8u tag being created) this sounds like a reasonable fix and it hopefully won't make it any worse so I approve. (If possible please run a pipeline to verify this once merged before the nightlies are done)

karianna (approved):

PR #2259, opened 2020-11-24
Ensure build.sh getFirstTagFromOpenJDKGitRepo works in the workspace root git directory

andrew-m-leonard:
Some of the sbin/build.sh functions had specific dependency on being the the correct directory, but also didn't locate correctly to getJdkVersion().

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -790,6 +790,15 @@ makeACopyOfLibFreeFontForMacOSX() {
 #
 # Excluding "openj9" tag names as they have other ones for milestones etc. that get in the way
 getFirstTagFromOpenJDKGitRepo() {
+
+  # Save current directory of caller so we can return to that directory at the end of this function
+  # Some caller's are not in the git repo root, but instead build/*/images directory like the archive functions
+  # and any function called after removingUnnecessaryFiles()
+  local savePwd="${PWD}"
+
+  # Change to openjdk git repo root to find build tag
+  cd "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+
   # JDK8 tag sorting:
   # Tag Format "jdk8uLLL-bBB"
   # cut chars 1-5 => LLL-bBB
@@ -837,6 +846,9 @@ getFirstTagFromOpenJDKGitRepo() {
   else
     echo "$firstMatchingNameFromRepo"
   fi
+
+  # Restore pwd
+  cd "$savePwd"
 }
 
 createArchive() {
@@ -1024,7 +1036,7 @@ addSemVer() { # Pulls the semantic version from the tag associated with the open
   local fullVer=$(getOpenJdkVersion)
   SEM_VER="$fullVer"
   if [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK8_CORE_VERSION}" ]; then
-    SEM_VER=$(echo "$semVer" | cut -c4- | awk -F'[-b0]+' '{print $1"+"$2}' | sed 's/u/.0./')
+    SEM_VER=$(echo "$SEM_VER" | cut -c4- | awk -F'[-b0]+' '{print $1"+"$2}' | sed 's/u/.0./')
   else
     SEM_VER=$(echo "$SEM_VER" | cut -c5-) # i.e. 11.0.2+12
   fi


andrew-m-leonard:
https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-jdk8u-mac-x64-hotspot/9/console
https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-jdk8u-linux-arm-hotspot/7/

M-Davies:
run tests

M-Davies:
Tester failed due to an error generating the config
```
02:49:24  checking for cpio... /bin/cpio
02:49:24  checking build system type... Invalid configuration `have': machine `have' not recognized
02:49:24  configure: error: /bin/bash /home/adoptopenjdk/workspace/build-scripts-pr-tester/build-test/jobs/jdk8u/jdk8u-linux-x64-openj9/workspace/build/src/common/autoconf/build-aux/config.sub have failed
02:49:24  configure exiting with result code 1
02:49:24  No configurations found for /home/adoptopenjdk/workspace/build-scripts-pr-tester
```

sxa:
@andrew-m-leonard If this is a WIP can you mark it as a draft PR so no-one accidentally tries to merge it?

andrew-m-leonard:
Clean arm32 build: https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-jdk8u-linux-arm-hotspot/10/
Grinder sanity.openjdk: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5001/

andrew-m-leonard:
Mac: https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-jdk8u-mac-x64-hotspot/13/console
Grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5003/

sbin/build.sh
```diff
@@ -837,6 +843,9 @@ getFirstTagFromOpenJDKGitRepo() {
   else
     echo "$firstMatchingNameFromRepo"
   fi
+
+  # Restore pwd
+  cd "$savePwd"
```
sxa, line 848: You could probably just do `cd $OLDPWD` (or even `cd -`) instead of using a new variable to store the location.

sbin/build.sh
```diff
@@ -837,6 +843,9 @@ getFirstTagFromOpenJDKGitRepo() {
   else
     echo "$firstMatchingNameFromRepo"
   fi
+
+  # Restore pwd
```
sxa, line 847: Do you know which piece of code was in the wrong directory when the previous version went in? Would be useful to document it e.g. `# Without this, function blah is not in the right directory` especially if it's not possible for it to ensure it's in the right directory itself.

sxa (commented):
I think I'm ok with this but will leave formal approval til tomorrow. Ref one of my comments I would like to see some trapping of the piece of code that went wrong if at all possible - a 45 byte seemingly gibberish gzip output shouldn't be happening.

karianna (approved):
LGTM, (I agree with @sxa comments)

sbin/build.sh
andrew-m-leonard: Although that would work, I am cautious of using a logic that changes to the "previous", as what if someone later adds some intermediate function call that changes to some directory etc... this logic in this function would then not achieve the desired "spec", which is change back to the directory in which the getFirstTagFromOpenJDKGitRepo() function entered, cd - may not achieve that.

andrew-m-leonard: We can't use cd - as that echo's the pwd, which would then end up in the jdk version!

andrew-m-leonard: Not sure how reliant $OLDPWD is if the previous directory before was the same directory

andrew-m-leonard: Given the questions and risk of someone just adding an extra cd in this function, i'd rather leave it as it is

sbin/build.sh
andrew-m-leonard: i'll add a comment

sxa (approved):
While it still worries me that e.g. `createOpenJDKTarArchive` is not fully self-contained and requires the working directory set by removingUnnecessaryFiles (which is what I understand from your comment) I'm ok to merge this, but I would suggest it's something we might want to improve later in case this happens again. Heavily interdependent functions without well-defined inputs and output requirements negates the value of using functions somewhat.

sxa:
[arm JDK8](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-linux-arm-hotspot/) looking good tonight. Nice work!!!

PR #2269, opened 2020-11-30
doc: add explicit info on disabling nightly testing

sxa:
Add information about disabling nightly testing during release cycles.

Morgan, you mentioned this might need to be changed in the base files too but I wasn't clear on where that would be so is this enough?

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -101,7 +101,7 @@ Don't be scared off by this document! If you already understand the stuff in th
 During the week before release we lock down the openjdk-build repository to
 only include "critical" fixes (i.e. those which will otherwise cause a build
 break or other problem which will prevent shipping the release builds. This
-stops lask minute changes going in which may destabilise things.
+stops last minute changes going in which may destabilise things.
 
 If a change has to go in during this "lockdown" period it should be done by
 posting a comment saying "Requesting approval to merge during the lockdown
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -100,8 +100,9 @@ Don't be scared off by this document! If you already understand the stuff in th
 
 Here are the steps:
 
-1. If desired, add a banner to the website to indicate that the releases are coming in the near future ([Sample PR](https://github.com/AdoptOpenJDK/openjdk-website/pull/702/files))
-2. Build and Test the OpenJDK for "release" at AdoptOpenJDK using a build pipeline job as follows:
+1. Disabling nightly testing so the release builds aren't delayed by any nightly test runs (`ENABLE_TESTS: false` in [config_regeneration.groovy](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/build/common/config_regeneration.groovy#L292)
```
M-Davies, line 103: Ok, I think I can see a better way of doing this than what we've done before

Altering `config_regeneration.groovy` will only unset `ENABLE_TESTS` for the downstream jobs. We would also need to ensure that users remember to unset the Jenkins param `enableTests` (it's enabled by default) on the top level pipeline, otherwise the regenerated config will be overridden and tests will run.

I would suggest instead of trusting user memory,  we instead force the pipelines to regenerate with `runTests` in the [pipeline_job_template.groovy](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/jobs/pipeline_job_template.groovy) set to `false`, just like the PR Tester.

This would mean that `enableTests` is disabled by default on the top level jenkins job, meaning that a user would have to enable `enableTests` to override the regenerated config and allow tests to run.

RELEASING.md
```diff
@@ -96,12 +96,26 @@ Don't be scared off by this document! If you already understand the stuff in th
   
 6. Get permission to submit the release pipeline job from the AdoptOpenJDK TSC members, discussion is via the AdoptOpenJDK #release channel (https://adoptopenjdk.slack.com/messages/CLCFNV2JG).
 
+## Lockdown period
+
+During the week before release we lock down the openjdk-build repository to
+only include "critical" fixes (i.e. those which will otherwise cause a build
+break or other problem which will prevent shipping the release builds. This
+stops lask minute changes going in which may destabilise things.
```
M-Davies, line 104:
```suggestion
stops last minute changes going in which may destabilise things.
```

M-Davies (changes requested):

RELEASING.md
sxa: Thanks @M-Davies ... Comments addressed and (current) links added to the lines that need to be changed. During the next cycle we can modify this to point to the PR that we use instead as an example as I have for some of the other changes referred to in this doc.

M-Davies (approved):
Looks good! Thanks for that

karianna (approved):
LGTM

PR #2274, opened 2020-12-01
Perform openjdk-build git clean after scm checkout

andrew-m-leonard:
Nightly builds do not cleanWorkspace, thus the "workspace" folder can contain 1000s of files which on some machines can take many minutes, sometimes exceeding the Jenkins enforced 10minute timeout for git clean -fdx
This PR disables the git clean on the SCM checkout and explicitly performs the clean afterwards.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -909,6 +909,10 @@ class Build {
             try {
                 context.timeout(time: buildTimeouts.NODE_CHECKOUT_TIMEOUT, unit: "HOURS") {
                     context.checkout context.scm
+
+                    // Perform a git clean outside of checkout to avoid the Jenkins enforced 10 minute timeout
+                    // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1553
+                    context.sh(script: "git clean -fdx")
                 }
             } catch (FlowInterruptedException e) {
                 context.println "[ERROR] Node checkout workspace timeout (${buildTimeouts.NODE_CHECKOUT_TIMEOUT} HOURS) has been reached. Exiting..."
@@ -1083,6 +1087,10 @@ class Build {
                                 try {
                                     context.timeout(time: buildTimeouts.DOCKER_CHECKOUT_TIMEOUT, unit: "HOURS") {
                                         context.checkout context.scm
+
+                                        // Perform a git clean outside of checkout to avoid the Jenkins enforced 10 minute timeout
+                                        // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1553
+                                        context.sh(script: "git clean -fdx")
                                     }
                                 } catch (FlowInterruptedException e) {
                                     context.println "[ERROR] Master docker file scm checkout timeout (${buildTimeouts.DOCKER_CHECKOUT_TIMEOUT} HOURS) has been reached. Exiting..."
--- a/pipelines/build/common/create_job_from_template.groovy
+++ b/pipelines/build/common/create_job_from_template.groovy
@@ -40,7 +40,8 @@ pipelineJob("$buildFolder/$JOB_NAME") {
                     }
                     branch("${GIT_BRANCH}")
                     extensions {
-                        cleanBeforeCheckout()
+                        //repo clean is performed after scm checkout in pipelines/build/common/openjdk_build_pipeline.groovy
+
                         pruneStaleBranch()
                     }
                 }


sxa (approved):
So this is going to prevent us from caching the repository and therefore cause us to re-clone from scratch every time across every build? Seems like a sub-optimal solution.

sxa (changes requested):
So this is going to prevent us from caching the repository and therefore cause us to re-clone from scratch every time across every build? Seems like a sub-optimal solution.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -906,6 +906,13 @@ class Build {
                 }
             }
 
+            try {
+                context.echo("Cleaning any old build workspace from ${env.WORKSPACE}/workspace, as git clean -fdx can take a very long time to do the same.")
+                context.sh(script: "rm -rf ${env.WORKSPACE}/workspace")
```
M-Davies, line 911: Can we have a comment here with an issue link on why this is necessary for brevity (like the windows ones above)? Also, you should also be able to use cleanws instead of the script but that's just a suggestion, I won't be sad if you don't include that

M-Davies (changes requested):

andrew-m-leonard:
@sxa mmm so hadn't quite twigged the openjdk src repo was under there, I meant to clean just the build

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: cleanWs() cleans the whole existing repo, as per @sxa comment we should be leaving the existing cached git repo's

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -906,6 +906,15 @@ class Build {
                 }
             }
 
+            try {
```
karianna, line 909: Doesn't this contravene the `cleanWorkspace` flag?

Also, assuming this was OK to execute, will this fail on Windows as per the special case logic above?

karianna (dismissed):

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: i've added a comment pointing at the issue

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: clean Workspace is different, it remove the "whole" Jenkins workspace ie.removes the existing openjdk-build repo and it's sub openjdk src repo.
This removes just the workspace/build/src/build folder to aid the subsequent git clean -fdx which always gets called to remove non-git files.

andrew-m-leonard: The Windows special case logic is a workaround for the Process locked files issue.

andrew-m-leonard:
@sxa @karianna @M-Davies i've updated after comments, I had meant to only delete the old openjdk build folder not the src repo!

andrew-m-leonard:
On the pLinux machine it is now taking 6seconds to git clean -fdx as opposed to several minutes previously

M-Davies (approved):

sxa (approved):
Hmmmm OK I'll approve this for now, but if it turns out that the problem recurs regardless I'd want it backed out again. As mentioned elsewhere we had problems along these lines with gmake so I'm not sure if we're certain that this will be a permanent fix all the time

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -906,6 +906,15 @@ class Build {
...
+                // Clean any old workspace build folder as it contains 90k+ files which could slow git clean -fdx
+                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1553
+                context.echo("Cleaning any old workspace build from ${env.WORKSPACE}/workspace/build/src/build, as git clean -fdx can take a very long time to do the same.")
```
sxa, line 912: Can we add a link to the issue where this was reported? Also is this a better place to put this logic compared to prepareWorkspace in build.sh?

andrew-m-leonard: agree

andrew-m-leonard: This scm checkout of openjdk-build occurs right up front of the job, prior to the prepareWorkspace.sh being called... (form openjdk-build)

andrew-m-leonard:
@sxa i've changed the implementation so that the "git clean" is not performed during the scm.checkout where it is limited to 10mins, instead we explicitly call it afterwards.

PR #2276, opened 2020-12-01
Do not build mozilla cacerts bundle natively on Solaris/SPARC (too slow on our systems)

sxa:
Because of the slowness issues with java on the SPARC Solaris 10 machine, the [changes to run keytool multiple times during the build](https://github.com/AdoptOpenJDK/openjdk-build/pull/2113) have pushed it over the 8 hour timeout.

As a mitigation, this PR will stop the builds from generating cacerts on Solaris/SPARC and instaead download and use the cacerts file from the Solaris/x64 build. This is not intended to be a permanent solution, but it will allow us to get the JDK8/HotSpot/Solaris-sparc builds running clean again with valid output on this platform.

~(NOTE: This WIP PR currently includes a change to [exclude gradle](https://github.com/AdoptOpenJDK/openjdk-build/pull/2228) as it helps speed up testing, but that will NOT be part of this PR - only the change to prepareWorkspace.sh)~ Now removed again

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -515,7 +515,21 @@ prepareCacerts() {
     echo "Generating cacerts from Mozilla's bundle"
 
     cd "$SCRIPT_DIR/../security"
-    ./mk-cacerts.sh --keytool "${BUILD_CONFIG[JDK_BOOT_DIR]}/bin/keytool"
+    # Our Solaris/SPARC box exhibits slowness when running keytool
+    # Therefore pull the cacerts file from the latest Solaris/x64 build
+    if [ "$ARCHITECTURE" = "sparcv9" ]; then
+      mkdir tmp
+      cd tmp || exit 1
+      wget -q -O jdk8build.tgz "https://api.adoptopenjdk.net/v3/binary/latest/8/ea/solaris/sparcv9/jre/hotspot/normal/adoptopenjdk?project=jdk"
+      CACERTSFILE=`gzip -cd jdk8build.tgz | tar tf - | grep /cacerts$`
+      [ ! -z "$CACERTSFILE" ] && gzip -cd jdk8build.tgz | tar xf - "$CACERTSFILE"
+      cp "$CACERTSFILE" .. || exit 1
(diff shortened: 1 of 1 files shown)


sxa:
Looks like we may have a Solaris build failure in tonight's code, but after a few cycles I'm reasonably confident that this has worked around the problem and will allow the build to complete in less than the timeout period onc ethe code is working again.

(The exclude gradle PR mentioned in the description in addition to this would allow the build to get down to about 3h35)

sbin/prepareWorkspace.sh
```diff
@@ -515,7 +515,21 @@ prepareCacerts() {
     echo "Generating cacerts from Mozilla's bundle"
 
     cd "$SCRIPT_DIR/../security"
-    ./mk-cacerts.sh --keytool "${BUILD_CONFIG[JDK_BOOT_DIR]}/bin/keytool"
+    # Our Solaris/SPARC box exhibits slowness when running keytool
+    # Therefore pull the cacerts file from the latest Solaris/x64 build
+    if [ "$ARCHITECTURE" = "sparcv9" ]; then
+      mkdir tmp
+      cd tmp || exit 1
+      wget -q -O jdk8build.tgz "https://api.adoptopenjdk.net/v3/binary/latest/8/ea/solaris/sparcv9/jre/hotspot/normal/adoptopenjdk?project=jdk"
```
karianna, line 523: Are the certs bundle the same for each Java version?

karianna (changes requested):

sbin/prepareWorkspace.sh
sxa: By Java version do you mean 8 vs 11 vs 15? I haven't confirmed (although I'd hope so - I imagine @aahlenst could give a quick answer!) but since we're only building JDK8 on Solaris/SPARC just now it isn't relevant for the purposes of this fix.

aahlenst:
The contents are always the same. And the format is currently the same, too (Sun/JKS).

karianna (approved):

M-Davies (approved):

Issue #1707, opened 2020-04-24
`Makejdk-any-platform.sh --docker` runs configure on host, before making container


Willsparker:
When running `makejdk-any-platform.sh --docker`, the build scripts will run configure and attempt to set `JDK_BOOT_DIR` on the host machine before building the docker containers. This can lead to situations like this:
 https://ci.adoptopenjdk.net/job/DockerfileCheck/112/label=docker-aws-ubuntu1604-x64-1,variant=j9,version=jdk/console
where the machine doesn't have `javac` in the path, and https://github.com/AdoptOpenJDK/openjdk-build/blob/f9671c487711ab0ec8e8a8ce220572e30abb8d63/sbin/common/common.sh#L141 will fail.

karianna:
@Willsparker - I think this might be a case where we we need to force the docker builds to set the boot-jdk parameter.  I noticed building today that my jdk8u docker build picked up jdk14 as the boot jdk because that's the Java it detected on my host.

Thoughts?

Willsparker:
I agree - I would assume that configure build has to be run before the docker container is built to set some variables.
https://github.com/AdoptOpenJDK/openjdk-build/blob/c57ad18fb56e681892fbc662ee25aff117c0498b/makejdk-any-platform.sh#L53
From the looks, we could override the `JDK_BOOT_DIR` variable by adding it into the JDK specific Dockerfile configure script (i.e. https://github.com/AdoptOpenJDK/openjdk-build/blob/master/docker/jdk14/x86_64/ubuntu/dockerConfiguration.sh ) as the boot jdk will always be in a set place (i.e. `/usr/lib/jvm/jdkXX`).
However, that won't fix the issue I explained earlier, where a host machine may not have `java` or `javac` set- though that could probably be solved by adding an ` elif BUILD_CONFIG[docker] == "docker"` to https://github.com/AdoptOpenJDK/openjdk-build/blob/f9671c487711ab0ec8e8a8ce220572e30abb8d63/sbin/common/common.sh#L126

PR #2280, opened 2020-12-02
Stop setting bootjdk on host when running docker-build

Willsparker:
Fixes: #1707
Fixes: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1725

Stops attempting to find & setting the `JDK-BOOT-DIR` when running `configure_build` on the host machine of a docker-build. This works because the  `BUILD_CONFIG[DOCKER_FILE_PATH]` variable isn't set until right before the docker container is built, therefore it doesn't skip setting the `JDK-BOOT-DIR` when running the build within the docker container. 

Tested / currently testing at: https://ci.adoptopenjdk.net/job/DockerfileCheck/179/ & https://ci.adoptopenjdk.net/job/DockerfileCheck/180/

### DIFF
--- a/sbin/common/common.sh
+++ b/sbin/common/common.sh
@@ -155,48 +155,53 @@ createOpenJDKArchive()
 }
 
 function setBootJdk() {
-  if [ -z "${BUILD_CONFIG[JDK_BOOT_DIR]}" ] ; then
-    echo "Searching for JDK_BOOT_DIR"
+  # Stops setting the bootJDK on the host machine when running docker-build
+  if [ "${BUILD_CONFIG[DOCKER]}" != "docker" ] || ([ "${BUILD_CONFIG[DOCKER]}" == "docker" ] && [ "${BUILD_CONFIG[DOCKER_FILE_PATH]}" != "" ]) ; then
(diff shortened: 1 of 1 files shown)


M-Davies:
run tests

karianna:
@Willsparker - Looks like you have failing tests - are they related?

Willsparker:
That's not related to this fix :-) Looks like a config problem. It's JDK14 as well, which I don't think we _need_ to test for, necessarily.
The main thing is that it doesn't fail because it can't find JDK13 as a boot JDK on the host machine

M-Davies (approved):
LGTM!

karianna (approved):
lgtm

sxa:
@karianna FYI (ref discussion this morning where I mentioned different dockerfiles)

The DockerFileCheck job mentioned above currently tests three Dockerfiles which end-users might use in order to validate that they haven't suffered any bit rot:

- The HotSpot ones we generate for end-users in this repository
- The OpenJ9 ones we generate for end-users in this repository
- The OpenJ9 ones from Eclipse

It does NOT (currently) cover pulling the docker build images we have from dockerhub which we use to build on xLinux ant aarch64Linux

PR #2283, opened 2020-12-02
Check for missing folder shortnames where needed

adamfarley:
If a folder has a space in its name, then we need to substitute it
with its shortname on Windows.

If there isn't one for that folder, then we throw an error here.

It's OK not to find CUDA. It's not OK if the machine config is
*preventing* us from correctly identifying whether CUDA is there.

### DIFF
--- a/build-farm/platform-specific-configurations/windows.sh
+++ b/build-farm/platform-specific-configurations/windows.sh
@@ -149,12 +149,22 @@ then
     fi
 
     CUDA_VERSION=9.0
-    CUDA_HOME="C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v$CUDA_VERSION"
+    CUDA_HOME_FULL="C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v$CUDA_VERSION"
     # use cygpath to map to 'short' names (without spaces)
-    CUDA_HOME=$(cygpath -ms "$CUDA_HOME")
+    CUDA_HOME=$(cygpath -ms "$CUDA_HOME_FULL")
+    if [[ $CUDA_HOME == *" "* ]]; then
+      echo "[ERROR] All CUDA_HOME path folders must have either (a) no spaces, or (b) a shortened version configured in the environment."
+      echo "CUDA_HOME unshortened: ${CUDA_HOME_FULL}"
+      echo "CUDA_HOME shortened: ${CUDA_HOME}"
+      exit 1
+    fi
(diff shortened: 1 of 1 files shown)


adamfarley:
Note: Blocked on https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1729

Will advertise on slack for reviewers once that issue is resolved.

sxa:
> It's OK not to find CUDA

Hmmm I would dispute that. It's not OK for it to find it if we've included a `--with-cuda` parameter. The build should really be aborting if it can't find a suitable cuda to use if you specify the option.

adamfarley:
> 
> 
> > It's OK not to find CUDA
> 
> Hmmm I would dispute that. It's not OK for it to find it if we've included a `--with-cuda` parameter. The build should really be aborting if it can't find a suitable cuda to use if you specify the option.

The original code was designed to simply not enable cuda support if it couldn't find cuda. My statement was in the spirit of the original code's apparent intention.

This does seem erratic though, as some builds would have cuda support and some would not. Will add an error message for "could not find".

adamfarley:
Proof that the "short name" error works: https://ci.adoptopenjdk.net/view/work-in-progress/job/adamf_jdk11_openj9_windows/28/console

```
08:44:56  [ERROR] All CUDA_HOME path folders must have either (a) no spaces, or (b) a shortened version configured in the environment.
08:44:56  CUDA_HOME unshortened: C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v9.0
08:44:56  CUDA_HOME shortened: C:/PROGRA~1/NVIDIA GPU Computing Toolkit/CUDA/v9.0
```

M-Davies (approved):
LGTM

adamfarley:
Ok, Will added the short names earlier on. Here's to check we can actually pass the check stage and build:

https://ci.adoptopenjdk.net/view/work-in-progress/job/adamf_jdk11_openj9_windows/34/

Note that this'll fail at the configure step, as we'll know before then if this is all working.

adamfarley:
Ok, it works just fine. We pass the checks on a machine with valid short names for requisite folders, and now it runs fine.

Ready for the second review & merge.

karianna (approved):
LGTM - I'm going to assume our underlying infra has CUDA already installed properly.

adamfarley:
It's in the ansible playbooks, so it should be fine unless something's broken.

I tested prior forms of this change on several windows machines, and while the short name assignation was erratic, every one had cuda on it.

Thanks all for the reviews and the merge. :)

PR #2284, opened 2020-12-03
Allow speech marks in configure args

adamfarley:
Currently, we lose these speech marks when they get passed from
make-adopt-build-farm.sh to makejdk-any-platform.sh

This change prevents that.

### DIFF
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -152,5 +152,9 @@ export BUILD_ARGS="${BUILD_ARGS} --use-jep319-certs"
 
 echo "$PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh" --clean-git-repo --jdk-boot-dir "${JDK_BOOT_DIR}" --configure-args "${CONFIGURE_ARGS_FOR_ANY_PLATFORM}" --target-file-name "${FILENAME}" ${TAG_OPTION} ${OPTIONS} ${BUILD_ARGS} ${VARIANT_ARG} "${JAVA_TO_BUILD}"
 
+# Convert all speech marks in config args to make them safe to pass in.
+# These will be converted back into speech marks shortly before we use them, in build.sh.
+CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM//\"/temporary_speech_mark_placeholder}"
+
 # shellcheck disable=SC2086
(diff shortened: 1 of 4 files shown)


adamfarley:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/2279

Open to debate; the concept as well as the method. This method works, as demonstrated here:
https://ci.adoptopenjdk.net/view/work-in-progress/job/adamf_jdk11_openj9_windows/33/console

(Note that the job fails because I deliberately used a bad configure argument, to allow the job to shut down normally.)

adamfarley:
Well, previously if I started a Windows jdk11 job, and passed in the configure argument --with-jobs=\"0\" via jenkins, then the configure script would receive the argument --with-jobs=0.

However, this fix preserves the speech marks, so passing in --with-jobs=\"0\" puts --with-jobs="0" in the configure script command.

adamfarley:
- Before: 
Input: --with-jobs=\\"0\\"
Result: configure.sh --with-jobs=0

- After:
Input: --with-jobs=\\"0\\"
Result: configure.sh --with-jobs=\\"0\\"

build-farm/make-adopt-build-farm.sh
```diff
@@ -152,5 +152,8 @@ export BUILD_ARGS="${BUILD_ARGS} --use-jep319-certs"
 
 echo "$PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh" --clean-git-repo --jdk-boot-dir "${JDK_BOOT_DIR}" --configure-args "${CONFIGURE_ARGS_FOR_ANY_PLATFORM}" --target-file-name "${FILENAME}" ${TAG_OPTION} ${OPTIONS} ${BUILD_ARGS} ${VARIANT_ARG} "${JAVA_TO_BUILD}"
 
+# Convert all speech marks in config args to make them safe to pass in.
+CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM//\"/temporary_speech_mark_placeholder}"
```
karianna, line 156: @adamfarley OK I get the intent and your example is great - but I'm not grokking this source code.

Why do we need the long `temporary_speech_mark_placeholder` 'variable' here

adamfarley: It seemed the easiest way to preserve the speech marks as they were passed into the makejdk-any-platform.sh script without interfering with the speech marks around the wider CONFIGURE_ARGS_FOR_ANY_PLATFORM parameter.

I experimented with escaping the escapes of my escapes, but it seemed a convoluted way that was easy to break. This way we can pass the variable around as much as we like, and the speech marks survive until we use them in build.sh

karianna: OK, I get it now - can you add a comment to both lines telling the user where it's twin declaration is?

adamfarley: Sure thing. Uno momento.

adamfarley: Ok, meetings are over. Changes made.
I also updated the docs in a couple of places, to ensure people can use the new feature both pre-and-post repo split.

karianna (approved):
lgtm

joeyleeeeeee97:
@adamfarley  Thanks for your work!  I am starting to try this with our build. Later I will put a valid build URL here .

andrew-m-leonard (approved):

PR #2310, opened 2020-12-16
Migrate head to jdk17

gdams:


pipelines/jobs/configurations/jdk17_pipeline_config.groovy
```diff
@@ -0,0 +1,203 @@
...
+                        hotspot: 'win2016&&vs2019'
+                ],
+                test                : [
+                        nightly: [],
+                        weekly : []
+                ]
+        ],
+
+
+        x32Windows: [
```
aahlenst, line 103: Does that still make sense? i was under the impression that 32 bit on x64 is unmaintained.

aahlenst (approved):

pipelines/jobs/configurations/jdk17_pipeline_config.groovy
gdams: If it builds, I suggest we keep it for now

gdams: We can make a formal support decision when 17 goes GA

andrew-m-leonard (approved):

PR #2312, opened 2020-12-17
fix issue from introducing CROSS_COMPILE env var

gdams:
converts the cross_compile env var to be a build arguement

### DIFF
--- a/pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
@@ -46,13 +46,10 @@ class Config15 {
                 configureArgs       : '--with-noncompressedrefs --enable-dtrace --enable-jitserver'
         ],
 
-        // Currently we have to be quite specific about which windows to use as not all of them have freetype installed
         x64Windows: [
                 os                  : 'windows',
                 arch                : 'x64',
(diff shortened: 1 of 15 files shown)


pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
```diff
@@ -46,13 +46,10 @@ class Config15 {
...
         ],
 
-        // Currently we have to be quite specific about which windows to use as not all of them have freetype installed
         x64Windows: [
                 os                  : 'windows',
                 arch                : 'x64',
-                additionalNodeLabels: [
-                        hotspot: 'win2012&&vs2017'
-                ],
+                additionalNodeLabels: 'win2012&&vs2017',
```
andrew-m-leonard, line 52: i’m a little unsure about the additionalNodeLabels changes? openj9 could run on any build&&windows&&x64 before, this would limit them?

karianna (approved):
lgtm

pipelines/jobs/configurations/jdk15u_pipeline_config.groovy
gdams: We've always built openj9 on win2012, I'm just forcing it to happen

andrew-m-leonard (approved):
good

Issue #2325, opened 2020-12-21
JDK11/J9/RISC-V builds broken


sxa:
Thank you for reporting a broken build! Please fill in the following details:
- **Jenkins link to the failing log:** https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-riscv64-openj9/
- **Which platforms (OS+arch) are affected:** JDK/J9/RISC-V
- **If you look at the historic logs, does the problem appear to be specific to one machine?** No
- **Link to any upstream project bug report (e.g. [OpenJ9](https://github.com/eclipse/openj9/issues)/[JBS](https://bugs.openjdk.java.net/browse))** N/A

https://ci.adoptopenjdk.net/job/build-scripts/job/utils/job/cross-compiled-version-out/ failes like this in the error case:
```

00:08:12  Copied 1 artifact from "build-scripts » jobs » jdk11u » jdk11u-linux-riscv64-openj9" build number 235
[Pipeline] echo
00:08:12  [INFO] Unzipping...
[Pipeline] sh
00:08:23  + tar -zxvf **/OpenJDK*-jdk*.tar.gz
00:08:23  tar (child): **/OpenJDK*-jdk*.tar.gz: Cannot open: No such file or directory
00:08:23  tar (child): Error is not recoverable: exiting now
00:08:23  tar: Child returned status 2
00:08:23  tar: Error is not recoverable: exiting now
[Pipeline] }
```
FYA @M-Davies and @gdams since it looks like we're using the same job for Win/aarch64 too and [this PR](https://github.com/AdoptOpenJDK/openjdk-build/pull/2117) went in around the time it started failing

PR #2327, opened 2020-12-21
Revert to the old linux cross compile filter

M-Davies:
Fixes : #2325 

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -862,7 +862,7 @@ class Build {
         if (buildConfig.TARGET_OS == "windows") {
             filter = "**\\OpenJDK*-jdk*_windows_*.zip"
         } else {
-            filter = "**/OpenJDK*-jdk*.tar.gz"
+            filter = "OpenJDK*-jdk*_linux_*.tar.gz"
         }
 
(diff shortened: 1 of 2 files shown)


sxa (approved):
Approving for test/debug purposes

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -862,7 +862,7 @@ class Build {
         if (buildConfig.TARGET_OS == "windows") {
             filter = "**\\OpenJDK*-jdk*_windows_*.zip"
         } else {
-            filter = "**/OpenJDK*-jdk*.tar.gz"
+            filter = "OpenJDK*-jdk*_linux_*.tar.gz"
```
karianna, line 865: This wouldn't be correct on Mac (for Mac OS X Apple Silicon)

karianna (changes requested):

sxa:
Looks like it's using that `filter` as a parameter to the `tar xf` command. The behaviour has changed in https://github.com/AdoptOpenJDK/openjdk-build/pull/2117 to modify the filter used in all cases to include the `**/` prefix, so at the moment this seems a viable fix. @andrew-m-leonard Are you able to approve this given that @karianna is on vacation (and his concerns - related to if we were to cross-compile on macos/aarch64 in the future - have now been addressed)

gdams (approved):

PR #2336, opened 2020-12-23
jdk16 x32Windows test pipeline config not valid

andrew-m-leonard:

### DIFF
--- a/pipelines/jobs/configurations/jdk16_pipeline_config.groovy
+++ b/pipelines/jobs/configurations/jdk16_pipeline_config.groovy
@@ -103,7 +103,10 @@ class Config16 {
                 buildArgs : [
                         hotspot : '--jvm-variant client,server'
                 ],
-                test                : ['sanity.openjdk']
+                test                : [
+                        nightly: [],
+                        weekly : ['sanity.openjdk', 'sanity.perf', 'sanity.system', 'extended.system']
+                ]
         ],
 
         ppc64Aix    : [


aahlenst (commented):
Why is it not valid? And why is the solution the correct one?

andrew-m-leonard:
test has to be either 'default' or a map list: https://github.com/AdoptOpenJDK/openjdk-build/blob/238ad36cb85d19ce443f6f0a16ea2153ff4c10ce/pipelines/build/common/build_base_file.groovy#L190

aahlenst:
To be more helpful, what I'm looking for:

* A comment in the commit message that somehow explains why the previous one was "not valid" (looks like it's structurally wrong).
* A justification in the pipeline config why we're deviating from the standard test config.

andrew-m-leonard:
@aahlenst commit message updated

sxa (approved):
Approved on the assumption that this will be modified so we're not only running these tests weekly going forward

gdams (approved):

pipelines/jobs/configurations/jdk16_pipeline_config.groovy
```diff
@@ -103,7 +103,10 @@ class Config16 {
                 buildArgs : [
                         hotspot : '--jvm-variant client,server'
                 ],
-                test                : ['sanity.openjdk']
+                test                : [
+                        nightly: [],
+                        weekly : ['sanity.openjdk', 'sanity.perf', 'sanity.system', 'extended.system']
```
smlambert, line 108: You could remove weekly override here completely (though its benign unless we actually had a weekly build pipeline scheduled that would run them or tried to run a release build, which would run both nightly and weekly lists), but given the discussion in Slack, we will plan to address this and other platforms with another issue/PR to enable testing on jdk16

smlambert (approved):

pipelines/jobs/configurations/jdk16_pipeline_config.groovy
andrew-m-leonard: sure, I was just trying to make it consistent with x64Windows

PR #2340, opened 2020-12-24
Adjust JDK15+/J9/Windows builds to pick up correct openssl

sxa:
Update to https://github.com/AdoptOpenJDK/openjdk-build/pull/2339 to pick up the correct OpenSSL build


aahlenst (approved):

sxa:
Pushed another update into here as the original logic was flawed (I'd added it to a `>=11` isntead of `>11` section.

I'm going to force this through without two approvals as it's a build break fix (FYI @karianna)

PR #2348, opened 2021-01-03
More robust looping of timestamp servers

karianna:


gdams (approved):

serverTimestamp.properties
```diff
@@ -0,0 +1,7 @@
+comodaca=http://timestamp.comodoca.com/authenticode
+isectigo=http://timestamp.sectigo.com
+comodocarfc31611=http://timestamp.comodoca.com/rfc3161
```
adamfarley, line 3: Is the final "1" a typo? Should it be "comodocarfc3161"?

karianna: Good find! Fixed :-)

sign.sh
```diff
@@ -62,23 +92,27 @@ signRelease()
...
-        if ! "$signToolPath" sign /f "${SIGNING_CERTIFICATE}" /p "$SIGN_PASSWORD" /fd SHA256 /t http://timestamp.globalsign.com/scripts/timestamp.dll "$f"; then
-          echo "RETRYWARNING: Failed to sign ${f} at $(date +%T): Possible timestamp server error - RC $? ... Retrying in 10 seconds"
-          sleep 10s
-          "$signToolPath" sign /f "${SIGNING_CERTIFICATE}" /p "$SIGN_PASSWORD" /fd SHA256 /t http://timestamp.globalsign.com/scripts/timestamp.dll "$f"
-        fi        
+        for ((i = 0; i < ${#TIMESTAMPKEYS[@]}; i++))
+        do
+          if ! "$signToolPath" sign /f "${SIGNING_CERTIFICATE}" /p "$SIGN_PASSWORD" /fd SHA256 /t "${TIMESTAMPVALUES[i]}" "$f"; then
+            echo "RETRYWARNING: Failed to sign ${f} at $(date +%T): Possible timestamp server error at \"${TIMESTAMPVALUES[i]}\" - RC $? ... Trying new server in 5 seconds"
+            sleep 5s
```
sxa, line 99: If we're just looping through different ones I'd argue this retry delay is probably unneccesary

karianna: I kept that in there in case of temporary localhost or local network issue - if that makes sense I can add a comment to that effect?

sxa: Yep happy with that, although at only seconds it would have to be very temporary :-)

karianna:
rebase sucks

PR #2354, opened 2021-01-04
Ensure dragonwell jdk8 windows64 builds using dragonwell label

andrew-m-leonard:


sxa (approved):
OK, but this is NOT the right answer going forward - we need to ensure that all build machines are able to build all variants and need to make sure we're working towards that so we can revert this.

gdams (approved):

PR #2362, opened 2021-01-06
Clean up the artifacts on master after archiving them elsewhere

adamfarley:
This to prevent artifact accumulation, and eventual issues due
to an over-filled master node.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -508,6 +508,7 @@ class Build {
     /*
     Build installer master function. This builds the downstream installer jobs on completion of the sign and test jobs.
     The installers create our rpm, msi and pkg files that allow for an easier installation of the jdk binaries over a compressed archive.
+    For Mac, we also clean up pkgs on master node from previous runs, if needed (Ref openjdk-build#2350).
     */
     def buildInstaller(VersionInfo versionData) {
         if (versionData == null || versionData.major == null) {
@@ -518,7 +519,7 @@ class Build {
         context.node('master') {
             context.stage("installer") {
                 switch (buildConfig.TARGET_OS) {
-                    case "mac": buildMacInstaller(versionData); break
+                    case "mac": context.sh 'rm -f workspace/target/*.pkg; done'; buildMacInstaller(versionData); break
                     case "linux": buildLinuxInstaller(versionData); break
                     case "windows": buildWindowsInstaller(versionData); break
                     default: return; break


sxa:
Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/2350

Questions
- This PR mentions Windows MSIs as well, but I haven't (so far) seen any indication that those are affected. Do we believe it's a problem there too?
- Removing them after completion would seem to be a change that would make it harder to debug as the "current" one will disappear. Can we perform this delete earlier in the process somewhere so tha the most recent artefacts are retained?

adamfarley:
>     * This PR mentions Windows MSIs as well, but I haven't (so far) seen any indication that those are affected. Do we believe it's a problem there too?

No, that was me covering all bases. The change has been modified to only affect Mac packages.

>     * Removing them after completion would seem to be a change that would make it harder to debug as the "current" one will disappear. Can we perform this delete earlier in the process somewhere so tha the most recent artefacts are retained? Perhaps before the copy from the installer job?

Good idea. Change made.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -508,6 +508,7 @@ class Build {
     /*
     Build installer master function. This builds the downstream installer jobs on completion of the sign and test jobs.
     The installers create our rpm, msi and pkg files that allow for an easier installation of the jdk binaries over a compressed archive.
+    For Mac, we also clean up pkgs on master node from previous runs, if needed.
```
sxa, line 511:
```suggestion
    For Mac, we also clean up pkgs on master node from previous runs, if needed (Ref openjdk-build#2350
```

sxa (approved):
Suggested change to include a reference to the issue that explains why this change is required, but otherwise LGTM. Hopefully we can run a quick pipeline without tests to verify once this is merged.

pipelines/build/common/openjdk_build_pipeline.groovy
adamfarley: Done.

adamfarley: Will remember to do this in future.

M-Davies (approved):
LGTM

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -518,7 +519,7 @@ class Build {
         context.node('master') {
             context.stage("installer") {
                 switch (buildConfig.TARGET_OS) {
-                    case "mac": buildMacInstaller(versionData); break
+                    case "mac": context.sh 'rm -f workspace/target/*.pkg; done'; buildMacInstaller(versionData); break
```
karianna, line 522: Shouldn't this be in a cleanup function of somesort?  I'm curious as to why we don't explicitly do this for say, Windows.

adamfarley: I'm honestly uncertain as to why we don't need to do this for Windows. The msi files don't seem to accumulate like the pkg files do.

I don't think this needs an entire function. I think the function should be consistent in how much trash it leaves behind, and since we don't create trash from a Linux run, we shouldn't do for a Mac run either.

That said, I see no reason we couldn't have a "cleanWs" at the start/end of a run. Perhaps an enhancement?

karianna: Yeah we can open a separate issue for that I guess.

adamfarley:
Ok, the checks have passed, but we should wait until tomorrow morning for the merge so we can give it a run without risking the nightlies.

adamfarley:
Doesn't look like any of the mac nightlies are queued, so let's get this merged and a test run done.

adamfarley:
Post-merge testing seems to be blocked on a networking issue.

https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk15u/job/jdk15u-mac-x64-hotspot/165/console
11:01:49  * What went wrong:
11:01:49  Could not determine the dependencies of task ':shadowJar'.
11:01:49  > Could not resolve all dependencies for configuration ':runtimeClasspath'.
11:01:49     > Could not resolve org.jenkins-ci.main:jenkins-core:2.85.
11:01:49       Required by:
11:01:49           project :
11:01:49        > Could not resolve org.jenkins-ci.main:jenkins-core:2.85.
11:01:49           > Could not get resource 'https://repository.mulesoft.org/nexus/content/repositories/public/org/jenkins-ci/main/jenkins-core/2.85/jenkins-core-2.85.pom'.
11:01:49              > Could not GET 'https://repository.mulesoft.org/nexus/content/repositories/public/org/jenkins-ci/main/jenkins-core/2.85/jenkins-core-2.85.pom'.
11:01:49                 > Read timed out
11:01:49     > Could not resolve org.jenkins-ci.plugins.workflow:workflow-step-api:2.22.
11:01:49       Required by:
11:01:49           project :
11:01:49        > Could not resolve org.jenkins-ci.plugins.workflow:workflow-step-api:2.22.
11:01:49           > Could not get resource 'https://repository.mulesoft.org/nexus/content/repositories/public/org/jenkins-ci/plugins/workflow/workflow-step-api/2.22/workflow-step-api-2.22.pom'.
11:01:49              > Could not GET 'https://repository.mulesoft.org/nexus/content/repositories/public/org/jenkins-ci/plugins/workflow/workflow-step-api/2.22/workflow-step-api-2.22.pom'.
11:01:49                 > Read timed out

sxa:
Above error does not appear to have stopped the build progressing

adamfarley:
Stewart is right. It prevented the last build from progressing (164), so I made assumptions when 165 saw the same error. The file we timeout trying to access now seems accessible though (and it wasn't earlier), so I'm putting 164 down as "upstream server glitch" and moving on.

[Here's an extra PR to remove the json and sha256 files as well.](https://github.com/AdoptOpenJDK/openjdk-build/pull/2363)

adamfarley:
The only pkg files left [(in this build run)](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk15u/job/jdk15u-mac-x64-hotspot/165/console) are the ones from this build. Success!

```
11:41:07  + sha256sum workspace/target/OpenJDK15U-jdk_x64_mac_hotspot_2021-01-07-10-59.pkg

11:41:09  + sha256sum workspace/target/OpenJDK15U-jre_x64_mac_hotspot_2021-01-07-10-59.pkg
```

PR #2374, opened 2021-01-12
Add more robust cycling through Timestamp servers for signing windows artifacts

karianna:


M-Davies (approved):
Code looks good but I imagine this will need testing in a live environment

sxa (commented):
Has this been tested using a list with a known broken one at the start (possible the one we removed in 2374 :-) )? I'd say that's worth doing and if it works then we should be able to merge as the logic will be working :-)

sxa (changes requested):
Can you modify the commit message, otherwise looking through the history and seeing just a reference to a PR number (that won't be elsewhere in the commit logs) means this will be rather opaque

PR #2379, opened 2021-01-13
Add scmReference config for weekly release builds

andrew-m-leonard:
Add into the pipeline jdk job config weekly release build scmReference configuration, allowing the weekend weekly build to actually build the release tags/branches.

### DIFF
--- a/pipelines/build/common/weekly_release_pipeline.groovy
+++ b/pipelines/build/common/weekly_release_pipeline.groovy
@@ -1,3 +1,4 @@
+import groovy.json.*
 
 /*
 Licensed under the Apache License, Version 2.0 (the "License");
@@ -13,11 +14,43 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+stage("Submit Release Pipelines") {
+    // Map of <variant> : <scmRef>
+    def Map<String, String> scmRefs = new JsonSlurper().parseText("${params.scmReferences}") as Map
 
-stage("invoke_pipeline_as_release") {
-    build job: "${params.buildPipeline}",
-          parameters: [
-                          [$class: 'StringParameterValue', name: 'releaseType', value: 'Release']
+    // Map of <platform> : [<variant>,<variant>,..]
+    def Map<String, List<String>> targetConfigurations = new JsonSlurper().parseText("${params.targetConfigurations}") as Map
+
+    def jobs = [:]
+
+    // For each variant create a release pipeline job
+    scmRefs.each{ variant ->
+        def variantName = variant.key
+        def scmRef = variant.value
+        def Map<String, List<String>> targetConfig = [:]
+
+        targetConfigurations.each{ target ->
+            if (target.value.contains(variantName)) {
+                targetConfig.put(target.key,[variantName])
+            }
+        }
+
+        if (!targetConfig.isEmpty()) {
+            echo("Creating ${params.buildPipeline} - ${variantName}")
+            jobs[variantName] = {
+                stage("Build - ${params.buildPipeline} - ${variantName}") {
+                  build job: "${params.buildPipeline}",
+                      parameters: [
+                          string(name: 'releaseType',        value: 'Release'),
+                          string(name: 'scmReference',       value: scmRef),
+                          text(name: 'targetConfigurations', value: JsonOutput.prettyPrint(JsonOutput.toJson(targetConfig)))
                       ]
+                }
+            }
+        }
+    }
+
+    // Submit jobs
+    parallel jobs
 }
 
--- a/pipelines/build/regeneration/build_pipeline_generator.groovy
+++ b/pipelines/build/regeneration/build_pipeline_generator.groovy
@@ -60,6 +60,7 @@ node('master') {
     config.JOB_NAME = "weekly-openjdk${javaVersion}-pipeline"
     config.SCRIPT   = "pipelines/build/common/weekly_release_pipeline.groovy"
     config.PIPELINE = "openjdk${javaVersion}-pipeline"
+    config.weekly_release_scmReferences = target.weekly_release_scmReferences
 
     if (Boolean.parseBoolean(enablePipelineSchedule) == true) {
       try {
--- a/pipelines/jobs/configurations/jdk11u.groovy
+++ b/pipelines/jobs/configurations/jdk11u.groovy
@@ -64,4 +64,12 @@ triggerSchedule_nightly="TZ=UTC\n05 18 * * 2,4"
 // 17:05 Sat
 triggerSchedule_weekly="TZ=UTC\n05 17 * * 6"
 
+// scmReferences to use for weekly release build
+weekly_release_scmReferences=[
+        "hotspot"        : "jdk-11.0.10+8_adopt",
+        "openj9"         : "v0.24.0-release",
+        "corretto"       : "",
+        "dragonwell"     : ""
+]
+
 return this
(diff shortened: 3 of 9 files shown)


andrew-m-leonard:
Test:
https://ci.adoptopenjdk.net/job/build-scripts-testing/job/weekly-openjdk11-pipeline/16/parameters/
https://ci.adoptopenjdk.net/job/build-scripts-testing/job/weekly-openjdk11-pipeline/16/console

pipelines/jobs/configurations/jdk11u.groovy
```diff
@@ -64,4 +64,12 @@ triggerSchedule_nightly="TZ=UTC\n05 18 * * 2,4"
 // 17:05 Sat
 triggerSchedule_weekly="TZ=UTC\n05 17 * * 6"
 
+// scmReferences to use for weekly release build
+weekly_release_scmReferences=[
+        "hotspot"        : "jdk-11.0.10+8_adopt",
```
karianna, line 69: Should that be 11.0.9.1+1 now?

pipelines/build/common/weekly_release_pipeline.groovy
```diff
@@ -13,11 +14,45 @@ See the License for the specific language governing permissions and
...
+            }
+        }
+    }
+
+    // Submit jobs
+    parallel jobs
 }
 
+
+
```
karianna, line 58: Some extra whitespace

karianna (changes requested):
Some nitpicks.

smlambert (approved):
👍 (assume you've verified the scmReference values are the ones you wish to build this weekend).

pipelines/jobs/configurations/jdk11u.groovy
andrew-m-leonard: we already have 11.0.10+8

andrew-m-leonard:
The release scmReferences to be built this weekend, latest "tags" for hotspot and 0.24.0 release branch for openj9:
```
jdk8:
- hotspot : jdk8u282-b07
- openj9: v0.24.0-release
jdk11:
- hotspot : jdk-11.0.10+8_adopt
- openj9 : v0.24.0-release
jdk15:
- hotspot : jdk-15.0.1+9_adopt
- openj9 : v0.24.0-release
```
```
jdk16, jdk17: normal "head"
corretto: "head", no new code or tags since last release
dragonwell: "head", no new code or tags since last release
```

sxa (approved):
Hmmm - I'm not sure whether having the values coded into github requiring a PR to update it is ideal, but we can revisit that later and get this in for now.

Moving forward (i.e. not the run before a release week!) do we want to build it from head (thus running the full set of tests from the most recent code) or the latest tag (to fully test the release process correctly)? I think latest tag is probably the right decision (although it could mean we end up building the same build multiple times!) so couldn't we autodetect it instead of requiring us to submit a PR every time something is tagged?

sxa:
@karianna Happy with the updates?

andrew-m-leonard:
@sxa so detecting latest tag/branch is not quite easy, for example for openj9 it is not a tag, but a branch (v0.24.0-release), so maybe (if possible) determine the latest "created" branch for openj9? (Is suspect sort by v0.??0-release)
hotspot definitely possible
corretto/dragonwell also have different tagging methods, but again just picking "latest" tag possible, but that is not trivial either because of merge commits, which means you can pick the wrong tag, hence our tag sort logic...

I think the philosophy for standard weekend runs is to run the full "release" test set against the latest code, but i'm open for discussion on this?

pipelines/jobs/configurations/jdk17.groovy
```diff
@@ -61,4 +61,12 @@ triggerSchedule_nightly="TZ=UTC\n30 03 * * 3,5"
 // 12:05 Sun
 triggerSchedule_weekly="TZ=UTC\n05 12 * * 7"
 
+// scmReferences to use for weekly release build
+weekly_release_scmReferences=[
```
M-Davies, line 65: Can we add a note to `RELEASING.md` that this variable will need to be adjusted when JDK head is bumped?

M-Davies (changes requested):

sxa:
> maybe (if possible) determine the latest "created" branch for openj9? 

Yep that's whan I envisaged at least once the milestone builds are available. Between the release and the next milestone I guess we'd probably just want to build master's HEAD to get full test coverage on the new code

andrew-m-leonard:
> > maybe (if possible) determine the latest "created" branch for openj9?
> 
> Yep that's whan I envisaged at least once the milestone builds are available. Between the release and the next milestone I guess we'd probably just want to build master's HEAD to get full test coverage on the new code

That's the problem, how do we automatically know to switch from v0.24.0-release branch back to openj9 branch ?

smlambert:
We arrived at this approach for weekly 'release' builds from the need for 2 separate things... 
1) vet the release process (original plan was to do this once, one week before a release)
2) run the larger suite of tests more often (original plan to use head)

I think what @andrew-m-leonard has here is great, as it allows for regular weekly runs to use head, and then the week before any given release, to reuse the mechanism for the tagged branches.  Does this need to be automated, that would be 'neat', but I wouldn't say its required if we use the approach (head mostly, tagged branches just before a release).

andrew-m-leonard:
> We arrived at this approach for weekly 'release' builds from the need for 2 separate things...
> 
>     1. vet the release process (original plan was to do this once, one week before a release)
> 
>     2. run the larger suite of tests more often (original plan to use head)
> 
> 
> I think what @andrew-m-leonard has here is great, as it allows for regular weekly runs to use head, and then the week before any given release, to reuse the mechanism for the tagged branches. Does this need to be automated, that would be 'neat', but I wouldn't say its required if we use the approach (head mostly, tagged branches just before a release).

Agree, I think let's keep it simple, I know from many issues that trying to find latest jdk tags and sorting is a nightmare!!

pipelines/jobs/configurations/jdk17.groovy
andrew-m-leonard: Do you mean reset back to "" HEAD ?

andrew-m-leonard: added

M-Davies (approved):

pipelines/build/common/weekly_release_pipeline.groovy
```diff
@@ -1,3 +1,4 @@
+import groovy.json.*
```
karianna, line 1: Still prefer explicit imports :-)

pipelines/build/regeneration/build_pipeline_generator.groovy
```diff
@@ -60,6 +60,7 @@ node('master') {
     config.JOB_NAME = "weekly-openjdk${javaVersion}-pipeline"
     config.SCRIPT   = "pipelines/build/common/weekly_release_pipeline.groovy"
     config.PIPELINE = "openjdk${javaVersion}-pipeline"
+    config.weekly_release_scmReferences = target.weekly_release_scmReferences
```
karianna, line 63: The variables above a set globally (if we're following CAPITAL convention) - why not this one?

karianna (approved):
Two more nitpicks but they'll do.

pipelines/build/regeneration/build_pipeline_generator.groovy
andrew-m-leonard: yes, I was following the convention of targetConfigurations.. I saw these as well

sxa:
> Agree, I think let's keep it simple, I know from many issues that trying to find latest jdk tags and sorting is a nightmare!!

True, although since we've already done it for `getFirstTagFromOpenJDKGitRepo` we shoulkd be able to re-use it without it being too much of a nightmare :-D

PR #2392, opened 2021-01-19
Disable nightly testing during January 2021 release cycle

sxa:
Disabled as per the instructions now in https://github.com/AdoptOpenJDK/openjdk-build/blob/master/RELEASING.md

@andrew-m-leonard  I assume none of your recent scheduling changes will invalidate this way of disabling the tests

Signed-off-by: Stewart X Addison <sxa@redhat.com>

PR #2401, opened 2021-01-22
Revert "Disable nightly testing during January 2021 release cycle (#2392)

sxa:
GAs are mostly out at the time of writing so we should be safe to re-enable.

This reverts commit ea53ac3d1623de5ecefdcd9aee3379a776be84b2.


M-Davies (approved):

andrew-m-leonard (approved):
lgtm

PR #2403, opened 2021-01-25
Disable jdk15u pipeline now that it is retired

andrew-m-leonard:

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -262,6 +262,8 @@ When the repo has been created, a few changes to the codebase will be necessary 
 ### Post Release Tasks
 Once all the release binaries have been published the following tasks should be completed:
 1. Reset the "weekly_release_scmReferences" (change to "") for the weekend release test build so it is using HEAD streams: https://github.com/AdoptOpenJDK/openjdk-build/tree/master/pipelines/jobs/configurations
+2. If the latest version just released has come to the end of its non-LTS lifecycle (2 CPU updates, eg.jdk-15.0.2), then disable and retire that version form the Nightly pipeline builds:
+   - eg.https://github.com/AdoptOpenJDK/openjdk-build/pull/2403/files
 
 ## Summary on point releases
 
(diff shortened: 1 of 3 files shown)


sxa (changes requested):
Looking through RELEASING.md I don't see any references to this part of the process, so I'd like to that added before merging this as it seems like a reasonable thing to do once the final version of any release stream has gone out

RELEASING.md
```diff
@@ -262,6 +262,8 @@ When the repo has been created, a few changes to the codebase will be necessary
 ### Post Release Tasks
 Once all the release binaries have been published the following tasks should be completed:
 1. Reset the "weekly_release_scmReferences" (change to "") for the weekend release test build so it is using HEAD streams: https://github.com/AdoptOpenJDK/openjdk-build/tree/master/pipelines/jobs/configurations
+2. If the latest version just released has come to the end of its non-LTS lifecycle (2 CPU updates, eg.jdk-15.0.2), then disable and retire that version form the Nightly pipeline builds:
+   - eg.https://github.com/AdoptOpenJDK/openjdk-build/pull/2403/files
```
aahlenst, line 266: Does anyone know the upstream policy here? We had some irregular releases with 8 and 11. Is it impossible that OpenJDK releases 15.0.2.1? I'm not talking about Azul's MTS work here.

karianna (approved):
lgtm

RELEASING.md
karianna: Hi Andreas - there should not be a 15.0.2.1 (Oracle is pretty strict about saying 'tough luck', just go to 16).  We can always enable if need be.

gdams (approved):

sxa (approved):

M-Davies:
For future reference, the [jdk15 job generator](https://ci.adoptopenjdk.net/job/build-scripts/job/utils/job/pipeline_jobs_generator_jdk15/) should also be disabled as part of this PR.

PR #2405, opened 2021-01-25
More robust Timestamp Server cycling for Windows Signing

karianna:


gdams (approved):

sxa (changes requested):
Not sure why this PR was recreated but it means my previous objection was lost (and hasn't been addressed):

_Can you modify the commit message, otherwise looking through the history and seeing just a reference to a PR number (that won't be elsewhere in the commit logs) means this will be rather opaque_

karianna:
@sxa done

sxa (approved):
Superficially seems logically sound, although I'd recommend verifying with a "broken" timestamp server (such as the original one we had to remove) before merging this if that hasn't already been done.

sxa (changes requested):
I'm not convinced this works - when I add some debug in it doesn't look like it's going round the for loop at all. Needs to be tested

```
+ read -r f
+ echo 'Signing ./bin/api-ms-win-core-console-l1-1-0.dll'
Signing ./bin/api-ms-win-core-console-l1-1-0.dll
+ (( i = 0 ))
+ (( i < 0 ))
+ read -r f
```

sxa:
Maybe it's just late in the day, but isn't this going to sign the file multiple times - one per timestamp server rather than retry with different ones?

sxa:
Given that the key is irrelevant I'd be inclined to go with **_something_** like this which (at least to me, you may disagree) looks a lot simpler although I'd even be tempted to inline the list of servers instead of using the properties file ;-)
```
TIMESTAMPSERVERS= `cut -d= -f2 < $TIMESTAMP_SERVER_CONFIG`

STAMPED=false
for SERVER in $TIMESTAMPSERVERS; do
  "$signToolPath" sign /f "${SIGNING_CERTIFICATE}" /p "$SIGN_PASSWORD" /fd SHA256 /t "${TIMESTAMPVALUES[i]}" "$f"
  if [ $? -eq 0 ]; then
      STAMPED=true;
  else
      echo "RETRYWARNING: Failed to sign ${f} at $(date +%T): Possible timestamp server error at ${SEVER} - Trying new server in 5 seconds"
      sleep 5
   fi
done
```

sxa:
Getting quite a lot of failures in the [sign_build](https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/sign_build/) job today.

andrew-m-leonard (approved):
lgtm

sign.sh
```diff
@@ -62,24 +92,30 @@ signRelease()
...
-      echo "$FILES" | while read -r f;
-      do
-        echo "Signing ${f}"
-        if ! "$signToolPath" sign /f "${SIGNING_CERTIFICATE}" /p "$SIGN_PASSWORD" /fd SHA256 /t http://timestamp.globalsign.com/scripts/timestamp.dll "$f"; then
-          echo "RETRYWARNING: Failed to sign ${f} at $(date +%T): Possible timestamp server error - RC $? ... Retrying in 10 seconds"
-          sleep 10s
-          "$signToolPath" sign /f "${SIGNING_CERTIFICATE}" /p "$SIGN_PASSWORD" /fd SHA256 /t http://timestamp.globalsign.com/scripts/timestamp.dll "$f"
-        fi
-      done
+        echo "$FILES" | while read -r f;
```
andrew-m-leonard, line 107: Shouldn't this be cat "$FILES" | ... ?

andrew-m-leonard (changes requested):

sign.sh
andrew-m-leonard: Actually hmm, that's not right either

sign.sh
```diff
@@ -62,24 +92,30 @@ signRelease()
...
+        do
+          echo "Signing ${f}"
+          for ((i = 0; i < ${#TIMESTAMPKEYS[@]}; i++))
+          do
+            # We still use a 5 second delay between attempts in case of a temporary localhost and/or local network condition
+            if ! "$signToolPath" sign /f "${SIGNING_CERTIFICATE}" /p "$SIGN_PASSWORD" /fd SHA256 /t "${TIMESTAMPVALUES[i]}" "$f"; then
+              echo "RETRYWARNING: Failed to sign ${f} at $(date +%T): Possible timestamp server error at \"${TIMESTAMPVALUES[i]}\" - RC $? ... Retrying in 5 seconds"
+              sleep 5s
+            fi
```
andrew-m-leonard, line 116: Does need to break if it signs successfully?

sign.sh
sxa: No, otherwise you'd be catting DLLs which gets rather messy

sign.sh
sxa: Almost certainly and it's why I made the comment about it looping signing multiple times https://github.com/AdoptOpenJDK/openjdk-build/pull/2405#issuecomment-766997573

sxa:
I'm working on redoing this under https://github.com/AdoptOpenJDK/openjdk-build/pull/2409

karianna:
Closing in favour of Stewarts (his is less code - that is good.

PR #2409, opened 2021-01-26
Use multiple backup timeservers for Windows signing

sxa:
This does the following:

- Implements timestamp server fallback as originally attempted in https://github.com/AdoptOpenJDK/openjdk-build/pull/2405 and https://github.com/AdoptOpenJDK/openjdk-build/pull/2374 - this supercedes both
- Adjusts the ordering so that the "known good" server is first in the list, reducing the likelihood of failure (globalsign is no longer working today)
- Unifies the `exe` and `dll` parts in a single `find`
- Adjusts the way the JDK directory is detected, thus removing the `find: 'basename' terminated by signal 13` that have been polluting the logs
- Uses `-q` on the `unzip` command in order to avoid filling the log with a list of all files in the JDK

The latest runs of [sign_build](https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/sign_build) have been from this branch

I am somewhat worried that none of the timestamp servers we have are reliable though...

### DIFF
--- a/sign.sh
+++ b/sign.sh
@@ -33,6 +33,16 @@ WORKSPACE=$(pwd)
 TMP_DIR_NAME="tmp"
 TMP_DIR="${WORKSPACE}/${TMP_DIR_NAME}/"
 
+# List of valid timestamp servers:
+# http://timestamp.comodoca.com/authenticode -> OK 02/08/2030 -> Sectigo RSA Time Stamping Signer #1
+# http://timestamp.sectigo.com -> OK 02/08/2030 -> Sectigo RSA Time Stamping Signer #1 .. same as previous but with another url
+# http://timestamp.comodoca.com/rfc3161 -> OK 02/08/2030 -> Sectigo RSA Time Stamping Signer #1 .. same as previous but with another url
+# http://tsa.startssl.com/rfc3161 -> OK 15/08/2028 -> WoSign Time Stamping Signer ( buyed by WoTrus )
+# http://tsa.starfieldtech.com -> OK 17/09/2027 -> Starfield Timestamp Authority - G2
+# http://timestamp.globalsign.com/scripts/timstamp.dll -> OK 24/06/2027 -> GlobalSign TSA for MS Authenticode - G2
+# http://timestamp.digicert.com -> OK 22/10/2024 -> DigiCert Timestamp Responder
+TIMESTAMP_SERVER_CONFIG="./serverTimestamp.properties"
+
 checkSignConfiguration() {
   if [[ "${OPERATING_SYSTEM}" == "windows" ]] ; then
     if [ ! -f "${SIGNING_CERTIFICATE}" ]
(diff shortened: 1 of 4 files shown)


sxa:
Based on https://ci.adoptopenjdk.net/job/build-scripts/job/release/job/sign_build/14493/console the comodoca servers seems really slow but hopefully it'll get there so we can merge this. Need to verify whether any occurrences of `Failed to sign ${f} using any time server - aborting` show up as I've had to make some adjustments to stop the exit for not aborting completely.

andrew-m-leonard (approved):
good

sign.sh
```diff
@@ -52,35 +62,37 @@ checkSignConfiguration() {
...
+      for f in $FILES
       do
         echo "Signing ${f}"
-        if ! "$signToolPath" sign /f "${SIGNING_CERTIFICATE}" /p "$SIGN_PASSWORD" /fd SHA256 /t http://timestamp.globalsign.com/scripts/timestamp.dll "$f"; then
-          echo "RETRYWARNING: Failed to sign ${f} at $(date +%T): Possible timestamp server error - RC $? ... Retrying in 10 seconds"
-          sleep 10s
-          "$signToolPath" sign /f "${SIGNING_CERTIFICATE}" /p "$SIGN_PASSWORD" /fd SHA256 /t http://timestamp.globalsign.com/scripts/timestamp.dll "$f"
+        STAMPED=false
+        for SERVER in $TIMESTAMPSERVERS; do
+          if [ "$STAMPED" = "false" ]; then
```
karianna, line 79: Does that need to be `==`? and two `[[`?

karianna (changes requested):

sign.sh
sxa: Simple answer: No

Long answer: What I've written is the correct POSIX-compliant version for a basic string compare without specific bash-isms, therefore none of those are necessary

sxa: Simple answer: No

What I've written is the correct POSIX-compliant version without specific bash-isms

M-Davies (changes requested):
Shellcheck issues
```
In ./sign.sh line 65:
  TIMESTAMPSERVERS=`cut -d= -f2 < $WORKSPACE/$TIMESTAMP_SERVER_CONFIG | tr -d \\\\r`
                   ^-- SC2006: Use $(...) notation instead of legacy backticked `...`.
                                  ^--------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
  TIMESTAMPSERVERS=$(cut -d= -f2 < "$WORKSPACE"/$TIMESTAMP_SERVER_CONFIG | tr -d \\\\r)


In ./sign.sh line 80:
            echo Signing $f using $SERVER
                         ^-- SC2086: Double quote to prevent globbing and word splitting.
                                  ^-----^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
            echo Signing "$f" using "$SERVER"


In ./sign.sh line 90:
          echo Failed to sign ${f} using any time server - aborting
                              ^--^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
          echo Failed to sign "${f}" using any time server - aborting


In ./sign.sh line 151:
jdkDir=$(ls -1d $TMP_DIR/* | head -1 | xargs basename)
         ^---------------^ SC2012: Use find instead of ls to better handle non-alphanumeric filenames.
                ^------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
jdkDir=$(ls -1d "$TMP_DIR"/* | head -1 | xargs basename)
```

sxa:
Hmmm I'm not convinced about the 5 second pause either as it could mean we wait for up to about 30 seconds if only the final one in the list is working

karianna (approved):

karianna:
> Hmmm I'm not convinced about the 5-second pause either as it could mean we wait for up to about 30 seconds if only the final one in the list is working

That may well trigger pipeline timeouts so perhaps 2s?

M-Davies (approved):

sxa:
Switched to 2s - will merge once the checks complete again - hopefully before the JDK15 pipeline starts

PR #2411, opened 2021-01-27
Use the correct openj9 extensions repo name to runTests

andrew-m-leonard:
The openj9 extensions repo to runTests against was being determined incorrectly for the jdk(head) scenario.

### DIFF
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -189,7 +189,12 @@ class Build {
         if (buildConfig.VARIANT == "corretto") {
             suffix="corretto/corretto-${javaNumber}"
         } else if (buildConfig.VARIANT == "openj9") {
-            suffix = "ibmruntimes/openj9-openjdk-jdk${javaNumber}"
+            def openj9JavaToBuild = buildConfig.JAVA_TO_BUILD
+            if (openj9JavaToBuild.endsWith("u")) {
+                // OpenJ9 extensions repo does not use the "u" suffix
+                openj9JavaToBuild = openj9JavaToBuild.substring(0, openj9JavaToBuild.length() - 1)
+            }
+            suffix = "ibmruntimes/openj9-openjdk-${openj9JavaToBuild}"
         } else if (buildConfig.VARIANT == "hotspot") {
             suffix = "adoptopenjdk/openjdk-${buildConfig.JAVA_TO_BUILD}"
         } else if (buildConfig.VARIANT == "dragonwell") {


pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -189,7 +189,12 @@ class Build {
         if (buildConfig.VARIANT == "corretto") {
             suffix="corretto/corretto-${javaNumber}"
         } else if (buildConfig.VARIANT == "openj9") {
-            suffix = "ibmruntimes/openj9-openjdk-jdk${javaNumber}"
+            def openj9JavaToBuild = buildConfig.JAVA_TO_BUILD
+            if (openj9JavaToBuild.endsWith("u")) {
+                // OpenJ9 extensions repo does not use the "u" suffix
+                openj9JavaToBuild=openj9JavaToBuild.substring(0,openj9JavaToBuild.length()-1)
```
sxa, line 195: Could this not use `${javaNumber}` as it did previously or will the substring return somethign different in some cases?

sxa (approved):
For simplicity I'd lean slightly towards `if (buildConfig.JAVA_TO_BUILD.endsWith("u")` then set `suffix` accordingly thus removing the need for `openj9JavaToBuild` but this will work too so I won't block it going in either way :-)

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: That was the problem javaNumber is 17 for JAVA_TO_BUILD=jdk, if your building jdk you want the jdk extensions repo

andrew-m-leonard:
Good test run : https://ci.adoptopenjdk.net/job/Test_openjdk17_j9_sanity.openjdk_x86-64_linux/2/console

pipelines/build/common/openjdk_build_pipeline.groovy
M-Davies, line 195: One little nitpick for readability 🙂 
```suggestion
                openj9JavaToBuild = openj9JavaToBuild.substring(0, openj9JavaToBuild.length() - 1)
```

M-Davies (changes requested):

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: yes, fixed

M-Davies (approved):

AdamBrousseau (commented):
I think this also works
```
suffix = "ibmruntimes/openj9-openjdk-jdk" + ("${javaNumber}" - "u")
```

sxa:
> I think this also works `suffix = "ibmruntimes/openj9-openjdk-jdk" + ("${javaNumber}" - "u")`

That looks even simpler than my suggestion if it works :-)

sophia-guo:
Looks like it is a common issue for other vendors' ( corretto, dragwell) jdk (head). If that's the case probably we should update the logic for all other vendors.

PR #2416, opened 2021-01-28
Add cleanWorkspaceAfterBuild and cleanWorkspaceBuildOutputAfterBuild

andrew-m-leonard:
Add the facility to clean the build workspace or just the workspace/build/src/build and workspace/target binary output at the end of the build. This provides a large saving of disk space, up to 30Gb for Build Output alone and an additional 6Gb for the whole workspace. The default settings for generated pipelines are:
- cleanWorkspaceAfterBuild : false
- cleanWorkspaceBuildOutputAfterBuild : true
Additionally, cleanWorkspaceAfterBuild can be specified in the jdk platform configuration if a specific platform requires cleaning only.

### DIFF
--- a/pipelines/build/common/build_base_file.groovy
+++ b/pipelines/build/common/build_base_file.groovy
@@ -51,6 +51,8 @@ class Builder implements Serializable {
     boolean enableInstallers
     boolean enableSigner
     boolean cleanWorkspaceBeforeBuild
+    boolean cleanWorkspaceAfterBuild
+    boolean cleanWorkspaceBuildOutputAfterBuild
     boolean propagateFailures
     boolean keepTestReportDir
     boolean keepReleaseLogs
@@ -119,10 +121,18 @@ class Builder implements Serializable {
 
         def testList = getTestList(platformConfig)
 
+        def platformCleanWorkspaceAfterBuild = getCleanWorkspaceAfterBuild(platformConfig)
+
         // Always clean on mac due to https://github.com/AdoptOpenJDK/openjdk-build/issues/1980
         def cleanWorkspace = cleanWorkspaceBeforeBuild
         if (platformConfig.os == "mac") {
             cleanWorkspace = true
+        }
+
+        def cleanWsAfter = cleanWorkspaceAfterBuild
+        if (platformCleanWorkspaceAfterBuild) {
+            // Platform override specified
+            cleanWsAfter = platformCleanWorkspaceAfterBuild
         }
 
         return new IndividualBuildConfig(
@@ -151,7 +161,9 @@ class Builder implements Serializable {
                 ENABLE_TESTS: enableTests,
                 ENABLE_INSTALLERS: enableInstallers,
                 ENABLE_SIGNER: enableSigner,
-                CLEAN_WORKSPACE: cleanWorkspace
+                CLEAN_WORKSPACE: cleanWorkspace,
+                CLEAN_WORKSPACE_AFTER: cleanWsAfter,
+                CLEAN_WORKSPACE_BUILD_OUTPUT_ONLY_AFTER: cleanWorkspaceBuildOutputAfterBuild
         )
     }
 
@@ -218,6 +230,18 @@ class Builder implements Serializable {
 
         testList.unique()
         return testList
+    }
+
+    /*
+    Get the cleanWorkspaceAfterBuild override for this platform configuration
+    */
+    Boolean getCleanWorkspaceAfterBuild(Map<String, ?> configuration) {
+        Boolean cleanWorkspaceAfterBuild = null
+        if (configuration.containsKey("cleanWorkspaceAfterBuild") && configuration.get("cleanWorkspaceAfterBuild")) {
+            cleanWorkspaceAfterBuild = configuration.cleanWorkspaceAfterBuild as Boolean
+        }
+
+        return cleanWorkspaceAfterBuild
     }
 
     /*
@@ -683,6 +707,8 @@ return {
     String additionalBuildArgs,
     String overrideFileNameVersion,
     String cleanWorkspaceBeforeBuild,
+    String cleanWorkspaceAfterBuild,
+    String cleanWorkspaceBuildOutputAfterBuild,
     String adoptBuildNumber,
     String propagateFailures,
     String keepTestReportDir,
@@ -734,6 +760,8 @@ return {
             additionalBuildArgs: additionalBuildArgs,
             overrideFileNameVersion: overrideFileNameVersion,
             cleanWorkspaceBeforeBuild: Boolean.parseBoolean(cleanWorkspaceBeforeBuild),
+            cleanWorkspaceAfterBuild: Boolean.parseBoolean(cleanWorkspaceAfterBuild),
+            cleanWorkspaceBuildOutputAfterBuild: Boolean.parseBoolean(cleanWorkspaceBuildOutputAfterBuild),
             adoptBuildNumber: adoptBuildNumber,
             propagateFailures: Boolean.parseBoolean(propagateFailures),
             keepTestReportDir: Boolean.parseBoolean(keepTestReportDir),
--- a/pipelines/build/common/create_job_from_template.groovy
+++ b/pipelines/build/common/create_job_from_template.groovy
@@ -87,6 +87,8 @@ pipelineJob("$buildFolder/$JOB_NAME") {
                 <dt><strong>ENABLE_INSTALLERS</strong></dt><dd>Run installers</dd>
                 <dt><strong>ENABLE_SIGNER</strong></dt><dd>Run signer</dd>
                 <dt><strong>CLEAN_WORKSPACE</strong></dt><dd>Wipe out workspace before build</dd>
+                <dt><strong>CLEAN_WORKSPACE_AFTER</strong></dt><dd>Wipe out workspace after build</dd>
+                <dt><strong>CLEAN_WORKSPACE_BUILD_OUTPUT_ONLY_AFTER</strong></dt><dd>Wipe out workspace build output only, after build</dd>
             </dl>
         """)
     }
--- a/pipelines/build/common/openjdk_build_pipeline.groovy
+++ b/pipelines/build/common/openjdk_build_pipeline.groovy
@@ -67,7 +67,6 @@ class Build {
         NODE_CHECKOUT_TIMEOUT : 1,
         BUILD_JDK_TIMEOUT : 8,
         BUILD_ARCHIVE_TIMEOUT : 3,
-        AIX_CLEAN_TIMEOUT : 1,
         MASTER_CLEAN_TIMEOUT : 1,
         DOCKER_CHECKOUT_TIMEOUT : 1,
         DOCKER_PULL_TIMEOUT : 2
@@ -906,27 +905,20 @@ class Build {
     Executed on a build node, the function checks out the repository and executes the build via ./make-adopt-build-farm.sh
     Once the build completes, it will calculate its version output, commit the first metadata writeout, and archive the build results.
     */
-    def buildScripts(cleanWorkspace, filename) {
+    def buildScripts(cleanWorkspace, cleanWorkspaceAfter, cleanWorkspaceBuildOutputAfter, filename) {
         return context.stage("build") {
-
             if (cleanWorkspace) {
                 try {
 
                     try {
                         context.timeout(time: buildTimeouts.NODE_CLEAN_TIMEOUT, unit: "HOURS") {
-                            if (buildConfig.TARGET_OS == "windows") {
-                                // Windows machines struggle to clean themselves, see:
-                                // https://github.com/AdoptOpenJDK/openjdk-build/issues/1855
-                                context.sh(script: "rm -rf C:/workspace/openjdk-build/workspace/build/src/build/*/jdk/gensrc")
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1419
-                                context.sh(script: "rm -rf J:/jenkins/tmp/workspace/build/src/build/*/jdk/gensrc")
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1662
-                                context.sh(script: "rm -rf E:/jenkins/tmp/workspace/build/src/build/*/jdk/gensrc")
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1818
-                                context.sh(script: "rm -rf C:/Jenkins/temp/workspace/build/src/build/*/jdk/gensrc")
-                                context.cleanWs notFailBuild: true, disableDeferredWipeout: true, deleteDirs: true
+                            // Note: Underlying org.apache DirectoryScanner used by cleanWs has a bug scanning where it misses files containing ".." so use rm -rf instead
+                            // Issue: https://issues.jenkins.io/browse/JENKINS-64779
+                            if (context.WORKSPACE != null && !context.WORKSPACE.isEmpty()) {
+                                context.println "Cleaning workspace files: " + context.WORKSPACE + "/*"
+                                context.sh(script: "rm -rf " + context.WORKSPACE + "/*")
                             } else {
-                                context.cleanWs notFailBuild: true
+                                context.println "Warning: Unable to clean workspace as context.WORKSPACE is null/empty"
                             }
                         }
                     } catch (FlowInterruptedException e) {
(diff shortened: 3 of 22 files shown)


andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

AdamBrousseau:
I believe this or addresses https://github.com/AdoptOpenJDK/openjdk-build/issues/2296

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
runt tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -906,28 +905,20 @@ class Build {
...
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1818
-                                context.sh(script: "rm -rf C:/Jenkins/temp/workspace/build/src/build/*/jdk/gensrc")
-                                context.cleanWs notFailBuild: true, disableDeferredWipeout: true, deleteDirs: true
-                            } else {
-                                context.cleanWs notFailBuild: true
-                            }
+                            // Note: Underlying org.apache DirectoryScanner has a bug scanning files containing ".." so must be deleted explicitly
+                            context.println "Cleaning workspace files containing '..'"
+                            context.sh(script: "find . -name \"*..*\"") 
+                            context.sh(script: "find . -name \"*..*\" | xargs rm -f")
```
sxa, line 919: While it seems to work this line scares the ?@#& out of me. I'm nervous about the possibility that it'll find a `..` directory somewhere and point `rm -f` at it ... I suppose it's not recursive though

sxa (commented):
I really don't like that find construct ... If this is likely to only affect one file can we just handle it explicitly or perhaps `find . -type f -name \*..\*`? Also I'd recommend removing the verbosity here - maybe just keep the `rm` lines instead of the bit printing them all out?

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: yeah I thought about that too, but as I tested it carefully, and as you said it is not recursive

andrew-m-leonard:
> I really don't like that find construct ... If this is likely to only affect one file can we just handle it explicitly or perhaps `find . -type f -name \*..\*`? Also I'd recommend removing the verbosity here - maybe just keep the `rm` lines instead of the bit printing them all out?

It's any file that has ".." in it, the 2 i've discovered so far:
```
17:51:00  workspace/build/src/build/windows-x86_64-normal-clientANDserver-release/support/gensrc/java.base/java/nio/_the...marker
17:31:58  workspace/build/src/build/windows-x86_64-normal-server-release/jdk/gensrc/java/nio/_the..
```

andrew-m-leonard:
> I really don't like that find construct ... If this is likely to only affect one file can we just handle it explicitly or perhaps `find . -type f -name \*..\*`? Also I'd recommend removing the verbosity here - maybe just keep the `rm` lines instead of the bit printing them all out?

Yeah, I thought the print was useful debug for the moment, but i'll remove

andrew-m-leonard:
> I really don't like that find construct ... If this is likely to only affect one file can we just handle it explicitly or perhaps `find . -type f -name \*..\*`? Also I'd recommend removing the verbosity here - maybe just keep the `rm` lines instead of the bit printing them all out?

Yes, I like the -type f, good idea

andrew-m-leonard:
@sxa The other solution is not use CleanWs, just use rm -rf on the specific folders?
The logic behind CleanWs using DirectoryScanner is just overkill, especially for what we're doing

sxa:
> @sxa The other solution is not use CleanWs, just use rm -rf on the specific folders?

i'm thinking that's probably the best option - keeps it simple, but if we do that let's add a comment pointing to the issue as an explanation of why we're not using `cleanWs`

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
```diff
@@ -80,14 +80,16 @@ class Config11 {
...
+                cleanWorkspaceAfterBuild: true
         ],
 
         s390xLinux    : [
                 os                  : 'linux',
                 arch                : 's390x',
                 test                : 'default',
-                configureArgs       : '--enable-dtrace=auto'
+                configureArgs       : '--enable-dtrace=auto',
+                cleanWorkspaceAfterBuild: true
```
sxa, line 92: Can we make this clean the buid output only? Should be safe enough since the machines have a reasonable amount of space compared to the AIX ones. (We should also aim to work towards doing the same on AIX)

sxa (changes requested):
Mostly looks ok other than my comment about `s390x`.

Also, can we add an entry to https://github.com/AdoptOpenJDK/openjdk-build/blob/master/FAQ.md pointing to this as an example of how to add a new parameter to build build pipelines?

pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
andrew-m-leonard: Done. Removed s390x so will use the default of BuildOutput only.

andrew-m-leonard:
run tests

pipelines/build/common/build_base_file.groovy
```diff
@@ -220,6 +232,17 @@ class Builder implements Serializable {
         return testList
     }
 
+    /* Get the cleanWorkspaceAfterBuild override for this platform configuration
```
karianna, line 235: Format of comment:

/*
 *
 */

pipelines/build/common/create_job_from_template.groovy
```diff
@@ -87,6 +87,8 @@ pipelineJob("$buildFolder/$JOB_NAME") {
                 <dt><strong>ENABLE_INSTALLERS</strong></dt><dd>Run installers</dd>
                 <dt><strong>ENABLE_SIGNER</strong></dt><dd>Run signer</dd>
                 <dt><strong>CLEAN_WORKSPACE</strong></dt><dd>Wipe out workspace before build</dd>
+                <dt><strong>CLEAN_WORKSPACE_AFTER</strong></dt><dd>Wipe out workspace after build</dd>
+                <dt><strong>CLEAN_WORKSPACE_BUILD_OUTPUT_AFTER</strong></dt><dd>Wipe out workspace build output only, after build</dd>
```
karianna, line 91: Recommend rename of CLEAN_WORKSPACE_BUILD_OUTPUT_AFTER to CLEAN_WORKSPACE_BUILD_OUTPUT_ONLY

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -906,28 +905,16 @@ class Build {
...
-                                context.sh(script: "rm -rf J:/jenkins/tmp/workspace/build/src/build/*/jdk/gensrc")
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1662
-                                context.sh(script: "rm -rf E:/jenkins/tmp/workspace/build/src/build/*/jdk/gensrc")
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1818
-                                context.sh(script: "rm -rf C:/Jenkins/temp/workspace/build/src/build/*/jdk/gensrc")
-                                context.cleanWs notFailBuild: true, disableDeferredWipeout: true, deleteDirs: true
-                            } else {
-                                context.cleanWs notFailBuild: true
-                            }
+                            // Note: Underlying org.apache DirectoryScanner used by cleanWs has a bug scanning where it misses files containing ".." so use rm -rf instead
```
karianna, line 915: Link to bug for that issue?

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -999,15 +986,22 @@ class Build {
...
+                if (cleanWorkspaceAfter || cleanWorkspaceBuildOutputAfter) {
                     try {
-                        context.timeout(time: buildTimeouts.AIX_CLEAN_TIMEOUT, unit: "HOURS") {
-                            context.cleanWs notFailBuild: true
+                        context.timeout(time: buildTimeouts.NODE_CLEAN_TIMEOUT, unit: "HOURS") {
+                            // Note: Underlying org.apache DirectoryScanner used by cleanWs has a bug scanning where it misses files containing ".." so use rm -rf instead
+                            if (cleanWorkspaceAfter) {
+                                context.println "Cleaning workspace files: " + context.WORKSPACE + "/*"
+                                context.sh(script: "rm -rf " + context.WORKSPACE + "/*")
+                            } else {
```
karianna, line 996: let's make that an explicit else if :-)

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -999,15 +986,22 @@ class Build {
...
+                                context.println "Cleaning workspace build output files: " + context.WORKSPACE + "/workspace/build/src/build"
+                                context.sh(script: "rm -rf " + context.WORKSPACE + "/workspace/build/src/build")
+                                context.println "Cleaning workspace build output files: " + context.WORKSPACE + "/workspace/target"
+                                context.sh(script: "rm -rf " + context.WORKSPACE + "/workspace/target")
+                            }
                         }
                     } catch (FlowInterruptedException e) {
-                        context.println "[ERROR] AIX clean workspace timeout (${buildTimeouts.AIX_CLEAN_TIMEOUT} HOURS) has been reached. Exiting..."
+                        context.println "[ERROR] clean workspace timeout (${buildTimeouts.NODE_CLEAN_TIMEOUT} HOURS) has been reached. Exiting..."
```
karianna, line 1004: Is it always a timeout?

pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
```diff
@@ -80,7 +80,8 @@ class Config11 {
                         hotspot: 'xlc13&&aix710',
                         openj9:  'xlc13&&aix715'
                 ],
-                test                : 'default'
+                test                : 'default',
+                cleanWorkspaceAfterBuild: true
```
karianna, line 84: Does this not default to true in build_base_file.groovy already?

karianna (changes requested):

pipelines/jobs/configurations/jdk11u_pipeline_config.groovy
andrew-m-leonard: No, this is the "pipeline" config, in build_base_file.groovy it sets "CLEAN_WORKSPACE": true, which means the generated build "job" config has it as true, so if you were to submit a build job: https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk16/job/jdk16-aix-ppc64-openj9/build?delay=0sec
But, pipelines have a default of "false", eg.when you submit https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/openjdk11-pipeline/build?delay=0sec
The pipeline generated default is here: https://github.com/AdoptOpenJDK/openjdk-build/blob/0b12dd8c8ad8a7153363bd965b5d5d37690bb3de/pipelines/jobs/pipeline_job_template.groovy#L74

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: I think that was originally added if cleanWs could hang/loop, maybe on Windows? not sure really. Can't see an awful lot of point of it. If the workspace clean fails, we really should fail the job, as the workspace could be in an odd state.
I'll change that.

andrew-m-leonard: ah just looked again, so yes, FlowInterruptedException is a Interrupt from the context.timeout(), so it is correct. A different failure/Exception is caught by the outer catch(Exception e)...

pipelines/build/common/build_base_file.groovy
andrew-m-leonard: fixed

pipelines/build/common/create_job_from_template.groovy
andrew-m-leonard: fixed

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: fixed

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -906,28 +905,16 @@ class Build {
...
-                                context.sh(script: "rm -rf E:/jenkins/tmp/workspace/build/src/build/*/jdk/gensrc")
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1818
-                                context.sh(script: "rm -rf C:/Jenkins/temp/workspace/build/src/build/*/jdk/gensrc")
-                                context.cleanWs notFailBuild: true, disableDeferredWipeout: true, deleteDirs: true
-                            } else {
-                                context.cleanWs notFailBuild: true
-                            }
+                            // Note: Underlying org.apache DirectoryScanner used by cleanWs has a bug scanning where it misses files containing ".." so use rm -rf instead
+                            context.println "Cleaning workspace files: " + context.WORKSPACE + "/*"
+                            context.sh(script: "rm -rf " + context.WORKSPACE + "/*")
```
AdamBrousseau, line 917: Is the same result achieved if you remove the `/*`? i.e just rm the workspace? I am just worried if WORKSPACE ever comes back empty.

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: Created issue and added: https://issues.jenkins.io/browse/JENKINS-64779

pipelines/build/common/openjdk_build_pipeline.groovy
andrew-m-leonard: I was thinking about that, and I see your point, I hadn't thought about it that way round. Without the /* it will remove the workspace folder as well, which I think will cause a problem. I can add a check that WORKSPACE is not empty...

andrew-m-leonard:
run tests

pipelines/build/common/openjdk_build_pipeline.groovy
AdamBrousseau: Ok that's probably a good idea. You might be right removing the folder might not work. We have a cleanup job that runs to rm all our workspaces but it ignores the cleanup job folder even though there's nothing in it. I think the pipeline complains.

pipelines/build/common/openjdk_build_pipeline.groovy
sxa: That should probably be added into the comment

andrew-m-leonard:
run tests

andrew-m-leonard:
run tests

sxa (commented):
Looking good now but I won't formally approve til I have another look over it in the morning (unless two other people want to approve in the meantime!)

Would be good to check a few random machines to see how much space `workspace/build-scripts` is taking and compare it to the same numbers a week after merging this when it should have taken effect across most of the machines. Should particularly help the scaleway arm32 build systems

sxa (dismissed):

pipelines/build/common/build_base_file.groovy
```diff
@@ -151,7 +161,9 @@ class Builder implements Serializable {
                 ENABLE_TESTS: enableTests,
                 ENABLE_INSTALLERS: enableInstallers,
                 ENABLE_SIGNER: enableSigner,
-                CLEAN_WORKSPACE: cleanWorkspace
+                CLEAN_WORKSPACE: cleanWorkspace,
+                CLEAN_WORKSPACE_AFTER: cleanWsAfter,
+                CLEAN_WORKSPACE_BUILD_OUTPUT_ONLY: cleanWorkspaceBuildOutputAfterBuild
```
sxa, line 166: Hmm it worries me that after the response to @karianna 's comment these two names are now inconsistent. I quite liked the"after" in the name for consistency with other variables but either way we do need to consider how new people looking at the scripts will see this.

sxa (commented):
Other than the naming which I think it's now confusing I think this is ready

pipelines/build/common/build_base_file.groovy
andrew-m-leonard: @sxa do we want CLEAN_WORKSPACE_BUILD_OUTPUT_ONLY_AFTER ?

andrew-m-leonard: quite a mouthful!

andrew-m-leonard: BUILD_OUTPUT_ONLY sort of implies at the end of the build

andrew-m-leonard: i'll change it to: CLEAN_WORKSPACE_BUILD_OUTPUT_ONLY_AFTER

andrew-m-leonard: changed

andrew-m-leonard:
run tests

sxa (approved):
Yep I'll take that - thanks. Clear enough now I reckon :-)

sxa:
ping @karianna as you are currently blocking this

andrew-m-leonard:
run tests

karianna (approved):

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -906,27 +905,20 @@ class Build {
...
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1419
-                                context.sh(script: "rm -rf J:/jenkins/tmp/workspace/build/src/build/*/jdk/gensrc")
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1662
-                                context.sh(script: "rm -rf E:/jenkins/tmp/workspace/build/src/build/*/jdk/gensrc")
-                                // https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1818
-                                context.sh(script: "rm -rf C:/Jenkins/temp/workspace/build/src/build/*/jdk/gensrc")
-                                context.cleanWs notFailBuild: true, disableDeferredWipeout: true, deleteDirs: true
+                            // Note: Underlying org.apache DirectoryScanner used by cleanWs has a bug scanning where it misses files containing ".." so use rm -rf instead
+                            // Issue: https://issues.jenkins.io/browse/JENKINS-64779
+                            if (context.WORKSPACE != null && !context.WORKSPACE.isEmpty()) {
```
AdamBrousseau, line 917: i think `if (context.WORKSPACE)` achieves the same thing.

pipelines/build/common/openjdk_build_pipeline.groovy
```diff
@@ -906,27 +905,20 @@ class Build {
...
+                                context.println "Cleaning workspace files: " + context.WORKSPACE + "/*"
+                                context.sh(script: "rm -rf " + context.WORKSPACE + "/*")
```
AdamBrousseau, line 919: I think this ignored the `.` hidden directories (eg. `.git` folder). Which is causing me issues on my win boxes.

PR #2417, opened 2021-01-28
build.sh: Add eyecatcher timestamps to check build performance

sxa:
Found this useful while running multiple builds in succession so I feel it could be beneficial here if we want to see which sections are fast/slow in different environments (also highlights [CaCerts](https://github.com/AdoptOpenJDK/openjdk-build/pull/2113) and [gradle timings](https://github.com/AdoptOpenJDK/openjdk-build/issues/2139))

This may be useful in optimising for third-party users of our build scripts

(If anyone feels particularly strongly about the removal of `showCompletionMessage()` let me know but it seems to be extra complexity wrapping a function around an echo statement :-) )

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -892,11 +892,6 @@ createOpenJDKTarArchive() {
   createArchive "${jdkTargetPath}" "${BUILD_CONFIG[TARGET_FILE_NAME]}"
 }
 
-# Echo success
-showCompletionMessage() {
-  echo "All done!"
-}
-
 copyFreeFontForMacOS() {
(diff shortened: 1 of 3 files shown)


M-Davies (approved):
LGTM. More feedback is good

sxa:
(Related: With the source already extracted, cacerts generation disabled, and not using gradle, my quickest machine can build `makejdk-ant-platform.sh` for HotSpot in about 4 minutes, reducing to about 2½ with a pre-populated ccache. Those sorts of times definitely make tests of source code changes quite feasible :-) )

andrew-m-leonard (approved):

PR #2425, opened 2021-02-01
Stop creating mozilla cacerts bundle if it is overridden

sxa:
When using `--custom-cacerts false` as a parameter to `makejdk-any-platform.sh` (e.g. with `BUILD_ARGS` in the pipeline set to `--custom-cacerts false`) the mk-cacerts.sh ([added in this PR](https://github.com/AdoptOpenJDK/openjdk-build/pull/2113/files)) still generates the new cacerts file even when it is not required. This can add over a minute to the builds unecessarily.

Also testing in [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/998)

[Slack discussion before I created this PR](https://adoptopenjdk.slack.com/archives/C09NW3L2J/p1611945898053100)


karianna (approved):
LGTM

aahlenst (approved):

aahlenst (changes requested):
We've an inconsistency here. The gate for the configure arg is:

```
 if [ "${BUILD_CONFIG[CUSTOM_CACERTS]}" != "false" ] ; then
```

aahlenst:
I prefer `[ "${BUILD_CONFIG[CUSTOM_CACERTS]}" != "false" ]` because it being on is the default. But I can live with both.

karianna:
> I prefer `[ "${BUILD_CONFIG[CUSTOM_CACERTS]}" != "false" ]` because it being on is the default. But I can live with both.

It's a double negative aren't we better off saying `[ "${BUILD_CONFIG[CUSTOM_CACERTS]}" == "true" ]`?

aahlenst:
@karianna `if [ "${BUILD_CONFIG[CUSTOM_CACERTS]}" = "true" ] ; then` is developer convenience, `if [ "${BUILD_CONFIG[CUSTOM_CACERTS]}" != "false" ] ; then` is user convenience :grin: 

Whatever is chosen, it should be consistent.

sxa:
> true



> > I prefer `[ "${BUILD_CONFIG[CUSTOM_CACERTS]}" != "false" ]` because it being on is the default. But I can live with both.
> 
> It's a double negative aren't we better off saying `[ "${BUILD_CONFIG[CUSTOM_CACERTS]}" == "true" ]`?

That's a philosophical question of whether you consider "false" to be negative and whether there may be other options other than `true` in the future for sourcing the cacerts.

@aahlenst You said you can live with both but you've actively marked it as "changes requested" instead of a "comment" which blocks approval - was that intentional? Saying that I agree with your comment from a consisteny perspective - I suppose the other option would be to change https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh#L381 to be `= "true"` too.

This is getting a bit abstract, but if there was any chance of it being set to blank or something else then that would be a driving decision maker, although to me the absence of a value would generally be considered "false" so checking for `true` is a better option.

aahlenst:
> You said you can live with both but you've actively marked it as "changes requested"

Yes, because currently the gates are different. A typo brings us into "both are true" territory, something we don't want I guess, especially if we say "OpenJDK, include our custom trust store" while we don't generate one.

sxa:
> > You said you can live with both but you've actively marked it as "changes requested"
> 
> Yes, because currently the gates are different. A typo brings us into "both are true" territory, something we don't want I guess, especially if we say "OpenJDK, include our custom trust store" while we don't generate one.

OK so you mean you'd be ok if https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh#L381 was modified to be the same - in that case I'm happy to make that change

sxa:
@aahlenst @karianna Both happy with the last update?

aahlenst (approved):
Great, thanks @sxa.

sxa:
`--custom-cacerts false` now seems to have the desired effect of not generating the bundle - verified on [JDK11/xLinux](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-x64-openj9/920/consoleFull) and [JDK11/Windows](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-windows-x64-hotspot/906/consoleFull)

PR #2458, opened 2021-02-08
Supersede local file over network download

M-Davies:
* Local platform configs will now superceed network locations
* Build will fail if a local file doesn't exist and PLATFORM_CONFIG_LOCATION is unset


karianna (approved):
lgtm

sxa (approved):
Certainly works for me so I'll approve (as expected, since it was in response to a concern I raised!) although we should possibly make it explicit in a comment to
```
PLATFORM_CONFIG_FILEPATH="$SCRIPT_DIR/platform-specific-configurations/${OPERATING_SYSTEM}.sh"     
if [ ! -f "${PLATFORM_CONFIG_FILEPATH}" ]
```
to make it clear that if someone wants to auto-download to take effect then they are expected to delete the local ones from the clone - there is no way to force it to use downloaded ones instead of local ones.

PR #2463, opened 2021-02-09
pipeline: move to separate repo

gdams:
Part of the process at https://github.com/AdoptOpenJDK/openjdk-build/issues/2455


M-Davies (changes requested):
Have you checked if the [docker dir](https://github.com/gdams/openjdk-build/tree/remove_pipeline/docker) needs to be moved over too?

The [Metadata](https://github.com/gdams/openjdk-build/tree/remove_pipeline#metadata) section of the README can also be moved over as that is handled by the Jenkins scripts.

gdams:
@M-Davies docker dir is not needed. I've moved the Metadata section over to ci-jenkins-pipelines

M-Davies (approved):
Approved but people should probably be double notified of this happening (I'll do that now in Slack)

M-Davies (changes requested):
Just a thought, we may also want to re-evaluate what checks we have in each repo.

I.e. the [Compile / Groovy](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/.github/workflows/compile.yml) job may as well be moved to ci.jenkins.pipelines and removed from here since that's where all the groovy code is going (if we don't intend on removing it as part of https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/issues/14).

gdams:
I've already removed the groovy compile job from this PR?

M-Davies:
> I've already removed the groovy compile job from this PR?

Ah so I see now. I missed it in the file search

pipelines/build/prTester/README.md
```diff
@@ -1,123 +0,0 @@
-# PR Testers
```
M-Davies, line 1: One last query, since we're keeping the linter and build jobs, can we keep the documentation in this repo? Could be copied and pasted into the root Readme?

M-Davies: By keep documentation, I mean keep the linter and build parts but remove the other parts

karianna:
@gdams needs rebase and one Q answered

sxa (changes requested):
I can't see any changes to RELEASING.md in here and it still contains references to the pipelines and git-hg stuff in this repository so they will need to be updated too (Which will also bring your "files changed" count to 100 in this PR :-)

We should get these updates done before the JDK16 release so that the documentation is correct for the person who does any updates relating to the brand new release.

M-Davies (commented):
Linting failures but other than that this looks good to me now! Thanks for doing this 🙂

sxa (approved):
Thanks - I think that's probably all that's needed unless there are other references to some of these deleted files elsewhere across our repositories but those will need separate PRs :-)

M-Davies (approved):

PR #2464, opened 2021-02-10
Archive failed openjdk make failure logs

andrew-m-leonard:
Create an archive of openjdk make failure logs and dumps if any.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -444,11 +444,15 @@ executeTemplatedFile() {
 
   echo "Currently at '${PWD}'"
 
+  # We need the exitcode from the configure-and-build.sh script
+  set +eu
+
   # Execute the build passing the workspace dir and target dir as params for configure.txt
   bash "${BUILD_CONFIG[WORKSPACE_DIR]}/config/configure-and-build.sh" ${BUILD_CONFIG[WORKSPACE_DIR]} ${BUILD_CONFIG[TARGET_DIR]}
   exitCode=$?
 
-  if [ "${exitCode}" -eq 1 ]; then
+  if [ "${exitCode}" -eq 3 ]; then
+    createOpenJDKFailureLogsArchive
     echo "Failed to make the JDK, exiting"
     exit 1
   elif [ "${exitCode}" -eq 2 ]; then
@@ -459,6 +463,42 @@ executeTemplatedFile() {
     exit 2
   fi
 
+  # Restore exit behavior
+  set -eu
+}
+
+createOpenJDKFailureLogsArchive() {
+    echo "OpenJDK make failed, archiving make failed logs"
+    cd build/*
+
+    local adoptLogArchiveDir="AdoptOpenJDKLogsArchive"
+
+    # Create new folder for failure logs
+    rm -rf ${adoptLogArchiveDir}
+    mkdir ${adoptLogArchiveDir}
+
+    # Copy build and failure logs
+    if [[ -f "build.log" ]]; then
+      echo "Copying build.log to ${adoptLogArchiveDir}"
+      cp build.log ${adoptLogArchiveDir}
+    fi
+    if [[ -d "make-support/failure-logs" ]]; then
+      echo "Copying make-support/failure-logs to ${adoptLogArchiveDir}"
+      mkdir -p "${adoptLogArchiveDir}/make-support"
+      cp -r "make-support/failure-logs" "${adoptLogArchiveDir}/make-support"
+    fi
+
+    # Find any cores, dumps, ..
+    find . -name 'core.*' -name 'core.*.dmp' -o -name 'javacore.*.txt' -o -name 'Snap.*.trc' -o -name 'jitdump.*.dmp' | sed 's#^./##' | while read -r dump ; do
+      filedir=$(dirname "${dump}")
+      echo "Copying ${dump} to ${adoptLogArchiveDir}/${filedir}"
+      mkdir -p "${adoptLogArchiveDir}/${filedir}"
+      cp "${dump}" "${adoptLogArchiveDir}/${filedir}"
+    done
+
+    # Archive logs
+    local makeFailureLogsName=$(echo "${BUILD_CONFIG[TARGET_FILE_NAME]//-jdk/-makefailurelogs}")
+    createArchive "${adoptLogArchiveDir}" "${makeFailureLogsName}"
 }
 
 getGradleJavaHome() {


andrew-m-leonard:
run tests

karianna:
> run tests

I think we don't need to do that anymore CC @gdams for confirmation

sbin/build.sh
```diff
@@ -459,6 +463,41 @@ executeTemplatedFile() {
     exit 2
   fi
 
+  # Restore exit behavior
+  set -eu
+}
+
+createOpenJDKFailureLogsArchive() {
+    echo "OpenJDK make failed, archiving make failed logs"
+    cd build/*
```
karianna, line 472: Are we sure we're in the directory that build hangs off?

sbin/build.sh
```diff
@@ -459,6 +463,41 @@ executeTemplatedFile() {
...
+
+    local adoptLogArchiveDir="AdoptOpenJDKLogsArchive"
+
+    # Create new folder for failure logs
+    rm -rf ${adoptLogArchiveDir}
```
karianna, line 477: Removing the old archive first on purpose?

sbin/build.sh
```diff
@@ -459,6 +463,41 @@ executeTemplatedFile() {
...
+    fi
+    if [[ -d "make-support/failure-logs" ]]; then
+      echo "Copying make-support/failure-logs to ${adoptLogArchiveDir}"
+      mkdir -p "${adoptLogArchiveDir}/make-support"
+      cp -r "make-support/failure-logs" "${adoptLogArchiveDir}/make-support"
+    fi
+
+    # Find any cores, dumps, ..
+    find . -name 'core.*' -name 'core.*.dmp' -o -name 'javacore.*.txt' -o -name 'Snap.*.trc' -o -name 'jitdump.*.dmp' | sed 's#^./##' | while read -r dump ; do
+      filedir=$(dirname "${dump}")
```
karianna, line 493: worth adding an echo here as well?

sbin/build.sh
```diff
@@ -459,6 +463,41 @@ executeTemplatedFile() {
...
+      mkdir -p "${adoptLogArchiveDir}/${filedir}"
+      cp "${dump}" "${adoptLogArchiveDir}/${filedir}"
+    done
+
+    # Archive logs
+    local makeFailureLogsName=$(echo "${BUILD_CONFIG[TARGET_FILE_NAME]//-jdk/-makefailurelogs}")
+    createArchive "${adoptLogArchiveDir}" "${makeFailureLogsName}"
```
karianna, line 500: where does that live? workspace?

karianna (changes requested):

sbin/build.sh
andrew-m-leonard: Yes called from executeTemplatedFile() which steps into working dir

sbin/build.sh
andrew-m-leonard: yes, just to ensure there isn't any remnant from a dirty workspace

sbin/build.sh
andrew-m-leonard: yes good idea

sbin/build.sh
andrew-m-leonard: createArchive() creates the tar.gz in the workspace/target with the other archives, from the temporary AdoptOpenJDKLogsArchive folder just created in the build dir

sbin/build.sh
andrew-m-leonard: added

M-Davies:
> > run tests
> 
> I think we don't need to do that anymore CC @gdams for confirmation

The standard .github checks jobs should be sufficient to test the changes in this case anyway (unless andrew intends on adding additional groovy code to expose it to the Jenkins side of things)

andrew-m-leonard:
> > > run tests
> > 
> > 
> > I think we don't need to do that anymore CC @gdams for confirmation
> 
> The standard .github checks jobs should be sufficient to test the changes in this case anyway (unless andrew intends on adding additional groovy code to expose it to the Jenkins side of things)

No that's fine, I just tried it to see what would happen!

M-Davies (approved):

gdams (approved):

karianna (approved):

sbin/build.sh
```diff
@@ -459,6 +463,42 @@ executeTemplatedFile() {
...
+      cp build.log ${adoptLogArchiveDir}
+    fi
+    if [[ -d "make-support/failure-logs" ]]; then
+      echo "Copying make-support/failure-logs to ${adoptLogArchiveDir}"
+      mkdir -p "${adoptLogArchiveDir}/make-support"
+      cp -r "make-support/failure-logs" "${adoptLogArchiveDir}/make-support"
+    fi
+
+    # Find any cores, dumps, ..
+    find . -name 'core.*' -name 'core.*.dmp' -o -name 'javacore.*.txt' -o -name 'Snap.*.trc' -o -name 'jitdump.*.dmp' | sed 's#^./##' | while read -r dump ; do
```
keithc-ca, line 492: I expect the list of filename patterns here is incomplete. For the issue this was intended to address, it will be helpful to see files such as the following for `jdk.jshell`:
```
jdk/modules/jdk.jshell/_the.jdk.jshell_batch
jdk/modules/jdk.jshell/_the.jdk.jshell_batch.cmdline
jdk/modules/jdk.jshell/_the.jdk.jshell_batch.filelist
jdk/modules/jdk.jshell/_the.jdk.jshell_batch.log
jdk/modules/jdk.jshell/_the.jdk.jshell.config_vardeps
jdk/modules/jdk.jshell/_the.jdk.jshell_pubapi
jdk/modules/jdk.jshell/_the.jdk.jshell-server.conf
jdk/modules/jdk.jshell/_the.jdk.jshell.vardeps
```

andrew-m-leonard: It does capture the whole of make-support/failure-logs, which will contain the above if it fails, I confirmed that with a test I did

andrew-m-leonard: the above "dump" list is generic for hotspot/openj9/corretto...

andrew-m-leonard: See test example here: https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-x64-openj9-linuxXL/835/

keithc-ca: Ok, that will capture the log, but not `.cmdline` or `.filelist`.
Also, I don't see any clues in the log where one would find the resulting archive.

andrew-m-leonard: The .cmdline is captured as that is placed in failure-logs:  
[OpenJDK11U-makefailurelogs_x64_linux_openj9_linuxXL_2021-02-12-15-16(1).tar.gz](https://github.com/AdoptOpenJDK/openjdk-build/files/5972992/OpenJDK11U-makefailurelogs_x64_linux_openj9_linuxXL_2021-02-12-15-16.1.tar.gz)

andrew-m-leonard: The archive is archived same as where images etc would be:
image

keithc-ca: Your earlier comment referenced build 835, which doesn't have a corresponding artifact that I could find. Did something change in between that build and 838?

Also, the `.cmdline` file available for 838 makes reference to
```
@/home/adoptopenjdk/workspace/build-scripts/jobs/jdk11u/jdk11u-linux-x64-openj9-linuxXL/workspace/build/src/build/linux-x86_64-normal-server-release/jdk/modules/jdk.jshell/_the.jdk.jshell_batch.tmp
```
but `_the.jdk.jshell_batch.tmp` is not included in the archive.

andrew-m-leonard: Yes, sorry Adopt only keeps 1 artifact history on the build job and a subsequent one was submitted, we're thinking of increasing this

andrew-m-leonard: we'd need an upstream change if something is not being copied to the make-support/failed-logs folder

keithc-ca: > we'd need an upstream change if something is not being copied to the make-support/failed-logs folder

Not if you capture file matching `jdk/modules/*/_the.*`.

andrew-m-leonard: ah see what you mean, thanks Keith

PR #2478, opened 2021-02-17
autodetect as much as possible in make-adopt-build-farm.sh

sxa:
Follow on from https://github.com/AdoptOpenJDK/openjdk-build/commit/238ad36cb85d19ce443f6f0a16ea2153ff4c10ce - autodetect default values (but tell user what is being defaulted so they can override easily if required ...

Testing with [VPC](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/1058/) and [QPC](https://ci.adoptopenjdk.net/view/Tooling/job/QEMUPlaybookCheck/203/)

### DIFF
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -18,16 +18,40 @@ set -e
 
 PLATFORM_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-## Sanity check as this script requires various environment variables so fail
-## fast if one or more are not present. MAYBE autodetect defaults in future?
+## autodetect defaults to improve usability when running this for debugging/testing
 
-SANEVARS=0
+if [ -z "$ARCHITECTURE"  ]; then
+   ARCHITECTURE=`uname -p`
+   if [       "$OSTYPE" = "cygwin"  ]; then ARCHITECTURE=`uname -m`; fi
+   if [ "$ARCHITECTURE" = "x86_64"  ]; then ARCHITECTURE=x64;        fi
+   if [ "$ARCHITECTURE" = "powerpc" ]; then ARCHITECTURE=ppc64;      fi
+   echo ARCHITECTURE not defined - assuming $ARCHITECTURE
+   export ARCHITECTURE
+fi
+
+if [ -z "$TARGET_OS" ]; then
+  if [ "$OSTYPE" = "cygwin" ]; then
+    TARGET_OS=windows
+  else
+    TARGET_OS=`uname`
+  fi
+  echo TARGET_OS not defined - assuming you want "$TARGET_OS"
+  export TARGET_OS
+fi
+
+if [ -z "$JAVA_TO_BUILD" ]; then
+  if [ "$1" != "${1##jdk}" ]; then
+    echo Setting JAVA_TO_BUILD to "$1" from the parameter supplied
+    export JAVA_TO_BUILD="$1"
+  else
+    echo JAVA_TO_BUILD not defined - defaulting to jdk11u
+    export JAVA_TO_BUILD=jdk11u
+  fi
+fi
+
 [ -z "$JAVA_TO_BUILD" ] && echo JAVA_TO_BUILD not defined - set to e.g. jdk8u && SANEVARS=1
-[ -z "$TARGET_OS"     ] && echo TARGET_OS not defined - set to e.g. linux     && SANEVARS=1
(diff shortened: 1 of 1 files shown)


sxa:
@Willsparker @Haroon-Khel This change should mean that most of the stuff for VPC and QPC in https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/97ecbc8e0d26d75b06cc38b3ca01d737aa706af4/ansible/pbTestScripts/buildJDK.sh#L125 can go as it will all default to reasonable values (Albeit `VARIANT=hotspot`, and this change defaults to `jdk11u` instead of `jdk8u` but that shouldn't be a problem.

build-farm/make-adopt-build-farm.sh
```diff
@@ -18,16 +18,40 @@ set -e
 
 PLATFORM_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-## Sanity check as this script requires various environment variables so fail
-## fast if one or more are not present. MAYBE autodetect defaults in future?
+## autodetect defaults to improve usability when running this for debugging/testing
+
+if [ -z "$ARCHITECTURE"  ]; then
+   ARCHITECTURE=`uname -p`
+   if [       "$OSTYPE" = "cygwin"  ]; then ARCHITECTURE=`uname -m`; fi
```
M-Davies, line 25:
```suggestion
   if [ "$OSTYPE" = "cygwin"  ]; then ARCHITECTURE=`uname -m`; fi
```

sxa: Hmmm I disagree with this - better to have the `[ ]; then ARCHITECTURE=` lined up for readability

M-Davies: Why not increase the indentation of the others to match for it in that case? I'll approve this regardless as it's just a suggestion

M-Davies (approved):

build-farm/make-adopt-build-farm.sh
```diff
@@ -18,16 +18,40 @@ set -e
...
+   if [ "$ARCHITECTURE" = "x86_64"  ]; then ARCHITECTURE=x64;        fi
+   if [ "$ARCHITECTURE" = "powerpc" ]; then ARCHITECTURE=ppc64;      fi
+   echo ARCHITECTURE not defined - assuming $ARCHITECTURE
+   export ARCHITECTURE
+fi
+
+if [ -z "$TARGET_OS" ]; then
```
karianna, line 32: Some comments on these if blocks (or extracting them to functions if possible) will help

sxa: I thought they were self-explanatory but I can add a comment sayting "translate uname output to the names which adoptopenjdk uses"

sxa: Added commands (and other special cases for Solaris which I didn't include)

build-farm/make-adopt-build-farm.sh
sxa: Personal preference I guess - I definitely prefer:
```
   if [       "$OSTYPE" = "cygwin"  ]; then ARCHITECTURE=`uname -m`; fi
   if [ "$ARCHITECTURE" = "x86_64"  ]; then ARCHITECTURE=x64;        fi
   if [ "$ARCHITECTURE" = "powerpc" ]; then ARCHITECTURE=ppc64;      fi
```
over 
```
if [ "$OSTYPE" = "cygwin"        ]; then ARCHITECTURE=`uname -m`; fi
if [ "$ARCHITECTURE" = "x86_64"  ]; then ARCHITECTURE=x64;        fi
if [ "$ARCHITECTURE" = "powerpc" ]; then ARCHITECTURE=ppc64;      fi
```
because the `=` are lined up, so I think I'll leave it, but noted for the future ;-)

sxa: (I suppose the third option would be:
```
if [ "$OSTYPE"       = "cygwin"  ]; then ARCHITECTURE=`uname -m`; fi
if [ "$ARCHITECTURE" = "x86_64"  ]; then ARCHITECTURE=x64;        fi
if [ "$ARCHITECTURE" = "powerpc" ]; then ARCHITECTURE=ppc64;      fi
```
Hmmmm ....

sxa: OK I'll compromise on that last one :-) Updated

karianna (approved):
LGTM

Issue #1409, opened 2019-11-28
HotSpot JDK11 builds failing to bootstrap with JDK10


sxa555:
Ref: http://mail.openjdk.java.net/pipermail/jdk-updates-dev/2019-November/002172.html

As per the normal rule, we currently build JDK11 using JDK10 as the bootstrap, but this has been broken. Based on the mailing list we expect a fix to be forthcoming.

```
=== Output from failing command(s) repeated here ===
* For target buildtools_jdk_tools_classes__the.BUILD_TOOLS_JDK_batch:
/ramdisk0/build/workspace/build-scripts/jobs/jdk11u/jdk11u-aix-ppc64-hotspot/workspace/build/src/make/jdk/src/classes/build/tools/generatecacerts/GenerateCacerts.java:84: error: cannot find symbol
        List<String> entries = Files.list(Path.of(dir))
                                              ^
  symbol:   method of(String)
  location: interface Path
/ramdisk0/build/workspace/build-scripts/jobs/jdk11u/jdk11u-aix-ppc64-hotspot/workspace/build/src/make/jdk/src/classes/build/tools/generatecacerts/GenerateCacerts.java:97: error: cannot find symbol
            try (InputStream fis = Files.newInputStream(Path.of(dir, entry))) {
                                                            ^
  symbol:   method of(String,String)
  location: interface Path
```

karianna:
Holding @sxa555 responsible to revert this at the right time ;-)

PR #2479, opened 2021-02-17
Remove java 11 boot JDK restriction (Reverts PR 1416 and then some!)

sxa:
Reverts #1409 and allows a JDK10 boot JDK (For JDK11) to be auto-detected and downloaded, which https://github.com/AdoptOpenJDK/openjdk-build/pull/2167 was not able to do.

JDK7 (for JDK8) will still be detected if present, but will not be auto-downloaded (highly unlikely any machine would not have it set up anyway...)

Summary:
- Removes the conditional surrounding the detection/download
- Adjusts the "Downloading" message to print the API target we are downloading from (means moving that line to after `apiURL is set`)
- Changes indentation and the wrapping of one comment (which is why this looks more substantial than it is (Suggest reviewers do the review with whitespace ignored)
- Moves the printing of the `BOOT JDK` `java -version` output little earlier, and suppresses the printing on the `java -version` which checks that it works (moving the visible one earlier means that any "odd" errors won't get hidden from view now we're suppressing the second output

Testing at https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/1053/

### DIFF
--- a/build-farm/platform-specific-configurations/aix.sh
+++ b/build-farm/platform-specific-configurations/aix.sh
@@ -46,56 +46,52 @@ if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]; then
 fi
 echo LDR_CNTRL=$LDR_CNTRL
 
-# Any version above 8 (11 for now due to openjdk-build#1409
-if [ "$JAVA_FEATURE_VERSION" -gt 11 ]; then 
-  BOOT_JDK_VERSION="$((JAVA_FEATURE_VERSION-1))"
-  BOOT_JDK_VARIABLE="JDK$(echo $BOOT_JDK_VERSION)_BOOT_DIR"
-  if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
-    bootDir="$PWD/jdk-$BOOT_JDK_VERSION"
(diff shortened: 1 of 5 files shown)


karianna (approved):
LGTM

build-farm/platform-specific-configurations/aix.sh
```diff
@@ -46,56 +46,53 @@ if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]; then
...
+        echo "Downloading GA release of boot JDK version ${BOOT_JDK_VERSION} failed."
+        echo "Attempting to download EA release of boot JDK version ${BOOT_JDK_VERSION} from ${apiURL}"
+        # shellcheck disable=SC2034
+        releaseType="ea"
         apiURL=$(eval echo ${apiUrlTemplate})
-        # make-adopt-build-farm.sh has 'set -e'. We need to disable that
-        # for the fallback mechanism, as downloading of the GA binary might
-        # fail.
-        set +e
+        echo "Attempting to download EA release of boot JDK version ${BOOT_JDK_VERSION} from ${apiURL}"
```
M-Davies, line 81: Why do state we're downloading twice?

sxa: Cut and paste error - fixed. Good spot

M-Davies (approved):

karianna (approved):
LGTM

PR #2482, opened 2021-02-19
Fix for boot JDK autodetection in windows.sh

sxa:
echo statement was correct, but the following line was incorrect due to a cut & paste error meaning it was set to the PATH on Linux/AIX systems instead of windows.


M-Davies (changes requested):
Conflicts 😢

karianna (approved):
LGTM

M-Davies (approved):

PR #2489, opened 2021-02-23
AIX: LIBPATH to include /opt/freeware/lib/pthread/ppc64 at the front

Haroon-Khel:
ref https://github.com/eclipse/openj9/issues/12018 and https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1952

Ive updated the LIBPATH variable for AIX so that it provides a path to the library /opt/freeware/lib/gcc/powerpc-ibm-aix7.1.0.0/8/pthread/libstdc++.a, which has a symlink in /opt/freeware/lib/pthread/ppc64.

This solves a cmake issue we seem to be getting, which causes the following error
```
13:56:15  Could not load program cmake:
13:56:15  Symbol resolution failed for cmake because:
13:56:15  	Symbol _ZTINSt6thread6_StateE (number 358) is not exported from dependent
13:56:15  	  module /opt/freeware/lib/libstdc++.a[libstdc++.so.6].
13:56:15  	Symbol _ZTVNSt6thread6_StateE (number 359) is not exported from dependent
13:56:15  	  module /opt/freeware/lib/libstdc++.a[libstdc++.so.6].
13:56:15  	Symbol _ZNSt6thread4joinEv (number 379) is not exported from dependent
13:56:15  	  module /opt/freeware/lib/libstdc++.a[libstdc++.so.6].
...
```

### DIFF
--- a/build-farm/platform-specific-configurations/aix.sh
+++ b/build-farm/platform-specific-configurations/aix.sh
@@ -30,7 +30,7 @@ source "$SCRIPT_DIR/../../sbin/common/constants.sh"
 export PATH="/opt/freeware/bin:/usr/local/bin:/opt/IBM/xlC/13.1.3/bin:/opt/IBM/xlc/13.1.3/bin:$PATH"
 # Without this, java adds /usr/lib to the LIBPATH and it's own library
 # directories of anything it forks which breaks linkage
-export LIBPATH=/opt/freeware/lib:/usr/lib
+export LIBPATH=/opt/freeware/lib/pthread/ppc64:/opt/freeware/lib:/usr/lib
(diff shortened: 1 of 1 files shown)


M-Davies (approved):

sxa (approved):
I think this will work, but not 100% sure it won't have side effects so I'm going to merge and kick off another JDK11u pipeline with AIX only and no tests)

jdekonin:
If this apart of a compile/build job that creates binaries, you should check to make sure it isn't baking in any dependencies.  If the SDK cannot run on a system that doesn't have this installed `/opt/freeware/lib/pthread/ppc64` it will need to be undone.

Issue #2466, opened 2021-02-12
Intermittent AIX jvmti.h build failure


andrew-m-leonard:
See: https://github.com/AdoptOpenJDK/openjdk-build/issues/1450#issuecomment-778061361
```
11:01:20  In file included from ./src/hotspot/share/interpreter/abstractInterpreter.cpp:45:
11:01:20  In file included from ./src/hotspot/share/prims/jvmtiExport.hpp:28:
11:01:20  ./build/aix-ppc64-server-release/hotspot/variant-server/gensrc/jvmtifiles/jvmti.h:1766:5: error: expected parameter declarator
11:01:20      );
11:01:20      ^
11:01:20  ./build/aix-ppc64-server-release/hotspot/variant-server/gensrc/jvmtifiles/jvmti.h:1770:5: error: expected parameter declarator
11:01:20      );
11:01:20      ^
11:01:20  ./build/aix-ppc64-server-release/hotspot/variant-server/gensrc/jvmtifiles/jvmti.h:1774:5: error: expected parameter declarator
11:01:20      );
11:01:20      ^
11:01:20  ./build/aix-ppc64-server-release/hotspot/variant-server/gensrc/jvmtifiles/jvmti.h:1778:5: error: expected parameter declarator
11:01:20      );
11:01:20      ^
11:01:20  ./build/aix-ppc64-server-release/hotspot/variant-server/gensrc/jvmtifiles/jvmti.h:1782:5: error: expected parameter declarator
11:01:20      );
11:01:20      ^
11:01:20  ./build/aix-ppc64-server-release/hotspot/variant-server/gensrc/jvmtifiles/jvmti.h:1786:5: error: expected parameter declarator
11:01:20      );
11:01:20      ^
11:01:20  ./build/aix-ppc64-server-release/hotspot/variant-server/gensrc/jvmtifiles/jvmti.h:1790:5: error: expected parameter declarator
11:01:20      );
11:01:20      ^
11:01:20  ./build/aix-ppc64-server-release/hotspot/variant-server/gensrc/jvmtifiles/jvmti.h:1794:5: error: expected parameter declarator
11:01:20      );
11:01:20      ^
11:01:20  ./build/aix-ppc64-server-release/hotspot/variant-server/gensrc/jvmtifiles/jvmti.h:1798:5: error: expected parameter declarator
11:01:20      );
11:01:20      ^
[...]

PR #2498, opened 2021-02-26
Use AIX hotspot bootjdk to avoid issue #2466

andrew-m-leonard:
Issue https://github.com/AdoptOpenJDK/openjdk-build/issues/2466 when building jdk16/17 AIX hotspot, using a hotspot boot jdk instead avoids the problem.


M-Davies (approved):

karianna (approved):
LGTM

PR #2507, opened 2021-03-02
Fix gcc-9.3 build error and limit gcc 9.3 on dragonwell aarch64

joeyleeeeeee97:
Related issue: https://github.com/AdoptOpenJDK/openjdk-build/issues/2501
In https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-x64-dragonwell/164/console , we should be using 
```
CC=/usr/local/gcc9/bin/gcc-9.3
```
Also, limit GCC only on  aarch64 platform

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -137,10 +137,10 @@ if [ $executedJavaVersion -ne 0 ]; then
     exit 1
 fi
 
-if [ "${VARIANT}" == "${BUILD_VARIANT_DRAGONWELL}" ] && [ "$JAVA_FEATURE_VERSION" -eq 11 ] && [ -r /usr/local/gcc9/ ]; then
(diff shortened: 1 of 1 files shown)


joeyleeeeeee97:
@sxa Thanks for adding GCC 9.3.0 on ' adoptopenjdk/centos6_build_image', please also add them on adoptopenjdk/centos7_build_image? Thanks

sxa (approved):
I looked at that block of code several times when I was working on making gcc9 available but didn't spot that it said `CC` instead of `CXX`. Yes I definitely approve of this.

I hadn't realised you only wanted GCC9 on aarch64 - is that intentional? So we no longer need it on the x64 systems? Or is the intention to use GCC9 on x64 at some point in the future?

sxa:
> @sxa Thanks for adding GCC 9.3.0 on ' adoptopenjdk/centos6_build_image', please also add them on adoptopenjdk/centos7_build_image? Thanks

Hmmm both should be automatic. If it hasn't gone into both we'll need to investigate why ...

M-Davies (approved):

sxa:
Verifying at https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk11-pipeline/1414/

PR #2516, opened 2021-03-04
Fixup RISC-V cross-compiler check so it doesn't fail with set -e

sxa:
The original check of `[ ! -x "$CXX" ]` returns false when the location exists, so the `set -e` causes the script to fail at that point which is not the desired behaviour. This reverses the logic so it works as expected. I'm on a roll today ...

This needs to be put in before the nightly builds


andrew-m-leonard (approved):
lgtm

adamfarley (approved):
LGTM

PR #2506, opened 2021-03-01
Update FAQ to account for #2463

M-Davies:
* Twin PR for ci-jenkins-pipelines -> https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/pull/67

Signed-off-by: Morgan Davies <morgandavies2020@gmail.com>

Issue #2129, opened 2020-10-08
Clarity on where config parameters should be defined


sxa:
Spun out from discussion in https://github.com/AdoptOpenJDK/openjdk-build/pull/2125#pullrequestreview-504661752

The location for changes to config parameters is something we need to provide guidance on. I came across it when putting [this section of the FAQ](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/FAQ.md#how-do-i-change-the-parameters-such-as-configure-flags-for-a-jenkins-build) together as things are currently in one of three places. Which to use depends on who we want to be affected and to me knowledge we've avoided being clear on where changes should be made which has already lead to confusion in the past. Quick summary:

| Location | Impact |
| --- | --- |
| groovy files (as per this PR) | Only when run through our jenkins pipelines |
| [platform-specific-configurations](https://github.com/AdoptOpenJDK/openjdk-build/tree/master/build-farm/platform-specific-configurations) scripts | Those using [build-farm/make-adopt-build-farm.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/build-farm/make-adopt-build-farm.sh) (inc. our pipelines) - should be stuff specific to our machines |
| [build.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh) | Anyone (including end users) who are running [makejdk-any-platform.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/makejdk-any-platform.sh) |

So it depends what we want the last line of those to be. If it's to allow users to replicate the adopt builds with the same configuration options as us as far as possible, then I think it ought to be in build.sh but if we want it to be optional for users building it themselves then the jenkins pipelines aren't a bad choice. But we should really make it clear for new people to the project where changes should be made e.g. via an update to the FAQ.

We should discuss when to use each type, citing examples of when things should be added in each of the above three places.

I would suggest:
- The platform scripts when it affects environment variables for the bulids, and including config parameters to override memory usage for our specific build machines. We do currently have things like CUDA and OpenSSL enablement for OpenJ9 in there - they could be moved into build.sh is we want to have VM specifics in there (Putting it into the groovy would result in lots more editing if they change).
[...]

PR #2518, opened 2021-03-05
Expand config parameter location explanation

M-Davies:
* Further explains how and where shell script parameters should be injected
* Follow on from #2506
* Part 2 of #2129

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -63,12 +63,17 @@ All machines at AdoptOpenJDK are set up using the ansible playbooks from the
 
 ## How do I change the parameters, such as configure flags, for a Jenkins build
 
-Either:
+Where you change them depends on the scope of the parameter or flag:
 
-- Modify the environment files in [platform-specific-configurations](https://github.com/AdoptOpenJDK/openjdk-build/tree/master/build-farm/platform-specific-configurations)
-- Modify the [pipeline files](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/tree/master/pipelines/build), although this is normally only done for configuration differences such as OpenJ9 Large Heap builds. Please only do this if the changes you wish to make will effect the Jenkins builds.
-
-[Example PR - Adding a new configure flag for OpenJ9](https://github.com/AdoptOpenJDK/openjdk-build/pull/1442/files)
+- *If the parameter will affect all users, regardless of environment or OS/Arch*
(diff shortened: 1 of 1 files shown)


sxa (commented):
This superficially looks like a good set of recommendations, but I'll hold off on approving until Monday as I'm quite tired just now. I do think it's something that most of the build team shoul dhave input to so might be good to circulate in #build for wider review and opinions, but this will be a great addition to the documentation so thanks for kicking it off!

karianna (approved):
LGTM

FAQ.md
```diff
@@ -63,12 +63,17 @@ All machines at AdoptOpenJDK are set up using the ansible playbooks from the
...
-- Modify the [pipeline files](https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/tree/master/pipelines/build), although this is normally only done for configuration differences such as OpenJ9 Large Heap builds. Please only do this if the changes you wish to make will effect the Jenkins builds.
-
-[Example PR - Adding a new configure flag for OpenJ9](https://github.com/AdoptOpenJDK/openjdk-build/pull/1442/files)
+Where you change them depends on the scope of the parameter or flag:
+
+- *If the parameter will affect all users, regardless of environment or OS/Arch*
+  - [build.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/sbin/build.sh) OR [makejdk-any-platform.sh](https://github.com/AdoptOpenJDK/openjdk-build/blob/master/makejdk-any-platform.sh) depending on how high up in the execution stack it needs to be.
+  - [Example PR - Adding a new archival feature for OpenJ9 memory dumps](https://github.com/AdoptOpenJDK/openjdk-build/pull/2464)
+- *If the parameter will affect all machines of a specific OS OR will affect a jenkins machine style environment at the shell script level*
```
sxa, line 71:
```suggestion
- *If the parameter will affect all machines of a specific OS OR related to the environment set up by [our ansible scripts](https://github.com/AdoptOpenJDK/openjdk-infrastructure) at the shell script level*
```

sxa (commented):
I'm not a fan of "jenkins machine style environment" as that's a bit vague. I'd suggest that the intention was to reference machines in the jenkins farm which were set up by ansible specifically so I'd recommend different wording explicitly pointing to the infra repo as per my suggested change.

We possibly have another piece of work to do to ensure that all variables are set in the "right" place according to these guidelines but this is a good start and we badly need this sort of guidance :-)

sxa (approved):
:+1:

PR #2520, opened 2021-03-05
Unbreak RISC-V OpenJ9 build

sxa:
We lost the `--disable-ddr` and build JDK options required for OpenJ9 in https://github.com/AdoptOpenJDK/openjdk-build/pull/2509/files - this reinstates them to allow it to build properly again.

Running some tests to verify but I'm confident this will fix everything. Will also need https://github.com/AdoptOpenJDK/ci-jenkins-pipelines/pull/81 but I'm confident that this will end up with the simplest code that will build OpenJ9 and the Bisheng RISC-V builds in the desired tested environments and configurations.

Please leave me to merge this once after review to ensure I've completed testing.


M-Davies (approved):
LGTM

andrew-m-leonard (approved):
lgtm

Issue #2508, opened 2021-03-02
make-adopt-build-farm.sh does not handle mistyped names


sxa:
**What are you trying to do?** Run make-adopt-build-farm with an alternate VARIANT set

**Expected behaviour:** If I mistype the VARIANT value it aborts

**Observed behaviour:** If I mistype VARIANT it defaults to hotspot

**Any other comments:**

M-Davies:
Is that not what we want? When running without params set, it informs you that it is defaulting to Hotspot (like the other params):

```shell
(base) ➜  build-farm git:(master) ./make-adopt-build-farm.sh 
ARCHITECTURE not defined - assuming x64
TARGET_OS not defined - assuming you want Darwin
JAVA_TO_BUILD not defined - defaulting to jdk11u
VARIANT not defined - assuming hotspot
FILENAME not defined - assuming jdk11u-hotspot.tar.gz
BUILD TYPE: 
VERSION: jdk11u
ARCHITECTURE x64
VARIANT: hotspot
OS: darwin
SCM_REF: 
Detecting boot jdk for: jdk11u
Found build version: 11
Required boot JDK version: 10
[ERROR] No local file detected at /Users/morgan/Documents/Repos/openjdk-build/build-farm/platform-specific-configurations/darwin.sh and PLATFORM_CONFIG_LOCATION is not set. Please set PLATFORM_CONFIG_LOCATION to a repository path of a platform config file (e.g. AdoptOpenJDK/openjdk-build/master/build-farm/platform-specific-configurations).
```

Would you prefer to induce a "hard" failure where the script will fail if it does not detect one or more of these params?

PR #2521, opened 2021-03-05
Terminate build on unknown variant type

M-Davies:
Fixes: #2508 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ pipelines/.gradle
 pipelines/gradle-cache
 pipelines/target
 **/.DS_Store
+build-farm/platform-specific-configurations/platformConfigFile.sh
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -55,9 +55,19 @@ if [ -z "$JAVA_TO_BUILD" ]; then
   fi
 fi
 
-[ -z "$JAVA_TO_BUILD" ] && echo JAVA_TO_BUILD not defined - set to e.g. jdk8u && SANEVARS=1
+[ -z "$JAVA_TO_BUILD" ] && echo JAVA_TO_BUILD not defined - set to e.g. jdk8u
 [ -z "$VARIANT"       ] && echo VARIANT not defined - assuming hotspot && export VARIANT=hotspot
 [ -z "$FILENAME"      ] && echo FILENAME not defined - assuming "${JAVA_TO_BUILD}-${VARIANT}.tar.gz" && export FILENAME="${JAVA_TO_BUILD}-${VARIANT}.tar.gz"
+
+# shellcheck source=sbin/common/constants.sh
+source "$PLATFORM_SCRIPT_DIR/../sbin/common/constants.sh"
+
+# Check that the given variant is in our list of common variants
+# shellcheck disable=SC2086,SC2143
+if [ -z "$(echo ${BUILD_VARIANTS} | grep -w ${VARIANT})" ]; then
+  echo "[ERROR] ${VARIANT} is not a recognised build variant. Valid Variants = ${BUILD_VARIANTS}"
+  exit 1
+fi
 
 ## Very very build farm specific configuration
 export OPERATING_SYSTEM
@@ -105,7 +115,6 @@ echo "OS: ${OPERATING_SYSTEM}"
 echo "SCM_REF: ${SCM_REF}"
 OPTIONS=""
 
-EXTENSION=""
 # shellcheck disable=SC2034
 CONFIGURE_ARGS_FOR_ANY_PLATFORM=""
 CONFIGURE_ARGS=${CONFIGURE_ARGS:-""}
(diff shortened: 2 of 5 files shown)


sxa (commented):
Have I ever mentioned how much I hate both arrays and regex in shell scripts? :-) (I'd personally have done something like `case $VARIANT in; hotspot|openj9|bisheng|coretto|dragonwell) ;; *) echo $VARIANT is not a valid variant && exit 1;; esac` or even `if [ -z "echo $LISTOFVARIANTS | grep $VARIANT" ]; then echo Invalid variant; exit 1; fi`
Regardless, in principle this will work (and don't feel compelled to change how it's done) but from the checks it's not letting `hotspot` through at the moment so will need tweaked.

M-Davies:
That logic implies more hardcoding of variants in places that are not so obvious. What is the problem with using regexes on arrays in shell?

sxa:
> That logic implies more hardcoding of variants in places that are not so obvious. What is the problem with using regexes on arrays in shell?

Two reasons:
1. I find them quite a bit more difficult to read and parse than alternatives (The array is being defined in your PR only for the purpose of this at the moment - it could just as easily being a space-separated string to be used in my `grep` example)
2. (Personal one) I've always preferred to keep as close to standard POSIX shell syntax as far as possible as opposed to relying on `bash`-specifics as possible for portability

But as I said, not something I'd insist on being changed - we've filled the scripts with lots of bash-isms already and they're not going away ;-)

build-farm/make-adopt-build-farm.sh
```diff
@@ -55,10 +55,19 @@ if [ -z "$JAVA_TO_BUILD" ]; then
...
-[ -z "$JAVA_TO_BUILD" ] && echo JAVA_TO_BUILD not defined - set to e.g. jdk8u && SANEVARS=1
+[ -z "$JAVA_TO_BUILD" ] && echo JAVA_TO_BUILD not defined - set to e.g. jdk8u
 [ -z "$VARIANT"       ] && echo VARIANT not defined - assuming hotspot && export VARIANT=hotspot
 [ -z "$FILENAME"      ] && echo FILENAME not defined - assuming "${JAVA_TO_BUILD}-${VARIANT}.tar.gz" && export FILENAME="${JAVA_TO_BUILD}-${VARIANT}.tar.gz"
 
+# shellcheck source=sbin/common/constants.sh
+source "$PLATFORM_SCRIPT_DIR/../sbin/common/constants.sh"
+
+# Check that the given variant is in our list of common variants
+if [ -z "$(echo "$BUILD_VARIANTS" | grep -w "$VARIANT")" ]; then
```
sxa, line 66: The quotes inside another set of quotes looks a bit odd, but I presume it's been tested to work inside the `$()`

build-farm/make-adopt-build-farm.sh
```diff
@@ -105,7 +114,6 @@ echo "OS: ${OPERATING_SYSTEM}"
 echo "SCM_REF: ${SCM_REF}"
 OPTIONS=""
 
-EXTENSION=""
```
sxa, line 108: As long as the removal of this doesn't cause a problem when the variable is used within `sign-releases.sh` I'm ok with this - and I don't think it will.

sxa (approved):
Subject to the other two comments not being things that will cause problems I approve this.

build-farm/make-adopt-build-farm.sh
M-Davies: This was mostly because shellcheck complained about them not being in quotes. It still picks up an invalid VARIANT but I can look into removing them or disabling the shellcheck rule if it's annoying to see?

build-farm/make-adopt-build-farm.sh
M-Davies: I checked and it doesn't seem to as EXTENSION is always set by `sign-releases.sh`:
https://github.com/AdoptOpenJDK/openjdk-build/blob/cf47cbc28665a8707485c49c58095c740a80fa43/build-farm/sign-releases.sh#L23-L30

build-farm/make-adopt-build-farm.sh
sxa: No I'm not overly concerned about it - happy for it to be left unless the next reviewer objects :-)
Thanks for the other changes though!

M-Davies: It has already been implemented 🙂

.gitignore
```diff
@@ -6,3 +6,4 @@ pipelines/.gradle
 pipelines/gradle-cache
 pipelines/target
 **/.DS_Store
+build-farm/platform-specific-configurations/platformConfigFile.sh
```
karianna, line 9: Why is this file being ignored?

karianna (approved):
LGTM

.gitignore
M-Davies: It's an automatically generated file by `set-platform-config.sh` and shouldn't be commited during builds

PR #2525, opened 2021-03-10
Allow other sign tools via param SIGN_TOOL

AdamBrousseau:
Add param SIGN_TOOL that will allow other signing
tools to be used.
Add case for Windows "ucl" sign tool.
Default to signtool.exe on Windows.

### DIFF
--- a/build-farm/sign-releases.sh
+++ b/build-farm/sign-releases.sh
@@ -18,6 +18,7 @@
 BUILD_ARGS=${BUILD_ARGS:-""}
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+export SIGN_TOOL
 export OPERATING_SYSTEM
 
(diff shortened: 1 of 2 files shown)


M-Davies (commented):
How is `SIGN_TOOL` to be specified? Is it by an environment variable or pass via `--sign`?

AdamBrousseau:
`SIGN_TOOL` would be a default configured Jenkins param on the sign_build job. I have added an export to `signRelease.sh` if that's what you're looking for, like is done for `OPERATING_SYSTEM`. Although when running through Jenkins it does not seem to be necessary to export it.

To be honest I didn't think about the mac use case but I won't be using the job for mac for the foreseeable future.

karianna (approved):
LGTM

M-Davies (approved):
Looks good but I don't think many people will realise this exists as it isn't immediately obvious

PR #2529, opened 2021-03-15
Add warning if AIX ulimit -n is too low

sxa:
Warn user if AIX ulimit is too low to perform the git checkout operations.
For all but JDK11, the default AIX limit of `1048575` should (only just for JDK8 - it needs 860Mb at the time of writing) work. Our jenkins users have this limit larger by defualt, but for anyone running our scripts themselves, this will warn them and pause to try and encourage to read the message otherwise they are likely to end up with things like this:
```
Cloning into 'openjdk-jdk11u'...
fatal: write error: A file cannot be larger than the value set by ulimit.
fatal: index-pack failed
```

### DIFF
--- a/build-farm/platform-specific-configurations/aix.sh
+++ b/build-farm/platform-specific-configurations/aix.sh
@@ -20,6 +20,7 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # AIX default ulimit is frequently less than we need to clone the LTS JDK repositories
 FILESIZELIMIT=$(ulimit)
 if [ "$FILESIZELIMIT" != "unlimited" ]; then
+  # Set to ~2GB as this works for the AIX hosts we have as of April 2021
   if [ "$FILESIZELIMIT" -lt 2097150 ]; then
     echo "WARNING: MAXIMUM USER FILE SIZE (ulimit -n) IS $FILESIZELIMIT (<2097150) - GIT MAY HAVE PROBLEMS CLONING"
     sleep 5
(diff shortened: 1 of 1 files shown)


aixtools:
Just a bit surprised by the value in the comment above, compared to the actual code.

Further, I would not stare at the (current) default - as a test value - but take a value we already know to be a minimum.
e.g. "Say unlimited is preferred, XYZ (blocks) is (approximate) minimum."

And maybe - rather than sleep(5) - a bit more 'invasive' and `press return to continue`

sxa:
> Just a bit surprised by the value in the comment above, compared to the actual code.

Can you clarify that? What value are you surprised by?

> And maybe - rather than sleep(5) - a bit more 'invasive' and press return to continue

I disagree - I don't want anything in this to be interactive.

sxa:
(Linter checks are unrelated to this PR. Not sure why it's suddenly become a problem but I'm not going to fix all of those instances in various scripts in this PR)

build-farm/platform-specific-configurations/aix.sh
```diff
@@ -16,6 +16,16 @@
 
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# AIX default ulimit is frequently less than we need to clone the LTS JDK repositories
```
karianna, line 20: Why the value of 2097150?

karianna (changes requested):

karianna (approved):

smlambert (approved):
LGTM

PR #2539, opened 2021-03-20
checks for processes on Linux/arm32 builds for scaleway

sxa:
We often have scaleway build machines offline - most recently [here](https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/2061) - this is to add some extra checks that will be performed on those systems before a build is executed to see if there are any obvious problems. I do not anticipate this being permanent.


karianna (approved):
LGTM - I'll create an issue to reverse it later post investigation

gdams (approved):

M-Davies (approved):

PR #2541, opened 2021-03-20
make-adopt-build-farm: Use arm instead of armv7l

sxa:
Found while verifying https://github.com/AdoptOpenJDK/openjdk-build/pull/2539. Introduced with the new autodetection in https://github.com/AdoptOpenJDK/openjdk-build/commit/3ec1ae9aa9d66114ee07be15205a2b6064c85101 (which isn't used by our pipelines)

While we use `armv7l` for `OS_ARCHITECTURE`, the `ARCHITECTURE` variables should be `arm` for arm32 builds.


M-Davies (changes requested):
Linter issues

sxa:
> Linter issues

Those seem to be linter issue that have nothing to do with this PR - I would suggest they need to be part of a separate PR as it will make it less obvious what was really changed to fix the problem that this PR is addressing.

M-Davies (approved):

sxa:
(Raised https://github.com/AdoptOpenJDK/openjdk-build/issues/2542 for the linter problems)

andrew-m-leonard (approved):

PR #2547, opened 2021-03-24
Stop use of 'basename -s' which does not work on Cent/RHEL6

sxa:
Linux/x64 build logs have:
```
17:00:21  ===RELEASE FILE GENERATED===
17:00:21  basename: invalid option -- 's'
17:00:21  Try `basename --help' for more information.
```
I believe this is equivalent to what the `-s` was intending to trap.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -1125,7 +1125,7 @@ addVendorToJson(){
 }
 
 addSourceToJson(){ # Pulls the basename of the origin repo, or uses 'openjdk-build' in rare cases of failure
-  local repoName=$(basename -s .git $(cd ${BUILD_CONFIG[WORKSPACE_DIR]} && git config --get remote.origin.url 2>/dev/null))
+  local repoName=$(basename $(cd ${BUILD_CONFIG[WORKSPACE_DIR]%.git} && git config --get remote.origin.url 2>/dev/null))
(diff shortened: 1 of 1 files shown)


sbin/build.sh
```diff
@@ -1125,7 +1125,7 @@ addVendorToJson(){
 }
 
 addSourceToJson(){ # Pulls the basename of the origin repo, or uses 'openjdk-build' in rare cases of failure
-  local repoName=$(basename -s .git $(cd ${BUILD_CONFIG[WORKSPACE_DIR]} && git config --get remote.origin.url 2>/dev/null))
+  local repoName=$(basename $(cd ${BUILD_CONFIG[WORKSPACE_DIR]%.git} && git config --get remote.origin.url 2>/dev/null))
```
M-Davies, line 1128: Should this not be contained in quotes to avoid the % splitting out?
```suggestion
  local repoName=$(basename $(cd "${BUILD_CONFIG[WORKSPACE_DIR]%.git}" && git config --get remote.origin.url 2>/dev/null))
```

sxa: I don't think there's any situation in which that specifically would be a problem given that it's within the `${ }` (interested to see if you can make one up though!) .. And as far as I can see even our new aggressive linting hasn't objected to this, but I can change it

M-Davies (approved):

gdams (approved):

Issue #2530, opened 2021-03-16
link failures in platform-specific-configuration scripts


sxa:
**What are you trying to do?** Submit a PR

**Expected behaviour:** LInt checks pass in areas I haven'' touched

**Observed behaviour:**
```
2021-03-15 18:14:02 [INFO]   Linting [BASH] files...
2021-03-15 18:14:02 [INFO]   ----------------------------------------------
2021-03-15 18:14:02 [INFO]   ----------------------------------------------
2021-03-15 18:14:02 [INFO]   ---------------------------
2021-03-15 18:14:02 [INFO]   File:[build-farm/platform-specific-configurations/aix.sh]
2021-03-15 18:14:02 [ERROR]   Found errors in [shellcheck] linter!
2021-03-15 18:14:02 [ERROR]   Error code: 1. Command output:
------

In build-farm/platform-specific-configurations/aix.sh line 33:
  echo Found $(find $TMPDIR -type f -print | wc -l) items in it - removing them
             ^-- SC2046: Quote this to prevent word splitting.


In build-farm/platform-specific-configurations/aix.sh line 60:
BOOT_JDK_VARIABLE="JDK$(echo $BOOT_JDK_VERSION)_BOOT_DIR"
                      ^-----------------------^ SC2116: Useless echo? Instead of 'cmd $(echo foo)', just use 'cmd foo'.


In build-farm/platform-specific-configurations/aix.sh line 65:
  export ${BOOT_JDK_VARIABLE}="${bootDir}"
         ^------------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
  export "${BOOT_JDK_VARIABLE}"="${bootDir}"


In build-farm/platform-specific-configurations/aix.sh line 69:
      echo Could not use ${BOOT_JDK_VARIABLE} - using /usr/java${BOOT_JDK_VERSION}_64
                         ^------------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
      echo Could not use "${BOOT_JDK_VARIABLE}" - using /usr/java${BOOT_JDK_VERSION}_64


In build-farm/platform-specific-configurations/aix.sh line 70:
      export ${BOOT_JDK_VARIABLE}="/usr/java${BOOT_JDK_VERSION}_64"
             ^------------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
[...]

Issue #2542, opened 2021-03-22
linter failures in make-adopt-build-farm.sh


sxa:
**What are you trying to do?** Submit PRs

**Expected behaviour:** lint checks run successfully on areas of code that aren't modified by my PRs

**Observed behaviour:** linker fails: e.g. https://github.com/AdoptOpenJDK/openjdk-build/pull/2541/checks?check_run_id=2155314328

**Any other comments:** Since this wasn't previously a problem I would imagine the linter has been updated with new rules that have caused a problem.

If we don't already have them, we should perhaps have clearer guidelines in the CONTRIBUTING guide to help people run the checks on their own system.

sxa:
> The new rules was probably me adding the super linter in #2513 . I'll look into this and your twin issue #2530

I was going to mention that one but I thought I'd found a change that had gone into this file since then. But it does seem the most likely option :-)

PR #2548, opened 2021-03-24
Shellcheck fixes and prtester doc move

M-Davies:
* Fixes shellchecker linter issues inside super linter job
* Moves the pr tester documentation out of redundant pipelines dir

Fixes: #2530 , #2542 


M-Davies:
Please do not merge if any checks fail as I would like to validate that my changes did not cause them

karianna (approved):
LGTM

Willsparker (approved):

build-farm/make-adopt-build-farm.sh
```diff
@@ -126,6 +127,7 @@ if [ -z "${JDK_BOOT_VERSION}" ]
 then
   echo "Detecting boot jdk for: ${JAVA_TO_BUILD}"
   echo "Found build version: ${JAVA_FEATURE_VERSION}"
+  # shellcheck disable=SC2004
   JDK_BOOT_VERSION=$(($JAVA_FEATURE_VERSION-1))
```
sxa, line 131: Might as well fix this one instead of bypassing the check

```suggestion
  JDK_BOOT_VERSION=$(( JAVA_FEATURE_VERSION - 1 ))
```

build-farm/make-adopt-build-farm.sh
```diff
@@ -206,7 +208,7 @@ if [ "${OPERATING_SYSTEM}" != "aix" ] ; then
     export BUILD_ARGS="${BUILD_ARGS} --create-debug-image"
 fi
 
-echo "$PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh" --clean-git-repo --jdk-boot-dir "${JDK_BOOT_DIR}" --configure-args "${CONFIGURE_ARGS_FOR_ANY_PLATFORM}" --target-file-name "${FILENAME}" ${TAG_OPTION} ${OPTIONS} ${BUILD_ARGS} ${VARIANT_ARG} "${JAVA_TO_BUILD}"
+echo "$PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh" --clean-git-repo --jdk-boot-dir "${JDK_BOOT_DIR}" --configure-args "${CONFIGURE_ARGS_FOR_ANY_PLATFORM}" --target-file-name "${FILENAME}" "${TAG_OPTION}" "${OPTIONS}" "${BUILD_ARGS}" "${VARIANT_ARG}" "${JAVA_TO_BUILD}"
```
sxa, line 211: I'd be slightly tempted to do the quotes over the whole thing instead, but happy either way so don't feel compelled to accept it :-)
```suggestion
echo "$PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh --clean-gitrepo --jdk-boot-dir ${JDK_BOOT_DIR} --configure-args ${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --target-file-name ${FILENAME} ${TAG_OPTION} ${OPTIONS} ${BUILD_ARGS} ${VARIANT_ARG} ${JAVA_TO_BUILD}"
```

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -156,10 +160,11 @@ if which ccache 2> /dev/null; then
 fi
 
 # If we are in a cross compilation environment for RISC-V
-if [ "${ARCHITECTURE}" == "riscv64" -a "`uname -m`" == "x86_64" ]; then
+# shellcheck disable=SC2166
+if [ "${ARCHITECTURE}" == "riscv64" -a "$(uname -m)" == "x86_64" ]; then
```
sxa, line 164:
```suggestion
if [ "${ARCHITECTURE}" == "riscv64" ] && [ "$(uname -m)" == "x86_64" ]; then
```

build-farm/platform-specific-configurations/aix.sh
```diff
@@ -47,17 +48,19 @@ fi
 echo LDR_CNTRL=$LDR_CNTRL
 
 BOOT_JDK_VERSION="$((JAVA_FEATURE_VERSION-1))"
+# shellcheck disable=SC2116
 BOOT_JDK_VARIABLE="JDK$(echo $BOOT_JDK_VERSION)_BOOT_DIR"
```
sxa, line 52: We have a few other versions of this where it has been replaced with e.g.
```suggestion
BOOT_JDK_VARIABLE="JDK${BOOT_JDK_VERSION}_BOOT_DIR"
```

Is thre a reason not to do the same here?

sbin/common/common.sh
```diff
@@ -147,16 +148,16 @@ createOpenJDKArchive()
...
-          time tar -cf - "${repoDir}"/ | GZIP=-9 $COMPRESS -c > $fileName.tar.gz
+          time tar -cf - "${repoDir}"/ | GZIP=-9 $COMPRESS -c > "$fileName.tar.gz"
       fi
   fi
 }
 
 function setBootJdk() {
   # Stops setting the bootJDK on the host machine when running docker-build
-  if [ "${BUILD_CONFIG[DOCKER]}" != "docker" ] || ([ "${BUILD_CONFIG[DOCKER]}" == "docker" ] && [ "${BUILD_CONFIG[DOCKER_FILE_PATH]}" != "" ]) ; then
+  if [ "${BUILD_CONFIG[DOCKER]}" != "docker" ] || { [ "${BUILD_CONFIG[DOCKER]}" == "docker" ] && [ "${BUILD_CONFIG[DOCKER_FILE_PATH]}" != "" ]; } ; then
```
sxa, line 160: I'm worried about this change? If I run
```
if false || { false && true }; then echo true; else echo false; fi
```
but
```
if false || ( false && true ); then echo true; else echo false; fi
```
works

sbin/prepareWorkspace.sh
```diff
@@ -49,17 +50,18 @@ checkoutAndCloneOpenJDKGitRepo() {
...
     # Ensure cached origin fetch remote repo is correct version and repo (eg.jdk11u, or jdk), remember "jdk" sub-string of jdk11u hence grep with "\s"
     # eg. origin https://github.com/adoptopenjdk/openjdk-jdk11u (fetch)
     # eg. origin https://github.com/adoptopenjdk/openjdk-jdk (fetch)
     # eg. origin git@github.com:adoptopenjdk/openjdk-jdk.git (fetch)
     # eg. origin https://github.com/alibaba/dragonwell8.git (fetch)
     # eg. origin https://github.com/feilongjiang/bishengjdk-11-mirror.git (fetch)
+    # shellcheck disable=SC2166
     if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_DRAGONWELL}" -o "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_BISHENG}" ]; then
-      git --git-dir "${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/.git" remote -v | grep "origin.*fetch" | egrep "${BUILD_CONFIG[REPOSITORY]}.git|${BUILD_CONFIG[REPOSITORY]}\s"
+      git --git-dir "${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/.git" remote -v | grep "origin.*fetch" | grep -E "${BUILD_CONFIG[REPOSITORY]}.git|${BUILD_CONFIG[REPOSITORY]}\s"
```
sxa, line 62: Surprised it objected to this since egrep is pretty standard UNIX, but ok...

M-Davies:
Checks look good but seeing as @sxa said he was reviewing this PR in Slack I will hold off merging until that review is in

sbin/prepareWorkspace.sh
```diff
@@ -252,7 +255,7 @@ setGitCloneArguments() {
   cd "${BUILD_CONFIG[WORKSPACE_DIR]}"
   local git_remote_repo_address="${BUILD_CONFIG[REPOSITORY]}.git"
 
-  GIT_CLONE_ARGUMENTS=(${BUILD_CONFIG[SHALLOW_CLONE_OPTION]} "$git_remote_repo_address" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}")
+  GIT_CLONE_ARGUMENTS=("${BUILD_CONFIG[SHALLOW_CLONE_OPTION]}" "$git_remote_repo_address" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}")
```
sxa, line 258: I'll be honest I'm a bit worried about what the `()` are doing there, but that's not specific to this
```
bash-5.0$ GCA=(abc def ghi) 
bash-5.0$ echo $GCA
abc
```

sbin/prepareWorkspace.sh
```diff
@@ -326,9 +329,9 @@ checkingAndDownloadingAlsa() {
 
 sha256File() {
   if [ -x "$(command -v shasum)" ]; then
-    (shasum -a 256 | cut -f1 -d' ') <$1
+    (shasum -a 256 | cut -f1 -d' ') <"$1"
```
sxa, line 332: Again not an issue for this PR, but I'm curious as to how we ended up with this as opposed to either:
`shasum -a 256 "$1" | cut -f1 -d' '`
or if that cann't work somewhere:
`cat "$1" | shasum -a 256 | cut -f1 -d' '`

sxa (changes requested):
A few comments - some real changes I think we should make, some general comments (occasionally on the underlying scripts, not this PR - feel free to skip those!)

I do worry that we've dumped so many shellcheck exceptions in here. If we're choosing to use the tools we should really be trying to make it pass, or change our check policies.

build-farm/make-adopt-build-farm.sh
M-Davies: Fair. I didn't implement this one as I was unsure on when `JAVA_FEATURE_VERSION` is a string and when it is a int

build-farm/make-adopt-build-farm.sh
M-Davies: That works too. Much neater

build-farm/platform-specific-configurations/aix.sh
M-Davies: I think that one slipped through the net. I'll go through and clean up these manually

gdams (approved):

sbin/common/common.sh
M-Davies: For reference, [SC2235](https://github.com/koalaman/shellcheck/wiki/SC2235) is the directive. It's mostly to speed up the processing time. I am confused what the problem is, both statements and the alternative syntax
```bash
if false || { false && true; }; then echo true; else echo false; fi
```
amount to the same result for me?

sbin/prepareWorkspace.sh
M-Davies: Apparently it's because [egrep is deprecated](https://github.com/koalaman/shellcheck/wiki/SC2196)

sbin/prepareWorkspace.sh
M-Davies: > shasum -a 256 "$1" | cut -f1 -d' '

I'm not sure either but that seems like a more viable one. Much more readable anyway

build-farm/make-adopt-build-farm.sh
```diff
@@ -208,7 +207,7 @@ if [ "${OPERATING_SYSTEM}" != "aix" ] ; then
     export BUILD_ARGS="${BUILD_ARGS} --create-debug-image"
 fi
 
-echo "$PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh" --clean-git-repo --jdk-boot-dir "${JDK_BOOT_DIR}" --configure-args "${CONFIGURE_ARGS_FOR_ANY_PLATFORM}" --target-file-name "${FILENAME}" "${TAG_OPTION}" "${OPTIONS}" "${BUILD_ARGS}" "${VARIANT_ARG}" "${JAVA_TO_BUILD}"
+echo "$PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh --clean-gitrepo --jdk-boot-dir ${JDK_BOOT_DIR} --configure-args ${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --target-file-name ${FILENAME} ${TAG_OPTION} ${OPTIONS} ${BUILD_ARGS} ${VARIANT_ARG} ${JAVA_TO_BUILD}"
```
sxa, line 210: Should that have removed the hyphen from `--clean-gitrepo`? (Or was that an error in my suggestion?)

sbin/common/common.sh
sxa: Hmmm whichever shell I was using at the time definitely wouldn't work with that one

andrew-m-leonard: The first ";" is important, ie.:
```
if false || { false && true; } ; then
```

sbin/prepareWorkspace.sh
andrew-m-leonard: Can you check grep -E works on Solaris?

sbin/prepareWorkspace.sh
```diff
@@ -86,10 +87,11 @@ checkoutAndCloneOpenJDKGitRepo() {
   fi
 
   checkoutRequiredCodeToBuild
+  # shellcheck disable=SC2086
   if [ $checkoutRc -ne 0 ]; then
     echo "RETRYWARNING: Checkout required source failed, cleaning workspace and retrying..."
     cd "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}"
-    rm -rf "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+    rm -rf "${BUILD_CONFIG[WORKSPACE_DIR]:?}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
```
andrew-m-leonard, line 94: Why is the ? needed?

sbin/prepareWorkspace.sh
sxa: It'll be ok

sbin/common/common.sh
M-Davies: I think
```bash
if false || { false && true; } ; then
```
would produce a syntax error because of the whitespace after `}`

build-farm/make-adopt-build-farm.sh
M-Davies: No it shouldn't have. Good catch

M-Davies: No it shouldn't have. Good catch

sbin/prepareWorkspace.sh
M-Davies: If `WORKSPACE_DIR` is empty or not set, we may end up starting from the root dir. If `OPENJDK_SOURCE_DIR` is also empty or not set, everything in the root dir could be deleted. `:?` adds in an additional check which will force the script to fail if `WORKSPACE_DIR` is empty

https://github.com/koalaman/shellcheck/wiki/SC2115

The alternative would be doing `:-` with some default value

sxa (approved):
I think[*] I'm happy with this now

[*] - With these many changes in one PR it's impossible to be sure but hopefully I've picked up the more obvious ones ;-)

M-Davies:
Will wait for checks to complete then merge

PR #2550, opened 2021-03-26
OpenJ9 linux always configure --with-openssl=fetched

AdamBrousseau:
OpenJ9 get_source.sh will fetch the latest OpenSSL version
during the build. We should compile against that version
rather than 1.0.2r on the system.

Related AdoptOpenJDK/openjdk-infrastructure#2084

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -50,20 +50,9 @@ fi
 
 if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]
 then
-  # CentOS 6 has openssl 1.0.1 so we use a self-installed 1.0.2 from the playbooks
-  if grep 'release 6' /etc/redhat-release >/dev/null 2>&1 || grep 'jessie' /etc/os-release >/dev/null 2>&1 || grep 'SUSE' /etc/os-release >/dev/null 2>&1; then
-    if [ -r /usr/local/openssl-1.0.2/include/openssl/opensslconf.h ]; then
-      export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-openssl=/usr/local/openssl-1.0.2"
-    else
-      export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-openssl=fetched"
-    fi
-  else
-    export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-openssl=system"
-  fi
-fi
+  # OpenJ9 fetches the latest OpenSSL in their get_source.sh
+  export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-openssl=fetched"
 
-if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]
-then
   if [ "${ARCHITECTURE}" == "ppc64le" ] || [ "${ARCHITECTURE}" == "x64" ]
   then
     CUDA_VERSION=9.0


AdamBrousseau:
Looking for reviews from @pshipton and @sxa

sxa (commented):
My concerns would be twofold:
- If we do this will it still run against systems with 1.0.2 as a default e.g. RHEL7? Is it worth the risk of changing it?
- Is there any reason to download and use openssl when all of our machines are configured with the playbooks which will ensure a suitable version is available on platforms wher we are not bundling it? It introduces extra risk if the openssl source download fails.
- Doing the check then setting it to the `/usr/local` directory OR `fetched` runs the risk of behavior changing based on whether it is present. If, for whatever reason, the installation of openssl into `/usr/local` failed that failure will be hidden and it'll just use a differnet one, which is always risky for ensuring consistency

(There are valid counter-reasons but we need to at least consider the three above before making this change)

pshipton:
> If we do this will it still run against systems with 1.0.2 as a default e.g. RHEL7?

Yes. The version used for compile shouldn't affect the version used at runtime.

> Is it worth the risk of changing it?

Questionable. In theory it shouldn't matter. The OpenJ9 builds fetch and compile against the latest, and that is my preference as doing this will discover a problem if one is ever introduced, which should never happen.

> Is there any reason to download and use openssl when all of our machines are configured with the playbooks which will ensure a suitable version is available on platforms where we are not bundling it? It introduces extra risk if the openssl source download fails.

Note the older versions contain known security vulnerabilities if that matters for the maintenance of the machines.

> Doing the check then setting it to the /usr/local directory OR fetched runs the risk of behavior changing based on whether it is present. If, for whatever reason, the installation of openssl into /usr/local failed that failure will be hidden and it'll just use a differnet one, which is always risky for ensuring consistency

In theory the version used for compile doesn't matter.

pshipton:
I should note that we need to ensure the correct and latest version (1.1.1k atm) is bundled for Windows and Mac.

keithc-ca:
As Peter says, it should not matter which version of openssl we use at compile time.

Only [one source file](https://github.com/eclipse/openj9/blob/master/runtime/compiler/net/LoadSSLLibs.hpp#L26-L27) (currently) _needs_ to include any openssl header files, the others are redundant. With a little work, I expect we can remove those includes as well in which case *no* version of openssl would be required at build time.

AdamBrousseau:
I think Win and Mac do the right thing
https://github.com/AdoptOpenJDK/openjdk-build/blob/041c33c6a56ff3474133714c6d0ef6d7d9bce233/build-farm/platform-specific-configurations/mac.sh#L51
```
Running ./configure with arguments 'bash ./configure --verbose  --with-vendor-name=AdoptOpenJDK --with-vendor-url=https://adoptopenjdk.net/ --with-vendor-bug-url=https://github.com/AdoptOpenJDK/openjdk-support/issues --with-vendor-vm-bug-url=https://github.com/eclipse/openj9/issues --with-vendor-version-string=AdoptOpenJDK --with-version-opt=202103241343 --without-version-pre --with-version-build=7 --with-boot-jdk=/Library/Java/JavaVirtualMachines/adoptopenjdk-10.jdk/Contents/Home --with-macosx-codesign-identity="Developer ID Application: London Jamocha Community CIC" --with-debug-level=release --with-native-debug-symbols=external --enable-ccache --with-freemarker-jar=/Users/jenkins/workspace/build-scripts/jobs/jdk11u/jdk11u-mac-x64-openj9/workspace/./build//freemarker-2.3.31/freemarker.jar --with-jvm-variants=server --with-cacerts-file=/Users/jenkins/workspace/build-scripts/jobs/jdk11u/jdk11u-mac-x64-openj9/sbin/../security/cacerts --disable-ccache  --disable-warnings-as-errors --with-openssl=fetched --enable-openssl-bundling --with-sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/ --enable-dtrace=auto --with-cmake --with-freetype=bundled'
```
https://github.com/AdoptOpenJDK/openjdk-build/blob/041c33c6a56ff3474133714c6d0ef6d7d9bce233/build-farm/platform-specific-configurations/windows.sh#L148
```
Running ./configure with arguments 'bash ./configure --verbose  --with-vendor-name=AdoptOpenJDK --with-vendor-url=https://adoptopenjdk.net/ --with-vendor-bug-url=https://github.com/AdoptOpenJDK/openjdk-support/issues --with-vendor-vm-bug-url=https://github.com/eclipse/openj9/issues --with-vendor-version-string=AdoptOpenJDK --with-version-opt=202103251815 --without-version-pre --with-version-build=7 --with-boot-jdk=/cygdrive/c/openjdk/jdk-10 --with-debug-level=release --with-native-debug-symbols=external --with-jvm-variants=server --with-cacerts-file=/cygdrive/e/jenkins/tmp/sbin/../security/cacerts  --disable-warnings-as-errors --with-freemarker-jar=/cygdrive/c/openjdk/freemarker.jar --with-openssl=/cygdrive/c/openjdk/OpenSSL-1.1.1j-x86_64-VS2017 --enable-openssl-bundling --enable-cuda --with-cuda=C:/PROGRA~2/NVIDIA~2/CUDA/v9.0 --with-toolchain-version=2017 '
```

sxa:
> Note the older versions contain known security vulnerabilities if that matters for the maintenance of the machines.

We don't change the default on the machine, so any system apps will always be using the one supplied with the OS, not the one we store in `/usr/local/openssl-xxx` purely for the purposes of building OpenJ9 against.

sxa:
> I should note that we need to ensure the correct and latest version (1.1.1k atm) is bundled for Windows and Mac.

Yep I'm finalising that under https://github.com/AdoptOpenJDK/openjdk-build/pull/2553 which is separate from this PR

pshipton:
I didn't consider the JITServer use of openssl in my answers. I'll ask @mpirvu to comment on that.

AdamBrousseau:
Note there is another playbook for openssl 1.1.1b for jitserver.
https://github.com/AdoptOpenJDK/openjdk-infrastructure/blob/5ffd313d178b3ca4bbbf67d98d1f34f3c7c2ab03/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/OpenSSL/tasks/main.yml

mpirvu:
Same answer for JITServer, it should not matter which version of OpenSSL is present at compile time.
With respect to different variants of 1.1.x, as long as the API for various routines remains the same, as it should, we should be fine.

AdamBrousseau:
Can we have the jitserver use the version that we pull in get_source.sh? That way we could also remove the second playbook.

mpirvu:
> Can we have the jitserver use the version that we pull in get_source.sh? That way we could also remove the second playbook.

yes, that should work

keithc-ca:
> As Peter says, it should not matter which version of openssl we use at compile time.
> 
> Only [one source file](https://github.com/eclipse/openj9/blob/master/runtime/compiler/net/LoadSSLLibs.hpp#L26-L27) (currently) _needs_ to include any openssl header files, the others are redundant. With a little work, I expect we can remove those includes as well in which case _no_ version of openssl would be required at build time.

When I wrote the above, I forgot about `jdk/crypto/jniprovider/NativeCrypto.c` in the extension repos, but, again, compiles against any version should be fine.

karianna:
Any concerns about reproducing older builds with this change?

pshipton:
> Any concerns about reproducing older builds with this change?

From my perspective, it shouldn't make any difference.

karianna (approved):
LGTM

karianna:
Failure on GH Action

```
Compiling 224 properties into resource bundles for jdk.localedata
engines/e_afalg.c:31:4: warning: #warning "AFALG ENGINE requires Kernel Headers >= 4.1.0" [-Wcpp]
 #  warning "AFALG ENGINE requires Kernel Headers >= 4.1.0"
    ^~~~~~~
engines/e_afalg.c:32:4: warning: #warning "Skipping Compilation of AFALG engine" [-Wcpp]
 #  warning "Skipping Compilation of AFALG engine"
    ^~~~~~~
Compiling 3180 files for java.base
test/rsa_test.c: In function 'test_rsa_oaep':
test/rsa_test.c:349:5: warning: 'key' may be used uninitialized in this function [-Wmaybe-uninitialized]
     RSA_free(key);
     ^~~~~~~~~~~~~
ar: creating apps/libapps.a
ar: creating libssl.a
ar: creating test/libtestutil.a
/bin/sh: CCACHE_COMPRESS=1: command not found
Makefile:6863: recipe for target 'fuzz/asn1parse-test' failed
gmake[5]: *** [fuzz/asn1parse-test] Error 127
gmake[4]: *** [all] Error 2
Makefile:174: recipe for target 'all' failed
```

sxa:
I still don't see a compelling reason for changing this. The comments above seem to suggest there's no technical reason to use a later version on platforms where it is not bundled and it only has the potential to introduce instability by downloading things from third party websites during the build process.

Questions:
- What is the disadvantage of keeping it as-is?
- @keithc-ca suggests it might be fairly easy to remove this requirement anyway - is that something that there is a desire for?
- For @mpirvu: Does JITServer still require 1.1.1 to build against?
- The playbook role that @AdamBrousseau referred to earlier is not run on any of the AdoptOpenJDK buils systems (and given some of the things in there relating to adjusting default environments I am disinclined to do so) yet we have jitserver enabled on some platforms already e.g. [pLinux JDK11](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-ppc64le-openj9/894/consoleFull) has `-enable-jitserver` and ` --with-openssl=system`

mpirvu:
> Does JITServer still require 1.1.1 to build against?

JITServer does not need SSL library at build time. At runtime, given the options to force message encryption, JITServer looks for a library named:
```

**Comment by keithc-ca** [comment]
> > Does JITServer still require 1.1.1 to build against?
> 
> JITServer does not need SSL library at build time.

Correction: it *does* need the header files. The question is whether the 1.0.2 header files will suffice.

**Comment by AdamBrousseau** [comment]
>has the potential to introduce instability by downloading things from third party websites during the build process

Note that we have been cloning the latest openssl code every build as long as this req has been around.

**Comment by mpirvu** [comment]
> The question is whether the 1.0.2 header files will suffice.

Yes, 1.0.2 headers will suffice. When we implemented SSL support for JITServer we tested the following scenarios:
```
building against 1.1.1, loaded with 1.0.2g
building against 1.0.2g, loaded with 1.1.1
building against 1.1.1, loaded with 1.1.1
```

**Comment by pshipton** [comment]
> What is the disadvantage of keeping it as-is?

Is there is ever an API compatibility problem with a new version of openssl it won't be noticed in the Adopt builds. However it will be in the OpenJ9 builds since they build against the latest.

> has the potential to introduce instability by downloading things from third party websites during the build process

Atm it comes from github, as does the source for the other repos. If it doesn't work, probably nothing works so you can't build anyway.

**Comment by karianna** [comment]
GH Action needs updating.

```
Compiling 224 properties into resource bundles for jdk.localedata
engines/e_afalg.c:31:4: warning: #warning "AFALG ENGINE requires Kernel Headers >= 4.1.0" [-Wcpp]
 #  warning "AFALG ENGINE requires Kernel Headers >= 4.1.0"
    ^~~~~~~
engines/e_afalg.c:32:4: warning: #warning "Skipping Compilation of AFALG engine" [-Wcpp]
 #  warning "Skipping Compilation of AFALG engine"
    ^~~~~~~
Compiling 3180 files for java.base
test/rsa_test.c: In function 'test_rsa_oaep':
test/rsa_test.c:349:5: warning: 'key' may be used uninitialized in this function [-Wmaybe-uninitialized]
     RSA_free(key);
     ^~~~~~~~~~~~~
ar: creating apps/libapps.a
ar: creating libssl.a
ar: creating test/libtestutil.a
/bin/sh: CCACHE_COMPRESS=1: command not found
Makefile:6863: recipe for target 'fuzz/asn1parse-test' failed
```

sxa:
> GH Action needs updating.

@karianna Can you advise - is that something this PR author needs to be responsible for?

pshipton:
>  If we were to build against the latest headers, would we be confident that it still work out of the box on distributions like RHEL7 which have OpenSSL 1.0.2?

Yes, it's designed to work this way.

> Surely there is a risk that if there is a breaking change in a later 1.1.1 then using the latest one would mask the fact it won't execute on supported platforms if OpenJ9 were to use of new functionality without additional work since we're relying on the system SSL libraries on most platforms and that would be a customer visible break.

I don't believe so. Perhaps if there was new work done and it wasn't properly implemented or tested then Adopt building against 1.0.2 may catch that fact, but perhaps not. However, while building against 1.0.2 is supported today for no particular reason, it's not a benefit or something we need to keep working. If there is related new work done and any reason to do so, I can see support for building against 1.0.2 being dropped. It's more important that 1.0.2 is supported at runtime, which is supported by runtime testing, not by what happens at build time.

> Unless you plan to prereq 1.1.1 then it seems a risky strategy not to build against the earlier supported version.

1.1.1 is actually a prereq for some features, and 1.1.1 can be used at runtime even if 1.0.2 is used to compile (and vice versa).


In summary, basically it doesn't matter what you compile against. You can take this change, or not. The benefits are very nebulous.

gdams (approved):

gdams:
run tests

keithc-ca:
Is this not redundant now that #2917 is merged?

AdamBrousseau:
I wanted to take another look at this because we (Semeeu) still use "system" on p/z linux.

tellison:
No activity for a while. Can this be closed?

AdamBrousseau:
I've re-read all the comments. I don't think there is anything of value I can add to the conversation. I don't think the issue is settled one way or another.

pshipton:
I'm approving the change. I think it will be beneficial in the long term. 1.0.2 is already out of support, and 1.1.1 goes out of support next year. OpenJ9 builds will be switching to build against 3.0.x sometime next year, and sometime in the future may depend on this version. Having the builds fetch and build against the latest will avoid having to make this change in the future.

pshipton (approved):

karianna:
@AdamBrousseau Looks like a bunch of failures (I forced a rerun)

AdamBrousseau:
@karianna any way to relaunch just the 1 failure, or do we ignore?

sxa:
> @karianna any way to relaunch just the 1 failure, or do we ignore?

Re-running now. I don't expect any real problem with it given that it's an Alpine build.

PR #2551, opened 2021-03-26
Disable shellcheck and quotes on JRE mirror

M-Davies:
* Causes issues on mac with ls


andrew-m-leonard (approved):

karianna (approved):
LGTM

sxa (approved):
I believe this is the correct fix  With `${BUILD_CONFIG[JRE_PATH]}` in quotes, if that value has a wildcard in it (and from the log it looks like it's set to `jre-*-jre` then the quotes will inhibit the wildcard expansion

PR #2555, opened 2021-03-31
Remove quotes from BUILD_CONFIG[SHALLOW_CLONE_OPTION]

AdamBrousseau:
When this value is empty ("") it will cause git commands
to fail because "" will be picked up as an argument

Introduced by 962e6febfeb


karianna (approved):
LGTM

M-Davies (approved):

PR #2561, opened 2021-04-09
macos: use full xcode for jdk17+

gdams:
Fixes the following configure error:

```bash
17:52:59  checking for metal... [not found]
17:52:59  checking if metal can be run using xcrun... no
17:52:59  configure: error: XCode tool 'metal' neither found in path nor with xcrun
17:52:59  configure exiting with result code 1
```


smlambert (approved):
LGTM

Willsparker (approved):
LGTM

PR #2573, opened 2021-04-16
macOS: prevent m1 builds from cross compiling x64

gdams:
Fixes the bug reported in https://github.com/AdoptOpenJDK/openjdk-build/issues/1922#issuecomment-820747121

Essentially the arm64 macs were still cross-compiling for x64 rather than building on arm64

Confirmed to be fixed:

```bash
file java
java: Mach-O 64-bit executable arm64
```

### DIFF
--- a/build-farm/platform-specific-configurations/mac.sh
+++ b/build-farm/platform-specific-configurations/mac.sh
@@ -58,6 +58,8 @@ else
     if [ "${ARCHITECTURE}" == "x64" ]; then
       # We can only target 10.9 on intel macs
       export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-extra-cxxflags=-mmacosx-version-min=10.9"
+    elif [ "${ARCHITECTURE}" == "arm64" ]; then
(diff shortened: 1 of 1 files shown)


M-Davies (approved):

gdams:
build failure is unrelated

gdams:
@AdoptOpenJDK/tsc Approval to merge during the lockdown cycle please

sxa:
Approved during freeze on the assumption that this will be verified through a build pipeline today after it is merged and the output checked.

build-farm/platform-specific-configurations/mac.sh
```diff
@@ -58,6 +58,8 @@ else
     if [ "${ARCHITECTURE}" == "x64" ]; then
       # We can only target 10.9 on intel macs
       export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-extra-cxxflags=-mmacosx-version-min=10.9"
+    elif [ "${ARCHITECTURE}" == "arm64" ]; then
```
tellison, line 61: `arm64` and not `aarch64` correct? (just based on the appended command-line arg this modifies)

gdams: arm64 is correct. It's what Apple refers to it as (along with all the other JVM vendors)

sxa (approved):

sxa:
Somewhat worried by the inconsistence of using `arm64` or macos for `ARCHITECTURE` VS `aarch64` on Linux, but that's not related to this PR

Issue #2574, opened 2021-04-16
Introduce labeller, project manager and improved comment bot from ci


M-Davies:
Propose porting over the changes implemented in https://github.com/adoptium/ci-jenkins-pipelines/issues/59 & https://github.com/adoptium/ci-jenkins-pipelines/issues/123 as well as the [labeller.yml](https://github.com/adoptium/ci-jenkins-pipelines/tree/master/.github) job to ease manual project management efforts

PR #2575, opened 2021-04-16
Add comment bot, labeller and project manager

M-Davies:
Closes: #2574 

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,37 @@
+# Config file used by .github/workflows/labeler.yml
+
+# This file matches file/folder changes to labels. See https://github.com/actions/labeler for more information.
+
+# e.g to add the test label to any changes in the /tests directory:
+
+# test:
+#   - tests/*
+
+ghActions:
+  - .github/*
+documentation:
+  - '**/*.md'
+jenkins:
+  - build-farm/*
+aix:
+  - build-farm/platform-specific-configurations/aix.sh
+alpine-linux:
+  - build-farm/platform-specific-configurations/alpine-linux.sh
+macos:
+  - build-farm/platform-specific-configurations/mac.sh
+solaris:
+  - build-farm/platform-specific-configurations/solaris.sh
+windows:
+  - build-farm/platform-specific-configurations/windows.sh
+docker:
+  - docker/*
+  - docker-build.sh
+testing:
+  - tests/*
+security:
+  - security/*
+  - SECURITY.md
+  - sbin/sig_check/*
+installer:
+  - build-farm/sign-releases.sh
+  - sign.sh
--- a/.github/regex_labeler.yml
+++ b/.github/regex_labeler.yml
@@ -35,63 +35,78 @@ labels:
   - label: "macos"
     matcher:
       title: "mac"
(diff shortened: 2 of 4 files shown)


karianna (approved):
LGTM, we can always fine tune it

sxa (commented):
I'll be compeltely honest I'm not a fan of the comment bot. Is there a reason we can't cover the same information in a PR template? Otherwise it feels like just spamming my notifications with a comment that doesn't tell me anything as a regular contributor. For non-regular contributors a template with the same information would seem to be adequate? Perhaps this should be discussed this at the next [community call](https://github.com/AdoptOpenJDK/TSC/issues/228) as a minimum before proceeding?

karianna:
For me, we can never over-communicate, so I'd be happy to have that information presented twice in rapid order to encourage folks to 'do the right thing'

M-Davies:
> For me, we can never over-communicate, so I'd be happy to have that information presented twice in rapid order to encourage folks to 'do the right thing'

I agree. Templates are useful but I find that people only really read or use the parts that allow them to populate the template, so having it reinforced in a comment is more assertive. Furthermore, the only notification you would get is at the same time the pull request is created, so you would be getting a notification anyway for the pull request that has been created (the comment gets merged into that same notification) I wouldn't be against adding similar info to the templates too but that's a conversation for another issue I think.

sxa:
> For me, we can never over-communicate, so I'd be happy to have that information presented twice in rapid order to encourage folks to 'do the right thing'

I'm not prepared to block this (which is why I reviewed only as a comment) but I personally consider that if someone has made the effort to "do the right thing" and read our contributing document, follow all the procedures and then gets a large block of MAKE SURE YOU'VE DONE THIS text stuck in front of them that it could be seen as quite insulting, but I do understand that others might not see it that way ...

The comment says "if you have not made a pull request here before" - could we dad in logic to only add it for new contributors to the repo?

andrew-m-leonard:
So I have similar views, templates are great for people unsure what to do to follow, but can be patronizing to more experienced people who probably know exactly what's needed for what they are raising. I'm guilty when raising some Issues, of the first thing I do to the template is block mark, delete! It just clutters my view of what I want to raise/discuss/create...

gdams:
so here is a proposal, why don't we show the message to first-time contributors only. We should be able to use the github API to work that out. Other changes in this PR (labelled and project manager) are useful to all users.

sxa:
> so here is a proposal, why don't we show the message to first-time contributors only. We should be able to use the github API to work that out. Other changes in this PR (labelled and project manager) are useful to all users.

That's the proposal I made above - do it for new contributors's first PR to the repo, so I'm 👍🏻 if it can be implemented.

sxa:
Bear in mind this may be subject the [new github rules on actions and new contributors](https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/)

M-Davies:
> That's the proposal I made above - do it for new contributors's first PR to the repo, so I'm 👍🏻 if it can be implemented.

This has been done. Please re-review @gdams @sxa

gdams (approved):

PR #2580, opened 2021-04-16
Shenandoah present on win32

adamfarley:
As per the Shenandoah project webpage, if shenandoah is set as a
with-jvm-features option during pre-build configuration (which Adopt
does), then we should expect Shenandoah to be both present and
functional in the build.

My assumption is that Windows 32bit was simply skipped during smoke
test pre-deployment testing, and this problem was simply not seen
as a result.


adamfarley:
@AdoptOpenJDK/tsc - Approval to merge during the lockdown cycle please

sophia-guo:
LGTM

smlambert (approved):
thanks @adamfarley !

karianna (approved):
LGTM

sxa:
I can't check the linter failure from here at the moment but that should be addressed first.

smlambert:
Linter failure described [here](https://github.com/AdoptOpenJDK/openjdk-build/issues/2566#issuecomment-821493395), and we should revisit whether some rules are overly strict (have created https://github.com/AdoptOpenJDK/openjdk-build/issues/2581 for linter scrutiny going forward).

PR #2587, opened 2021-04-19
Don't bundle freetype with alpine linux

adamfarley:
This makes us consistent with other linux builds, which also
do not bundle freetype to avoid various problems.

### DIFF
--- a/build-farm/platform-specific-configurations/alpine-linux.sh
+++ b/build-farm/platform-specific-configurations/alpine-linux.sh
@@ -22,6 +22,9 @@ source "$SCRIPT_DIR/../../sbin/common/constants.sh"
 # ccache seems flaky on alpine
 export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --disable-ccache"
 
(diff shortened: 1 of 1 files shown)


adamfarley:
Breakdown of the reasons for this PR can be found [here](https://github.com/AdoptOpenJDK/openjdk-build/issues/2579#issuecomment-822451663). Please add any counter-arguments there.

adamfarley:
Alpine linux build (in the PR checks) correctly uses the system freetype and does not bundle.

adamfarley:
@AdoptOpenJDK/tsc - Approval to merge during the lockdown cycle please

Note: This change brings alpine linux in-line with non-alpine xLinux. Also, based on [this comment](https://github.com/AdoptOpenJDK/openjdk-docker/issues/529#issuecomment-811403244), I'm assuming the freetype bundling was not deliberate, and that I'm not removing anything we want to keep. Do let me know if I'm wrong.

smlambert (approved):

sxa:
I think I'm in favour of this for consistency, and also I suspect one of the main uses of Alpine is in docker containers where we have an issue already regarding adding freetype to those I believe.

karianna (approved):
LGTM

PR #2605, opened 2021-04-28
Remove invalid debuginfo symlink

sxa:
Fixes https://github.com/adoptium/adoptium-support/issues/298 (Rogue symlink to something that gets deleted)

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -769,7 +769,7 @@ deleteDebugSymbols() {
     ;;
   *)
     # on other platforms, we want to remove .debuginfo files
-    find "${jdkTargetPath}" "${jreTargetPath}" -type f -name "*.debuginfo" | xargs rm -f || true
+    find "${jdkTargetPath}" "${jreTargetPath}" -name "*.debuginfo" | xargs rm -f || true
     ;;
   esac
 }


karianna (approved):
LGTM

andrew-m-leonard (approved):
good

M-Davies (approved):
LGTM

sxa:
(Note to committers: While this change should be safe, I haven't checked that this definitely works ... Will aim to do that tomorrow so I recommend holding off merging for now then we can verify before the change his the nightlies)

gdams (approved):

sxa:
This looks to be good - merging

breun:
I think that instead of `| xargs rm -f` you could also use the `-delete` flag of the `find` command?

```
     -delete
             Delete found files and/or directories.  Always returns true.  This executes from the current working directory as find recurses down the tree.  It will not attempt to
             delete a filename with a ``/'' character in its pathname relative to ``.'' for security reasons.  Depth-first traversal processing is implied by this option.  Following
             symlinks is incompatible with this option.
```

sxa:
> I think that instead of `| xargs rm -f` you could also use the `-delete` flag of the `find` command?

It's not a standard option across all UNIX-based platforms.

Issue #2608, opened 2021-04-30
Precision enhancement to vendor version smoke test


adamfarley:
***Summary***
We automatically skip the check for the vendor name in the -version output from non-Hotspot JDK8 builds in [VendorPropertiesTest.java](https://github.com/adoptium/temurin-build/blob/95b8cddd15457a115e887bace5a8faa47a89de2d/test/functional/buildAndPackage/src/net/adoptopenjdk/test/VendorPropertiesTest.java#L71).

@keith-ca has requested a modification here to make the "skip" logic more precise.

***Original Comment:***

If it's (the code that provides vendor name in JDK8 -version) only missing from OpenJ9 and JDK8, then I suggest a clearer (and more accurate) condition would be
```
    jdkVersion.usesVM(VM.OPEN9) && !jdkVersion.isNewerOrEqual(9)
```

_Originally posted by @keithc-ca in https://github.com/adoptium/temurin-build/pull/2602#discussion_r623251847_

keithc-ca:
This seems appropriate given https://github.com/adoptium/temurin-build/pull/2602#discussion_r623245682:

> The vendor name is only absent from builds which are both OpenJ9 and JDK8.

adamfarley:
Seems reasonable. PR raised.

https://github.com/adoptium/temurin-build/pull/2610

PR #2610, opened 2021-04-30
Clarify test exception for JDK8 on OpenJ9

adamfarley:
This exception is specific to OpenJ9, so the exception should mention
it by name.

### DIFF
--- a/test/functional/buildAndPackage/src/net/adoptium/test/JdkVersion.java
+++ b/test/functional/buildAndPackage/src/net/adoptium/test/JdkVersion.java
@@ -158,6 +158,16 @@ public class JdkVersion {
 
     /**
      * Are we running on an OpenJDK feature version
+     * that is older than the supplied int?
+     * @param  featureParam  supplied int mentioned above
+     * @return boolean       result
+     */
+    public boolean isOlderThan(final int featureParam) {
+        return feature < featureParam;
+    }
+
+    /**
+     * Are we running on an OpenJDK feature version
      * that is greater than the supplied feature int.
      * OR
      * Are we running on the same OpenJDK feature
--- a/test/functional/buildAndPackage/src/net/adoptium/test/VendorPropertiesTest.java
+++ b/test/functional/buildAndPackage/src/net/adoptium/test/VendorPropertiesTest.java
@@ -67,11 +67,6 @@ public class VendorPropertiesTest {
      */
     @Test
     public void javaVersionPrintsVendor() {
-        // Skip test on JDK8 for non-Hotspot JDKs.
-        if (!jdkVersion.isNewerOrEqual(9) && !jdkVersion.usesVM(VM.HOTSPOT)) {
-            return;
-        }
-
         String testJdkHome = System.getenv("TEST_JDK_HOME");
         if (testJdkHome == null) {
             throw new AssertionError("TEST_JDK_HOME is not set");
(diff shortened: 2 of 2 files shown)


adamfarley:
Closes https://github.com/adoptium/temurin-build/issues/2608

adamfarley:
Note: Ignoring the three types of linter failure (line too long, magic number, & TODO) for the same reason I ignored it here:

M-Davies (approved):
Looks good. Imo, it's easier to read this as the below but I won't block it on that account
```java
        if (jdkVersion.usesVM(VM.OPENJ9) && jdkVersion.isOlderOrEqual(9)) {
```

test/functional/buildAndPackage/src/net/adoptopenjdk/test/VendorPropertiesTest.java
```diff
@@ -67,8 +67,8 @@ public VendorPropertiesTest() {
      */
     @Test
     public void javaVersionPrintsVendor() {
-        // Skip test on JDK8 for non-Hotspot JDKs.
-        if (!jdkVersion.isNewerOrEqual(9) && !jdkVersion.usesVM(VM.HOTSPOT)) {
+        // Skip test on JDK8 with OpenJ9 VM, as java -version output will not include a vendor name by design.
+        if (jdkVersion.usesVM(VM.OPENJ9) && !jdkVersion.isNewerOrEqual(9)) {
```
karianna, line 71: Prefer `if (jdkVersion.usesVM(VM.OPENJ9) && jdkVersion.isOlderThan(9)) {`

karianna (changes requested):

llxia:
If the test is using TKG, then I think we should use playlist feature to limit JDK implementation and version. For example: 
https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml#L467
https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml#L476

smlambert:
Agree that is preferable (to use playlists to divide / exclude).  In order to do that, the tests will have to be refactored.

https://github.com/adoptium/temurin-build/blob/master/test/functional/buildAndPackage/src/net/adoptopenjdk/test/VendorPropertiesTest.java#L69-L98 would be moved to its own class and into its own testng and playlist target (outside of [CommonFeatureTests](https://github.com/adoptium/temurin-build/blob/master/test/functional/buildAndPackage/testng.xml#L18-L23) and outside of https://github.com/adoptium/temurin-build/blob/master/test/functional/buildAndPackage/playlist.xml#L49 so it can be limited to 11+.

test/functional/buildAndPackage/src/net/adoptopenjdk/test/VendorPropertiesTest.java
adamfarley: Fixed, and I also tweaked the names of the other methods for consistency. Your thoughts?

karianna:
@adamfarley needs a rebase

adamfarley:
@karianna - Rebased. 

Also I removed the superfluous changes to isNewerOrEqual and isNewerOrEqualSameFeature because it made the method names unnecessarily long.

@sophia-guo and @smlambert - I agree that an exclude should happen at the playlist level.

On inspection, the rebase would require 
- the VmPropertiesChecks interface (along with implimentations) to be broken out, or made externally accessible.
- there to be two entries in the playlist for one test, in order to enable both OpenJ9 (11+) testing and Hotspot (8+) testing.

Preliminary work towards this end seemed to add complexity, so I had a rethink and moved the logic inside AdoptiumPropertiesChecks. That way, instead of skipping the test, we explicitly test for the different behaviour on OpenJ9, JDK8. We do this in other tests, explicitly testing for different behaviour, and it seemed to simplify the whole business.

What do you think?

karianna (approved):

smlambert (approved):
Approving this approach as a 'temporary' measure.  As discussed in previous comments, we should ideally restructure the smoke tests and leverage the playlist filtering for jdkVersion checks, rather than expect each type of test to manage its own checking (which then requires us to maintain code for version checking in multiple places).  

Ultimately, a follow-up refactoring PR will be preferential and help simplify and reduce tech debt.

llxia (approved):

PR #2628, opened 2021-05-18
skip codesigning testimage and debugimage

gdams:
Using the EF signing tool we lose over 45 minutes signing these two archives


smlambert (approved):

karianna (approved):
LGTM

PR #2630, opened 2021-05-19
Handle get jdk tag sort behaviour on Alpine for empty field behaviour

andrew-m-leonard:
On Alpine the "sort" behaviour for empty numeric sort fields is different, eg.jdk-16<.0.0>+36
Pad field fully before sorting

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -883,7 +883,7 @@ getFirstTagFromOpenJDKGitRepo() {
   local jdk11plus_tag_sort1="sort -t+ -k2,2n"
   # Second, (stable) sort on (V), (W), (X), (P): P(Patch) is optional and defaulted to "0"
   local jdk11plus_tag_sort2="sort -t. -k2,2n -k3,3n -k4,4n -k5,5n"
-  jdk11plus_get_tag_cmd="grep -v _openj9 | grep -v _adopt | sed 's/jdk-/jdk./g' | sed 's/+/.0.+/g' | $jdk11plus_tag_sort1 | nl | $jdk11plus_tag_sort2 | sed 's/\.0\.+/+/g' | cut -f2- | sed 's/jdk./jdk-/g' | tail -1"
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
run tests

sxa (commented):
I can see how this might resolve it, but do you have an example of how this fails? With two lines in a file:
```jdk-16.0.1+9
jdk-16+36
```
the following command appears to already show the correct result `jdk-16.0.1+9` - am I using the correct command (i.e. I can't seem to make it fail as it seems to in the bulid)? `cat tags | sed 's/+/.0+/g' | sort -t+ -k2,2n | nl | sort -t. -k2,2n -k3,3n -k4,4n -k5,5n | sed 's/\.0+/+/g' | tail -1`

sxa (approved):
It does appear to have produced the correct result in the test build so I'll tentatively approve :-)

andrew-m-leonard:
Test to replicate:
```
#!/bin/bash

echo -e "jdk-16.0.1+9\njdk-16+36\n" > /tmp/versions
jdk11plus_tag_sort1="sort -t+ -k2,2n"
jdk11plus_tag_sort2="sort -t. -k2,2n -k3,3n -k4,4n -k5,5n"
jdk11plus_get_tag_cmd="grep -v _openj9 | grep -v _adopt | sed 's/jdk-/jdk./g' | sed 's/+/.0.0+/g' | $jdk11plus_tag_sort1 | nl | $jdk11plus_tag_sort2 | sed 's/\.0\.0+/+/g' | cut -f2- | sed 's/jdk./jdk-/g' | tail -1"
eval "cat /tmp/versions | $jdk11plus_get_tag_cmd"
```

sxa:
Ah - I had tested by switchinging out the `.0` instead of just the `0` - that explains why I didn't replicate it. Thanks. Happy for the merge :-)

johnoliver (approved):

PR #2633, opened 2021-05-25
modify plist values as part of build instead of installer

gdams:
This is required as part of the EF codesign work

These commands originally came from https://github.com/adoptium/installer/pull/316


tellison (approved):
LGTM

Willsparker (approved):
LGTM

PR #2641, opened 2021-06-02
Remove showCompletionMessage

gdams:
It was removed as part of https://github.com/adoptium/temurin-build/pull/2417 but it was clearly missed for the exploded image


sxa (approved):
Surprised that hasn't been spotted before ...

johnoliver (approved):

PR #2649, opened 2021-06-03
actions: add macOS back to the pool to include exploded build testing

gdams:
This should allow us to better test the changes that have recently gone in to support EF codesigning.

We took macOS out because we were starving the pool and builds were taking too long. I've reduced this to only build JDK8 and JDK11 hotspot so it should cope much better


karianna (approved):
LGTM

johnoliver (approved):

PR #2660, opened 2021-06-11
Add smoke test to PR workflow

sophia-guo:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -254,7 +254,7 @@ jobs:
 
      - name: Set root of jdk image dir
        run: |
-         imageroot=$(find ${HOME}/${{matrix.version}}-${{matrix.os}}-${{matrix.vm}} -name release -type f)
+         $imageroot = find ${HOME}/${{matrix.version}}-${{matrix.os}}-${{matrix.vm}} -name release -type f
          echo "TEST_JDK_HOME=$(dirname ${imageroot})" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
      - name: Smoke test
        uses: adoptium/run-aqa@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,7 @@ jobs:
     container:
       image: ${{ matrix.image }}
     strategy:
+      fail-fast: false
       matrix:
         os: [linux]
         version: [jdk8u, jdk11u, jdk16u, jdk17, jdk]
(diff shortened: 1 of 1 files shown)


gdams:
I'd suggest making the smoke test part of the build rather than redefining the matrix again and creating a separate job?

.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>openjdk-build</name>
```
gdams, line 3: is this file needed?

.github/workflows/build.yml
```diff
@@ -188,3 +279,46 @@ jobs:
        with:
          name: ${{matrix.version}}-${{matrix.os}}-${{matrix.vm}}
          path: workspace/target/*
+  
+  windows_smokeTest:
+    name: macos_smokeTest
+    runs-on: macos-latest
```
gdams, line 285: assume this is meant to run on `windows-latest`?

.project
sophia-guo: no, I will remove it.

sophia-guo:
> I'd suggest making the smoke test part of the build rather than redefining the matrix again and creating a separate job?

Yes, we can do like that. The benefit of separating build and test job is to make it clear if test job fails while build job passes. However if we'd like smoke tests block the build jobs if there is a failure making tests part of the build would make more sense.

.github/workflows/build.yml
```diff
@@ -192,3 +285,47 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Restore build artifacts
+      uses: actions/download-artifact@v2
+      with:
+        name: ${{matrix.version}}-${{matrix.os}}-${{matrix.vm}}
+        path: ~/${{matrix.version}}-${{matrix.os}}-${{matrix.vm}}
+    - name: Unpack jdk
+      run: |
+        tar -xf "${HOME}/${{matrix.version}}-${{matrix.os}}-${{matrix.vm}}/OpenJDK.tar.gz" -C "${HOME}/${{matrix.version}}-${{matrix.os}}-${{matrix.vm}}"
```
gdams, line 313: this is going to need to be an unzip command

gdams:
> Yes, we can do like that. The benefit of separating build and test job is to make it clear if test job fails while build job passes. However if we'd like smoke tests block the build jobs if there is a failure making tests part of the build would make more sense.

Yeah i'd really prefer that it was part of the build, it will make it super clear when we make breaking changes to the build scripts if you're happy to merge the smoke jobs into the build ones?

.github/workflows/build.yml
gdams: You likely need a rebase to pick up https://github.com/adoptium/temurin-build/commit/631f4e3d41b2c33fec83fad88a9264c8130b73d0

sophia-guo:
re https://github.com/adoptium/temurin-build/pull/2660#issuecomment-863036615

Sure, will update to combine the test as a part of the build.

gdams:
@sophia-guo this should fix the centos issues (looking for `apt-get` https://github.com/adoptium/run-aqa/pull/91

gdams (approved):

gdams:
Linter failure can be ignored - there is no way to avoid the duplicate code

smlambert (approved):
LGTM

PR #2662, opened 2021-06-14
Allow building of arm32 target on aarch64 kernels with default 32-bit userspace (and use >4 cores)

sxa:
This will allow us to sped up the arm32 builds by running them in docker containers on the larger aarch64 hosts which we have. (Note that at present the main build will work but `make images` in the JDK currently requires a modified `uname` command which gives the correct result)

Related: https://github.com/adoptium/infrastructure/pull/2201

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -81,12 +81,15 @@ then
   if lscpu | grep aarch64; then
      echo Validating 32-bit environment on 64-bit host - perhaps it is a docker container ...
      if ! file /bin/ls | grep "32-bit.*ARM"; then
-       echo /bin/ls is not a 32-bit system. This configuration is invalid without extra work
+       echo /bin/ls is not a 32-bit binary but ARCHITECTURE=arm. Non-32-bit userland is invalid without extra work
        exit 1
      fi
      echo Looks reasonable - configuring to allow building that way ...
      export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --target=armv7l-unknown-linux-gnueabihf --host=armv7l-unknown-linux-gnueabihf"
   fi
+  # "4" is a temporary override which allows us to utilise all build
+  # cores on our scaleway systems while they are still in use but still
+  # allow more on larger machines. Can be revisited post-Scaleway
   if [ "$(lscpu|awk '/^CPU\(s\)/{print$2}')" = "4" ]; then
     export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-jobs=4 --with-memory-size=2000"
   fi
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -78,15 +78,16 @@ fi
 
 if [ "${ARCHITECTURE}" == "arm" ]
 then
-  if [ "$(uname -m)" == "aarch64" ]; then
+  if lscpu | grep aarch64; then
      echo Validating 32-bit environment on 64-bit host - perhaps it is a docker container ...
-     if ! file /bin/ls | grep 32-bit.*ARM; then
+     if ! file /bin/ls | grep "32-bit.*ARM"; then
        echo /bin/ls is not a 32-bit system. This configuration is invalid without extra work
        exit 1
      fi
      echo Looks reasonable - configuring to allow building that way ...
      export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --target=armv7l-unknown-linux-gnueabihf --host=armv7l-unknown-linux-gnueabihf"
-  else
+  fi
+  if [ "$(lscpu|awk '/^CPU\(s\)/{print$2}')" = "4" ]; then
     export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-jobs=4 --with-memory-size=2000"
   fi
   if [ "$JAVA_FEATURE_VERSION" -eq 8 ]; then
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -44,15 +44,6 @@ then
   fi
 fi
 
-if [ "$(uname -m)" == "aarch64" ] && [ "$ARCHITECTURE" = "arm" ]
-then
-   echo Validating 32-bit environment on 64-bit host - perhaps it is a docker container ...
-   file /bin/ls | grep 32-bit.*ARM
-   [ $? -ne 0 ] && echo /bin/ls is not a 32-bit system. This configuration is invalid without extra work && exit 1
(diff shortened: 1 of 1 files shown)


build-farm/platform-specific-configurations/linux.sh
```diff
@@ -78,7 +78,18 @@ fi
 
 if [ "${ARCHITECTURE}" == "arm" ]
 then
-  export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-jobs=4 --with-memory-size=2000"
+  if lscpu | grep aarch64; then
```
karianna, line 81: Should we `[` bracket this if statement to match style?

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -78,7 +78,18 @@ fi
...
+     echo Validating 32-bit environment on 64-bit host - perhaps it is a docker container ...
+     if ! file /bin/ls | grep "32-bit.*ARM"; then
```
karianna, line 83: Same comment as above

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -78,7 +78,18 @@ fi
 
 if [ "${ARCHITECTURE}" == "arm" ]
 then
-  export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-jobs=4 --with-memory-size=2000"
+  if lscpu | grep aarch64; then
+     echo Validating 32-bit environment on 64-bit host - perhaps it is a docker container ...
```
karianna, line 82: Can we detect if it is a Docker container to add clarity?

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -78,7 +78,18 @@ fi
...
+     if ! file /bin/ls | grep "32-bit.*ARM"; then
+       echo /bin/ls is not a 32-bit system. This configuration is invalid without extra work
```
karianna, line 84: Maybe: `is not configured as a 32-bit system`?

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -78,7 +78,18 @@ fi
...
+       exit 1
+     fi
+     echo Looks reasonable - configuring to allow building that way ...
+     export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --target=armv7l-unknown-linux-gnueabihf --host=armv7l-unknown-linux-gnueabihf"
+  fi
+  if [ "$(lscpu|awk '/^CPU\(s\)/{print$2}')" = "4" ]; then
```
karianna, line 90: 4 seems a very specific number, what if there's 2 or 8 or whatever?

karianna (changes requested):

build-farm/platform-specific-configurations/linux.sh
sxa: No - this is an external commend which we are testing the return code is (as `[` is when doing logical operations) and is not doing a `test` operaion, so that would not have the desired effect.

build-farm/platform-specific-configurations/linux.sh
sxa: Same reply as above ;-)

build-farm/platform-specific-configurations/linux.sh
sxa: I'm not sure there is significant value to detecting it explicitrly, and bear in mind it would also be true if you were running a 32-bit userspace on a 64-bit kernel (which is a valid configuratoin - I've used it on the raspberry pi OS before) - I could remove that comment, but the docker environment is where I intend to use it in our build environment.

build-farm/platform-specific-configurations/linux.sh
sxa: Hmmm. it's not a system configuration option really, although reading it back I don't like "32-bit system" either since it's explicitly checking `/bin/ls` only. Perhaps `/bin/ls is not 32-bit - 32-bit userland tools are needed to be able to build natively for ARCHITECTURE=arm`? Happy for other suggestions though

build-farm/platform-specific-configurations/linux.sh
sxa: Honeslty this was always a hack to make it use all the CPU cores on the restricted RAM scaleway systems that we have as we were ending up with the builds taking a lot longer because the build system was actively dropping the number of cores it would otherwise use. I'm propoing leaving that stuff as is to allow maximum use of the Scaleway systems when builds end up on there, and then probably revisit it later.

Bear in mind that the previous code ALWAYS locked the bulid system to 4 cores/2Gb on ARM so this change just allows it to use more if they are available. I'll be delighted to strip this out once we're running on better systems :-) If anyone wants to go back and verify the implications on the existing scaleway systems of removing this I'm happy for them to check it and potentially remove it ASAP but for now I think this is the right approach to ensure stability.

karianna: If you add a comment explaining this situation then I'm good :-)

gdams (approved):
LGTM once Martijn's comments are resolved

andrew-m-leonard (approved):
seems reasonable

build-farm/platform-specific-configurations/linux.sh
sxa: Done - shortened the above essay to indicate that it is a temporary override for the Scaleway machine capacities.

build-farm/platform-specific-configurations/linux.sh
sxa: Updated

sxa:
Note that with the introduction of my `uname` hack this method building of building does now work without this PR, but still has the core count restriction. So it means that despite using a 32-core system, https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-linux-arm-hotspot/673/console still has:
```
18:40:08  Build performance summary:
18:40:08  * Cores to use:   4
18:40:08  * Memory limit:   2000 MB
18:40:08  * ccache status:  disabled
```
So we still want this PR :-)

sxa:
NOTE: Unclear why at the moment but the docker containers on at least one of the aarch64 hosts (as is visible in the link above) is crashing when running `java -version`. This is NOT specific to this change, as the GA AdoptOpenJDK 8u292 exhibits the same behaviour in the same environment. The code which was built in the new environment does NOT crash when run natively on one of my other arm32 OSs.

karianna (approved):

karianna:
@sxa Good to merge when you like

sxa:
Thanks - awaiting getting a solution to https://github.com/adoptium/adoptium-support/issues/319 before merging as otherwise it's a little pointless (at least for JDK8!)

sxa:
Running one final test across three versions of the JDK before merging ..

sxa:
Going to hold off on merging this for now. For some reason (despite the default detection in the openjdk configure process ) something is locking it to a single core with this change.

sxa:
OK I think this is ok ... Something in our build scripts is forcing the value back to 1 unless it gets a `NUM_PROCESSORS` value from the jenkins executor, which is somethig we ought to resolve separately for when people are using this in other environments. The default value should be the autodetected one.

PR #2665, opened 2021-06-16
Use Ubuntu-supplied GCC on JDK8u arm32 builds

sxa:
Fixes https://github.com/adoptium/adoptium-support/issues/319 (As long as we're happy building with gcc 5.4.0 instead of 7.5.0 on this build only)


sxa:
Taking this back to draft in light of the problems that were seen previously with the default gcc - https://github.com/adoptium/adoptium-support/issues/33

sxa:
Superceded by new GCC installed as per https://github.com/adoptium/adoptium-support/issues/319

Issue #2594, opened 2021-04-21
Prototype "Reproducible Build" using current available Adopt jenkins job & infra framework and "build info"


andrew-m-leonard:
This is a work item for EPIC https://github.com/AdoptOpenJDK/openjdk-build/issues/2522.
The aim being to "prototype" an initial attempt at being able to "recreate" an Adopt build based on specifying the basic "build info" currently available for an existing "release", and assuming the current available "infra" setup.

### Design:
**Assumptions for 1st prototype:**
1. Hotspot only
2. no installer
3. no signing

**Spec:**
Extend the current openjdk-pipeline & build jobs to take as extra inputs basic "build info":
- Openjdk "mirror" source git "commit SHA" or "tag"
- openjdk-build repo SHA
- ci-jenkins-pipelines repo SHA
-

andrew-m-leonard:
The Jenkins build jobs are currently "generated" whenever openjdk-build or ci-jenkins-pipelines changes occur. The contents of these jobs also contain an extended amount of JSON configuration/settings, both the job parameters and the JSON settings are specific to the build setup at the time of the release/build.
It thus makes sense to architect the prototype as a two step process:
1. "Generate" required level of openjdk-pipeline & build jobs based on "build info"
2. Execute the generated jobs..

andrew-m-leonard:
So utilizing the Adopt Build job USER build scripts settings to point at a SHA of a level of the openjdk-build scripts you want to use (eg.January 19th CPU Update), is quite easy to do. Here is a build of January jdk-11.0.10+9 Hotspot Release: https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-x64-hotspot/960/parameters/
So this is a re-build using the january build scripts of the january source tag, using the current infrastructure and tooling.
Q: Is this "reproduced"? A: Maybe!
In reality all this has done has re-built that source SHA using the current infrastructure and tooling levels. So as such, it's probably using different versions of numerous dependencies, especially for non-container build platforms.
The build above was on xLinux, so the build container it was using was from the current adoptopenjdk/centos6_build_image image. Improvement could be achieved if we cached along side each release we mark as reproducible the docker build images used to build them, so that would improve reproducibility of platforms:
- xLinux
- pLinux
- aarch64Linux
but not with:
- Mac
- Windows
- zLinux
- AIX
Docker build containers could be created with some work for Mac, Windows and zLinux, but AIX would be a problem.

[...]

PR #2666, opened 2021-06-21
Add BUILD_CONFIG[OS_FULL_VERSION]

andrew-m-leonard:
For the "reproducible builds" issue: https://github.com/adoptium/temurin-build/issues/2594#issuecomment-841199132
the key piece of information missing from the built "BUILD_CONFIG" was the full os version level the binary was built on, this adds that info.

### DIFF
--- a/sbin/common/config_init.sh
+++ b/sbin/common/config_init.sh
@@ -79,6 +79,7 @@ OPENJDK_FOREST_NAME
 OPENJDK_SOURCE_DIR
 OPENJDK_UPDATE_VERSION
 OS_KERNEL_NAME
+OS_FULL_VERSION
 OS_ARCHITECTURE
 PATCHES
 RELEASE
@@ -365,6 +366,24 @@ function configDefaults() {
 
   # The OS kernel name, e.g. 'darwin' for Mac OS X
   BUILD_CONFIG[OS_KERNEL_NAME]=$(uname | awk '{print tolower($0)}')
+
+  # Determine OS full system version
+  local unameSys=$(uname -s)
+  local unameOSSysVer=$(uname -sr)
+  if [ "${unameSys}" == "Linux" ]; then
+    # Linux distribs add more useful distrib in the single line file /etc/system-release
+    if [ -f "/etc/system-release" ]; then
+      local linuxName=$(tr -d '"' < /etc/system-release)
+      unameOSSysVer="${unameOSSysVer} : ${linuxName}"
+    fi
+  elif [ "${unameSys}" == "AIX" ]; then
+    # AIX provides full version info using oslevel
+    aixVer=$(oslevel -r)
+    unameOSSysVer="${unameSys} ${aixVer}"
+  fi
+
+  # Store the OS full version name, eg.Darwin 20.4.0
+  BUILD_CONFIG[OS_FULL_VERSION]="${unameOSSysVer}"
 
   local arch=$(uname -m)
 


andrew-m-leonard:
run tests

sbin/common/config_init.sh
```diff
@@ -366,6 +367,35 @@ function configDefaults() {
...
 
+  # Determine OS full system version
+  local unameSys=$(uname -s)
+  local unameOSSysVer=$(uname -sr)
+  if [ "${unameSys}" == "Linux" ]; then
+    # Linux distribs add more useful distrib /etc/*-release file content
+    set +e
+    cat /etc/*-release > /dev/null
+    local cmdRc="$?"
+    if [ "${cmdRc}" == "0" ]; then
```
sxa, line 378: How about:
```
if cat /etc/*-release` > /dev/null; then
```
instead of 
```
 cat /etc/*-release > /dev/null
    local cmdRc="$?"
    if [ "${cmdRc}" == "0" ]; then
```
Seems quite a bit simpler to read - at least to me.

sbin/common/config_init.sh
```diff
@@ -366,6 +367,35 @@ function configDefaults() {
...
+      local linuxName=""
+      cat /etc/*-release | grep PRETTY_NAME > /dev/null
+      cmdRc="$?"
+      if [ "${cmdRc}" == "0" ]; then
```
sxa, line 382: As above - `if grep PRETTY_NAME /etc/*-release; then`

sbin/common/config_init.sh
```diff
@@ -366,6 +367,35 @@ function configDefaults() {
...
+        linuxName=$(cat /etc/*-release | grep PRETTY_NAME | cut -d'=' -f2 | tr -d '"')
```
sxa, line 383: Maybe:
```
linuxName=$(grep PRETTY_NAME /etc/*-release | cut -d= -f2 | tr -d '"')
```
Or even this, but it's a bit less clear what's happening so the grep might be better:
```
linuxName=$(awk -F= "/PRETTY_NAME/{print$2}" | tr -d '"')
```

sbin/common/config_init.sh
```diff
@@ -366,6 +367,35 @@ function configDefaults() {
...
+      else
+        linuxName=$(cat /etc/*-release | tr -d '"')
+      fi
+      unameOSSysVer="${unameOSSysVer} : ${linuxName}"
+    fi
+    set -e
+  elif [ "${unameSys}" == "AIX" ]; then
+    # AIX provides full version info using oslevel
+    aixVer=$(oslevel)
```
sxa, line 392: This gives a more useful definition of the OS level
```suggestion
    aixVer=$(oslevel -r)
```

sxa (changes requested):
AIX one is probably most critical in terms of the changes here but the simplifications in the other comments are up to you whether to take or not :-)

sbin/common/config_init.sh
andrew-m-leonard: old habits! changed

sbin/common/config_init.sh
andrew-m-leonard: yep

sbin/common/config_init.sh
andrew-m-leonard: prefer grep

sbin/common/config_init.sh
andrew-m-leonard: yep

sxa (approved):

sxa:
Approving on the assumption that my suggestions have been tested first ;-)

andrew-m-leonard:
> Approving on the assumption that my suggestions have been tested first ;-)
@sax They have!

sbin/common/config_init.sh
```diff
@@ -366,6 +367,31 @@ function configDefaults() {
...
+  local unameOSSysVer=$(uname -sr)
+  if [ "${unameSys}" == "Linux" ]; then
+    # Linux distribs add more useful distrib /etc/*-release file content
+    set +e
+    if cat /etc/*-release > /dev/null; then
+      local linuxName=""
+      if grep PRETTY_NAME /etc/*-release; then
+        linuxName=$(grep PRETTY_NAME /etc/*-release | cut -d= -f2 | tr -d '"')
+      else
+        linuxName=$(cat /etc/*-release | tr -d '"')
```
sxa, line 381: I assume this isn't going to cause any problems if it's returning multi-line output?

gdams (approved):

sbin/common/config_init.sh
andrew-m-leonard: mmm possibly, will check

sxa: Where does it return just `Red Hat`? Not on my RHEL7 and 8 systems ...

andrew-m-leonard: @sxa here: test-ibmcloud-rhel7-x64-1
```
NAME="Red Hat Enterprise Linux Server"
VERSION="7.9 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.9"
PRETTY_NAME="Red Hat"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.9:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.9
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.9"
```

PR #2667, opened 2021-06-23
Arm machines do not have /etc/system-release use /etc/os-release

andrew-m-leonard:
Arm machines do not have /etc/system-release, but do have /etc/os-release, default to NAME & VERSION from that if no system-release


andrew-m-leonard:
run tests

PR #2668, opened 2021-06-23
Arm machines do not have /etc/system-release use /etc/os-release

andrew-m-leonard:
Arm machines do not have /etc/system-release, but do have /etc/os-release, default to NAME & VERSION from that if no system-release

### DIFF
--- a/sbin/common/config_init.sh
+++ b/sbin/common/config_init.sh
@@ -371,10 +371,20 @@ function configDefaults() {
   local unameSys=$(uname -s)
   local unameOSSysVer=$(uname -sr)
   if [ "${unameSys}" == "Linux" ]; then
-    # Linux distribs add more useful distrib in the single line file /etc/system-release
+    # Linux distribs add more useful distrib in the single line file /etc/system-release,
+    # or property file /etc/os-release
     if [ -f "/etc/system-release" ]; then
       local linuxName=$(tr -d '"' < /etc/system-release)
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
run tests

sbin/common/config_init.sh
```diff
@@ -371,10 +371,20 @@ function configDefaults() {
   local unameSys=$(uname -s)
   local unameOSSysVer=$(uname -sr)
   if [ "${unameSys}" == "Linux" ]; then
-    # Linux distribs add more useful distrib in the single line file /etc/system-release
+    # Linux distribs add more useful distrib in the single line file /etc/system-release,
+    # or property file /etc/os-release
     if [ -f "/etc/system-release" ]; then
       local linuxName=$(tr -d '"' < /etc/system-release)
       unameOSSysVer="${unameOSSysVer} : ${linuxName}"
+    elif [ -f "/etc/os-release" ]; then
```
karianna, line 379: Is it possible that an O/S would ever have both of these folders?

karianna (approved):
LGTM

sbin/common/config_init.sh
andrew-m-leonard: Yes, some do, but they have the same os name value. most have system-release, some also have os-release and quite often both are symlinks to a vendor release file, eg."redhat-release"

gdams (approved):

PR #2669, opened 2021-06-23
Ensure correct boot jdk is used for openj9 and correct required boot jdk console msg

andrew-m-leonard:
OpenJ9 must use jdk-11 bootjdk for building jdk-11.
Also currently build-farm issues msg:
```
19:38:27  Required boot JDK version: 10
```
but does not actually always use this version.
Correct this logic and pass settings down to where boot jdk download logic is.

### DIFF
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -128,8 +128,15 @@ then
   echo "Detecting boot jdk for: ${JAVA_TO_BUILD}"
   echo "Found build version: ${JAVA_FEATURE_VERSION}"
   JDK_BOOT_VERSION=$(( JAVA_FEATURE_VERSION - 1 ))
+  if [ "${JAVA_FEATURE_VERSION}" == "11" ] && [ "${VARIANT}" == "openj9" ]; then
+    # OpenJ9 only supports building jdk-11 with jdk-11
+    JDK_BOOT_VERSION="11"
+  fi
 fi
 echo "Required boot JDK version: ${JDK_BOOT_VERSION}"
+
+# export for platform specific scripts
+export JDK_BOOT_VERSION
 
 # shellcheck source=build-farm/set-platform-specific-configurations.sh
 source "${PLATFORM_SCRIPT_DIR}/set-platform-specific-configurations.sh"
--- a/build-farm/platform-specific-configurations/mac.sh
+++ b/build-farm/platform-specific-configurations/mac.sh
@@ -81,42 +81,41 @@ fi
 echo "[WARNING] You may be asked for your su user password, attempting to switch Xcode version to ${XCODE_SWITCH_PATH}"
 sudo xcode-select --switch "${XCODE_SWITCH_PATH}"
 
-# Adopt does not have builds of OpenJDK 7, OpenJDK 8 can boot itself just fine.
-if [ "${JAVA_FEATURE_VERSION}" = "8" ]; then
-  BOOT_JDK_VERSION="${JAVA_FEATURE_VERSION}"
-else
-  BOOT_JDK_VERSION="$((JAVA_FEATURE_VERSION-1))"
+# No MacOS builds available of OpenJDK 7, OpenJDK 8 can boot itself just fine.
+if [ "${JDK_BOOT_VERSION}" == "7" ]; then
+  echo "No jdk7 boot JDK available on MacOS using jdk8"
+  JDK_BOOT_VERSION="8"
 fi
-BOOT_JDK_VARIABLE="JDK${BOOT_JDK_VERSION}_BOOT_DIR"
+BOOT_JDK_VARIABLE="JDK${JDK_BOOT_VERSION}_BOOT_DIR"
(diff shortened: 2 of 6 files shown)


andrew-m-leonard:
run tests

build-farm/make-adopt-build-farm.sh
```diff
@@ -128,9 +128,16 @@ then
...
   JDK_BOOT_VERSION=$(( JAVA_FEATURE_VERSION - 1 ))
+  if [ "${JAVA_FEATURE_VERSION}" == "11" ] && [ "${VARIANT}" == "openj9" ]; then
+    # OpenJ9 only supports building jdk-11 with jdk-11
+    JDK_BOOT_VERSION="11"
+  fi
 fi
 echo "Required boot JDK version: ${JDK_BOOT_VERSION}"
 
+export REQUIRED_BOOT_JDK_VERSION
+REQUIRED_BOOT_JDK_VERSION="${JDK_BOOT_VERSION}"
```
sxa, line 139: What's the reason for creating a new variable instead of just using `JDK_BOOT_VERSION`?

andrew-m-leonard: because it was being export'd and I wanted the name to be very clear as in "REQUIRED". "JDK_BOOT_VERSION" is rather anonymously named! IMO

andrew-m-leonard:
run tests

build-farm/make-adopt-build-farm.sh
sxa: Hmm not sure "because it was being exported" is a good reason to create a new variable (Does it even need to be exported?) which adds complexity. We should perhaps consider changing the name of the existing one.

andrew-m-leonard: ok no worries, i'll change

sophia-guo: Not sure if we need to change the existing one or add a new one. Yes, for some JDK we need different JDK_BOOT_VERSION from other but IMO JDK_BOOT_VERSION  still works. It is quite clear and straightforward IMO.

andrew-m-leonard: done

andrew-m-leonard:
@sophia-guo I think I might have not seen your review comment when I force updated the commit with Stewarts change...?

sxa (commented):
Hmmm we now have:
`BOOT_JDK_VERSION="${JDK_BOOT_VERSION}"`
in all the scripts which looks a bit odd - would it not make more sense to just use `JDK_BOOT_VERSION` in all of the `platform-specific-configurations` scripts?

andrew-m-leonard:
> Hmmm we now have:
> `BOOT_JDK_VERSION="${JDK_BOOT_VERSION}"`
> in all the scripts which looks a bit odd - would it not make more sense to just use `JDK_BOOT_VERSION` in all of the `platform-specific-configurations` scripts?

yeah I thought that too, but hesitated as quite a lot of lines to change... i'll take the plunge

sxa:
> yeah I thought that too, but hesitated as quite a lot of lines to change... i'll take the plunge

I think it's probably worthwhile otherwise it'll probably not get revisited in the future ... And since we're testing it one change at a time it should be ok. Sorry for the drip-fed comments on this. I'm fundamentally happy with the change (other than the fact it's a shame we can't still build with the older boot JDK!)

sophia-guo (approved):
LGTM

karianna (approved):
LGTM

build-farm/platform-specific-configurations/mac.sh
```diff
@@ -81,42 +81,43 @@ fi
 echo "[WARNING] You may be asked for your su user password, attempting to switch Xcode version to ${XCODE_SWITCH_PATH}"
 sudo xcode-select --switch "${XCODE_SWITCH_PATH}"
 
-# Adopt does not have builds of OpenJDK 7, OpenJDK 8 can boot itself just fine.
-if [ "${JAVA_FEATURE_VERSION}" = "8" ]; then
-  BOOT_JDK_VERSION="${JAVA_FEATURE_VERSION}"
+# No MacOS builds available of OpenJDK 7, OpenJDK 8 can boot itself just fine.
+if [ "${JDK_BOOT_VERSION}" == "7" ]; then
+  echo "No jdk7 boot JDK available on MacOS using jdk8"
+  MAC_BOOT_JDK_VERSION="8"
```
sxa, line 87: Can we just set `JDK_BOOT_VERSION=8` here and use that through the script to be consistent with the other platforms?

sxa: (Obviously since you have two approvals without me you could merge it anyway, but it would be nice to change this ;-) )

andrew-m-leonard: we could, my thinking was that was changing the upstream script value, say other scripts were being called that... but i'm just following a not changing the caller's value principle when it's actually not relevant here anyway...

andrew-m-leonard: nw, changed

sxa (approved):
Looks good to go 👍🏻

PR #2689, opened 2021-08-18
Switch cert updater to use Temurin bot

gdams:
This will create the pull request as the correct user and pass the ECA checks


karianna (approved):
lgtm

sxa (approved):
ECA check appears to be passing anyway, but I'll approve this if it's using the wrong token and we can verify tonight whether there are any other side effects of this change.

PR #2703, opened 2021-09-08
Temporarily remove building JDK (tip) for GH Action

karianna:
Since we can't yet download the latest 17 it means we can't build tip properly due to sealed classes support.  We will enable this after we release 17 ourselves and use setupjava-2 action


karianna:
Closing this PR, I'd have to do more YAML surgery on Mac and Windows sections and I don't understand it well enough

PR #2705, opened 2021-09-10
Change vendor from Eclipse Foundation to Eclipse Adoptium

sxa:
As discussed in the marketing group meeting and in the Adoptium Workgroup Steering Committee meetings this week. This should remove ambiguity to end users by clarifying that this is the "Adoptium" build.

Signed-off-by: Stewart X Addison <sxa@redhat.com>

PR #2710, opened 2021-09-10
Revert removal of Temurin name from JDK8 java -version output

sxa:
This partially reverts https://github.com/adoptium/temurin-build/pull/2705 and makes the version output say `(Temurin)` instead of `(Eclipse Adoptium)` to make it consistent with other versions and vendors

NOTE: Does not currently have the desired effect as this also causes `java.vendor` to become `Temurin` again on JDK8

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -254,7 +254,7 @@ configureVersionStringParameter() {
 
     if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_HOTSPOT}" ]; then
 
-      addConfigureArg "--with-company-name=" "\"Eclipse Adoptium\""
+      addConfigureArg "--with-company-name=" "\"Temurin\""
 
       # No JFR support in AIX or zero builds (s390 or armv7l)
       if [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" != "s390x" ] && [ "${BUILD_CONFIG[OS_KERNEL_NAME]}" != "aix" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" != "armv7l" ]; then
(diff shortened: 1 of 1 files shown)


karianna (approved):
LGTM - we can try to resolve again later.

johnoliver (approved):

karianna:
@sxa LMK if this needs to come out of draft before the release

gdams (approved):

jerboaa:
Is there a way to retest this PR? With #2711 in, results should look better.

karianna:
> Is there a way to retest this PR? With #2711 in, results should look better.

I tried re-running the GH Actions, but I'm not sure if that is a rerun with the latest config or with the cached GH Actions config.  One of the issues I think is that this change may not work for `main` but will work for `dev`.

karianna:
> > Is there a way to retest this PR? With #2711 in, results should look better.
> 
> I tried re-running the GH Actions, but I'm not sure if that is a rerun with the latest GH config (now that #2711 has landed) or with the cached GH Actions config. Judging by the jdk tip failures (which didn't happen on #2711), I'd say that a rerun still uses old config.
> 
> One of the others issues I think is that this change may not work for `main` but will work for `dev` WRT to Java 8.

I'll try editing this PR with a comment which should force a rerun with the latest GH Action config.

gdams:
@karianna I think the PR will need an actual rebase

karianna:
> @karianna I think the PR will need an actual rebase

I think my editing of the file seemed to do the trick.  So the tests pass now but I'm wondering if it's the right thing (tm)

karianna:
We can delay landing this until Oct PSU, so taking out of Sept Release train

sxa:
> I think my editing of the file seemed to do the trick. So the tests pass now but I'm wondering if it's the right thing (tm)

Merging. It should let the tests (smoke in jenkins and github) pass and we can discuss subsequently if there is an alternate way forward for this output.

Issue #2704, opened 2021-09-08
Upgrade GH Actions to use setup-java 2 and grab latest Java 17 for tip builds


karianna:
@karianna Thanks for the review!

Build failures seem to be caused by:
```
warning: unknown enum constant Feature.SEALED_CLASSES
warning: unknown enum constant Feature.SEALED_CLASSES
```
That suggests the JDK 18 build uses a too old JDK 17 boot JDK. Uses 17+20. Should be 17+24 or better. See:
https://bugs.openjdk.java.net/browse/JDK-8260517

Either way, unrelated failures.

_Originally posted by @jerboaa in https://github.com/adoptium/temurin-build/issues/2702#issuecomment-915043130_

gdams:
The build jobs actually use the API. I'll do some digging but I think we just need to switch to the Adoptium api to get the latest nightly builds.

gdams:
Yup, see all the references in https://github.com/adoptium/temurin-build/search?q=api.adoptopenjdk.net&type= that need updating.

karianna:
> Yup, see all the references in https://github.com/adoptium/temurin-build/search?q=api.adoptopenjdk.net&type= that need updating.

Will do this post-release - I think much of that YAML needs updating to use the `setup-java` action

PR #2711, opened 2021-09-13
switch bootJDK download scripts to use adoptium binaries

gdams:
This fixes the failing JDK tip builds

fixes: https://github.com/adoptium/temurin-build/issues/2704

I really want to split these horrible blocks of code into a shared function but will work on that in a follow-up PR

### DIFF
--- a/.azure-devops/build/steps/macOS/before.yml
+++ b/.azure-devops/build/steps/macOS/before.yml
@@ -38,12 +38,12 @@ steps:
 
   # download JDKs from AdoptOpenJDK
   - bash: |
-      wget -O "${{ parameters.dependenciesDir }}/openjdk8.tar.gz" "https://api.adoptopenjdk.net/v3/binary/latest/8/ga/mac/x64/jdk/hotspot/normal/adoptopenjdk?project=jdk"
+      wget -O "${{ parameters.dependenciesDir }}/openjdk8.tar.gz" "https://api.adoptium.net/v3/binary/latest/8/ga/mac/x64/jdk/hotspot/normal/adoptium?project=jdk"
       if [ "${BOOTJDK_VERSION}" == "15" ]
       then
-          wget -O "${{ parameters.dependenciesDir }}/openjdk.tar.gz" "https://api.adoptopenjdk.net/v3/binary/latest/${BOOTJDK_VERSION}/ea/mac/x64/jdk/hotspot/normal/adoptopenjdk?project=jdk"
+          wget -O "${{ parameters.dependenciesDir }}/openjdk.tar.gz" "https://api.adoptium.net/v3/binary/latest/${BOOTJDK_VERSION}/ea/mac/x64/jdk/hotspot/normal/adoptium?project=jdk"
       else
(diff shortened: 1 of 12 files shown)


jerboaa (approved):
Looks fine to me.

.azure-devops/build/steps/macOS/before.yml
```diff
@@ -38,12 +38,12 @@ steps:
 
   # download JDKs from AdoptOpenJDK
   - bash: |
-      wget -O "${{ parameters.dependenciesDir }}/openjdk8.tar.gz" "https://api.adoptopenjdk.net/v3/binary/latest/8/ga/mac/x64/jdk/hotspot/normal/adoptopenjdk?project=jdk"
+      wget -O "${{ parameters.dependenciesDir }}/openjdk8.tar.gz" "https://api.adoptium.net/v3/binary/latest/8/ga/mac/x64/jdk/hotspot/normal/adoptium?project=jdk"
       if [ "${BOOTJDK_VERSION}" == "15" ]
       then
-          wget -O "${{ parameters.dependenciesDir }}/openjdk.tar.gz" "https://api.adoptopenjdk.net/v3/binary/latest/${BOOTJDK_VERSION}/ea/mac/x64/jdk/hotspot/normal/adoptopenjdk?project=jdk"
+          wget -O "${{ parameters.dependenciesDir }}/openjdk.tar.gz" "https://api.adoptium.net/v3/binary/latest/${BOOTJDK_VERSION}/ea/mac/x64/jdk/hotspot/normal/adoptium?project=jdk"
```
karianna, line 44: I don't think adoptium can sever up 15 binaries no?

karianna (changes requested):
Should we be using GH setup-java action for this?  Am also worried that adoptium only pulls down subset of adoptopenjdk API

gdams:
> Should we be using GH setup-java action for this? Am also worried that adoptium only pulls down subset of adoptopenjdk API

No one of the main reasons we don't is because we want to test the code that pulls down boot JDK's. I've added another check which will fall back to adoptopenjdk if required.

gdams:
Test failures are unrelated to this change

karianna (approved):

karianna:
I'm going to merge this as it'll help with testing any emergency PRs WRT to java 17

sxa:
@karianna @gdams I can see the fallback to adoptopenjdk in the linux script here but not the others. Isn't this introducing a bit of a risk? Given that we have the forwarding from adoptopenjdk->adoptium I'm not clear on why we're pushing this through during a lockdown period.

karianna:
> @karianna @gdams I can see the fallback to adoptopenjdk in the linux script here but not the others. Isn't this introducing a bit of a risk? Given that we have the forwarding from adoptopenjdk->adoptium I'm not clear on why we're pushing this through during a lockdown period.

17/tip builds were failing for GH Actions which meant we didn't have PR testing on any changes we might need to push through during release week, so this was worth getting in (it certainly couldn't make it worse).

Issue #2688, opened 2021-08-17
Alpine musl builds of Temurin


brunoborges:
It seems that Project Portola (JEP 386) had very minor changes to make it possible to build OpenJDK for musl libc instead of GNU glibc. Most of the changes were about steps required to build.

Currently, the following OpenJDK builds provide packages for Alpine Linux:

. Amazon Corretto
. Azul Systems' Zulu
. BellSoft Liberica JDK

It would be good to have a build of Eclipse Temurin for Alpine Linux so that the ecosystem could leverage AQAVit's value.

jerboaa:
https://openjdk.java.net/jeps/386 is JDK 16 and better. So I'm for providing those for JDK 16+ builds only (JDK 17 now) unless changes get backported upstream first.

gdams:
The plan is almost certainly to produce MUSL based images. We already produce nightly builds of them.

More information will be available shortly

brunoborges:
> unless changes get backported upstream first.

I used to think that too, @jerboaa, but after looking at JEP 386 and the work that got done, I couldn't see any actual changes other than build instructions. Can you point me to a changeset/PR that shows significant changes ?

brunoborges:
Ok, after digging a bit more with some help from a co-worker, I found this commit: https://github.com/openjdk/jdk/commit/63009f90/

I'm curious as to whether other vendors have backported these changes into their pre-16 branches in order to produce musl binaries, and whether there was ever an attempt to backport these changes to upstream (jdk11u at the very least).

karianna:
We've got Alpine musl based images in the CI/CD pipeline + compliance project so we can now test whatever we build.  The PMC is currently discussing backporting to 11 and 8.

karianna:
PMC has agreed to allow backports to temurin 8 and 11 in parallel to working on the backport upstream at OpenJDK.  You'll soon see one Stephanie Crater working on this starting next week :-).

jerboaa:
@karianna The consensus on the PMC was to get an overall feel for the JDK 8/11 backport upstream first. Then, if the general agreement is affirmative to get it in upstream, proceed with the downstream work. That way it would only be a temporary downstream divergence. Has this been socialized on upstream mailing lists?

karianna:
[...]

PR #2714, opened 2021-09-13
enable headless for alpine builds

gdams:


gdams:
A simple test application to test if the binary is headless:

```java
import java.awt.*;

public class HeadlessBasics
{
    public static void main(String[] args)
    {
        // Check whether the application is
        // running in headless mode.
        GraphicsEnvironment ge = 
        GraphicsEnvironment.getLocalGraphicsEnvironment();
        System.out.println("Headless mode: " + ge.isHeadless());
    }
}
```

gdams:
Seems to have worked:

```bash
/working-dir # ./jdk-16.0.2\+7/bin/java HeadlessBasics
Headless mode: true
```

karianna (approved):
LGTM

karianna:
PMC approved - failed tests are for unrelated reasons (we have another PR in the wings for that).

sxa (commented):
Hmmm - can we not do this in our pipelines and leave the build scripts capable of building both as an option - this change will likely make it a bit more difficult for people to build a non-headless version from our scripts going forwards if they wish to do so.

gdams:
> Hmmm - can we not do this in our pipelines and leave the build scripts capable of building both as an option - this change will likely make it a bit more difficult for people to build a non-headless version from our scripts going forwards if they wish to do so.

@sxa I'd prefer to make a change like that in our pipeline. This sits in our platform-specific configurations directory which is not the default path for other people using our scripts.

sxa:
> @sxa I'd prefer to make a change like that in our pipeline. It creates a disconnect between the two repos and creates issues like the arm64 macOS bugs that I've just had to fix. This sits in our platform-specific configurations directory which is not the default path for other people using our scripts.

I presume your first sentence should have had a 'not' in it?

That disconnect was introduced when we split out piplines from the build repo (which I wasn't a fan of if I'm honest) and we have several other configure arguments for the different variants and platforms in places like https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy now. I'd be reluctant to add more things into what really ought to be configuration files specific to our ansible machine environment as opposed to configurable options affecting the JVM which it should be possible to override. It's not ideal having it two repositories but it's where we are and I don't want to put things in a different place just because of that, and certainly not just because a mistake was made with changing one such option.

(As an aside, when I build my own JDKs with our scripts I tend to use `make-adopt-build-farm.sh` in preference to `makejdk-any-platform.sh` because it's MUCH more usable for an end-user. I'd be quite surprised if many people actively chose to just use makejdk-any-platform on its own unless they didn't know about the other ones because the additional value it provides is quite low for other users IMHO so I don't think we should remove the ability to switch this on and off in those files without really good reason)

karianna:
> A simple test application to test if the binary is headless:
> 
> ```java
> import java.awt.*;
> 
> public class HeadlessBasics
> {
>     public static void main(String[] args)
>     {
>         // Check whether the application is
>         // running in headless mode.
>         GraphicsEnvironment ge = 
>         GraphicsEnvironment.getLocalGraphicsEnvironment();
>         System.out.println("Headless mode: " + ge.isHeadless());
>     }
> }
> ```

Can we add that as an actual Smoke Test?

karianna:
> > @sxa I'd prefer to make a change like that in our pipeline. It creates a disconnect between the two repos and creates issues like the arm64 macOS bugs that I've just had to fix. This sits in our platform-specific configurations directory which is not the default path for other people using our scripts.
> 
> I presume your first sentence should have had a 'not' in it?
> 
> That disconnect was introduced when we split out piplines from the build repo (which I wasn't a fan of if I'm honest) and we have several other configure arguments for the different variants and platforms in places like https://github.com/adoptium/ci-jenkins-pipelines/blob/master/pipelines/jobs/configurations/jdk11u_pipeline_config.groovy now. I'd be reluctant to add more things into what really ought to be configuration files specific to our ansible machine environment as opposed to configurable options affecting the JVM which it should be possible to override. It's not ideal having it two repositories but it's where we are and I don't want to put things in a different place just because of that, and certainly not just because a mistake was made with changing one such option.
> 
> (As an aside, when I build my own JDKs with our scripts I tend to use `make-adopt-build-farm.sh` in preference to `makejdk-any-platform.sh` because it's MUCH more usable for an end-user. I'd be quite surprised if many people actively chose to just use makejdk-any-platform on its own unless they didn't know about the other ones because the additional value it provides is quite low for other users IMHO so I don't think we should remove the ability to switch this on and off in those files without really good reason)

So I think this boils down to:  Do we think headless is a sensible option for *all* users of our build scripts or only for our actual temurin builds?

I personally think it's OK to be headless for all users. I've yet to see any Alpine users using a Headful (is that the right term?) build...

Side note - I wonder if we should take temurin off the name of this repo and just have it as build scripts sometime in the future to make it clearer that anyone can use them.

sxa:
> So I think this boils down to: Do we think headless is a sensible option for _all_ users of our build scripts or only for our actual temurin builds?

My view - absolutely only for temurin. There is no reason to actively disable this for anyone else, or even people using our scripts to build their own Temurin-equivalent binary. We have already had opinions expressed from at least two partiies that certain APIs not related to GUIs - image handling as I recall - may well not be available in such a build, so let's not make this something that we don't have an easy override mechanism for if others are using our build scripts and want a complete build, and we should ensure that we can switch back if required in the least intrusive way, which is in the Temurin build pipelines. This would be consistent with how we enable other JDK options such as dtrace enablement.

> Side note - I wonder if we should take temurin off the name of this repo and just have it as build scripts sometime in the future to make it clearer that anyone can use them.

Maybe - but they're maintained primarily for and by temurin, so I'm ok with it. But we absolutely must not forget when making changes that others may be using them (such as OpenJ9!) and take that into consideration when making changes - we have to have some responsibility to the user community and these scripts were intended to be used by others who want them. Dragonwell have already expressed an interest in having Alpine builds and we shouldn't force Temurin decisions such as this upon them.

gdams:
related to https://github.com/adoptium/temurin-build/issues/2688

gdams:
closing in favour of https://github.com/adoptium/ci-jenkins-pipelines/pull/190

PR #2723, opened 2021-09-24
Fix source archive creation

### LINKED ISSUE #2722: --create-source-archive may fail to translate the target name to something unique
When performing a build with `--create-source-archive` and specifying a target file name, for example `--target-file-name OpenJDK11U-jdk_x64_linux_hotspot_2021-09-23-16-24.tar.gz`, then the source archive translation fails. This code:

```
  if echo ${BUILD_CONFIG[TARGET_FILE_NAME]} | grep -q hotspot; then
    srcArchiveName=$(echo "${BUILD_CONFIG[TARGET_FILE_NAME]//-jdk_*(?)_hotspot_/-sources_}")
  else
[...]
```
... doesn't work. While it translates properly on a regular shell, it doesn't seem to work in scripts. What's worse, this kind of model is being used in CI.

The net effect is that `sourceArchiveName` will remain `${BUILD_CONFIG[TARGET_FILE_NAME]}`, in the above example `OpenJDK11U-jdk_x64_linux_hotspot_2021-09-23-16-24.tar.gz`. Then later on that filename gets replaced by the actual JDK build and it looks as if no source archive got created during the build at all.

This issue doesn't surface if no target file name is being passed as it enters the `else` branch in the above code.

jerboaa:
- For one the target filename gets now properly replaced by using
  sed over Bash variable substitution.
- For two, add an extra check to fail the build if the resulting
  tarball name doesn't contain string `-sources`
- For three, add '-' file argument passed to grep so as to properly
  set the return code as if expressions rely on them

Closes #2722

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -508,14 +508,21 @@ createSourceArchive() {
   local sourceArchiveTargetPath="$(getSourceArchivePath)"
   local tmpSourceVCS="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/tmp-openjdk-git"
   local srcArchiveName
-  if echo ${BUILD_CONFIG[TARGET_FILE_NAME]} | grep -q hotspot; then
+  if echo ${BUILD_CONFIG[TARGET_FILE_NAME]} | grep -q hotspot -; then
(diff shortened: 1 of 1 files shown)


jerboaa:
OK, remaining build failures seem unrelated. Smoketests seem to expect shenandoah and zgc to be there for 8 (But they aren't in 8u).

karianna (approved):
LGTM

gdams (approved):

tellison (approved):

PR #2724, opened 2021-09-25
Fix macOS build info containing newline

octylFractal:
Possibly due to changes in `awk`, current builds have the following content in the `release` file, which is hard to parse:
```
BUILD_INFO="OS: Mac OS X Version: 10.14.6
18G84"
```
This change fixes the shell command to no longer produce newlines, which should restore the normal nature of this entry. I was unable to test this change as it was reported to me by a user of one of my programs, and I don't have a Mac to verify on, but they ran the command that is now there and confirmed it does not produce a newline-separator anymore.


gdams:
confirmed the fix as:

```bash
BUILD_INFO="OS: Mac OS X Version: 10.15.7 19H1417"
```

gdams (approved):

karianna (approved):
LGTM

PR #2725, opened 2021-09-27
Implement building of OpenJDK static libs

### LINKED ISSUE #2699: Build static-libs image
OpenJDK 11+ provides the ability to build statically linked versions of OpenJDK libraries. Those libraries are required to build Graal VM CE (or [Mandrel](https://github.com/graalvm/mandrel)) from source. We should allow for optionally building those artefacts in build scripts.

jerboaa:
This implements building of OpenJDK static libs via the
'static-libs-image' make target. It supports building them
for glibc Linux distributions as well as musl-based C-library.

Closes #2699


karianna (approved):
LGTM - I assume this is to support GraalVM deployments?

jerboaa:
> LGTM

Thanks.

> I assume this is to support GraalVM deployments?

It's more like for users who want to build GraalVM from source to give them a boot JDK for compiling GraalVM. The JDK bits plus the static-libs brought into it, enables that. Users can pick and choose.

jerboaa:
> > I assume this is to support GraalVM deployments?
> 
> It's more like for users who want to build GraalVM from source to give them a boot JDK for compiling GraalVM. The JDK bits plus the static-libs brought into it, enables that. Users can pick and choose.

Also, it would allow us for [mandrel releases](https://github.com/graalvm/mandrel/releases/) to switch to Temurin builds for the boot JDK, rather than using old [openjdk11-upstream-binaries](https://github.com/AdoptOpenJDK/openjdk11-upstream-binaries).

gdams (approved):

Issue #2683, opened 2021-08-03
Should Adoptium be producing Temurin JREs for releases beyond version 8?


tellison:
We all agree that Temurin 8 should have both a JDK and JRE build available, but what about later versions?

With jlink being introduced alongside modularity, there is a new way to strip down the Java runtime to minimize the delivery in a way that specifically supports the application.  Users can also jlink a 'general purpose' runtime environment for multiple applications to share.

AdoptOpenJDK produced JREs and JDKs based OpenJDK 11.  Should Adoptium continue this for Temurin, or drop the concept of JREs beyond Temurin 8? This would be a regression for users of the AdoptOpenJDK JREs when moving to Temurin builds.

sxa:
I think we should ideally give users some warning before deprecating something, however in principle I agree that not providing JREs for 11+ seems "reasonable" particularly if we give people a guide on how to do it themselves, however I think we should probably check with companies supporting the Adopt binaries before finalising such as decision.

johanjanssen:
We are using AdoptOpenJDK Java 11 Docker JRE images and hope to switch to Java 17 JRE images once/if available.

For us, they're a lot better than JDK's as they're a lot smaller. It's also easy to maintain as we can simply change the Docker image tag once a new minor/major version of Java is released.

I've looked at jlink before, but it's less straightforward. If I look at for instance https://medium.com/azulsystems/using-jlink-to-build-java-runtimes-for-non-modular-applications-9568c5e70ef4 it's still some work to use, especially if the application doesn't use the module system. 

Instead of simply using the latest Docker image, we would have to create our own Docker image. Preferably one for each application to have them as small as possible. Then we would have to run the steps every time a new minor or major version of Java is released.

Of course it's all possible and we might be able to automate some parts. But it's still a bigger effort then simply using a JRE. I'm afraid that if the JRE's are no longer available, then people will switch to a JDK or a JRE build from another vendor instead of jlink as it's easier and less maintenance. 

Also if I look at the number of downloads (see below) for some of the Chocolatey packages (Windows package manager) I maintain, then I see the JRE is still quite popular. 

`
adoptopenjdk	   139288
adoptopenjdkjre 121536
adoptopenjdk11jre	26051
adoptopenjdk11	34622
openjdk8	        165571
openjdk8jre	330620
adoptopenjdk8	191632
adoptopenjdk8jre	36497
`

[...]

PR #2736, opened 2021-09-30
add shell/powershell script to root dir for easy JRE generation

gdams:


sbin/build.sh
```diff
@@ -988,6 +988,28 @@ createNoticeFile() {
   fi
 }
 
+# Adds a script for easy JRE generation using jlink
+addCreateJREScript() {
```
karianna, line 992: Let's not call it a JRE as that's an official designation.  I'd change it to Runtime or something like that

karianna (changes requested):

sxa (commented):
We do need to carefully look at the comments/feedback we've had in https://github.com/adoptium/temurin-build/issues/2683 and the [slack](https://adoptium.slack.com/archives/C09NLQQAV/p1632733549158800) [conversations](https://adoptium.slack.com/archives/C09NLQQAV/p1632955929226600) to see if this would alleviate people's objections. There has obviously been some resistance to just running the one liner we have provided them with. Would an extra script that (fundamentally) just runs that one-liner really alleviate those concerns for most people (it doesn't for the download size issue, but likely does for the bundling scneario) and means we're basically shipping an Adoptium specific tool with the JDK which is a new precedent to be setting, so we do need to be careful about this decision and it's perception given how we claim to be a generally "pure" OpenJDK. We did in the past come up with a name for a "JRE alternative" too which we could use for this.

sbin/createJRE.sh
```diff
@@ -0,0 +1,51 @@
...
+    --no-man-pages \
+    --no-header-files \
+    --compress=2 \
+    --output "$jreDirectory"
+
+echo "Testing generated JRE"
+"$jreDirectory/bin/java" --version
+echo -e "Java Version test passed ✅\n"
+
+echo "Success: You're JRE runtime is available at $jreDirectory"
```
sxa, line 51:
```suggestion
echo "Success: Your JRE runtime is available at $jreDirectory"
```

sbin/createJRE.sh
```diff
@@ -0,0 +1,51 @@
...
+    --strip-debug \
+    --strip-debug \
+    --no-man-pages \
+    --no-header-files \
+    --compress=2 \
+    --output "$jreDirectory"
+
+echo "Testing generated JRE"
+"$jreDirectory/bin/java" --version
+echo -e "Java Version test passed ✅\n"
```
sxa, line 49: This really needs to check the return code from the java command, as it displays the message even when it fails. Also same comment for `jlink`

sxa (changes requested):

sbin/createJRE.sh
```diff
@@ -0,0 +1,51 @@
...
+    echo "e.g ./makeJRE.sh /home/user/jreruntime"
+    exit 1
+fi
+
+jreDirectory="$1"
+binPath="$jdkDirectory/bin"
+
+"$binPath/jlink" --add-modules ALL-MODULE-PATH \
+    --strip-debug \
+    --strip-debug \
```
sxa, line 41: Duplicate line?

sbin/createJRE.sh
gdams: added some

sxa (dismissed):

PR #2737, opened 2021-09-30
Actions: auto cancel builds if user pushes another commit

gdams:
Currently, if a user pushes several commits to a PR the CI will try to build each commit. This wastes both time and resources.

This PR adds a task that will cancel any other builds for the same PR.


karianna (approved):
lgtm

smlambert (approved):

PR #2742, opened 2021-10-05
Use curl and not wget by default to avoid LetsEncrypt SSL handshake for freemarker and boot JDK downloads

sxa:
This is impacting the OpenJ9 buids which use freemarker and I believe `curl` will work on the affected systems (or at least many of them!) as it uses it's own CA store.

FYI @AdamBrousseau 

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -137,7 +137,7 @@ if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
       # make-adopt-build-farm.sh has 'set -e'. We need to disable that for
       # the fallback mechanism, as downloading of the GA binary might fail.
       set +e
-      wget -q -O - "${apiURL}" | tar xpzf - --strip-components=1 -C "$bootDir"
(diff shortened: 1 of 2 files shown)


sxa:
[VPC run to check this](https://ci.adoptopenjdk.net/view/Tooling/job/VagrantPlaybookCheck/1247/) (along with https://github.com/adoptium/infrastructure/pull/2339 which is failing with the CentOS site)

AdamBrousseau:
I drafted https://github.com/ibmruntimes/ci-jenkins-pipelines/pull/28 and https://github.com/ibmruntimes/temurin-build/pull/3 which are untested at the moment. Since the only remaining platform that requires freemarker is riscv, @pshipton was ok with breaking it until cmake support was ready. Alternatively we could change the `if` to only download on that platform. I think the ideal check would be "if not cmake" or "if uma" but I'm unsure what info we have at the time of the check/download.

pshipton:
> ok with breaking it until cmake support was ready

Not what I meant. I meant I'm ok to break it for a week or two while we look for a workaround. Not sure when cmake will be ready, that might be too long.

sxa:
run tests

karianna (approved):
LGTM

smlambert (approved):
LGTM

PR #2746, opened 2021-10-06
Point scripts to correct bootjdk17 directory on Mac

Haroon-Khel:
ref https://adoptium.slack.com/archives/C53GHCXL4/p1633536423050000?thread_ts=1633443475.036300&cid=C53GHCXL4

The mac installer installs jdk17 into `/Library/Java/JavaVirtualMachines/temurin-17.jdk` while the build scripts anticipate `/Library/Java/JavaVirtualMachines/adoptopenjdk-*.jdk`, so I've made a special case for when the bootjdk is 17.

Im certain I have made the amendment in the correct place in the script, but please let me know if more needs to be added

### DIFF
--- a/build-farm/platform-specific-configurations/mac.sh
+++ b/build-farm/platform-specific-configurations/mac.sh
@@ -96,8 +96,8 @@ if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
     if [ -x "/Library/Java/JavaVirtualMachines/adoptopenjdk-${JDK_BOOT_VERSION}/Contents/Home/bin/javac" ]; then
       echo "Could not use ${BOOT_JDK_VARIABLE} - using /Library/Java/JavaVirtualMachines/adoptopenjdk-${JDK_BOOT_VERSION}/Contents/Home"
       export "${BOOT_JDK_VARIABLE}"="/Library/Java/JavaVirtualMachines/adoptopenjdk-${JDK_BOOT_VERSION}/Contents/Home"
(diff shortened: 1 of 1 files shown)


sxa (approved):
Yep I think that'll work for now :-) Once the checks are done we can merge and run another JDK18 build on mac to confirm.

We will need to change this since if you run the playbook on a new machine it'll presumably use the temurin diretories for 8 and 11 and 16 too (Although I think only 16 would be an issue for the builds!) but this will work to fix the immediately build failures that we have ;+1:

build-farm/platform-specific-configurations/mac.sh
```diff
@@ -96,6 +96,8 @@ if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
     if [ -x "/Library/Java/JavaVirtualMachines/adoptopenjdk-${JDK_BOOT_VERSION}/Contents/Home/bin/javac" ]; then
       echo "Could not use ${BOOT_JDK_VARIABLE} - using /Library/Java/JavaVirtualMachines/adoptopenjdk-${JDK_BOOT_VERSION}/Contents/Home"
       export "${BOOT_JDK_VARIABLE}"="/Library/Java/JavaVirtualMachines/adoptopenjdk-${JDK_BOOT_VERSION}/Contents/Home"
+    elif [[ ("$JDK_BOOT_VERSION" -eq 17) && ( -x "/Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home/bin/javac") ]]; then
```
karianna, line 99: perhaps equal or greater than?

karianna (changes requested):

Issue #2611, opened 2021-05-02
Documentation for smoke tests


M-Davies:
**Please describe what you want to be able to do that isn't documented and/or what you'd like to see in the documentation**
The [smoke tests](https://github.com/adoptium/temurin-build/tree/master/test/functional/buildAndPackage) while useful, do not have any obvious use or run-locally documentation dictating where and when the tests are run or why they are necessary.

Personally, while working on #2581 I have found them difficult to read and what the general purpose of them are so I would like a small FAQ entry or addition to [Testing.md](https://github.com/adoptium/temurin-build/blob/master/Testing.md) to explain the above.

@smlambert @sophia-guo Can either of you take a look at this?

sxa:
@sophia-guo @smlambert Do we have any specific documentation on the smoke tests we can point to in order to close this?  I had a quick look around and can't see any, but presumably it wouldn't be too hard (Worst case I can try and do it when I look at the RISC-V failures!)

SehrishHussain:
@smlambert I will wait on this one. Looks like it's going to take a lot of research to work on it?

smlambert:
I would suggest you trying to run smoke tests locally on your machine, but I believe you are running on Windows, which could mean some messy installation of prereqs.

I think you can create the instructions based on the existing ones in the Reproducing-Test-Failures-Locally doc and someone on a machine with prereqs installed can 'try out' the instructions from the doc.  

I would not add smoke test instructions to the ci-jenkins-pipelines FAQ at this time, for now just in this repo where the test material actually lives.

PR #2766, opened 2021-10-16
SmokeTesting.md file

SehrishHussain:
This file contains general instructions to perform Smoke Testing.
addresses issue: #2611


SehrishHussain:
@smlambert Not sure if I had to create a separate PR (for the file I created with general instructions) but here it is.

smlambert:
@SehrishHussain - it appears to be complaining about the 2nd _extended.functional in the sentence...

can you try wrapping it in backticks ` `  
so changing 
...we use _extended.functional and because... 
to
...we use `_extended.functional` and because...
where there are backticks around the word 

Think the linter is wrong here, but it may think you are trying to use markup when you have an _ underscore and no following underscore.

PR #2805, opened 2021-12-16
Add cyclonedx-lib build and --create-sbom option

andrew-m-leonard:
Download the necessary CycloneDX libraries and dependencies, and build a TemurinGenSBOM.java app framework.
Added a new BUILD_ARGS --create-sbom, default is "false", which when set builds the library and invokes TemurinGenSBOM at the end of the build.

**CycloneDX and dependent libraries and licenses:**
cyclonedx-core-java : v5.0.4 : Apache License, Version 2.0
jackson-core : v2.12.4 : Apache License, Version 2.0
jackson-dataformat-xml : v2.12.4 : Apache License, Version 2.0
jackson-databind : v2.12.4 : Apache License, Version 2.0
jackson-annotations : v2.12.4 : Apache License, Version 2.0
json-schema-validator : v1.0.58 : Apache License, Version 2.0
commons-codec : v1.15 : Apache License, Version 2.0
commons-io : v2.11.0 : Apache License, Version 2.0
github-package-url : v1.4.0 : MIT License

TemurinGenSOM is currently just a dummy implementation, that prints the args and creates a dummy test SBOM.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -651,29 +651,15 @@ generateSBoM() {
     exit 2
   fi
 
-  # We need the exitcode from the app 
-  set +eu
-
   # classpath to run CycloneDX java app TemurinGenSBOM
   classpath="${WORKSPACE}/cyclonedx-lib/build/jar/temurin-gen-sbom.jar:${WORKSPACE}/cyclonedx-lib/build/jar/cyclonedx-core-java.jar:${WORKSPACE}/cyclonedx-lib/build/jar/jackson-core.jar:${WORKSPACE}/cyclonedx-lib/build/jar/jackson-dataformat-xml.jar:${WORKSPACE}/cyclonedx-lib/build/jar/jackson-databind.jar:${WORKSPACE}/cyclonedx-lib/build/jar/jackson-annotations.jar:${WORKSPACE}/cyclonedx-lib/build/jar/json-schema.jar:${WORKSPACE}/cyclonedx-lib/build/jar/commons-codec.jar:${WORKSPACE}/cyclonedx-lib/build/jar/commons-io.jar:${WORKSPACE}/cyclonedx-lib/build/jar/github-package-url.jar"
 
   # Run app to generate SBoM
-  # List of commands
-  declare -a cmds=('--create temurin_sbom.json --name "Temurin SBOM" --version "1.2.3" --type "application" --author "Adoptium"'
-                   '--add_component temurin_sbom.json --name "openjdk" --version "jdk-19+12"'
-                  )
-  for cmd in "${cmds[@]}"
-  do
-    "${javaHome}"/bin/java -cp "${classpath}" temurin.sbom.TemurinGenSBOM "${cmd}"
-    exitCode=$?
-    if [ "${exitCode}" -ne 0 ]; then
-      echo "Failed to generate SBOM rc=${exitCode}, exiting"
-      set -eu
-      exit ${exitCode}
-    fi
-  done
 
-  set -eu
+  # Examples.. 
+  "${javaHome}"/bin/java -cp "${classpath}" temurin.sbom.TemurinGenSBOM --create         temurin_sbom.json --name "Temurin SBOM" --version "1.2.3" --type "application" --author "Adoptium"
+  "${javaHome}"/bin/java -cp "${classpath}" temurin.sbom.TemurinGenSBOM --add_component  temurin_sbom.json --name "openjdk" --version "1.0.0" --hash "abcdefg123456"
+  "${javaHome}"/bin/java -cp "${classpath}" temurin.sbom.TemurinGenSBOM --add_dependency temurin_sbom.json --name "gcc" --version "8.5.0"
 }
 
 getGradleJavaHome() {
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -606,6 +606,76 @@ createOpenJDKFailureLogsArchive() {
     createArchive "${adoptLogArchiveDir}" "${makeFailureLogsName}"
 }
 
+# Build the CycloneDX Java library and app used for SBoM generation
+buildCyclonedxLib() {
+  local javaHome=""
+
+  if [ ${JAVA_HOME+x} ] && [ -d "${JAVA_HOME}" ]; then
+    javaHome=${JAVA_HOME}
+  elif [ ${JDK8_BOOT_DIR+x} ] && [ -d "${JDK8_BOOT_DIR}" ]; then
+    javaHome=${JDK8_BOOT_DIR}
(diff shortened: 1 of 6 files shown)


andrew-m-leonard:
run with --create-sbom: https://ci.adoptopenjdk.net/view/work-in-progress/job/andrew-cyclonedx/27/console
```
20:45:18  TemurinGenSBOM: --create temurin_sbom.json --name "Temurin SBOM" --version "1.2.3" --type "application" --author "Adoptium"
20:45:19  SBOM: {
20:45:19    "bomFormat" : "CycloneDX",
20:45:19    "specVersion" : "1.3",
20:45:19    "version" : 1,
20:45:19    "components" : [
20:45:19      {
20:45:19        "author" : "Adoptium",
20:45:19        "name" : "TestComponent",
20:45:19        "version" : "1.0.0",
20:45:19        "type" : "application"
20:45:19      }
20:45:19    ]
20:45:19  }
```

andrew-m-leonard:
run tests

smlambert (approved):
LGTM (posed a question, but its not meant as a blocking question?)

andrew-m-leonard:
@smlambert thanks. Did you submit the review comment? I can't see the question?

sbin/build.sh
```diff
@@ -606,6 +606,76 @@ createOpenJDKFailureLogsArchive() {
...
+
+  # We need the exitcode from the app 
+  set +eu
+
+  # classpath to run CycloneDX java app TemurinGenSBOM
+  classpath="${WORKSPACE}/cyclonedx-lib/build/jar/temurin-gen-sbom.jar:${WORKSPACE}/cyclonedx-lib/build/jar/cyclonedx-core-java.jar:${WORKSPACE}/cyclonedx-lib/build/jar/jackson-core.jar:${WORKSPACE}/cyclonedx-lib/build/jar/jackson-dataformat-xml.jar:${WORKSPACE}/cyclonedx-lib/build/jar/jackson-databind.jar:${WORKSPACE}/cyclonedx-lib/build/jar/jackson-annotations.jar:${WORKSPACE}/cyclonedx-lib/build/jar/json-schema.jar:${WORKSPACE}/cyclonedx-lib/build/jar/commons-codec.jar:${WORKSPACE}/cyclonedx-lib/build/jar/commons-io.jar:${WORKSPACE}/cyclonedx-lib/build/jar/github-package-url.jar"
+
+  # Run app to generate SBoM
+  # List of commands
+  declare -a cmds=('--create temurin_sbom.json --name "Temurin SBOM" --version "1.2.3" --type "application" --author "Adoptium"'
```
sxa, line 662: I'd be tempted to unroll this loop since the use of an array just makes the flow harder to follow. Since we only have two operations I think it'd be simpler to just run them both, and give separate error messages for each section to make it clear which one failed e.g.
```
"${javaHome}"/bin/java -cp "${classpath}" temurin.sbom.TemurinGenSBOM -create temurin_sbom.json --name "Temurin SBOM" --version "1.2.3" --type "application" --author "Adoptium"
RC=$?
if [ "${RC}" -ne 0 ]; then
  echo Failed to ru TemurinGenSBOM. rc=${RC}, exiting
  exit $RC
fi

"${javaHome}"/bin/java -cp "${classpath}" --add_component temurin_sbom.json --name "openjdk" --version "jdk-19+12"
RC=$?
if [ "${RC}" -ne 0 ]; then
  echo Failed to add componenent temurin_sbom.json - rc=${existCode}, exiting
  exit $RC
fi
```
Also, can we avoid hardcoding `jdk-19+12`? - We should have access to the openjdk version number from within this script.

sbin/build.sh
```diff
@@ -606,6 +606,76 @@ createOpenJDKFailureLogsArchive() {
...
+    javaHome=${JDK8_BOOT_DIR}
+  elif [ ${JDK11_BOOT_DIR+x} ] && [ -d "${JDK11_BOOT_DIR}" ]; then
+    javaHome=${JDK11_BOOT_DIR}
+  else
+    echo "Unable to find a suitable JAVA_HOME to build the cyclonedx-lib"
+    exit 2
+  fi
+
+  # We need the exitcode from ant
+  set +eu
```
sxa, line 625: Is there any significant advantage to doing this and producing a custom error rather than letting it exit naturally? It would save quite a few lines of code.

sbin/build.sh
```diff
@@ -606,6 +606,76 @@ createOpenJDKFailureLogsArchive() {
...
+    javaHome=${JDK8_BOOT_DIR}
+  elif [ ${JDK11_BOOT_DIR+x} ] && [ -d "${JDK11_BOOT_DIR}" ]; then
+    javaHome=${JDK11_BOOT_DIR}
+  else
+    echo "Unable to find a suitable JAVA_HOME to run the TemurinGenSBOM app"
+    exit 2
+  fi
+
+  # We need the exitcode from the app 
+  set +eu
```
sxa, line 655: Similar to earlier coment, do we gain much by trapping this condition? I suppose you can capture the return code, although it's perhaps debatable how useful that is since ant will generally give more useful info in the failure message.

sbin/build.sh
andrew-m-leonard: The array of cmds is just an example...made up!
The TemurinGenSBOM scaffold just prints it out at the moment
In reality there will be probably be a lot of "operations" to generate SBOM components/dependencies/hashes/etc... hence lots of calls to the SBOM gen, hence I thought array of cmds better, but design TBD... @SehrishHussain 
i'll remove for now

sbin/build.sh
andrew-m-leonard: probably not

sbin/build.sh
andrew-m-leonard: changed, does look neater without loop

sbin/build.sh
andrew-m-leonard: removed

sbin/build.sh
andrew-m-leonard: agree removed

sxa (approved):
Thanks - as you said we can revisit if necessary in the future, but I think it is cleaner this way

smlambert:
Hmm, thought I submitted it, but it was a question on how generic can the class be... 
Whether it could be used by those who use our builds scripts as currently nothing in the class would need to be specific to Temurin (are passed in as cmds arg for specific Temurin SBOM) in which case is the Temurin specific name needed?

andrew-m-leonard:
> Hmm, thought I submitted it, but it was a question on how generic can the class be... Whether it could be used by those who use our builds scripts as currently nothing in the class would need to be specific to Temurin (are passed in as cmds arg for specific Temurin SBOM) in which case is the Temurin specific name needed?

@smlambert ah yes, I did wonder about the "name" myself for that reason... kept Temurin in the name as it's in a repo called temurin-build ...

Issue #2671, opened 2021-07-02
PROPOSAL: Use "temurin" as an alternate VARIANT option from hotspot


sxa:
Putting the feelers out for this one. Thoughts in here sprung out as part of the changes proposed in https://github.com/adoptium/temurin-build/pull/2670/files#diff-bf1a71b6b982bb100bd797f1aac5049b232b43398baf9494c8f2823c62c8aa82

### Reasons for doing it:

- It would make it simpler for third parties use our scripts to build alternate variants. At the moment they would have to use the disable-adopt-branch-safety option if they don't have Adopt/Termurin-proprietory things like our marker files, which it would be nice not to need for end-users trying our scripts.
- We could default the `hotspot` VARIANT to build from the upstream skara repositories
- At the moment anyone who uses our scripts to build from our repositories will produce a binary that calls itself `Temurin` in the version output and other vendor properties. This is something we should discourage to avoid confusion
- If we have other Adoptium project members who wish to support our binaries we should make it easy to use our build scripts to build and ship with the minimal amount of adjusting of parameters or potential trademark infringements, so having a "default HotSpot" option without any of the temurin-specifics would assist with that - it would come out as totally vanilla openjdk unless explicitly overriden - and we could provide good guidance on the 
- It would likely make building alternates like Loom, Panama with our scripts etc. a bit simpler

### Reasons against doing it:

- It's an extra codepath that should be maintained, although we can reasonably easily run it periodically to check, potentially just using `VagrantPlaybookCheck`

jerboaa:
A huge +1 from me. It would allow folks to use the scripts to produce vanilla OpenJDK builds fairly easily and that would be very beneficial. I could see myself help testing those code branches on a regular basis.

karianna:
Are we conflating the VM variant with the vendor here?  I'm pretty sure we had some old/long-running issues stating that we needed to add vendor and allow folks to differentiate on that.

sxa:
> A huge +1 from me. It would allow folks to use the scripts to produce vanilla OpenJDK builds fairly easily and that would be very beneficial. I could see myself help testing those code branches on a regular basis.

I remember you hitting some problems a few months ago and needed some of the additional options in such a situation :-)

sxa:
> Are we conflating the VM variant with the vendor here? I'm pretty sure we had some old/long-running issues stating that we needed to add vendor and allow folks to differentiate on that.

[...]

PR #2818, opened 2021-12-31
Create "temurin" variant distinct from "hotspot"

sxa:
Fixes https://github.com/adoptium/temurin-build/issues/2671

This is a stop-gap to using a better split of VARIANT/VENDOR that will let us test this in principle using the current way we use VARIANT, while allowing the default run of the build scripts to NOT include any Eclipse or Temurin strings which is better for anyone running the scripts themselves. It will also make it simpler to allow users to switch to an alternate openjdk repository that doesn't have out markers in them. Doing this as a stop-gap means we can work on the changes required with minimal risk of destabilising anything else

Will need further testing (possibly a comparison from @andrew-m-leonard to ensure that they are binary identical!) and a corresponding update to the pipelines repo for `s/hotspot/temurin/`, the smoke tests, and possibly the testing too (Also, what about the pipeline generators?)

WIP/Draft as I'm not going to be around for much of next week but putting this up for review and to see if anyone sees may be more that will be required :-)


andrew-m-leonard:
@sxa looks ok as a wip, just a couple of things I presume are draft, ie.the with-verion-pre-opt default and ignoring NUM_PROCESSORS. I suspect doing test builds with this will be needed to trial it...

sxa:
> @sxa looks ok as a wip, just a couple of things I presume are draft, ie.the with-verion-pre-opt default and ignoring NUM_PROCESSORS. I suspect doing test builds with this will be needed to trial it...

Yep those were just to force particular configurations on the system I was experimenting on.

sxa:
run tests

andrew-m-leonard:
run tests

sxa:
https://ci.adoptopenjdk.net/view/Failing%20Builds/job/build-scripts/job/jobs/job/jdk17u/job/jdk17u-linux-x64-hotspot/58/console is a `hotspot` build and produces this:
```
16:19:44  =JAVA VERSION OUTPUT=
16:19:44  openjdk version "17.0.1-beta" 2021-10-19
16:19:44  OpenJDK Runtime Environment (build 17.0.1-beta+12-202201061534)
16:19:44  OpenJDK 64-Bit Server VM (build 17.0.1-beta+12-202201061534, mixed mode, sharing)
16:19:44  =/JAVA VERSION OUTPUT=
```
Metadata for the above is: 
[metadata.hotspot.58.txt.gz](https://github.com/adoptium/temurin-build/files/7823377/metadata.hotspot.58.txt.gz)

59 is `temurin` and has:
```
16:20:00  =JAVA VERSION OUTPUT=
16:20:00  openjdk version "17.0.1-beta" 2021-10-19
16:20:00  OpenJDK Runtime Environment Temurin-17.0.1+12-202201061534 (build 17.0.1-beta+12-202201061534)
16:20:00  OpenJDK 64-Bit Server VM Temurin-17.0.1+12-202201061534 (build 17.0.1-beta+12-202201061534, mixed mode, sharing)
16:20:00  =/JAVA VERSION OUTPUT=
16:20:00  ===GENERATING RELEASE FILE===
```
Metadata is 
[metadata.temurin.59.txt.gz](https://github.com/adoptium/temurin-build/files/7823379/metadata.temurin.59.txt.gz)

Output and metadata look reasonable.

sxa:
Couple of notes:
- At the moment the code is producing a filename with `_temurin_` in it instead of `_hotspot_` - we could special case that and just use the output filename option to override that if we want to keep the same download name, however we currently don't have anything explicitly saying "Temurin" in the output.
- The pipelines are currently saying `Unrecognised build variant: temurin` after the smoke tests which is stopping the pipeline from starting the test runs. I've added in some changes for the `ci-jenkins-pipelines` repository in https://github.com/adoptium/ci-jenkins-pipelines/pull/226

Currently running tests at https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk17u/job/jdk17u-linux-x64-hotspot/62/ with the new pipeline changes to see if that resolves the last one. Also running JDK8 [arm32 hotspot](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-linux-arm-hotspot/809/) and [arm32 temurin](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-linux-arm-hotspot/811/) from the new code since those are from different repositories and I'm not yet sure what will happen ;-)

sxa:
Linter failures are covered under https://github.com/adoptium/temurin-build/issues/2831

gdams (approved):

configureBuild.sh
```diff
@@ -171,10 +171,14 @@ setRepository() {
     suffix="alibaba/dragonwell${BUILD_CONFIG[OPENJDK_CORE_VERSION]/jdk/}"
   elif [[ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_BISHENG}" ]]; then
     suffix="openeuler-mirror/bishengjdk-${BUILD_CONFIG[OPENJDK_CORE_VERSION]:3}"
-  elif [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK8_CORE_VERSION}" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "armv7l" ]; then
+  elif [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK8_CORE_VERSION}" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "armv7l" ] && [[ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_TEMURIN}" ]]; then
     suffix="adoptium/aarch32-jdk8u";
-  else
```
karianna, line 176: Is that arm source tree split still required?  I thought we had that down to the one extra repo now?

configureBuild.sh
```diff
@@ -255,6 +259,7 @@ processArgumentsforSpecificArchitectures() {
     fi
     if [[ ${BUILD_CONFIG[USER_SUPPLIED_CONFIGURE_ARGS]:-""} != *"--with-jobs"* ]]; then
       BUILD_CONFIG[USER_SUPPLIED_CONFIGURE_ARGS]="--with-jobs=${BUILD_CONFIG[NUM_PROCESSORS]} ${BUILD_CONFIG[USER_SUPPLIED_CONFIGURE_ARGS]:-''}"
+      BUILD_CONFIG[USER_SUPPLIED_CONFIGURE_ARGS]="--with-jobs=8 ${BUILD_CONFIG[USER_SUPPLIED_CONFIGURE_ARGS]:-''}"
```
karianna, line 262: why 8?

build-farm/make-adopt-build-farm.sh
```diff
@@ -59,7 +59,14 @@ fi
 
 [ -z "$JAVA_TO_BUILD" ] && echo JAVA_TO_BUILD not defined - set to e.g. jdk8u
 [ -z "$VARIANT"       ] && echo VARIANT not defined - assuming hotspot && export VARIANT=hotspot
-[ -z "$FILENAME"      ] && echo FILENAME not defined - assuming "${JAVA_TO_BUILD}-${VARIANT}.tar.gz" && export FILENAME="${JAVA_TO_BUILD}-${VARIANT}.tar.gz"
+if [ -z "$FILENAME"   ]; then
+  if [ "${VARIANT}" = "temurin" ]; then
+     # I don't like this - perhaps we should override elsewhere to keep consistency with existing release names
+     echo FILENAME not defined - assuming "${JAVA_TO_BUILD}-hotspot.tar.gz" && export FILENAME="${JAVA_TO_BUILD}-hotspot.tar.gz"
```
karianna, line 65: I wonder if we need to keep that naming scheme?  I guess so to be backward compatible?

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -91,7 +91,7 @@ then
   # cores on our scaleway systems while they are still in use but still
   # allow more on larger machines. Can be revisited post-Scaleway
   if [ "$(lscpu|awk '/^CPU\(s\)/{print$2}')" = "4" ]; then
-    export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-jobs=4 --with-memory-size=2000"
+    export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-jobs=5 --with-memory-size=2000"
   fi
```
karianna, line 95: Why 5?

sbin/build.sh
```diff
@@ -241,9 +246,11 @@ configureVersionStringParameter() {
   if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" != 8 ]; then
     addConfigureArg "--with-vendor-name=" "\"${BUILD_CONFIG[VENDOR]}\""
   fi
-  addConfigureArg "--with-vendor-url=" "${BUILD_CONFIG[VENDOR_URL]:-"https://adoptium.net/"}"
-  addConfigureArg "--with-vendor-bug-url=" "${BUILD_CONFIG[VENDOR_BUG_URL]:-"https://github.com/adoptium/adoptium-support/issues"}"
-  addConfigureArg "--with-vendor-vm-bug-url=" "${BUILD_CONFIG[VENDOR_VM_BUG_URL]:-"https://github.com/adoptium/adoptium-support/issues"}"
+  addConfigureArg "--with-vendor-url=" "${BUILD_CONFIG[VENDOR_URL]:-""}"
+  # This looks silly, but in the case where someone is building a plain hotspot
+  # This makes it a bit easier to find the code that sets it to override
+  addConfigureArg "--with-vendor-bug-url=" "${BUILD_CONFIG[VENDOR_BUG_URL]:-"file:///dev/null"}"
```
karianna, line 252: Perhaps could even add a sample "Please replace with your own Vendor URL e.g. ....."?

karianna (changes requested):

configureBuild.sh
sxa: Almost certainly true, although perhaps it's worth leaving it in until the next release just in case we hit a problem with the thing's we've GAd ... Should probably be removed as a separate PR to this one though, in order to avoid any confusion.

sbin/build.sh
sxa: My only concern what that is that ideally we wouldn't want people to modify the scripts but use a parameter/variable. but I'll add a comment in for now.

configureBuild.sh
sxa: I think that was in there because I'd hard coded for the machine I was testing on for maximum throughput. Removed.

build-farm/make-adopt-build-farm.sh
sxa: Yeah I'm going for "minimal viable product" at the moment - get the changes in with minmal risk of screwing anything up ... Which is still quite risky!

sxa: (Also bear in mind this is purely a default - normally it will be defined in our jenkins processes)

andrew-m-leonard (approved):

PR #2819, opened 2022-01-06
Enable jdk-19+ reproducible build options

andrew-m-leonard:
This PR enables "reproducible build" support for jdk-19+ builds by enabling:
- --with-source-date option enables openjdk make reproducible build option
- --with-build-user=adoptium ensure determinstic jvm build user
- --with-cacerts option to pass the mozilla cacerts folder into the openjdk build keystore tool which is determinsitic

### DIFF
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -132,6 +132,11 @@ then
     # OpenJ9 only supports building jdk-11 with jdk-11
     JDK_BOOT_VERSION="11"
   fi
+  if [ "${JAVA_FEATURE_VERSION}" == "19" ]; then
+    # To support reproducible-builds the jar/jmod --date option is required
+    # which is only available from jdk-19
+    JDK_BOOT_VERSION="19"
+  fi
 fi
 echo "Required boot JDK version: ${JDK_BOOT_VERSION}"
 
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -134,7 +134,7 @@ then
   fi
   if [ "${JAVA_FEATURE_VERSION}" == "19" ]; then
     # To support reproducible-builds the jar/jmod --date option is required
-    # which is only available from jdk-19
+    # which is only available from jdk-19 so we cannot bootstrap with JDK18
     JDK_BOOT_VERSION="19"
   fi
 fi
--- a/build-farm/set-platform-specific-configurations.sh
+++ b/build-farm/set-platform-specific-configurations.sh
@@ -24,6 +24,23 @@ then
     export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --disable-warnings-as-errors"
 fi
 
+# jdk-19 and above support reproducible builds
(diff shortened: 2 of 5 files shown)


andrew-m-leonard:
run tests

build-farm/make-adopt-build-farm.sh
```diff
@@ -132,6 +132,11 @@ then
     # OpenJ9 only supports building jdk-11 with jdk-11
     JDK_BOOT_VERSION="11"
   fi
+  if [ "${JAVA_FEATURE_VERSION}" == "19" ]; then
```
karianna, line 135: =>?

build-farm/make-adopt-build-farm.sh
```diff
@@ -132,6 +132,11 @@ then
...
+    # To support reproducible-builds the jar/jmod --date option is required
+    # which is only available from jdk-19
+    JDK_BOOT_VERSION="19"
```
karianna, line 138: Maybe future proof this as well?

karianna (changes requested):

build-farm/make-adopt-build-farm.sh
andrew-m-leonard: It's only building jdk-19 that we can't use N-1 as jdk-18 doesn't have the reproducible build options for jar&jmod.
Building jdk-20+ is fine, as jdk-19 does.

build-farm/make-adopt-build-farm.sh
andrew-m-leonard: It's only building jdk-19 that we can't use N-1 as jdk-18 doesn't have the reproducible build options for jar&jmod.
Building jdk-20+ is fine, as jdk-19 does.

karianna (approved):

andrew-m-leonard:
Alpine Linux failing to build

andrew-m-leonard:
Fails on Alpine Linux because it uses BusyBox "date" tool, but new openjdk reproducible build logic only detects GNU or BSD. I will fix upstream

build-farm/make-adopt-build-farm.sh
```diff
@@ -132,6 +132,11 @@ then
     # OpenJ9 only supports building jdk-11 with jdk-11
     JDK_BOOT_VERSION="11"
   fi
+  if [ "${JAVA_FEATURE_VERSION}" == "19" ]; then
+    # To support reproducible-builds the jar/jmod --date option is required
+    # which is only available from jdk-19
```
sxa, line 137: Worth adding that clarification into the code I think:

```suggestion
    # which is only available from jdk-19 so we cannot bootstrap with JDK18
```

sxa (commented):
To be clear, are the keystore / cacerts changes in here specific to reproducible builds, or something that's required for other changes that are in JDK19+

Also - it would be good (for historic record) to have a description on this PR summarising the changes.

andrew-m-leonard:
> To be clear, are the keystore / cacerts changes in here specific to reproducible builds, or something that's required for other changes that are in JDK19+
> 
@sxa they are required for reproducible builds as using keytool is not deterministic, so using GenerateCacerts jdk buil tool

andrew-m-leonard:
@sxa updated: https://github.com/adoptium/temurin-build/pull/2819#issue-815581840

sxa (approved):
Based on the responses LGTM

PR #2077, opened 2020-09-11
Add JVM Version, Vendor, and Build Source to Json Metadata

austin0:
Removes the `alpha: do not use` warnings and adds the following information to the Json metadata file:
- JVM Version (i.e. `openj9-0.22.0`)
- Vendor (i.e. `AdoptOpenJDK`)
- Build Source (i.e. `openjdk-build/b2a02bb`)

Also creates a `metadata` folder in the target_dir to contain the metadata files as future additions may cause the artifact list to become cluttered.

Resulting Json:
```
    "vendor": "AdoptOpenJDK",
    "os": "mac",
    "arch": "x64",
    "variant": "openj9",
    "jvm_version": "openj9-0.22.0",
    "version": {
        "minor": 0,
        "security": 0,
        "pre": null,
        "adopt_build_number": 0,
        "major": 15,
        "version": "15+29-202007070926",
        "semver": "15.0.0+29.0.202007070926",
        "build": 29,
        "opt": "202007070926"
    },
    "scmRef": "<output of git describe OR buildConfig.SCM_REF>",
    "buildRef": "<build-repo-name/build-commit-sha>",
    "version_data": "jdk15",
    "binary_type": "debugimage",
    "sha256": "<shasum>",
    "full_version_output": <output of java --version>,
    "configure_arguments": <output of bash configure>
```
Signed-off-by: Austin Bailey <Austin.Bailey@ibm.com>

PR #2832, opened 2022-01-20
Remove check that moves minor to tags in variant JSON

AdamBrousseau:
This "precaution" `if` has been there since it was authored.
I can't understand why it is there. The code seems to work
without it. It seems to unintentionally be moving those
values for every non-milestone release. That is, the code
works for milestones only but for releases it moves minor
to tags.

eg. 0.29.0-m1 becomes 0, 29, 0, m1 which is correct
but 0.29.0 becomes 0, 0, 0, 29 which is incorrect

Related #2077

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -1558,10 +1558,7 @@ addVariantVersionToJson(){
     local minor=$(echo "$variantJson" | awk -F[.] '{print $2}')
     local security=$(echo "$variantJson" | awk -F[.] '{print $3}')
     local tags=$(echo "$variantJson" | awk -F[.] '{print $4}')
-    if [[ $(echo "$variantJson" | tr -cd '.' | wc -c) -lt 3 ]]; then # Precaution for when OpenJ9 releases a 1.0.0 version
-      tags="$minor"
-      minor=""
-    fi
+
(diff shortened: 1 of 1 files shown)


karianna (approved):
LGTM - I trust the OpenJ9 analysis here :-)

AdamBrousseau:
It's also not clear to me from the original PR why this was added or what uses it. So I don't know if something will break with this change. I don't seem to find anything in the API related to `variant_version` other than some test code.

AdamBrousseau:
It would be nice to hear from original author @austin0 but I don't think he's active on the projects at the moment. If anyone else could shed some light it would be appreciated. Maybe if @johnoliver could confirm that the API doesn't use it, that would help my confidence.

austin0:
I can't figure out what I was trying to prevent there but it definitely seems like I missed that `0.29.0` and `1.0.0` would be handled the same way, looks like it's just a mistake on my behalf and can be removed.

AdamBrousseau:
Thanks for the response @austin0 
For the record, J9_TAG starts out as something like this
openj9-0.29.0 or this openj9-0.29.0-m1
I believe Adopt is in a code freeze for the release. I don't believe this will affect Temurin. If you'd rather wait, I can merge on my end as I need this for the release.

AdamBrousseau:
Is the code freeze from Jan lifted yet?

karianna:
Yes.

karianna:
Ping the build channel to get another reviewer would be my recommendation :-)

PR #2840, opened 2022-01-30
Initial prototype of Temurin SBOM metadata using CycloneDX core java

SehrishHussain:
Updated TemurinGenSBOM.java


SehrishHussain:
@andrew-m-leonard this is the PR from yesterday.

andrew-m-leonard:
The mysbom.json example I suggest just adding as a comment to this PR and removing from the commit.

andrew-m-leonard:
The "Linter" CI check is failing with various "formatting" and style issues: https://github.com/adoptium/temurin-build/runs/5004923153?check_suite_focus=true
It's possible your Eclipse IDE is formatting or inserting "tabs" causing issues, so you may want to check your Eclipse project settings for things like what it does with tabs... tabs are not good to use, have just spaces instead.

andrew-m-leonard:
@SehrishHussain a general comment on the Java code, in the "task" methods (eg.addMetadataProp), if you're hard-coding a string, then it either wants to be passed in as a possible variable or the method itself made more specific, eg.addTopLevelBOMMetadata(...)

smlambert:
hi @SehrishHussain - can you update the title of this PR to something a bit more descriptive, maybe something like "Populate Temurin SBOM with some initial build metadata".

SehrishHussain:
> @SehrishHussain so we probably need to remove the .classpath, .project files, and changes to the .gitignore files.

@andrew-m-leonard Should I rename these files to .gitignore? Also please let me know which IDE is preferred? I was using Eclipse and then Intellij IDEA.

andrew-m-leonard:
@SehrishHussain the .gitignore file is a specific file indicating to git what files are "not tracked"

The temurin-build repo has very little Java in it, so no one IDE is necessarily preferred. I in fact just use my favourite editor. But you can certainly keep using Eclipse, which would be useful for this plugin, simply don't add the .<eclipse project files> to your commit.

SehrishHussain:
@andrew-m-leonard all the above-suggested changes are done on the local repo on my machine. Shall I create a new commit with new files? or a new PR?

andrew-m-leonard:
@SehrishHussain ah that's great. Just push your new commit to this branch.

Issue #2831, opened 2022-01-19
GitHub linter checks failing for RELEASING.md


sxa:
Possibly as a result of https://github.com/adoptium/temurin-build/pull/2823 the RELEASING.md guide is failing as the linter is now looking for the following translations

- `html` -> `HTML`
- `website` -> `site`
- `github` - > `GitHub`
- `repo` -> `repository`

We should also modify the template to state that if linters or their configuration is updated, we have a checkbox that states that the existing code has been linted and issues resolved, as this is frustrating for regular committers.

sxa:
Ref https://github.com/adoptium/temurin-build/pull/2851#issuecomment-1039187872 I think we need to look at what it's checking for. I disagree that "site" is a suitably clear replacment for website in all cases.

PR #2845, opened 2022-02-11
Fix commit URL in RELEASING.md which looks like it got changed in an invalid way

sxa:
Fix is obvious - it is currently incorrect and therefore useless.


karianna:
Looks like https://github.com/adoptium/temurin-build/runs/5155472965?check_suite_focus=true has text linter issues, not sure how easy that is to fix/escape?

sxa:
> Looks like https://github.com/adoptium/temurin-build/runs/5155472965?check_suite_focus=true has text linter issues, not sure how easy that is to fix/escape?

I'd rather that was done under a standalone PR, but we should try and ensure that when the linter is upgraded that it is verified to pass so that people don't suddenly get unrelated failures. I raised an issue about this last month. https://github.com/adoptium/temurin-build/issues/2831

karianna (approved):
LGTM

sxa:
Closing as it looks like this got included in https://github.com/adoptium/temurin-build/pull/2905/files

Issue #2353, opened 2021-01-04
dragonwell8 requires jdk8 bootjdk


andrew-m-leonard:
dragonwell8 jdk8 requires a jdk8 bootjdk as it makes use of "::new" operator only found in jdk8+:
```
15:53:41  ## Starting langtools
15:53:42  Compiling 2 files for BUILD_TOOLS
15:53:43  e:\jenkins\tmp\workspace\build\src\langtools\src\share\classes\com\sun\tools\javac\file\ZipFileIndexCache.java:55: error: ')' expected
15:53:43                      .getFieldValue(ZipFileIndexCache.class, "sharedInstance", ZipFileIndexCache::new);
15:53:43                                                                                                 ^
15:53:43  e:\jenkins\tmp\workspace\build\src\langtools\src\share\classes\com\sun\tools\javac\file\ZipFileIndexCache.java:55: error: illegal start of expression
15:53:43                      .getFieldValue(ZipFileIndexCache.class, "sharedInstance", ZipFileIndexCache::new);
15:53:43                                                                                                  ^
```
dragonwell commit: https://github.com/alibaba/dragonwell8_langtools/commit/d09c9011a749833144a112172516274b50e4be5e

sxa:
It requires a Dragonwell8 boot JDK, and the builds should be configured to pick that up - it's a mistake if it isn't ...

andrew-m-leonard:
@sxa thanks. ah, ok it looks like the dragonwell11 builds are using the "dragonwell" label, but jdk8 is not hence it's picking any windows node... ok i'll create a PR to fix

sxa:
I want to fix it so that we don't have to lock it but that may do in the mean time. I'd hope it would fail if it couldn't find the right one though so it's a worry if it doesn't abort

andrew-m-leonard:
Fixed by https://github.com/AdoptOpenJDK/openjdk-build/pull/2354

andrew-m-leonard:
It will now be consistent with jdk11

PR #2853, opened 2022-02-15
When auto-downloading boot JDK, default to hotspot (Don't use temurin in API)

sxa:
I don't believe this will adversely affect anyone any more, so hard coding this in light of the changes made in https://github.com/adoptium/temurin-build/pull/2818 so that the API call does not try to use `temurin`

Alternate solution would be to use a bash specific construct of `${VARIANT/temurin/hotspot}`

(Causes JDK19 to fail to build as it won't be able to download a boot JDK and we don't install that one via the playbooks yet as it's not a GA level)

### DIFF
--- a/build-farm/platform-specific-configurations/alpine-linux.sh
+++ b/build-farm/platform-specific-configurations/alpine-linux.sh
@@ -34,7 +34,7 @@ if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
   if [ ! -d "$bootDir/bin" ]; then
     mkdir -p "$bootDir"
     releaseType="ga"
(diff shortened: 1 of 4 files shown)


sxa:
ping @AdamBrousseau I don't think this will negatively affect you (for production levels we should have them on the machine already, and the APIs won't download the latest Semeru anyway), but if you believe otherwise let me know.

andrew-m-leonard (approved):

karianna (approved):
LGTM

karianna:
@sxa worth checking with Alibaba?

sxa:
> @sxa worth checking with Alibaba?

Dragonwell is not available through the Adopt API so this will not affect them, but it's needed for Temurin JDK19 builds to succeed.

sxa:
NOTE: The Dragonwell boot JDK issue that lead to the previous two comments is referred to in #2353

PR #2855, opened 2022-02-23
Enable OpenJ9 CUDA support if the SDK is installed on aarch64

sxa:
We don't enable this as a matter of course, since our cloud machines do not have the CUDA development kit installed, but for anyone that has the CUDA 10.2 toolkit installed on a device such as the Jetson Nano ([thread](https://forums.developer.nvidia.com/t/cuda-toolkit-for-jetson-nano/122671)), this will enable the CUDA support to be built into OpenJ9.

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -218,15 +218,15 @@ if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]
 then
   # OpenJ9 fetches the latest OpenSSL in their get_source.sh
   export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --with-openssl=fetched"
-
-  if [ "${ARCHITECTURE}" == "ppc64le" ] || [ "${ARCHITECTURE}" == "x64" ]
+  if [ "${ARCHITECTURE}" == "aarch64" ]; then
+    echo PROBABLE JETSON NANO DEVELOPMENT ENVIRONMENT - Enabling CUDA ...
+    CUDA_HOME=/usr/local/cuda-10.2
+  else
+    CUDA_HOME=/usr/local/cuda-9.0
+  fi
+  if [ -f $CUDA_HOME/include/cuda.h ]
   then
-    CUDA_VERSION=9.0
-    CUDA_HOME=/usr/local/cuda-$CUDA_VERSION
-    if [ -f $CUDA_HOME/include/cuda.h ]
-    then
-      export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --enable-cuda --with-cuda=$CUDA_HOME"
-    fi
+    export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --enable-cuda --with-cuda=$CUDA_HOME"
   fi
 fi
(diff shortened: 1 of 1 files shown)


sxa:
This is a year old and seems to work (presumably it hasn't been reviewed because it was in draft while it underwent testing) so it would be good to push it in if no-one objects - tagging @andrew-m-leonard 
@AdamBrousseau @keithc-ca Any objections to enabling this to be picked up in our build scripts if it exists on aarch64? I seem to remember someone else in the J9 team possibly had one of the devices too so it would be good to allow it to be picked up if someone is using our scripts on a Jetson.

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -68,6 +68,9 @@ then
     then
       export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --enable-cuda --with-cuda=$CUDA_HOME"
     fi
+  elif [ "${ARCHITECTURE}" == "aarch64" ] && [ -f /usr/local/cuda-10.2/include/cuda.h ]; then
```
karianna, line 71: maybe extract / pass in version?

karianna (dismissed):

build-farm/platform-specific-configurations/linux.sh
sxa: I'm not sure. I think it might be better to put it in the section with x64/ppc64le and have a switch for 9.0/10.2 in there depending on arch. Bear in mind these scripts are mostly to set the correct ones to ensure consistent builds on our platforms and not intended for end users who can specify whatever they want to `makejdk-any-platform.sh` so having it fixed isn't such a bad thing here.

sxa: > I didn't look, but I would hope there's a way to arrange to use a different version, i.e. the hard-coded path doesn't override an explicitly specified `--with-code=/some/other/location`. Other than that, I have no objections.

There isn't, but that's no different from the other architectures :-)

sxa: Yeah that's what I was thinking with my [earlier comment](https://github.com/adoptium/temurin-build/pull/2855#discussion_r1132529085) too. Yeah that seems reasonable - I wasn't certain whether it would override or not so I guess we're good with that :-)

keithc-ca: `aarch64` needs to be included in the test a few lines up as well.
https://github.com/adoptium/temurin-build/blob/master/build-farm/platform-specific-configurations/linux.sh#L56

sxa: This is going to need a bit of a rebase after https://github.com/adoptium/temurin-build/pull/3350
@keithc-ca I'm not sure there's much point in having the architecture check you refer to in there at all to be honest. Worst case (And the only Linux it'll affect is s390x) it won't detect a cuda directory so won't build in the support.

sxa: OK I've rebased and force-pushed an update which removes that check. Also removes the specific `CUDA_VERSION` and just sets `CUDA_HOME` directly in the two cases. I feel that keeps it simpler but if people disagree LMK.

andrew-m-leonard:
this has been stale for over a year, converting to "draft", if no longer needed please close.

sxa:
I do want this in and working, although it may be superceded by the recent changes in https://github.com/adoptium/temurin-build/pull/3735 ...

keithc-ca:
Now that #3735 is in, this should just add aarch64 to the set of supported architectures for CUDA.

PR #2859, opened 2022-02-28
Add qemu into build process for OpenJ9 RISC-V cross-compiles

sxa:
May add this into the playbooks in future, but for now I'm going to download dynamically during the build. Required to execute `constgen`. Should fix https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-riscv64-openj9/

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -285,6 +285,16 @@ if [ "${ARCHITECTURE}" == "riscv64" ] && [ "${NATIVE_API_ARCH}" != "riscv64" ]; 
   if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]; then
     # shellcheck disable=SC2001
     CONFIGURE_ARGS_FOR_ANY_PLATFORM=$(echo "$CONFIGURE_ARGS_FOR_ANY_PLATFORM" | sed "s,with-openssl=[^ ]*,with-openssl=${RISCV_SYSROOT}/usr,g")
+    if ! which qemu-riscv64-static; then
+      # don't download it if we already have it from a previous build
+      if [ ! -x "$WORKSPACE/qemu-riscv64-static" ]; then
+        echo Download qemu-riscv64-static as it is required for the OpenJ9 cross build ...
+        curl https://ci.adoptopenjdk.net/userContent/riscv/qemu-riscv64-static.xz | xz -d > "$WORKSPACE/qemu-riscv64-static" && \
+        chmod 755 "$WORKSPACE/qemu-riscv64-static"
+      fi
+      export PATH="$PATH:$WORKSPACE" && \
+      qemu-riscv64-static --version
+    fi
   fi
 
   if [ "${VARIANT}" == "${BUILD_VARIANT_BISHENG}" ]; then


build-farm/platform-specific-configurations/linux.sh
```diff
@@ -68,6 +68,9 @@ then
     then
       export CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM} --enable-cuda --with-cuda=$CUDA_HOME"
     fi
+  elif [ "${ARCHITECTURE}" == "aarch64" ] && [ -f /usr/local/cuda-10.2/include/cuda.h ]; then
```
karianna, line 71: Is that 10.2 detection too brittle?

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -68,6 +68,9 @@ then
...
+    echo PROBABLE JETSON NANO DEVELOPMENT ENVIRONMENT - Enabling CUDA ...
```
karianna, line 72: wrap in quotes and probably doesn't need to be in CAPs

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -285,6 +288,16 @@ if [ "${ARCHITECTURE}" == "riscv64" ] && [ "${NATIVE_API_ARCH}" != "riscv64" ];
   if [ "${VARIANT}" == "${BUILD_VARIANT_OPENJ9}" ]; then
     # shellcheck disable=SC2001
     CONFIGURE_ARGS_FOR_ANY_PLATFORM=$(echo "$CONFIGURE_ARGS_FOR_ANY_PLATFORM" | sed "s,with-openssl=[^ ]*,with-openssl=${RISCV_SYSROOT}/usr,g")
+    if ! which qemu-riscv64-static; then
```
karianna, line 291: I assume the jenkins user will see that executable in its PATH

karianna (changes requested):

build-farm/platform-specific-configurations/linux.sh
sxa: Possibly, although I installed 10.2 on my machine pre-lockdown and as far as I can see it's still that version ;-)

sxa:
I'll rebase - this PR shouldn't have the CUDA stuff in it ... I accidentally based these changes off the wrong branch

build-farm/platform-specific-configurations/linux.sh
sxa: Not sure what your question is here. I put that line in to check if there was already a suitable executable installed on the machine somewhere, in which case that one can be used instead of the downloaded one. A likely precursor to adding something into the playbooks later as part of the default setup.

sxa:
Fixes https://github.com/adoptium/infrastructure/issues/2368

build-farm/platform-specific-configurations/linux.sh
sxa: (This was part of a different PR and should not have been included in this one - resolving)

sxa:
@karianna Are you ok with the above explanation? This is to resolve a current build break so would be good to get it resolved ASAP.

karianna (approved):

andrew-m-leonard (approved):

sxa:
Merged and testing at https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-riscv64-openj9/500/console

PR #2867, opened 2022-03-03
Update: CycloneDX SBOM (Added Component Property name-value pairs)

SehrishHussain:
Adds name-value pairs to Component.


cyclonedx-lib/src/temurin/sbom/TemurinGenSBOM.java
```diff
@@ -178,16 +180,20 @@ static Bom addComponent(final String compName, final String name, final String v
...
-        List<Property> prop = new ArrayList<>();
-        Component comp = new Component();
-        prop1.setName(name);
-        prop1.setValue(value);
-        prop.add(prop1);
-        comp.setProperties(prop);
-        bom.addProperty(prop1);
+        List<Component> componentArrayList = bom.getComponents();
+        for (Component item : componentArrayList) {
+            if (Objects.equals(item.getName(), compName)) {
```
andrew-m-leonard, line 187: Use of Objects.equals() is probably unnecessary here. Typically Objects.equals() is used where an object reference can be a valid null value, as this will return true for null == null.
This can simply be:
```
item.getName().equals(compName)
```

cyclonedx-lib/src/temurin/sbom/TemurinGenSBOM.java
```diff
@@ -178,16 +180,20 @@ static Bom addComponent(final String compName, final String name, final String v
...
+                    List<Property> propertiesList = item.getProperties();
+                    ArrayList<Property> newPropertyList = new ArrayList<>(propertiesList);
+                    Property prop1 = new Property();
+                    prop1.setName(name);
+                    prop1.setValue(value);
+                    newPropertyList.add(prop1);
+                    item.setProperties(newPropertyList);
```
andrew-m-leonard, line 194: Instead of getting the propertyList then adding to it, simply use the component.addProperty(Property) method, as you will actually be replacing the existing property list in the Component rather than just adding to it.
ie:
```
item.addProperty(prop1);
```

SehrishHussain: @andrew-m-leonard we have to make a list specifically to add more name-value pairs. I tried using `item.addProperty(prop1)` before. It doesn't add to the component property list.

andrew-m-leonard: @SehrishHussain the issue with addProperty() has been fixed in cyclonedx v7.1.0. So lets move up to that latest release..

cyclonedx-lib/src/temurin/sbom/TemurinGenSBOM.java
```diff
@@ -166,9 +166,9 @@ static Bom addMetadataProperty(final String name, final String value, final Stri
         return bom;
     }
 
-    static Bom addComponent(final String compName, final String name, final String value, final String description, final String fileName) {                    // Method to store Component --> name
+    static Bom addComponent(final String compName, final String name, final String value, final String description, final String fileName) {      // Method to store Component --> name & single name-value pair
         Bom bom = readJSONfile(fileName);
-    	Component comp = new Component();
+        Component comp = new Component();
```
andrew-m-leonard, line 171: I think from your comments, we need top remove the "prop2" setting in addComponent ?
At the moment the example in build.xml for --addComponent adds an empty property.

cyclonedx-lib/src/temurin/sbom/TemurinGenSBOM.java
```diff
@@ -154,40 +154,25 @@ static Bom addMetadata(final String name, final String fileName) {
         bom.setMetadata(meta);
         return bom;
     }
-    static Bom addMetadataProperty(final String name, final String value, final String fileName) {     // Method to store metadata --> Properties List --> name-values
```
andrew-m-leonard, line 157: We still need addMetadataProperty() for the top level BOM properties.

andrew-m-leonard (approved):
Looks good thank you @SehrishHussain

smlambert (approved):
LGTM (seems cyclonedx-core-java is evolving quickly)!

Issue #2849, opened 2022-02-14
Platform variable detection on MacOS in make-adopt-build-farm.sh does not work


sxa:
**What are you trying to do?** Run make-adopt-build.farm.sh on macos to build JDK17

**Expected behaviour:** Build runs

**Observed behaviour:** Build does not run for two reasons. (1) It's looking for `darwin.sh` instead of `mac.sh` by default (`TARGET_OS` is set to the result of `uname` by defualt I believe) and the boot JDK download didn't seem to work either

**Any other comments:** Should be fairly easy to fix ...

PR #2874, opened 2022-03-10
Auto-detect TARGET_OS for mac in make-adopt-build-farm.sh

sxa:
mac was not being auto-detected correctly. Also wrong variable was used for the solaris detection, so I'm fixing that here too.

### DIFF
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -39,8 +39,9 @@ fi
 ## so needs to be special cased - on everthing else "uname" is valid
 if [ -z "$TARGET_OS" ]; then
   TARGET_OS=$(uname)
-  if [ "$OSTYPE" = "cygwin" ]; then TARGET_OS=windows ; fi
-  if [ "$OSTYPE" = "SunOS"  ]; then TARGET_OS=solaris ; fi
+  if [ "$OSTYPE"    = "cygwin" ]; then TARGET_OS=windows ; fi
+  if [ "$TARGET_OS" = "SunOS"  ]; then TARGET_OS=solaris ; fi
+  if [ "$TARGET_OS" = "Darwin" ]; then TARGET_OS=mac     ; fi
   echo TARGET_OS not defined - assuming you want "$TARGET_OS"
   export TARGET_OS
 fi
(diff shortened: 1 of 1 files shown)


build-farm/make-adopt-build-farm.sh
```diff
@@ -39,8 +39,9 @@ fi
 ## so needs to be special cased - on everthing else "uname" is valid
 if [ -z "$TARGET_OS" ]; then
   TARGET_OS=$(uname)
-  if [ "$OSTYPE" = "cygwin" ]; then TARGET_OS=windows ; fi
-  if [ "$OSTYPE" = "SunOS"  ]; then TARGET_OS=solaris ; fi
+  if [ "$OSTYPE"    = "cygwin" ]; then TARGET_OS=windows ; fi
+  if [ "$TARGET_OS" = "SunOS"  ]; then TARGET_OS=solaris ; fi
```
karianna, line 43: This SunOs logic changed.  Was that deliberate?

karianna (changes requested):

build-farm/make-adopt-build-farm.sh
sxa: Yeah I considered doing that in a separate PR but (mostly becuase hardly anyone would care about Solaris) I put it in here. Happy to split it out again and get a free bonus PR out of it.

`$OSTYPE` won't be set to `SunOS` by default on Solaris so this needs to be using the output from `uname` - stored in `$TARGET_OS` - defined a few lines above to have the desired effect.

sxa: (Also note I did mention that I'd changed solaris in the PR description)

karianna (approved):

sxa:
Extra commit added with the slightly odd looking `ARCHITECTURE` conversion from `arm` to `aarch64` for macos. It looks quite odd since the next line converts `armv7l` to `arm` for 32-bit Linux/arm ...

sxa:
Fixes https://github.com/adoptium/temurin-build/issues/2849

sxa:
@andrew-m-leonard can you approve this one too (You're probably in a good position to check it)

FYI I noticed in https://github.com/adoptium/temurin-build/wiki/Building-OpenJDK-using-temurin-build-scripts-within-the-adopt-build-docker-container you mention setting a few variables like OS and architecture explicitly. In general those are now autodetected so the process is simpler now (Except on macos until this pr is merged ;-) )

andrew-m-leonard (approved):
lgtm

sxa:
check failures are unrelated and have issues elsewhere to deal with them, therefore merging

PR #2876, opened 2022-03-11
Remove xcode switch path commands from build scripts

Haroon-Khel:
Ref https://adoptium.slack.com/archives/C09NW3L2J/p1646743526526249

From the above slack thread, it was determined that a full switch path, `/Applications/Xcode.app`, can be used for JDK11. Since it is already being used for 8 and 17+, It will be made the default switch path on our build machines via the playbooks (pr incoming). Therefore these commands will not be needed in the build scripts


sxa (dismissed):
LGTM But please make sure the default is set correctly on all build machines before this is merged :-)

gdams (dismissed):

karianna (changes requested):
jdk8u build

configure: error: Xcode 6, 9-12 is required to build JDK 8, the version found was 13.2.1. Use --with-xcode-path to specify the location of Xcode or make Xcode active by using xcode-select.

sxa:
> jdk8u build
> 
> configure: error: Xcode 6, 9-12 is required to build JDK 8, the version found was 13.2.1. Use --with-xcode-path to specify the location of Xcode or make Xcode active by using xcode-select.

Where are you seeing that? [George checked and asserted earlier](https://adoptium.slack.com/archives/C09NW3L2J/p1646996110492689?thread_ts=1646743526.526249&cid=C09NW3L2J) that the two versions on the machine should have been the same on the build systems.

gdams:
> (Adjusted comment) @gdams is the github checks running in a completely differnet environment from the machies we have in jenkins in a way that this doesn't work for? Can we pre-select a default compiler in that environment?

Yeah I've just pushed a commit

gdams:
Looks like the JDK11u build is genuinely broken after switching:

```output
checking size of int *... 0
configure: error: Cannot continue.
configure: The tested number of bits in the target (0) differs from the number of bits expected to be found in the target (64)
configure exiting with result code 1
```

Haroon-Khel:
The build did not hit the same error as the github workflow

andrew-m-leonard:
this has been stale for over a year, converting to "draft", if no longer needed please close.

Issue #2882, opened 2022-03-25
Update xlc compiler version for jdk11 builds on AIX


backwaterred:
Due to some recent changes related to the Harfbuzz library version used to build jdk11, building jdk11 will now require xlc version 16.1. This change is expected to remedy the build break first observed [here](https://ci.adoptopenjdk.net/job/AIX-jdk11-dev-build/17/).

Further discussion on this issue and the changes required can be found at the links below:

- https://bugs.openjdk.java.net/browse/JDK-8282588
- https://mail.openjdk.java.net/pipermail/ppc-aix-port-dev/2022-March/004145.html

sxa:
I've managed to get this to build outside the build process but I had to adjust the memory limit below 7000 in `aix.sh` in order for it to build on the "adopt06" machine with XLC16 FP 11. We will now look at deploying that on the main temurin build machines in order to recover the builds.

backwaterred:
Good news. Thanks for your work on this @sxa!

sxa:
JDK11 is now building with XLC16. Closing.

PR #2886, opened 2022-03-31
Update compiler version for jdk11+ on AIX.

backwaterred:
Future builds of jdk11u will require xlc 16.1. This PR updates the version used to build those jdk versions.

https://github.com/adoptium/temurin-build/issues/2882

DCO 1.1 Signed-off-by: Tyler Steele

### DIFF
--- a/build-farm/platform-specific-configurations/aix.sh
+++ b/build-farm/platform-specific-configurations/aix.sh
@@ -118,11 +118,13 @@ fi
 
 if [ "$JAVA_FEATURE_VERSION" -ge 11 ]; then
   export LANG=C
-  if [ "$JAVA_FEATURE_VERSION" -ge 13 ]; then
+  if [ "$JAVA_FEATURE_VERSION" -ne 12 ]; then
+    # for jdk 11 & 13+
     export PATH=/opt/freeware/bin:$JAVA_HOME/bin:/usr/local/bin:/opt/IBM/xlC/16.1.0/bin:/opt/IBM/xlc/16.1.0/bin:$PATH
     export CC=xlclang
     export CXX=xlclang++
   else
+    # for jdk 12
     export PATH=/opt/freeware/bin:$JAVA_HOME/bin:/usr/local/bin:/opt/IBM/xlC/13.1.3/bin:/opt/IBM/xlc/13.1.3/bin:$PATH
   fi
 fi
--- a/build-farm/platform-specific-configurations/aix.sh
+++ b/build-farm/platform-specific-configurations/aix.sh
@@ -118,15 +118,9 @@ fi
 
 if [ "$JAVA_FEATURE_VERSION" -ge 11 ]; then
   export LANG=C
-  if [ "$JAVA_FEATURE_VERSION" -ne 12 ]; then
-    # for jdk 11 & 13+
-    export PATH=/opt/freeware/bin:$JAVA_HOME/bin:/usr/local/bin:/opt/IBM/xlC/16.1.0/bin:/opt/IBM/xlc/16.1.0/bin:$PATH
-    export CC=xlclang
-    export CXX=xlclang++
-  else
-    # for jdk 12
-    export PATH=/opt/freeware/bin:$JAVA_HOME/bin:/usr/local/bin:/opt/IBM/xlC/13.1.3/bin:/opt/IBM/xlc/13.1.3/bin:$PATH
-  fi
+  export PATH=/opt/freeware/bin:$JAVA_HOME/bin:/usr/local/bin:/opt/IBM/xlC/16.1.0/bin:/opt/IBM/xlc/16.1.0/bin:$PATH
+  export CC=xlclang
+  export CXX=xlclang++
 fi
 
 # J9 JDK14 builds seem to be chewing up more RAM than the others, so restrict it


backwaterred:
The automated build failures (on Windows and CentOS) are failing on another PR as well. I expect they are not related. 

The linter did not fail on the other PR, but the failure cause does not appear to be related to this change
```
/usr/bin/docker pull ghcr.io/github/super-linter:v4.9.2
Error response from daemon: manifest unknown
```

sxa (commented):
Plan to discuss on PMC call today to ensure the team are ok with this since I expect it to have an impact on end users that will need a later XLC runtime installed.

(Tagging @aixtools FYI)

aixtools:
IMHO - yet another reason to be looking - real soon - at the PATH variable - as jenkins will now be using the xlc16 more often than the xlc13 - my expectation.

As to later XLC runtime: that is not the key issue. The key issue _would_ have been the lowest level of AIX supported. However, since xlC16 is available on the build systems - no change to the lowest level of AIX supported is needed.

sxa:
> As to later XLC runtime: that is not the key issue.

It will be for customers.

aixtools:
> > As to later XLC runtime: that is not the key issue.
> 
> It will be for customers.

Actually, the XLC runtimes are provided with AIX updates. (:( just looked at several SP updates, only the xlC-13 updates are provided by default).

If you want extra `certainty` then packaging using installp would state the requirements needed. And there are two ways of doing this: just adding it to the `requirements` specification and the preview mode will report any missing RTE files, if any - OR - write a so-called pre_installation script that gets executed - and - while not installing, could give information (such as a URL) where the required (no-charge) updates could be acquired.

Anyway - I will help with this. Just need to get in-sync with the installer group.

aixtools:
For reference:
- IBM Doc Links: https://www.ibm.com/support/pages/ibm-xl-cc-runtime-aix-161
- https://www.ibm.com/support/pages/ibm-open-xl-cc-aix-and-xl-cc-aix-documentation-library#tab_133

Key element for xlC-16 - pre-requisites is AIX 7.1 TL4 - which is what the build systems are at.

tellison:
Just linking to Tyler's [discussion](https://mail.openjdk.java.net/pipermail/ppc-aix-port-dev/2022-March/004145.html) on this topic.

If the OpenJDK project is willing to change the supported platform version in an update release then we'll follow suit. Just ensuring we are in sync, and the change is done knowingly upstream.

sxa:
After discussion on the PMC call today we are happy with this change on the basis that IBM and SAP are ok with it, therefore we will merge it, but will delay until after the April release cycle.

@backwaterred I'd be ok with switching this for any JDK11+ to keep the changes simpler given that JDK12 is now out of support if you'd prefer that, but I'm ok with this as-is too.

backwaterred:
Happy news @sxa. Thanks for bringing this change to the team. 

I will make a change to simplify the build logic shortly.

backwaterred:
@sxa I have made the changes mentioned above.

aixtools:
> After discussion on the PMC call today we are happy with this change on the basis that IBM and SAP are ok with it, therefore we will merge it, but will delay until after the April release cycle.
> 
> @backwaterred I'd be ok with switching this for any JDK11+ to keep the changes simpler given that JDK12 is now out of support if you'd prefer that, but I'm ok with this as-is too.

re: upstream - and making life easier for consumers - may I interest anyone in a discussion on `packaging` of the AIX releases?

backwaterred:
> > After discussion on the PMC call today we are happy with this change on the basis that IBM and SAP are ok with it, therefore we will merge it, but will delay until after the April release cycle.
> > @backwaterred I'd be ok with switching this for any JDK11+ to keep the changes simpler given that JDK12 is now out of support if you'd prefer that, but I'm ok with this as-is too.
> 
> re: upstream - and making life easier for consumers - may I interest anyone in a discussion on `packaging` of the AIX releases?

I'd be happy to hear more about this. Are you talking about packaging for installp as one would for Linux package managers (rpm etc.)?

gdams:
> re: upstream - and making life easier for consumers - may I interest anyone in a discussion on packaging of the AIX releases?

@aixtools AIX packaging (assuming you're referring to an RPM file for YUM) would likely be best discussed in the [installer repo](https://github.com/adoptium/installer). I would certainly be in favour of producing them but we would need to understand the build requirements better. For reference the current Linux RPM's are produced in https://github.com/adoptium/installer/tree/master/linuxNew/jdk/redhat

aixtools:
Actually, I am thinking about `installp` format. I do alot of packaging (or did). Just made autoconf-2.71 and am testing a cups-2.4.X - that have no additional requirements (correction, autoconf requires gnu m4 as well). - but not 10 or more unrelated packages. IMHO: most of the .spec files written for `rpm -bb ...` have never been updated to the current situation and/or assume that because an OSS package was needed to build something - it is also needed for the same _something_ to run.

The packaging part is quite simple - but verification that the package is capable to report on missing requisites requires extensive testing. Again, not hard, but can be time consuming. `Less is more` - i.e., cannot test installation packages on build-capable systems. Ideal are `generic` (base) installs and then install the package (imho).

I'll try and get some traction in the installer repo. Having a partner who understands the current process for Linux would be a big help for me to do something similar for AIX.

sxa (approved):

sxa:
Build lockdown cycle for the April releases is over - merging.

PR #2904, opened 2022-04-20
Allow make-adopt-build-farm to detect alpine properly

sxa:
Autodetection for Alpine as distinct from a regular linux. The output of `uname -p` is `unknown` so I've worked around that by using `uname -m` instead like we do for cygwin.


build-farm/make-adopt-build-farm.sh~
```diff
@@ -0,0 +1,246 @@
+#!/bin/bash
```
karianna, line 1: file added by mistake?

karianna (changes requested):

build-farm/make-adopt-build-farm.sh~
sxa: Yep - my shell's been tab expanding to backup files as a priority recently ... I'll remove

karianna (approved):

sxa:
Looking for a second approver on this - @gdams @andrew-m-leonard ?

gdams (approved):

sxa:
Thanks @gdams

PR #2911, opened 2022-05-03
Update OpenJ9 builds to use OpenSSL 1.1.1o

keithc-ca:
See also eclipse-openj9/openj9#14994.

PR #2917, opened 2022-05-09
OpenJ9: Change Windows to OpenSSL fetched

AdamBrousseau:
Windows build fixed to be able to compile
openssl on the fly rather than using system
pre-compiled binaries.

### DIFF
--- a/build-farm/platform-specific-configurations/windows.sh
+++ b/build-farm/platform-specific-configurations/windows.sh
@@ -24,7 +24,6 @@ export DRAGONWELL8_BOOTSTRAP=/cygdrive/C/openjdk/dragonwell-bootstrap/jdk8u272-g
 export ALLOW_DOWNLOADS=true
 export LANG=C
 export OPENJ9_NASM_VERSION=2.13.03
-export OPENSSL_VERSION=1.1.1m
 
 TOOLCHAIN_VERSION=""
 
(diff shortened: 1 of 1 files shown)


sxa (approved):
LGTM

build-farm/platform-specific-configurations/windows.sh
```diff
@@ -24,7 +24,6 @@ export DRAGONWELL8_BOOTSTRAP=/cygdrive/C/openjdk/dragonwell-bootstrap/jdk8u272-g
 export ALLOW_DOWNLOADS=true
 export LANG=C
 export OPENJ9_NASM_VERSION=2.13.03
-export OPENSSL_VERSION=1.1.1m
```
karianna, line 27: Note sure about this change - was that variable only used by OpenJ9?

karianna (changes requested):

build-farm/platform-specific-configurations/windows.sh
AdamBrousseau: Yes, as far as I can tell it was only inside `if openj9`'s.

AdamBrousseau: Windows will now use [this line](https://github.com/adoptium/temurin-build/blob/36ce3640b6d5317b2282c83892ec6db7731dc170/sbin/prepareWorkspace.sh#L267) like all other platforms.

sxa: To be clear, it's the use of this variable that's within an OpenJ9 section. That variable  definition in the environment should not affect Temurin which does not use OpenSSL.

If you have a concern then I suppose the definition could be moved within another `if` block but we do have other variables outside such blocks which are not used by Temurin builds.

karianna (approved):

build-farm/platform-specific-configurations/windows.sh
keithc-ca: Yes, I think it was only used by OpenJ9. I don't see any occurrences of `OPENSSL_VERSION` in this repo after this change.

sxa:
@AdamBrousseau The JDK11u build I've just run seems to have tried to clone openssl 1.1.1m - is that expected?

keithc-ca:
Yes, that's expected because #2911 is not yet merged.

Issue #2900, opened 2022-04-14
Create CycloneDX SBOM by default as a Jenkins Artifact


andrew-m-leonard:
Jenkins builds currently produce a CycloneDX SBOM printed in the Console output, when a BUILD_ARG of "--create-sbom" is specified.
Extended this to actually produce the json as a Jenkins artifact by default in all builds.

sophia-guo:
@smlambert I can help this if you have other high priority jobs.

smlambert:
@sophia-guo - that will be great!  I have assigned us both (as I want to stay close on this, but I think there will be several pieces and I know you will be able to progress it faster than I will, thanks for the offer!).

smlambert:
FYI @sophia-guo that first step is in @zdtsw 's PR https://github.com/adoptium/ci-jenkins-pipelines/pull/296 enables the printing of SBOM to console 

@zdtsw - have you also put your name against SBOM work?  Here is one way it could be broken down and shared...

1. enable the commandline argument (via ci-jenkins-pipelines/pull/296)
2. make any necessary changes to have it create an actual json file in the event it just prints to console (code [here](https://github.com/adoptium/temurin-build/tree/master/cyclonedx-lib))
3. archive the file to the Jenkins job.  
4. can be done in tandem with the other steps is to ensure the contents of the SBOM are more complete (related: https://github.com/adoptium/temurin-build/issues/2813)
5. push the SBOM to the releases repository
6. update the API to be able to serve up the SBOM
7. additional enhancements to SBOM, such as, if latest version of cycloneDX available 1.5 will have a 'platform' type, if so, we should consider switching from using 'application' to 'platform' for Temurin

zdtsw:
> FYI @sophia-guo that first step is in @zdtsw 's PR [adoptium/ci-jenkins-pipelines#296](https://github.com/adoptium/ci-jenkins-pipelines/pull/296) enables the printing of SBOM to console
> 
> @zdtsw - have you also put your name against SBOM work? Here is one way it could be broken down and shared...
> 
> 1. enable the commandline argument (via ci-jenkins-pipelines/pull/296)
> 2. make any necessary changes to have it create an actual json file in the event it just prints to console (code [here](https://github.com/adoptium/temurin-build/tree/master/cyclonedx-lib))
> 3. archive the file to the Jenkins job.
[...]

PR #2920, opened 2022-05-12
feat: enable --create-sbom in pipeline to create CycloneDX SBOM

zdtsw:
enable flag `--create-sbom` to call buildCyclonedxLib() and generateSBoM()

this will be a default behavior for all build. 
it would be better to set in the ci-jenkins-pipeline as an extra flag for BUILD_CONFIGURATION then pass to temurin-build, so we can enable different jdk version or even os/arch per needs


smlambert:
Related: https://github.com/adoptium/temurin-build/issues/2900

zdtsw:
change solution to https://github.com/adoptium/ci-jenkins-pipelines/pull/296, to have better switch  passing from ci-jenkins-pipeline

zdtsw:
new PR: https://github.com/adoptium/temurin-build/pull/2926

PR #2924, opened 2022-05-13
fix: Upgrade git version on windows to 2.36.1 to bypass GH action checkout bug

zdtsw:
when introduce safe in git
	- https://github.com/actions/checkout/issues/760


karianna (approved):
LGTM

.github/workflows/build.yml
```diff
@@ -226,7 +226,7 @@ jobs:
      - name: Cygwin git configuration
        shell: bash
        # workaround for https://github.com/actions/checkout/issues/760
-       run: mkdir "$HOME" && git config --system core.autocrlf false && git config --global --add safe.directory "${GITHUB_WORKSPACE}"
+       run: mkdir "$HOME" && git config --system core.autocrlf false && git config --global --add safe.directory "/cygdrive/d/a/temurin-build/temurin-build"
```
karianna, line 229: Is that always going to be true?

karianna (changes requested):

.github/workflows/build.yml
zdtsw: it is running in windows so the path should be correct. 
but it is just a workaround till they have it fixed formally in GH Action, then can have line reverted

karianna (approved):
LGTM

zdtsw:
think it is finally working: https://github.com/adoptium/temurin-build/runs/6424692834?check_suite_focus=true#step:10:3

andrew-m-leonard (approved):

zdtsw:
could someone help to merge it?

PR #2926, opened 2022-05-13
feat: enable SBOM as default behavior

zdtsw:
Ref: https://github.com/adoptium/temurin-build/issues/2900

PR #2929, opened 2022-05-16
feat: enable --create-sbom to set in ci-jenkins-pipeline instead default

zdtsw:
Ref: https://github.com/adoptium/temurin-build/issues/2900

### DIFF
--- a/sbin/common/config_init.sh
+++ b/sbin/common/config_init.sh
@@ -467,8 +467,8 @@ function configDefaults() {
   # The default behavior of whether we want to create the legacy JRE
   BUILD_CONFIG[CREATE_JRE_IMAGE]="false"
 
-  # Set default value to "true. If we do not want this behavior, we can update buildArg per each config file instead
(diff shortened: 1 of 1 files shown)


sophia-guo:
This should be created as the reverse of https://github.com/adoptium/temurin-build/pull/2926/files?

zdtsw:
> This should be created as the revert of https://github.com/adoptium/temurin-build/pull/2926/files?

Yes, it is to revert back to the old logic -- default to "false"

sophia-guo:
Screen Shot 2022-05-16 at 1 23 00 PM

Using the revert button is probably an easier and correct way.

zdtsw:
> Screen Shot 2022-05-16 at 1 23 00 PM
> 
> Using the revert button is probably an easier and correct way.

made a new [PR](https://github.com/adoptium/temurin-build/pull/2930), revert old one and have this one closed

PR #2930, opened 2022-05-17
Revert "feat: enable SBOM as default behavior"

zdtsw:
Reverts adoptium/temurin-build#2926

Since we still have problem on Windows and AIX, till they are fixed, we cannot turn this as default feature, thus revert commit and fall back to add flag in [config groovy files ](https://github.com/adoptium/ci-jenkins-pipelines/pull/297)


zdtsw:
@smlambert  could you restart the GH action on linux? I dont think it related to my change

smlambert (approved):

karianna (approved):
LGTM

Issue #2949, opened 2022-05-26
Temurin Mac jdk11+ Info.plist has different JVMVendor values depending which build method you use


andrew-m-leonard:
jdk11+ Info.plist contains:
Built via Jenkins using --make-exploded/assemble-image two stage method:
```
                <key>JVMVendor</key>
                <string>Eclipse Adoptium</string>
```
Built standalone just using standard invocation:
```
                <key>JVMVendor</key>
                <string>Eclipse Temurin</string>
```

The reason being the make/codesign/assemble-exploded steps for jdk11+  in sbin/build.sh setPlistValueForMacOS() does nothing because BUILD_CONFIG[VENDOR] gets set in:
https://github.com/adoptium/temurin-build/blob/20b03b2ffcb660460a1a64d5b0a5577331c0778a/sbin/build.sh#L1816
which never gets called doing a 
https://github.com/adoptium/temurin-build/blob/20b03b2ffcb660460a1a64d5b0a5577331c0778a/sbin/build.sh#L1789

smlambert:
Related: https://github.com/adoptium/temurin-build/pull/2705, https://github.com/adoptium/temurin-build/issues/2664

andrew-m-leonard:
I suspect then this code needs an only if jdk8:
https://github.com/adoptium/temurin-build/blob/20b03b2ffcb660460a1a64d5b0a5577331c0778a/sbin/build.sh#L1215

andrew-m-leonard:
I've updated the description to reflect the problem.

PR #2951, opened 2022-05-27
Ensure Mac Plist update to Temurin is performed for only jdk8

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/2949

Ensures Mac Plist change to Eclipse Temurin is only done for jdk8, and logic is correctly checked for jdk11+.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -1207,8 +1207,8 @@ setPlistValueForMacOS() {
   local DIRECTORY="${1}"
   local TYPE="${2}"
 
-  # Only perform these steps for EF builds
-  if [[ "${BUILD_CONFIG[VENDOR]}" == "Eclipse Adoptium" ]]; then
+  # Only perform these steps for Eclipse Adoptium jdk8 builds
+  if [["${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK8_CORE_VERSION}"]] && [[ "${BUILD_CONFIG[VENDOR]}" == "Eclipse Adoptium" ]]; then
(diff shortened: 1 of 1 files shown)


karianna (approved):
LGTM - Are we not changing 8 because it would break backwards compat?

andrew-m-leonard:
@karianna I believe that's still up for discussion in the Steering group, @smlambert ?
This PR just fixes the build inconsistency with jdk11+

andrew-m-leonard:
Test builds:

Mac jdk17u:
```
                <key>JVMVendor</key>
                <string>Eclipse Adoptium</string>
```

Mac jdk8u:
```
                <key>JVMVendor</key>
                <string>Eclipse Temurin</string>
```

sxa (commented):
Without understanding the intricacies of the options that lead to the discovery of this I'm a little unclear on one thing: Does this change the defaults that will come out during the nightly and/or release pipelines, or will it only affect builds using non-default options in the pipelines?

andrew-m-leonard:
> Without understanding the intricacies of the options that lead to the discovery of this I'm a little unclear on one thing: Does this change the defaults that will come out during the nightly and/or release pipelines, or will it only affect builds using non-default options in the pipelines?

I found this issue when building on my Mac with a plain build-farm/make-adopt-build-farm.sh (no exploded/assemble two phase build), the end result being different to Jenkins in the Plist file. The reason being my build ran the configureCommandParameters and changed the Plist value.

@sxa The end result for nightlies and releases will be the same, due to the addition of the jdk8 check now that configureCommandParameters() is being called from the assemble-exploded phase.

Issue #2950, opened 2022-05-27
github PR CI check for build windows-2019 fails: ClassNotFoundException: temurin.sbom.TemurinGenSBOM


andrew-m-leonard:
SBOM created by default now, looks like the windows-2019 CI environment has a problem?
https://github.com/adoptium/temurin-build/runs/6617749655?check_suite_focus=true
```
ANT_HOME is set incorrectly or ant could not be located. Please set ANT_HOME.
ANT_HOME is set incorrectly or ant could not be located. Please set ANT_HOME.
Error: Could not find or load main class temurin.sbom.TemurinGenSBOM
Caused by: java.lang.ClassNotFoundException: temurin.sbom.TemurinGenSBOM
Error: Process completed with exit code 1.
```

karianna:
Correct, there's an issue with that GH Action

PR #2954, opened 2022-05-27
fix: add ANT_HOME to PATH and export

zdtsw:
- we set ANT_HOME in ansible for VM but not in GH action, we need to append it as env variable to GITHUB_ENV (which is a file , sourced by GH step)
	- windows-2019 pre-instealld ant, no need install again
	- after temurin-build we need to reset ANT_HOME back to its old value which is  used in run-aqa to run "ant" also build_tests.xml need it to find resource
	 
test run log:
https://github.com/adoptium/temurin-build/runs/6635265682?check_suite_focus=true

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -287,7 +287,7 @@ jobs:
          echo "TEST_JDK_HOME=$(dirname "${imageroot}")" | Out-File -FilePath "$env:GITHUB_ENV" -Encoding utf8 -Append
 
      - name: Unset ANT_HOME for smoke test
-       run: echo "ANT_HOME=''" | Out-File -FilePath "$env:GITHUB_ENV" -Encoding utf8 -Append
+       run: echo "ANT_HOME=" | Out-File -FilePath "$env:GITHUB_ENV" -Encoding utf8 -Append
 
      - name: Smoke test
        uses: adoptium/run-aqa@v1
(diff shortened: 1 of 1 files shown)


zdtsw:
@andrew-m-leonard @karianna if you have time for this

karianna (approved):
LGTM

andrew-m-leonard (approved):

sxa:
@zdtsw Since it's not clear from reading the PR changes why this was done, can you add a comment in this PR (or the description) explaining what the failure was that this addresses? i.e. what was the error in the actions previously?

sxa:
Thanks - I suggest using "fixes #2950" in the description in the future as that will automatically create the link between PR and issue and also automatically close the related issue when the PR is merged (See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

Interesting that it needs to be different for sbom generation and smoke tests though.

I'll close the issue now :-)

zdtsw:
> Thanks - I suggest using "fixes #2950" in the description in the future as that will automatically create the link between PR and issue and also automatically close the related issue when the PR is merged (See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
> 
> Interesting that it needs to be different for sbom generation and smoke tests though.
> 
> I'll close the issue now :-)

Thanks!
Yeah, I forgot to put the original issue number in the PR description, and GH does handle this afterwards :P

PR #2956, opened 2022-05-30
fix: overwrite JDK${JDK_BOOT_VERSION}_BOOT_DIR on solaris

zdtsw:
- solaris only build for jdk8, JDK7_BOOT_DIR should use jdk8+ for ant build, set to ansible value
	- set variable jar dir to make code simple
	- move into function setupAntEnv() to get return javajHome
	- remove JDK_BOOT_DIR, because it wont be working for ant build on jdk8

test run looks good: https://ci.adoptopenjdk.net/view/Failing%20Temurin%20jobs/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-temurin/87/
doing new run with final commit: https://ci.adoptopenjdk.net/view/Failing%20Temurin%20jobs/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-temurin/88/console

### DIFF
--- a/build-farm/platform-specific-configurations/solaris.sh
+++ b/build-farm/platform-specific-configurations/solaris.sh
@@ -37,3 +37,7 @@ export PATH=/opt/solarisstudio12.3/bin/:/opt/csw/bin/:/usr/ccs/bin:$PATH:/usr/sf
 export LC_ALL=C
 export HOTSPOT_DISABLE_DTRACE_PROBES=true
 export ENFORCE_CC_COMPILER_REV=5.12
+
+# sepcial handling for jdk8 build "ant" which must use jdk8+
+# to match the version we installed by ansible
+export JDK8_BOOT_DIR=/usr
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -189,7 +189,7 @@ then
   fi
 fi
 
-echo "Boot jdk directory: ${JDK_BOOT_DIR}:"
+echo "Boot jdk directory: ${JDK_BOOT_DIR}"
 "${JDK_BOOT_DIR}/bin/java" -version 2>&1 | sed 's/^/BOOT JDK: /'
 java -version 2>&1 | sed 's/^/JDK IN PATH: /g'
 
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -616,8 +616,10 @@ createOpenJDKFailureLogsArchive() {
     createArchive "${adoptLogArchiveDir}" "${makeFailureLogsName}"
 }
 
-# Build the CycloneDX Java library and app used for SBoM generation
-buildCyclonedxLib() {
+# Setup JAVA env to run "ant task"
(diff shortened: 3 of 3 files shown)


zdtsw:
GH action checks mostly failed in download package from api servers

build-farm/platform-specific-configurations/solaris.sh
```diff
@@ -37,3 +37,7 @@ export PATH=/opt/solarisstudio12.3/bin/:/opt/csw/bin/:/usr/ccs/bin:$PATH:/usr/sf
 export LC_ALL=C
 export HOTSPOT_DISABLE_DTRACE_PROBES=true
 export ENFORCE_CC_COMPILER_REV=5.12
+
+# sepcial handling for jdk8 build "ant" which must use jdk8+
+# to match the version we installed by ansible
+export JDK8_BOOT_DIR=/usr
```
karianna, line 43: Is that where we are storing the JDK 8 BOOT JDK? Seems like an odd location?

karianna (changes requested):

build-farm/platform-specific-configurations/solaris.sh
zdtsw: i think in order to not care about the patch version, we create symblink to /usr/bin/java on Solaris

https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/Solaris.yml#L278

karianna (approved):

andrew-m-leonard (approved):

build-farm/platform-specific-configurations/solaris.sh
sxa: I'm with @karianna on this one - the current location defined in jenkins should be correct i.e. `/opt/csw/java/jdk1.8.0_181`. At the moment the machines have a symlink from `/usr/bin/java` to that location so I'm unclear on why that change is required. If we want to update it we could do it in jenkins, but I really don't want it forced in the build scripts like this.

sxa (changes requested):
I don't agree with the `JDK8_BOOT_DIR` change going into solaris.sh.

It doesn't sound like this will actually change any of the behaviour here so I would leave that part out of it. Perhaps we should look at using /usr/lib/jvm/jdk8 for the temurin build that's installed, although again that should be defined in jenkins, not forced in `solaris.sh`

zdtsw:
@andrew-m-leonard helped to correct the value of JDK8_BOOT_DIR on that Solaris node from Jenkins node config (which I do not have permission to view it yet), now we started a new run (with all the code from master branch) 
Assume it should get the correct path this time and build should pass. If so, there is no need to the change of JDK8_BOOT_DIR from this PR

zdtsw:
the build is working https://ci.adoptopenjdk.net/view/Failing%20Temurin%20jobs/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-temurin/91/console

will have this PR closed, no need to update temurin-build.
if error happen again, ensure the value in Jenkins node setting is correct.

PR #2960, opened 2022-06-01
Correct CycloneDX JDK_BOOT_DIR variable reference for JAVA_HOME

andrew-m-leonard:
JDK_BOOT_DIR variable is not present in sbin/build.sh shell as it was only set in the make-adopt-build-farm.sh shell.
The BUILD_CONFIG[JDK_BOOT_DIR] variable contains the value passed from the make-adopt-build-farm.sh shell.


sxa (commented):
Does this run any risk of problems if people are running `make-adopt-build-farm.sh` in a standalone manner (which is what our instructions tell end-users to run)?

karianna (approved):
LGTM

PR #2962, opened 2022-06-01
Add --create-sbom to README.md

andrew-m-leonard:

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,9 @@ create a debug-image archive with the debug symbols.
 --create-jre-image
 create the legacy JRE image in addition to the JDK image.
 
+--create-sbom
+create the CycloneDX System Bill of Materials (JSON artifact).
+
 -d, --destination <path>
 specify the location for the built binary, e.g. /path/.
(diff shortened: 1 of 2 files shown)


zdtsw (approved):

karianna:
@andrew-m-leonard needs rebase

sxa (changes requested):
> Can this be added into https://github.com/adoptium/temurin-build/blob/master/makejdk-any-platform.1 too 

Ading a block on this as we don't want to have out of date information. Also in light of other conversations it looks like creating the SBOM is happening by default even without this option being specified which is probably not desired behavior

andrew-m-leonard:
@sxa added to [makejdk-any-platform.1](https://github.com/adoptium/temurin-build/blob/master/makejdk-any-platform.1)
and default being reverted to false: https://github.com/adoptium/temurin-build/pull/3053

andrew-m-leonard:
Alpine failures unrelated

andrew-m-leonard:
@sxa I think this is good now

sxa (approved):
LGTM

PR #2970, opened 2022-06-08
switch jobs to run on users pool rather than adoptium pool

gdams:
WIP until I can work out how to report the status back here...


zdtsw:
for saving adoptium GH cost or time per PR? 
I specifically disabled GH action on my forked repo

Issue #2983, opened 2022-06-17
Review current SBOM for obvious missing Linux and MacOS detail


andrew-m-leonard:
The current SBOM json needs reviewing to add obvious missing details, eg.gcc or clang version, Xcode version, ...?

andrew-m-leonard:
Closing out-of-date issue

PR #2989, opened 2022-06-20
cleanup: do not genereate component property "scmRef"

zdtsw:
- scmRef is the short SHA1 of source code commit on openjdk
	- it is duplicated as in "openjdkSourceCommit" which as fullpath
what we have now in sbom.json is like:
``` 
"components" : [
    {
      "author" : "Vendor: Eclipse",
      "group" : "Eclipse Temurin",
      "name" : "Temurin",
      "version" : "19-beta+27-202206180317",
      "type" : "framework"
    },
    {
      "name" : "JDK",
      "description" : "Temurin JDK Component",
      "properties" : [
        {
          "name" : "scmRef",
          "value" : "73b1da30674"
        },
        {
          "name" : "openjdkSourceCommit",
          "value" : "https://github.com/adoptium/jdk19/commit/73b1da30674"
        },
        {
          "name" : "buildRef",
          "value" : "https://github.com/adoptium/temurin-build/commit/6762709"
        },
`
``` 

Ref: https://github.com/adoptium/temurin-build/issues/2983

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -699,7 +699,7 @@ generateSBoM() {
   "${javaHome}"/bin/java -cp "${classpath}" temurin.sbom.TemurinGenSBOM --addComponent --jsonFile "$sbomJson" --compName "JDK" --description "${BUILD_CONFIG[BUILD_VARIANT]^} JDK Component"
 
   # Add scmRef JDK Component Property
-  addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "JDK" "scmRef" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/scmref.txt"
+  # addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "JDK" "scmRef" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/scmref.txt"
 
   # Add OpenJDK source ref commit JDK Component Property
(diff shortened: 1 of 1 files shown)


sbin/build.sh
```diff
@@ -699,7 +699,7 @@ generateSBoM() {
   "${javaHome}"/bin/java -cp "${classpath}" temurin.sbom.TemurinGenSBOM --addComponent --jsonFile "$sbomJson" --compName "JDK" --description "${BUILD_CONFIG[BUILD_VARIANT]^} JDK Component"
 
   # Add scmRef JDK Component Property
-  addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "JDK" "scmRef" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/scmref.txt"
+  # addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "JDK" "scmRef" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/scmref.txt"
```
karianna, line 702: Should we not just remove this line in that case?

karianna (changes requested):

sbin/build.sh
zdtsw: sure
want to get some option from @andrew-m-leonard first and if it is a good to go, i can remove these two lines

andrew-m-leonard:
@zdtsw give me 30mins, just want to try something

andrew-m-leonard:
@zdtsw yes as I thought, they appear the same in your example, but they are not actually the same thing, eg:
```
        {
          "name" : "scmRef",
          "value" : "jdk-17.0.4+7_adopt"
        },
        {
          "name" : "openjdkSourceCommit",
          "value" : "https://github.com/adoptium/jdk17u/commit/eb2197066f5"
        },
```

scmRef : The requested tag or ref to build
openjdkSourceCommit : The actual resulting SHA of the commit that was built

So in the above example, jdk-17.0.4+7_adopt tag commit has a SHA of eb2197066f5

PR #2991, opened 2022-06-21
fix: use OS_FULL_VERSION than OS_KERNEL in sbom to include more

zdtsw:
from conversation: https://adoptium.slack.com/archives/C09NW3L2J/p1655804491202619?thread_ts=1655802669.491939&cid=C09NW3L2J
we only store os kernal which does not help much for alpine, centos, rhel etc 

Ref: https://github.com/adoptium/temurin-build/issues/2983

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -719,8 +719,8 @@ generateSBoM() {
   # Add make_command_args JDK Component Property
   addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "JDK" "make_command_args" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/makeCommandArg.txt"
 
-  # Add OS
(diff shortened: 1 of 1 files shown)


karianna (approved):
LGTM

sxa (approved):
Yep I think this is a good change since as far as I can tell the kernel version is always included in this new definition.

sxa:
@andrew-m-leonard Are you happy with this? Some sample values:
- AIX 7100-04
- Linux 5.4.0-110-generic : Alpine Linux
- Linux 3.10.0-1160.62.1.el7.ppc64le : CentOS Linux release 7.9.2009 (AltArch)
- Darwin 20.1.0
And if we want to make it more comprehensive at some point we can modify the code in `unameOSSysVer` at https://github.com/adoptium/temurin-build/blob/abb4d36cfe18ceb3be70d8b30d28a1eae4262863/sbin/common/config_init.sh#L387

FYI @andrew-m-leonard in case you have any comments

andrew-m-leonard (approved):
lgtm

zdtsw:
think i need someone to merge it 
failed GH check on alpine is known issue at the moment

PR #2992, opened 2022-06-21
build: switch Windows JDK11 builds to use VS2019

gdams:
The Visual Studio 2019 backport is now complete for JDK11 which means we should be able to start building using VS2019


karianna:
LGTM

karianna (approved):
Lgtm

PR #3005, opened 2022-06-29
feat: change sbom content

zdtsw:
- read build_config from new construct string, not file
	- add USE_DOCKER and DOCKER_SHA1
	- set default to N.A if not available


sbin/build.sh
```diff
@@ -760,10 +768,11 @@ addSBOMMetadataPropertyFromFile() {
   local jsonFile="${3}"
   local name="${4}"
   local propFile="${5}"
+  value="N.A"
```
karianna, line 771: Should this be decalred local as well?  I forget my bash syntax...

sbin/common/config_init.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# shellcheck disable=SC2153,SC2155
+# shellcheck disable=SC2153,SC2155,SC2030,SC2031
```
karianna, line 2: I'd prefer not to ignore shellcheck where possible, what two rules are being broken?

sbin/build.sh
zdtsw: dont have to be in the function. 
but to avoid complication, i can set it to be a local variable.

sbin/common/config_init.sh
zdtsw: very strange.... now i removed disable on these two and linter did not complain.....

andrew-m-leonard (approved):

zdtsw:
Need help to merge this one too :D

PR #3019, opened 2022-07-01
aix: Remove /tmp/sh-np.* before building

sxa:
We are still seeing problems with errors in /tmp/ on AIX systems (Specifically build-osuosl-aix71*1)
This will clear out all such entries on AIX when running under jenkins (username=jenkins and `WORKSPACE` is defined).
Broadly copies what is in the crontab on the machines

Ref: https://github.com/adoptium/infrastructure/issues/929

### DIFF
--- a/build-farm/platform-specific-configurations/aix.sh
+++ b/build-farm/platform-specific-configurations/aix.sh
@@ -15,8 +15,16 @@
 # limitations under the License.
 ################################################################################
 
-
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Clear out /tmp/sh-np files if we're running in jenkins to avoid
+# "ambiguous redirect" pipe failures. This will be ok with only one executor
+# See https://github.com/adoptium/infrastructure/issues/929
+if [ "$(whoami)" = "jenkins" ] && [ -n "$WORKSPACE" ]; then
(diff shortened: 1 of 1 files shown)


build-farm/platform-specific-configurations/aix.sh
```diff
@@ -18,6 +18,14 @@
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+# Clear out /tmp/sh-np files if we're running in jenkins to avoid
+# "ambiguous redirect" pipe failures. This will be ok with only one executor
+# See https://github.com/adoptium/infrastructure/issues/929
+if [ "$(whoami)" = "jenkins" ] && [ -n "$WORKSPACE" ]; then
```
karianna, line 24:
```suggestion
if [ "$(whoami)" == "jenkins" ] && [ -n "$WORKSPACE" ]; then
```

karianna (changes requested):
i think the == is required

build-farm/platform-specific-configurations/aix.sh
sxa: > i think the == is required

No, single equals is standard POSIX shell syntax for string equality.

sxa:
(Will aim to fix ECA check tomorrow)

karianna (approved):

sxa:
Failing check was the labeller but it would almost certainly have just removed and re-added the same labels again :-)

PR #3021, opened 2022-07-04
reduce RAM usage during build on Linux systems with <4GB

sxa:
I've had issues on one of my 4GB aarch64 systems with two `ld` processes being started to perform linking of the final `libjvm` in parallel which use over 1.5Gb each and ultimately result in the system going heavily into swap and not completing in a reasonable amount of time. This change - which will only take effect on low memory systems - will use the option ([man page here](https://sourceware.org/binutils/docs-2.38/ld.html)) which will reduce the memory usage of the `ld` command at the expense of some performance (which is why I am not enabling it by default)

[Side notes: https://wiki.debian.org/ReduceBuildMemoryOverhead mentions other `ld` options that might reduce memory further, but the one in this PR should have enough of an effect]

### DIFF
--- a/build-farm/platform-specific-configurations/alpine-linux.sh
+++ b/build-farm/platform-specific-configurations/alpine-linux.sh
@@ -19,6 +19,12 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=sbin/common/constants.sh
 source "$SCRIPT_DIR/../../sbin/common/constants.sh"
 
+# Solves issues seen on 4GB HC4 systems with two large ld processes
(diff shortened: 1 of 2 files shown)


sxa:
For some reason the new parameter is not taking effect during the build process for `libjvm.so` -  continuing to investigate (With `LOG=debug` as a make parameter).

sxa:
Was having some odd behaviour where the response file passed into `ld` via a call to `g++` to do the link didn't seem to have all the options, but it seems to have taken effect since the RAM use is the "low" version, so I think this is good to merge as-is now. Unlikely to make any difference on our build machines as they typically have more RAM.

gdams (approved):

Issue #2973, opened 2022-06-08
Disable CCache for Release(Reproducible) builds jdk-19+


andrew-m-leonard:
For reproducible builds (currently jdk-19+), ccache must be disabled, automatically add --disable-ccache for jdk-19+ when RELEASE=true.

AdamBrousseau:
Will this be wrapped in "if temurin" or will it apply to all variants?

Is there compelling reason to not make this applicable to all versions right away?

@pshipton could confirm, but I think Semeru would be interested in this for all releases. I think we've talked about it in the past.

andrew-m-leonard:
@AdamBrousseau 
So I was going to make all variants, as it should apply. Basically, with ccache enabled the path substitution by CCache is not always deterministic in its use of ../../ in path substitutions.

pshipton:
Including Semeru sounds good.

sxa:
We should include JDK17 in this too especially since Linux/JDK17 should be able to be built in a reproducible way going forward too?

andrew-m-leonard:
> We should include JDK17 in this too especially since Linux/JDK17 should be able to be built in a reproducible way going forward too?

Yes, there's a few other things for that to be done as well

zdtsw:
related to https://github.com/adoptium/temurin-build/issues/2977

PR #3026, opened 2022-07-06
fix: move set "--enable-ccache" from build.sh to platform specific script

zdtsw:
- final "--disable-ccache" is called after these scripts to filter out "--enable-ccache"

Ref: https://github.com/adoptium/temurin-build/pull/3022/files#r914642300
Ref: https://github.com/adoptium/temurin-build/issues/2973

### DIFF
--- a/build-farm/platform-specific-configurations/mac.sh
+++ b/build-farm/platform-specific-configurations/mac.sh
@@ -38,6 +38,9 @@ else
     XCODE_SWITCH_PATH="/";
   fi
   export PATH="/Users/jenkins/ccache-3.2.4:$PATH"
+  if [ -n "$(which ccache)" ]; then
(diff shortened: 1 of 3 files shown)


andrew-m-leonard:
> @andrew-m-leonard I've just noticed that a lot of the reproducible build stuff is in the `build-farm` directory. I'd have typically said that `build-farm` should be used for things specific to the set up of our (Adopt's) machines, and options to enable functional configuration differences should be in the stuff in `sbin`. It seems a little strange to me to be activating different options for `RELEASE` within `build-farm` instead of e.g. `build.sh`. Was that a deliberate decision for some reason?

Interesting point, I hadn't really appreciated the difference, I think you're right. I'll raise an issue to move it

andrew-m-leonard:
@sxa so the bootjdk stuff is fine, but the with-source-date and ccache settings should move

zdtsw:
close this draft.
move flags into build.sh

PR #3033, opened 2022-07-08
Correct mk-cacerts.sh nokeystore alias file renaming to be unique

andrew-m-leonard:
Fixes https://github.com/adoptium/adoptium-support/issues/540

The nokeystore logic for reproducible cacerts creation had a uniqueness bug in the cert file renaming
for input to openjdk GenerateCacerts tool. The logic was removing digits from the alias which then resulted in
some aliases mapping to the same filename.

Corrected the alias file name uniqueness and updated logic to mirror both generation methods, and also verify
a cert filename clash does not occur.

Also corrected an existing bug with the detection of duplicate aliases, which was actually causing
certificate: "QuoVadis Root CA 3" to be missing as it is a substring of "QuoVadis Root CA 3 G3"

Corrected the invocation of mk-cacerts.sh with --nokeystore to be jdk-17+. --with-cacerts-src is backported to jdk17,18,19

### DIFF
--- a/security/mk-cacerts.sh
+++ b/security/mk-cacerts.sh
@@ -121,8 +121,8 @@ for FILE in certs/*.crt; do
 done
 
 if [ "$NO_KEYSTORE" = false ] ; then
-    num_certs=$("$KEYTOOL" -v -list -storepass changeit -keystore cacerts | grep "Alias name:" | wc -l)
+    num_certs=$("$KEYTOOL" -v -list -storepass changeit -keystore cacerts | grep -c "Alias name:")
 else
-    num_certs=$(ls certs | wc -l)
+    num_certs=$(find certs/* | wc -l)
 fi
 echo "Number of certs processed: $num_certs"
--- a/security/mk-cacerts.sh
+++ b/security/mk-cacerts.sh
@@ -85,18 +85,16 @@ IMPORTED=('null')
 for FILE in certs/*.crt; do
     SUBJECT=$(openssl x509 -subject -noout -in "$FILE")
     TRIMMED_SUBJECT="${SUBJECT#*subject= /}"
-    if [ "$NO_KEYSTORE" = false ] ; then
-        ALIAS="${TRIMMED_SUBJECT//\//,}"
-    else
-        # Remove/translate characters not valid in a filename
-        ALIAS_NO_INVALID="${TRIMMED_SUBJECT//[ :()]/_}"
-        ALIAS=$(echo "${ALIAS_NO_INVALID}" | tr -cd 'a-zA-Z,_' | tr A-Z a-z)
-    fi
+    ALIAS="${TRIMMED_SUBJECT//\//,}"
 
-    if [ "$NO_KEYSTORE" = false ] ; then
-         if printf '%s\n' "${IMPORTED[@]}" | grep "${ALIAS}"; then
-            echo "Skipping certificate with alias: $ALIAS as it already exists"
-        else
+    if printf '%s\n' "${IMPORTED[@]}" | grep "${ALIAS}"; then
+        echo "Skipping certificate file $FILE with alias: $ALIAS as it already exists"
+        if [ "$NO_KEYSTORE" = true ] ; then
+            # Remove duplicate $FILE so it is not imported using OpenJDK GenerateCacerts
+            rm $FILE
+        fi
+    else
+        if [ "$NO_KEYSTORE" = false ] ; then
             echo "Processing certificate with alias: $ALIAS"
             "$KEYTOOL" -noprompt \
             -import \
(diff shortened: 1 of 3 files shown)


security/mk-cacerts.sh
```diff
@@ -105,11 +103,26 @@ for FILE in certs/*.crt; do
             -file "$FILE" \
             -keystore "cacerts" \
             -storepass "changeit"
-
-            IMPORTED+=("${ALIAS}")
+        else
+            # Importing using OpenJDK GenerateCacerts, so must ensure alias is a valid filename
+            ALIAS_NO_INVALID="${ALIAS//[ :()]/_}"
+            ALIAS_FILENAME=$(echo "${ALIAS_NO_INVALID}" | tr -cd '0-9a-zA-Z,_' | tr '[:upper:]' '[:lower:]')
```
zdtsw, line 109: so the real fix is to have 0-9 not deleted

security/mk-cacerts.sh
```diff
@@ -105,11 +103,26 @@ for FILE in certs/*.crt; do
...
+                echo "security/mk-cacerts.sh needs ALIAS_FILENAME filter updating to make unique"
+                exit 1
+            fi
+            mv "$FILE" "certs/$ALIAS_FILENAME"
         fi
-    else
-         echo "Renaming $FILE to certs/$ALIAS"
-         mv "$FILE" "certs/$ALIAS"
+
+        IMPORTED+=("temurin_${ALIAS}_temurin")
```
zdtsw, line 119: guess to add termurin_ _temurin does not really make a difference? but for visibility

security/mk-cacerts.sh
```diff
@@ -85,18 +85,16 @@ IMPORTED=('null')
 for FILE in certs/*.crt; do
     SUBJECT=$(openssl x509 -subject -noout -in "$FILE")
```
zdtsw, line 86: i did a test
seems two crts got the same subject
-  certs/cert123.crt
-  certs/cert29.crt
subject=C = ES, CN = Autoridad de Certificacion Firmaprofesional CIF A62634068

andrew-m-leonard: Yes correct, that is a valid duplicate, the Mozilla cert query returns that same Certificate twice

security/mk-cacerts.sh
andrew-m-leonard: It does, as it prevents the already imported logic missing Certs:
```
    if printf '%s\n' "${IMPORTED[@]}" | grep "temurin_${ALIAS}_temurin"; then
```
As the current logic skips "QuoVadis Root CA 3" as it is a substring of "QuoVadis Root CA 3 G3"
See current build logs:
```
13:06:45  C=BM,O=QuoVadis Limited,CN=QuoVadis Root CA 3 G3
13:06:45  Skipping certificate with alias: C=BM,O=QuoVadis Limited,CN=QuoVadis Root CA 3 as it already exists
```

zdtsw: right! so if the "subject=C = BM, O = QuoVadis Limited, CN = QuoVadis Root CA 2 G3"  has been added before "subject=C = BM, O = QuoVadis Limited, CN = QuoVadis Root CA" it does make a difference

zdtsw (approved):
LGTM

karianna (approved):
LGTM

sxa (commented):
A few comments, mostly style/simplicity but the more eyeballs we have on this one the better - looks like (as per Wen's commetn) that there's a bit more shuffling around in here than just fixing the original problem.

karianna:
@andrew-m-leonard maybe follow up with a small PR for Stewart's suggestions?

security/mk-cacerts.sh
```diff
@@ -105,11 +103,26 @@ for FILE in certs/*.crt; do
...
             -keystore "cacerts" \
             -storepass "changeit"
-
-            IMPORTED+=("${ALIAS}")
+        else
+            # Importing using OpenJDK GenerateCacerts, so must ensure alias is a valid filename
+            ALIAS_NO_INVALID="${ALIAS//[ :()]/_}"
+            ALIAS_FILENAME=$(echo "${ALIAS_NO_INVALID}" | tr -cd '0-9a-zA-Z,_' | tr '[:upper:]' '[:lower:]')
+            echo "Renaming $FILE to certs/$ALIAS_FILENAME"
+            if test -f "certs/$ALIAS_FILENAME"; then
```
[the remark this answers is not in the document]
andrew-m-leonard: done

security/mk-cacerts.sh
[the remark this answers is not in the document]
andrew-m-leonard: done

Issue #3034, opened 2022-07-08
Need a SmokeTest to verify the number of CACerts in the built jdk is correct


andrew-m-leonard:
Verify that:
   num_in_keystore=$("$KEYTOOL" -v -list -storepass changeit -keystore jdk/lib/security/cacerts | grep -c "Alias name:")
matches the number of unique Certificates listed by security/mk-ca-bundle.pl

zdtsw:
so in this case we need to add a counter after `rm "$FILE"`

andrew-m-leonard:
@zdtsw  Yeah so the correct JDK image number of Certs should be "mk-ca-bundle.pl number" - "number of openssl alias duplicates"

andrew-m-leonard:
The SmokeTest needs some new code, the reason being because the output from this script:

    <jdk17 : keystore
    >=jdk17 : src file certs

Both get fed into the OpenJDK build that then import either into the image/jdk/lib/security/cacerts keystore, there is always the possibility that build fails or is incorrect, so we need to ensure it is good. One other issue that concerns me is if say a bug or network glitch occurs in [mk-ca-bundle.pl](https://github.com/adoptium/temurin-build/blob/master/security/mk-ca-bundle.pl) Perl script and say it returns 0 certs, how do we catch that? Should we hard code the expected number (ie.135) as it very rarely changes, or do we have a test that actually checks the expected list like OpenJDK do: https://github.com/adoptium/jdk17u/blob/5de8b78504df916d88c9e01941c4e27bfe49e949/test/jdk/sun/security/lib/cacerts/VerifyCACerts.java#L50

andrew-m-leonard:
I'm actually in favour of adding a similar test to [VeriftCACerts](https://github.com/adoptium/jdk17u/blob/5de8b78504df916d88c9e01941c4e27bfe49e949/test/jdk/sun/security/lib/cacerts/VerifyCACerts.java#L50) to our SmokeTest: https://github.com/adoptium/temurin-build/tree/master/test/functional/buildAndPackage

zdtsw:
on a newly built jdk19 (after the fix of cacerts): 
https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk19/job/jdk19-mac-x64-temurin_SmokeTests/17/console

and test failed on an old build 
[...]

PR #3036, opened 2022-07-11
feat: add post check for imported ca certs number

zdtsw:
- if enable --nokeysotre(reproducible build): current *.cert file nr. == return from perl - duplicated
	- if not set --nokeysotre: keytool -list == return from perl - duplicated

Fixes: https://github.com/adoptium/temurin-build/issues/3034

### DIFF
--- a/security/mk-cacerts.sh
+++ b/security/mk-cacerts.sh
@@ -125,7 +125,7 @@ else
 fi
 echo "Number of certs processed: $num_certs"
 
-# post verification: (nr.(mk-ca-bundle.pl) - nr.(already imported file) == (nr. (currnent files in certs folder)) || (nr. (alias in cacerts))
+# post verification: (nr.(mk-ca-bundle.pl) - nr.(already imported file) == (nr. (current files in certs folder)) || (nr. (alias in cacerts))
 certNum="$((certNum-alreadyExistsCounter))"
 if [ $certNum != $num_certs ]; then
     echo "Number of cert from mk-ca-bundle.pl: $certNum"
--- a/security/mk-cacerts.sh
+++ b/security/mk-cacerts.sh
@@ -55,7 +55,7 @@ fi
 # Convert Mozilla's list of certificates into a PEM file. The -n switch makes
 # it use the local certdata.txt in this folder.
 certNum=$(./mk-ca-bundle.pl -v -n ca-bundle.crt)
-echo "mk-ca-bundle.pl generates "$certNum "cerificates"
+echo "mk-ca-bundle.pl generates $certNum cerificates"
 
 # Split the PEM file into individual files because keytool cannot do it on its own.
 awk '
@@ -91,7 +91,7 @@ for FILE in certs/*.crt; do
             # Remove duplicate $FILE so it is not imported using OpenJDK GenerateCacerts
             rm "$FILE"
         fi
-        let "alreadyExistsCounter=alreadyExistsCounter+1"
+        (("alreadyExistsCounter=alreadyExistsCounter+1"))
     else
         if [ "$NO_KEYSTORE" = false ] ; then # for jdk8, jdk11
             echo "Processing certificate with alias: $ALIAS"
@@ -127,7 +127,7 @@ echo "Number of certs processed: $num_certs"
 
 # post verification: (nr.(mk-ca-bundle.pl) - nr.(already imported file) == (nr. (current files in certs folder)) || (nr. (alias in cacerts))
 certNum="$((certNum-alreadyExistsCounter))"
-if [ $certNum != $num_certs ]; then
+if [ "$certNum" != "$num_certs" ]; then
     echo "Number of cert from mk-ca-bundle.pl: $certNum"
     echo "Number imported to $KEYTOOL: $num_certs"
     exit 1
--- a/security/mk-cacerts.sh
+++ b/security/mk-cacerts.sh
@@ -80,7 +80,7 @@ awk '
 # The full subject needs to be used to prevent alias collisions.
 
 IMPORTED=('null')
-existCounter=0 # counter for duplicated file
+alreadyExistsCounter=0 # counter for duplicated file
 
 for FILE in certs/*.crt; do
     ALIAS=$(openssl x509 -subject -noout -in "$FILE" | sed 's/^subject=//' | tr '/' ',')
--- a/security/mk-cacerts.sh
+++ b/security/mk-cacerts.sh
@@ -91,7 +91,7 @@ for FILE in certs/*.crt; do
             # Remove duplicate $FILE so it is not imported using OpenJDK GenerateCacerts
             rm "$FILE"
         fi
-        let "existCounter=existCounter+1"
+        let "alreadyExistsCounter=alreadyExistsCounter+1"
     else
         if [ "$NO_KEYSTORE" = false ] ; then # for jdk8, jdk11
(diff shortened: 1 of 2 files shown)


security/mk-cacerts.sh
```diff
@@ -81,20 +80,20 @@ awk '
 # The full subject needs to be used to prevent alias collisions.
 
 IMPORTED=('null')
+existCounter=0 # counter for duplicated file
```
karianna, line 83:
```suggestion
alreadyExistsCounter=0 # counter for duplicated file
```

security/mk-cacerts.sh
```diff
@@ -81,20 +80,20 @@ awk '
...
         echo "Skipping certificate file $FILE with alias: $ALIAS as it already exists"
-        if [ "$NO_KEYSTORE" = true ] ; then
+        if [ "$NO_KEYSTORE" = true ] ; then # for jdk17+
             # Remove duplicate $FILE so it is not imported using OpenJDK GenerateCacerts
             rm "$FILE"
         fi
+        let "existCounter=existCounter+1"
     else
-        if [ "$NO_KEYSTORE" = false ] ; then
+        if [ "$NO_KEYSTORE" = false ] ; then # for jdk8,jdk11
```
karianna, line 96:
```suggestion
        if [ "$NO_KEYSTORE" = false ] ; then # for jdk8, jdk11
```

security/mk-cacerts.sh
```diff
@@ -120,9 +118,17 @@ for FILE in certs/*.crt; do
     fi
 done
 
-if [ "$NO_KEYSTORE" = false ] ; then
+if [ "$NO_KEYSTORE" = false ] ; then # for jdk8,11
```
karianna, line 121:
```suggestion
if [ "$NO_KEYSTORE" = false ] ; then # for jdk8, jdk11
```

security/mk-cacerts.sh
```diff
@@ -81,20 +80,20 @@ awk '
...
+    ALIAS=$(openssl x509 -subject -noout -in "$FILE" | sed 's/^subject=//' | tr -d /)
 
     if printf '%s\n' "${IMPORTED[@]}" | grep "temurin_${ALIAS}_temurin"; then
         echo "Skipping certificate file $FILE with alias: $ALIAS as it already exists"
-        if [ "$NO_KEYSTORE" = true ] ; then
+        if [ "$NO_KEYSTORE" = true ] ; then # for jdk17+
             # Remove duplicate $FILE so it is not imported using OpenJDK GenerateCacerts
             rm "$FILE"
         fi
+        let "existCounter=existCounter+1"
```
karianna, line 94:
```suggestion
        let "alreadyExistsCounter=alreadyExistsCounter+1"
```

security/mk-cacerts.sh
```diff
@@ -120,9 +118,17 @@ for FILE in certs/*.crt; do
...
 
-if [ "$NO_KEYSTORE" = false ] ; then
+if [ "$NO_KEYSTORE" = false ] ; then # for jdk8,11
     num_certs=$("$KEYTOOL" -v -list -storepass changeit -keystore cacerts | grep -c "Alias name:")
 else
     num_certs=$(find certs/* | wc -l)
 fi
 echo "Number of certs processed: $num_certs"
+
+# post verification: (nr.(mk-ca-bundle.pl) - nr.(already imported file) == (nr. (currnent files in certs folder)) || (nr. (alias in cacerts))
```
karianna, line 128:
```suggestion
# post verification: (nr.(mk-ca-bundle.pl) - nr.(already imported file) == (nr. (current files in certs folder)) || (nr. (alias in cacerts))
```

security/mk-cacerts.sh
```diff
@@ -120,9 +118,17 @@ for FILE in certs/*.crt; do
...
+certNum="$((certNum-existCounter))"
```
karianna, line 129:
```suggestion
certNum="$((certNum-alreadyExistsCounter))"
```

karianna (changes requested):
Small suggestions

karianna (approved):
lgtm

security/mk-cacerts.sh
```diff
@@ -55,7 +55,7 @@ fi
 # Convert Mozilla's list of certificates into a PEM file. The -n switch makes
 # it use the local certdata.txt in this folder.
 certNum=$(./mk-ca-bundle.pl -v -n ca-bundle.crt)
-echo "mk-ca-bundle.pl generates "$certNum "cerificates"
+echo "mk-ca-bundle.pl generates $certNum cerificates"
```
karianna, line 58:
```suggestion
echo "mk-ca-bundle.pl generates $certNum certificates"
```

karianna (approved):

karianna:
@zdtsw some conflicts with Andrew's recent fix to this file

security/mk-cacerts.sh
```diff
@@ -117,9 +118,17 @@ for FILE in certs/*.crt; do
...
+if [ "$NO_KEYSTORE" = false ] ; then # for jdk8, jdk11
     num_certs=$("$KEYTOOL" -v -list -storepass changeit -keystore cacerts | grep -c "Alias name:")
 else
     num_certs=$(find certs/* | wc -l)
 fi
 echo "Number of certs processed: $num_certs"
+
+# post verification: (nr.(mk-ca-bundle.pl) - nr.(already imported file) == (nr. (current files in certs folder)) || (nr. (alias in cacerts))
+certNum="$((certNum-alreadyExistsCounter))"
+if [ "$certNum" != "$num_certs" ]; then
```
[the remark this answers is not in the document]
andrew-m-leonard: https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk18u/job/jdk18u-linux-x64-temurin_SmokeTests/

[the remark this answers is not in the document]
zdtsw: ok, i see.
a new test need to perform after the compliation is done. 
something in https://github.com/adoptium/temurin-build/tree/master/test/functional/buildAndPackage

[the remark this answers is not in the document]
andrew-m-leonard: Yes, i'm happy for this PR to do the check it is, but for #3034 we need the SmokeTest change, the reason being because the output from this script:
- <jdk17 : keystore
- >=jdk17 : src file certs
Both get fed into the OpenJDK build that then import either into the image/jdk/lib/security/cacerts keystore, there is always the possibility that build fails or is incorrect, so we need to ensure it is good. One other issue that concerns me is if say a bug or network glitch occurs in [mk-ca-bundle.pl](https://github.com/adoptium/temurin-build/blob/master/security/mk-ca-bundle.pl) Perl script and say it returns 0 certs, how do we catch that? Should we hard code the expected number (ie.135) as it very rarely changes, or do we have a test that actually checks the expected list like OpenJDK do: https://github.com/adoptium/jdk17u/blob/5de8b78504df916d88c9e01941c4e27bfe49e949/test/jdk/sun/security/lib/cacerts/VerifyCACerts.java#L50

andrew-m-leonard (approved):

andrew-m-leonard:
@zdtsw lets merge this ones as-is anyway, as it's good anyway

zdtsw:
> @zdtsw lets merge this ones as-is anyway, as it's good anyway

sure, thanks!
i will keep the issue open and add more work for smoketest

PR #3039, opened 2022-07-11
Don't patch freetype for OpenJ9

keithc-ca:
The patch is only applicable for freetype version 2.8.1 and OpenJ9 doesn't need (or want) that patch applied (for any version).


karianna (approved):
LGTM

gdams (approved):

PR #3043, opened 2022-07-12
freetype: skip unnecessary download on JDK11+

gdams:
Ever since https://github.com/adoptium/temurin-build/pull/421/ was merged, we no longer use the downloaded version of freetype for JDK11+ (instead opting for the bundled version). This PR completely removes the unnecessary download of the freetype tarball.


zdtsw (approved):
LGTM

Issue #3048, opened 2022-07-18
Reproducible build refactor caused build.sh to use a variable JAVA_FEATURE_VERSION that isn't defined when you run makejdk-any-platform.sh on it's own.


karianna:
[Reproducible Build Refactor](https://github.com/adoptium/temurin-build/pull/3027) caused build.sh to use a variable that isn't defined when you run makejdk-any-platform.sh on it's own.

PR #3049, opened 2022-07-18
Fix java version logic for repro params so it does not use undefined variable

sxa:
Fixes https://github.com/adoptium/temurin-build/issues/3048

We should possible look at the use of `OPENJDK_FEATURE_NUMBER` vs `OPENJDK_CORE_VERSION` in `build.sh` more in the future, but while this is inconsistent with most of the other comparisons in the script that use the latter, the former allows us to do the `-ge` comparison as it's purely numeric. The alternative with `OPENJDK_CORE_VERSION` would be to use an explicit list of versions that are not allowed.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -91,9 +91,18 @@ configureShenandoahBuildParameter() {
 }
 
 # Configure reproducible build
-# jdk-17 and jdk-19+ support reproducible builds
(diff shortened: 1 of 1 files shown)


karianna (approved):
LGTM

zdtsw:
@sxa should we try to get this one merged?

PR #3050, opened 2022-07-18
build.sh: Use --with-build-user=VARIANT instead of hard coding to adoptium

sxa:
Currently the reproducible build option to define the build user (picked up by `gcc`) is hard coded in the scripts to `adoptium`. This happens on all values of VARIANT and so end users building their own openjdk with our scripts will have the `adoptium` user name being used. This is undesirable. This adjusts the code so that the default with no variant set will set the user to `hotspot` (The default VARIANT).

NOTE: This will change the builds of Eclipse Temurin fromusing a build user value of  `adoptium` to `temurin`. We could put in an extra if clause to change this if we want it to be left as adoptium.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -106,7 +106,7 @@ configureReproducibleBuildParameter() {
           addConfigureArg "--with-source-date=" "updated"
       fi
       # Ensure reproducible binary with a unique build user identifier
-      addConfigureArg "--with-build-user=" "adoptium"
+      addConfigureArg "--with-build-user=" "${BUILD_CONFIG[BUILD_VARIANT]}"
   fi
 }
 


karianna (approved):
LGTM

sxa:
Approval to ship during lockdown cycle please? (Committers can 👍🏻 for approval)

If we do not do this for the release cycle, then we will not be able to rebuild using the 'latest' scripts after this is merged, since they will different between adoptium (release) and temurin (Later builds)

sxa:
> Not the worst implication if not patched, but also looks quite safe. Would recommend an early check that this does actually work as expected, and the build user is duly set as 'temurin'.
> 
> I'm also assuming the build user is not used in any significant way, so we don't need to debate the chosen value (Eclipse/Adoptium/Temurin)

Yep been running multiple tests in the background before I merge - nearly finished :-)

Issue #3047, opened 2022-07-17
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory


kerner1000:
**What are you trying to do?**

````
temurin-build % ./makejdk-any-platform.sh --docker --clean-docker-build --create-jre-image jdk11u
````

**Expected behaviour:**

Build continues.

**Observed behaviour:**

Build breaks.

I get the following error:

```
Searching for JDK_BOOT_DIR
Guessing JDK_BOOT_DIR: /usr/lib/jvm/jdk11
If this is incorrect explicitly configure JDK_BOOT_DIR using --jdk-boot-dir
[ERROR] The BOOT_JDK_VERSION_NUMBER was not found. Likelihood is that the boot jdk settings properties don't contain a java.specification.version...
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

**Any other comments:**

**Comment by kerner1000** [comment]
Hi @karianna 
thank you for your response!

```
sw_vers
ProductName:	macOS
ProductVersion:	12.4
BuildVersion:	21F79
```

```
docker -v                                                                        
Docker version 20.10.17, build 100c701
```

**Comment by karianna** [comment]
I get further than this with the fix from #3049 but I hit:

```
===RELEASE FILE GENERATED===
Unable to fetch build SHA, does a work tree exist?...
Buildfile: /openjdk/sbin/../cyclonedx-lib/build.xml does not exist!
Build failed 
```

**Comment by sxa** [comment]
Oh this is on an M1 ... I wonder if it's not pulling down the correct docker image for that architecture

**Comment by kerner1000** [comment]
Hi @sxa its better now, but still fails.
This time with

```
Should have the source, I'm at /openjdk/build/src
From /openjdk//./build//src
 * branch                  HEAD       -> FETCH_HEAD
OpenJDK repo tag is jdk-11.0.16+7
Completed configuring the version string parameter, config args are now:  --with-vendor-name="Undefined Vendor" --with-vendor-url= --with-vendor-bug-url=file:///dev/null --with-vendor-vm-bug-url=file:///dev/null --with-version-opt=202207181607 --with-version-pre=beta --with-version-build=7
Building up the configure command...
Adjust configure for reproducible build
/openjdk/sbin/build.sh: line 96: JAVA_FEATURE_VERSION: unbound variable
```

sxa:
[...]

PR #3053, opened 2022-07-19
revert: back to "false" for BUILD_CONFIG[CREATE_SBOM]

zdtsw:
- disable create sbom by default (for build without using ci-jenkins-pipeline)
	
	- enable our CI build from https://github.com/adoptium/ci-jenkins-pipelines/pull/369 
Ref: https://github.com/adoptium/temurin-build/issues/3047

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -1759,7 +1759,7 @@ if [[ "${BUILD_CONFIG[ASSEMBLE_EXPLODED_IMAGE]}" == "true" ]]; then
   printJavaVersionString
   addInfoToReleaseFile
   addInfoToJson
-  if [[ "${BUILD_CONFIG[CREATE_SBOM]}" == "true" ]] && [[ -d "${CYCLONEDB_DIR}" ]]; then
+  if [[ "${BUILD_CONFIG[CREATE_SBOM]}" == "true" ]]; then
     javaHome="$(setupAntEnv)"
     buildCyclonedxLib "${javaHome}"
     generateSBoM "${javaHome}"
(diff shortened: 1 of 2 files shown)


karianna:
Would a better solution not be to pass in false to ${BUILD_CONFIG[CREATE_SBOM]} as a default?

zdtsw:
Think the steps will be
1) add all platform * version in ci-jenkins-pipeline json with --create-sbom   (If we all agree to go this way) 
2) remove `CREATE_SBOM = true `from temurin-build
then we do not really need this PR.

or set CREATE_SBOM to false or do not set it in temurin-build without 1) will cause our build does not generate sbom at the moment.

karianna:
Good PMC topic, I'll post it to them

sxa (commented):
> Would a better solution not be to pass in false to ${BUILD_CONFIG[CREATE_SBOM]} as a default?

+1 from me (as stated elsewhere!) and idefaulting it off would be consistent with what we do for jre creation. I'm not keen on having this PR as-is with checking for the presence of a directory.

karianna:
> Good PMC topic, I'll post it to them

PMC conversation seems to agree with a default of false with our ci-jenkins pipelines setting it to true where required.

sxa:
FYI @andrew-m-leonard

sxa:
NOTE: This must not be merged before https://github.com/adoptium/ci-jenkins-pipelines/pull/369

andrew-m-leonard (approved):

karianna (approved):
LGTM

sxa:
> NOTE: This must not be merged before [adoptium/ci-jenkins-pipelines#369](https://github.com/adoptium/ci-jenkins-pipelines/pull/369)

@karianna As per the above comment this should NOT have been merged yet as the above PR has the code to enable it for our regular builds in a different place and was not yet complete. I'll merge the other one now otherwise the SBOMs will be missing from any rebuilds we have to do this week.

Issue #3057, opened 2022-07-19
jdk17u temurin reproducible builds require a jdk17u bootjdk with jar -date support


andrew-m-leonard:
https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk17u/job/jdk17u-windows-x64-temurin/95/consoleFull
```
17:03:46  * Boot JDK:       openjdk version "16.0.2" 2021-07-20 OpenJDK Runtime Environment Temurin-16.0.2+7 (build 16.0.2+7) OpenJDK 64-Bit Server VM Temurin-16.0.2+7 (build 16.0.2+7, mixed mode, sharing) (at /cygdrive/e/jenkins/tmp/jdk-16)
```
==>
```
17:02:55  checking if Boot JDK jar supports --date=TIMESTAMP... false
```

To enable reproducible builds for jdk17u a jdk17u nightly/release bootjdk that has jar --date support in it.

zdtsw:
after merge, runs from https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk17u/job/jdk17u-windows-x64-temurin/118/consoleFull

```
0:09:53.027  Tools summary:
00:09:53.027  * Environment:    cygwin version 3.3.5-341.x86_64, 2022-05-13 12:27 UTC; windows version 6.3.9600; prefix "/cygdrive"; root "C:\cygwin64"
00:09:53.027  * Boot JDK:       openjdk version "17.0.4.1" 2022-08-12 OpenJDK Runtime Environment Temurin-17.0.4.1+1 (build 17.0.4.1+1) OpenJDK 64-Bit Server VM Temurin-17.0.4.1+1 (build 17.0.4.1+1, mixed mode, sharing) (at /cygdrive/c/openjdk/jdk-17)
```

PR #3076, opened 2022-08-22
feat: update to use jdk17 as boot jdk for jdk17 build to support reproducible builds

zdtsw:
- to enable reproducible build for jdk17 cannot use jdk16 without -date support
	- this will make jdk17 use jdk17 as boot
	- jdk18 use jdk17 as boot
	- jdk19 use jdk19 as boot etc

Ref: https://github.com/adoptium/temurin-build/issues/3057

### DIFF
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -144,8 +144,10 @@ then
   fi
   if [ "${JAVA_FEATURE_VERSION}" == "17" ]; then
     # To support reproducible-builds the jar/jmod --date option is required
-    # which is only available from jdk-17 so we cannot bootstrap with JDK16
+    # which is only available in jdk-17 and from jdk-19 so we cannot bootstrap with JDK16
     JDK_BOOT_VERSION="17"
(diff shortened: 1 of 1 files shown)


karianna (approved):
LGTM

build-farm/make-adopt-build-farm.sh
```diff
@@ -142,10 +142,10 @@ then
     # OpenJ9 only supports building jdk-11 with jdk-11
     JDK_BOOT_VERSION="11"
   fi
-  if [ "${JAVA_FEATURE_VERSION}" == "19" ]; then
+  if [ "${JAVA_FEATURE_VERSION}" == "17" ]; then
```
andrew-m-leonard, line 145: This is not quite right, as now when building jdk-19 it will try and use a jdk-18 bootjdk.
The required jar --date option is only available on jdk-17u and jdk-19.
This if needs to be if 17 OR 19.

andrew-m-leonard (changes requested):

andrew-m-leonard (approved):

PR #3079, opened 2022-08-23
doc: Update FAQ with correct compiler levels

sxa:
I haven't updated the macos ones as I'm not entirely sure of the best way to reference them, but I believe the rest of this information is now correct. We should endeavour to keep this table accurate when we make changes to the build scripts.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -39,12 +39,16 @@ In terms of compilers, these are what we currently use for each release:
 
 | Version | OS      | Compiler |
 |---------|---------|----------|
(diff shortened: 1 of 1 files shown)


karianna (approved):
LGTM - one day we'll just pull this from meta data

sxa:
> LGTM - one day we'll just pull this from meta data

Yeah particularly with the SBOM creation it's all around somewhere :-)

zdtsw (approved):

PR #3080, opened 2022-08-23
Docs: Add .sig regexp to RELEASING.md doc

Haroon-Khel:
.sig regexp is missing from the RELEASING.md doc, causing releases to go out without .sig files


sxa (approved):
Thanks

karianna (approved):
LGTM

Issue #3075, opened 2022-08-19
jdk.compiler.jmod ct.sym non-deterministic on s390xLinux


andrew-m-leonard:
jdk-17 s390xLinux jdk.compiler.jmod differs by ct.sym

Needs investigating:
1. Is same on jdk-19 ?
2. Is it consiostently different?
3.

sxa:
Couple of quick notes - the differences I saw (which resulted in Andrew raising this issue) were between running on two RHEL7 machines. The compiler was GCC 10.3 on both, and the boot JDK was 17.0.4 on both. The shared class cache was also different but that is expected. The following tarballs contain the `jdk.compiler.jmod`, `ct.sym`, `metadata` directory and build log from each of the two builds which show the differences:

[79.tar.gz](https://github.com/adoptium/temurin-build/files/9383846/79.tar.gz)
[77.tar.gz](https://github.com/adoptium/temurin-build/files/9383849/77.tar.gz)

sxa:
`ct.sym` is sensitive to the time zone on the machine which has an hour difference between the two systems. That likely explains the difference.

sxa:
Files that are different based on TZ differences:
```
[sxa@9757757beb16 build-farm]$ diff -r jdk-20+13.4 jdk-20+13
Binary files jdk-20+13.4/jmods/java.base.jmod and jdk-20+13/jmods/java.base.jmod differ
Binary files jdk-20+13.4/jmods/jdk.compiler.jmod and jdk-20+13/jmods/jdk.compiler.jmod differ
diff -r jdk-20+13.4/lib/classlist jdk-20+13/lib/classlist
1206a1207
> sun/util/calendar/ZoneInfoFile$Checksum
Binary files jdk-20+13.4/lib/ct.sym and jdk-20+13/lib/ct.sym differ
Binary files jdk-20+13.4/lib/modules and jdk-20+13/lib/modules differ
Binary files jdk-20+13.4/lib/server/classes.jsa and jdk-20+13/lib/server/classes.jsa differ
Binary files jdk-20+13.4/lib/server/classes_nocoops.jsa and jdk-20+13/lib/server/classes_nocoops.jsa differ
[sxa@9757757beb16 build-farm]$ 
```

PR #3090, opened 2022-09-02
Force TZ=UTC for reproducible release builds

sxa:
Fixes https://github.com/adoptium/temurin-build/issues/3075

While most timestamps in the builds are neutralised when some of the files are produced there are some files which have timestamps that are affected by the time zone setting, so if you build on a machine in a different time zone then the builds are not identical. I've seen this on xLinux (Timestamps within the `ct.sym` zipfile are different) as well as in the issue reference above between two Linux/s390x systems. This fix will set `TZ=UTC` alongside the other options when the reproducible builds are enabled.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -98,8 +98,8 @@ configureReproducibleBuildParameter() {
       # Enable reproducible builds implicitly with --with-source-date
       if [ "${BUILD_CONFIG[RELEASE]}" == "true" ]
       then
-           # TZ issue: https://github.com/adoptium/temurin-build/issues/3075
-           export TZ=UTC
+          # TZ issue: https://github.com/adoptium/temurin-build/issues/3075
+          export TZ=UTC
           # Use release date and disable CCache( remove --enable-ccache if exist)
           addConfigureArg "--with-source-date=version"  " --disable-ccache"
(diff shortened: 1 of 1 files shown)


sxa:
(Draft as the fix in here may not be working, although it does when I set the variable standalone in the calling shell)

sxa:
Smoke test failure raised at https://github.com/adoptium/temurin-build/issues/3093 but seems unlikely it's related to this PR

sxa:
> Smoke test failure raised at #3093 but seems unlikely it's related to this PR

Have done a revert on the change and the checks all appear to be passing ... Hmmm. Will re-push and try again with `TZ=UTC`

sxa:
It's looking ok this time - some of the non-Alpine linuxes are now passing. No idea what happened there. I added a change to revert what I did and have now removed the revert and it seems ok, so adding some explicit reviewers to this PR now.

sxa:
NOTE: I will not be around for the next week so if there are any problems when this is merged, others should feel empowered to revert/fix as appropriate.

sxa:
(Final push to do a whitespace adjustment)

karianna (approved):
LGTM

PR #3095, opened 2022-09-08
add jdk11u fixpath path for windows aarch64

gdams:
See https://github.com/microsoft/openjdk-aarch64#building-jdk-11 for context.

This will fix the failing windows aarch64 jdk11 builds

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -168,6 +168,14 @@ getOpenJDKUpdateAndBuildVersion() {
   cd "${BUILD_CONFIG[WORKSPACE_DIR]}"
 }
 
+addFixpathPatch() {
+  # Fixpath patch required for JDK11u Windows aarch64 builds
+  if [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" = "${JDK11_CORE_VERSION}" ] && [ "${ARCHITECTURE}" = "aarch64" ]; then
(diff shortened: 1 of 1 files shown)


gdams:
@zdtsw we are not using a devkit so that wouldn't be giving us a problem?

zdtsw (approved):

sxa:
From a secure development perspective I don't like this. To provide some context is this downloading a binary to replace a version that would normally be compiled (from fixpath.c) from source during the build? Is it possible to build it from modified source during the build?

I feel these points need to be clarified in this PR for historic reference as a minimum.

sxa:
And even if we are to leave this in place it should be checksummed on download to ensure it hasn't been tampered with.

jiekang:
@gdams @karianna just to be clear on my understanding, this patch is adding the download of an exe binary from Microsoft GH release page to the build process for windows aarch64, is that right?

I'd echo Stewart's sentiments that at the very least, this needs to be hardened as soon as possible. It would have been preferred to do that in the initial PR.

karianna:
> @gdams @karianna just to be clear on my understanding, this patch is adding the download of an exe binary from Microsoft GH release page to the build process for windows aarch64, is that right?
> 
> I'd echo Stewart's sentiments that at the very least, this needs to be hardened as soon as possible. It would have been preferred to do that in the initial PR.

Correct, @gdams will follow up with the appropriate fix.

PR #3098, opened 2022-09-13
Compile OpenJ9 Windows jdk11 with VS2019

pshipton:
This change is already used for OpenJ9/Semeru 11.0.17 builds via https://github.com/ibmruntimes/temurin-build/pull/25

FYI OpenJ9 also needed this OpenJDK change https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/574 related to using VS2019, to include vcruntime140_1.dll into the build. Perhaps Temurin would benefit from the same.


karianna (approved):
LGTM by @gdams to check as we had to fix an issue around this recently.

gdams:
> LGTM by @gdams to check as we had to fix an issue around this recently.

Yeah happy with this change, I actually backported this down into jdk11u-dev recently so just waiting for it to cut across to jdk11u

gdams (approved):

PR #3113, opened 2022-10-11
This is Liz

Pip7747:


smlambert:
Hi @Pip7747 - if you are following along on the directions from the [first-contributions repository](https://github.com/firstcontributions/first-contributions), this is one step that we do not require or use at our project, there is no need to change the Contributing.md file

I have also mentioned it in [this Slack message](https://adoptium.slack.com/archives/C01PUE50XQQ/p1665451093021779), but it is definitely easy to not have seen that message.

PR #3119, opened 2022-10-15
chore: remove jdk18 in GH action test

zdtsw:
dont waste resource test jdk18

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [linux]
-        version: [jdk8u, jdk11u, jdk17u, jdk18u, jdk19, jdk] #jdk head == jdk20
+        version: [jdk8u, jdk11u, jdk17u, jdk19, jdk] #jdk head == jdk20
         variant: [temurin]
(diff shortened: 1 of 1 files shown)


gdams (approved):

karianna (approved):

sxa:
@karianna since it's not critical and hasn't had the appropriate overrides to merge during lockdown we shouldn't really have merged this. (It's also explicltly labelled as `awaiting-merge-post-release`)

gdams:
> @karianna since it's not critical and hasn't had the appropriate overrides to merge during lockdown we shouldn't really have merged this. (It's also explicitly labelled as awaiting-merge-post-release)

I've toyed with the idea of having a GitHub action that we enable on the core repositories that explicitly blocks PR's while we are in lock-down until a PMC member approves

sxa:
> GitHub action that we enable on the core repositories that explicitly blocks PR's while we are in lock-down until a PMC member approves

I'd be a bit reluctant on that as a general rule as there may be situations where it's needed to resolve an urgent breakage where it can be justified to bypass it. Also in this case it was two PMC members who did the approvals + merge anyway :-)

PR #3130, opened 2022-10-21
Fix download of boot JDK when cross-compiling

luhenry:


karianna:
@luhenry are you able sign the ECA?  Actually I thought you already did...

gdams (approved):

karianna (approved):
LGTM

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -150,9 +150,14 @@ if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
       export "${BOOT_JDK_VARIABLE}"="/usr/lib/jvm/jdk-${JDK_BOOT_VERSION}"
     elif [ "$JDK_BOOT_VERSION" -ge 8 ]; then # Adoptium has no build pre-8
       mkdir -p "$bootDir"
+      export downloadArch
```
sxa, line 153: Why is this line needed - can we just remove it? And will this break native builds on the riscv64 boards?

luhenry: It's the same approach as https://github.com/adoptium/temurin-build/blob/fc76e274cf3de0e0740a03917a7fb7977e950c47/build-farm/platform-specific-configurations/windows.sh#L49-L54

sxa (approved):
This should be ok now that the book JDK architecture is fixed, thus preventing it failing on a native build. Merging now.

PR #3160, opened 2022-11-18
Fix actions to specific hashes

tellison:
Part of https://github.com/adoptium/adoptium/issues/187

Fixes actions used to specific version hashes, and specifies permissions where appropriate.
Most hashes captured by the StepSecurity tool, but all manually checked and version comment added. Some were updated to later versions.


tellison:
Grr, need to remove the original bot commit

PR #3161, opened 2022-11-21
Pin GitHub action dependencies to hashes

tellison:
Part of https://github.com/adoptium/adoptium/issues/187

Note this includes some dependency action version updates in addition to fixing to hash numbers.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,7 +60,7 @@ jobs:
             variant: bisheng
             image: adoptopenjdk/centos7_build_image
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
 
     - name: Build Linux
       run: ./build-farm/make-adopt-build-farm.sh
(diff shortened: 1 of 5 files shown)


karianna (approved):
LGTM

.github/workflows/build.yml
```diff
@@ -167,14 +167,14 @@ jobs:
         imageroot=$(find "${HOME}/JDK" -name release -type f)
         echo "TEST_JDK_HOME=$(dirname "${imageroot}")" >> "$GITHUB_ENV"
     - name: Smoke test
-      uses: adoptium/run-aqa@v2
+      uses: adoptium/run-aqa@6bacb4e732ad546eda1b09665b9067cdc87651f4 # v2
```
zdtsw, line 170: maybe good to pin https://github.com/adoptium/run-aqa/releases/tag/v2.0.1 which is the latest (last week) released version?

tellison: Yes, I did consider that, but wasn't sure that was intended as GitHub was still picking the existing v2.
@sophia-guo will it be OK for the smoke tests to use the run-aqa action at v2.0.1 ?

sophia-guo: v2.0.1 supports parallel, but it still needs improvement. For now I think v2 is good.

PR #3166, opened 2022-11-25
Add --build-reproducible-date=ISO8601_date build option to allow reproducible nightly builds

andrew-m-leonard:
Add --build-reproducible-date=ISO8601_date build build arg, which sets the build date & time used in the build
so for nightly builds the pre/post version output date, and all build dates are reproducible.

A follow on PR will update Temurin nightly builds to use this along with a "dual-build" with the same nightly "date", then invoking the reproducible comparison job, publishing the reproducible % to the metadata and subsequent publish to website... (ok, maybe more than one PR to follow!)

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -104,19 +104,28 @@ configureReproducibleBuildParameter() {
           addConfigureArg "--with-source-date=version"  " --disable-ccache"
           CONFIGURE_ARGS="${CONFIGURE_ARGS//--enable-ccache/}"
       else
-          # Use build date
-          addConfigureArg "--with-source-date=" "updated"
+          if [[ -n "${BUILD_CONFIG[BUILD_REPRODUCIBLE_DATE]}" ]]; then
+              # Use supplied date
+              addConfigureArg "--with-source-date=" "${BUILD_CONFIG[BUILD_REPRODUCIBLE_DATE]}"
 
-          # Specify --with-hotspot-build-time to ensure dual pass builds like MacOS use same time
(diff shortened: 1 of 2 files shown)


karianna (approved):
LGTM

karianna (changes requested):
Looks like GH Actions for the Alpine build couldn't take the new setting

andrew-m-leonard:
> Looks like GH Actions for the Alpine build couldn't take the new setting

@karianna thanks, I was using the wrong empty string test !

tellison (approved):

karianna (approved):

zdtsw:
bit late, but try to understand this PR:
In Jenkins,  we need to pass a new field `BUILD_REPRODUCIBLE_DATE` from jenkins job, which set the correct 'date'
The build result binary should be able to be comparison(as long as 2 builds have same `BUILD_REPRODUCIBLE_DATE` value, even the `releaseType` is "`Nightly`" if  it is `jdk17+` ?

andrew-m-leonard:
> bit late, but try to understand this PR: In Jenkins, we need to pass a new field `BUILD_REPRODUCIBLE_DATE` from jenkins job, which set the correct 'date' The build result binary should be able to be comparison(as long as 2 builds have same `BUILD_REPRODUCIBLE_DATE` value, even the `releaseType` is "`Nightly`" if it is `jdk17+` ?

Yes, the next PR logic to use this option will use it for jdk17, and jdk19+

PR #3176, opened 2022-12-08
github: only run GH action check "build" if it is scripts changes

zdtsw:
- no need run it if it is .md files :moneybag:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,10 @@ name: Build
 on:
   pull_request:
     branches: [ master ]
+    paths:
+      - "build-farm/**"
+      - "sbin/**"
+      - "**.sh"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -3,6 +3,10 @@ name: Build
 on:
   pull_request:
     branches: [ master ]
+    paths:
+      - "build-farm/**"
+      - "sbin/**"
+      - "**.sh"
```
gdams, line 9: You should also include this specific workflow file (so we can test changes made to it)

Should just be able to add .github/workflows/build.yml as a path.

gdams (changes requested):
1 change please and then this should be good to go

gdams (approved):

karianna (approved):
LGTM

PR #3180, opened 2022-12-14
Update branch name

tellison:
temurin-build `main` branch has the legacy name `master`. Update the filter to run there.

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -2,7 +2,7 @@ name: OSSF Scorecard
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


gdams (approved):

zdtsw:
unless I misunderstand, from https://github.com/adoptium/temurin-build/branches we do not have a "main" branch in temurin-build repo

smlambert:
re: https://github.com/adoptium/temurin-build/pull/3180#issuecomment-1357500565 - correct, it does not have a main branch and that is the point of this PR (to use the legacy name rather than main).

smlambert (approved):

PR #3188, opened 2022-12-20
switch freetype to clone source from GitHub

gdams:
Now that the FreeType source is available in [GitHub](https://github.com/freetype/freetype) we should switch to cloning the FreeType source from there rather than relying on our mirror on the Jenkins server. Not only does this mean we don't have to update the mirror each time we bump the version, but it also means that we can use an immutable SHA in the SBOM.

Checking with @andrew-m-leonard that I've set the details in the SBOM correctly. Ultimately I've switched to us storing `https://github.com/freetype/freetype/commit/${FREETYPE_SHA}` (matching the way we store other Git refs in the SBOM.

Note that compiling the FreeType source from GitHub adds a dependency on `automake` for the macOS builds. This doesn't seem to be present on our current build machines so will need adding (I can submit a PR to the infra repo)


sxa (commented):
A couple of points:
- This is potentially leaving us slightly more open to risk because while we are recording the SHA we're not using that for the clone so there is a possibility of undetected corruption of the branch in the upstream repository causing problems (unless the SBOMs are checked for changes), which is not a risk with the fixed SHA which we verify from our download server.
- If the version in github is the same as the one we're currently downloading, where has the extra macos dependency come from? That would suggest they're not the same.

gdams:
> This is potentially leaving us slightly more open to risk because while we are recording the SHA we're not using that for the clone so there is a possibility of undetected corruption of the branch in the upstream repository causing problems (unless the SBOMs are checked for changes), which is not a risk with the fixed SHA which we verify from our download server.

I was thinking about this, one option I'd considered was passing the git ref rather than the tag name, that way it cannot be corrupted, what do you think?

> If the version in github is the same as the one we're currently downloading, where has the extra macos dependency come from? That would suggest they're not the same.

It appears that the one we currently host has had autogen already run on it so we can skip straight to running configure/make. The GitHub source doesn't have this hence the change and added dependency.

sxa:
> > This is potentially leaving us slightly more open to risk because while we are recording the SHA we're not using that for the clone

> I was thinking about this, one option I'd considered was passing the git ref rather than the tag name, that way it cannot be corrupted, what do you think?

Yeah it's the only real way of protecting against an unexpected upstream change I think - or we could extract as you currently are and then just check that the list commit matches an "expected" SHA (Similar to what we were doing with the download) which could flag it if the upstream had done something silly. I think I'd prefer one of those two options. I'd lean a little more towards the second but ok with either.

> > If the version in github is the same as the one we're currently downloading, where has the extra macos dependency come from? That would suggest they're not the same.

> It appears that the one we currently host has had autogen already run on it so we can skip straight to running configure/make. The GitHub source doesn't have this hence the change and added dependency.

Interesting - ok that explains it. Hopefully that's the only difference and we won't end up with any functional differences as a result of us generating ourselves.

gdams:
> Yeah it's the only real way of protecting against an unexpected upstream change I think - or we could extract as you currently are and then just check that the list commit matches an "expected" SHA (Similar to what we were doing with the download) which could flag it if the upstream had done something silly. I think I'd prefer one of those two options. I'd lean a little more towards the second but ok with either.

Okay I'll look at setting that up

> Interesting - ok that explains it. Hopefully that's the only difference and we won't end up with any functional differences as a result of us generating ourselves.

I've done a diff and that's the only change between ours and theirs

sbin/prepareWorkspace.sh
```diff
@@ -452,25 +451,35 @@ checkingAndDownloadingFreeType() {
...
+      cd freetype || exit
+      git checkout "${BUILD_CONFIG[FREETYPE_FONT_VERSION]}" || exit
+      cd .. || exit
+      ;;
+    esac
+  
+    # Fetch the sha for the commit we just cloned
+    cd freetype || exit
+    FREETYPE_SHA=$(git rev-parse HEAD) || exit
+    FREETYPE_BUILD_INFO="https://github.com/freetype/freetype/commit/${FREETYPE_SHA}"
```
karianna, line 475: We get the sha but we don't check it?  Or am I mistaken?

gdams: the SHA is fetched for the SBOM (we don't need to check it IMO as I trust git to checkout the SHA we passed it originally)

karianna (approved):
LGTM

sxa (approved):
Thanks for the change. I'm ok with this now but it would be good to have an explicit 'ok from @andrew-m-leonard too on this if he's around today.

gdams:
Thanks @sxa, can you also approve https://github.com/adoptium/infrastructure/pull/2860 as it's a prerequisite?

andrew-m-leonard (approved):

PR #3211, opened 2023-01-10
add the --sign operation of TemurinGenSBOM.java

julian55455:
add the --sign operation of TemurinGenSBOM.java

Ref:  https://github.com/adoptium/temurin-build/issues/3158

PR #3215, opened 2023-01-12
Created a sign-sbom script to build cyclonedx library and sign the sbom

julian55455:
Created a sign-sbom script to build cyclonedx library and sign the sbom

### DIFF
--- a/build-farm/sign-sbom.sh
+++ b/build-farm/sign-sbom.sh
@@ -15,7 +15,7 @@
 ################################################################################
 
 # Import common SBOM functions
-source ../sbin/common/sbom.sh
+source ./sbin/common/sbom.sh
 
 # Get Jenkins job parameters
 privateKey=$1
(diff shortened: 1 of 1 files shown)


karianna:
@julian55455 Linter fails.

andrew-m-leonard:
@julian55455 This will be the jenkins called script, use https://github.com/adoptium/temurin-build/blob/master/build-farm/sign-releases.sh
as an example. so it will be called with the sbom JSON artifact to be signed copied from the upstream job to a folder called workspace/target/...json

eg:
```
workspace/target/OpenJDK17U-sbom_arm_linux_hotspot_17.0.6_10.json
```

You sign it, and then the Jenkins job itself will archive the signed artifact.

build-farm/sign-sbom.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
andrew-m-leonard, line 1: Use this example as the basis of what this script needs to do: https://github.com/adoptium/temurin-build/blob/master/build-farm/sign-releases.sh

andrew-m-leonard (changes requested):

zdtsw:
this script wont do any harm, as long as it is not get called.
but good to have #3211 fixed first then correct the script and add a caller

PR #3209, opened 2023-01-10
move buildCyclonedxLib to Build the CycloneDX Java library and app used for SBoM generation from build.sh to sbom.sh

julian55455:
move buildCyclonedxLib to Build the CycloneDX Java library and app used for SBoM generation from build.sh to sbom.sh

Ref: https://github.com/adoptium/temurin-build/issues/3158

PR #3236, opened 2023-02-03
Ensure TemurinGenSOM.java build with jdk-17+ for openkeystore support

andrew-m-leonard:
Fix build.sh to build TemurinGenSOM.java with jdk-17+ as openkeystore won't compile.

Fixes up: https://github.com/adoptium/temurin-build/pull/3209

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -696,7 +696,7 @@ setupAntEnv() {
   fi
 
   if [ ! -z "$javaHome" ]; then
-    javaVer=$("$javaHome}/bin/java -XshowSettings:properties -version 2>&1 | grep 'java.runtime.version' | sed 's/^.*= //' | tr -d '\r' | tr '+' '.' | cut -d'.' -f1")
+    javaVer=$("${javaHome}/bin/java -XshowSettings:properties -version 2>&1 | grep 'java.runtime.version' | sed 's/^.*= //' | tr -d '\r' | tr '+' '.' | cut -d'.' -f1")
   fi
 
   if [[ -z "$javaVer" ]] || [[ "$javaVer" -lt "17" ]]; then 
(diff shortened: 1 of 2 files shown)


smlambert (approved):
👍

sbin/build.sh
```diff
@@ -684,38 +684,18 @@ createOpenJDKFailureLogsArchive() {
...
-  # fall back to use JDK_BOOT_DIR which is set in make-adopt-build-farm.sh
-    javaHome="${BUILD_CONFIG[JDK_BOOT_DIR]}"
-  else
-    echo "Unable to find a suitable JAVA_HOME to build the cyclonedx-lib"
-    exit 2
-  fi
-  echo "${javaHome}"
-}
+  # TemurinGenSBOM.java requires jdk-17+ to build openkeystore
+  apiURL="https://api.adoptium.net/v3/binary/latest/17/ga/${OPERATING_SYSTEM}/${BUILD_CONFIG[OS_ARCHITECTURE]}/jdk/hotspot/normal/eclipse"
```
sxa, line 692: Do we really need to do this? Most of our build machines have java 17 installed in a known location so downloading it every time seems a bit over the top.

zdtsw: Solaris is a valid point.

andrew-m-leonard: @sxa What environment would point at it?
Also, the above is generic for user of this script

andrew-m-leonard: Yeah thought about Solaris, not sure what we do there....yet

andrew-m-leonard: I'd like to just get some thing building rather than none :-)

sxa: > Also, the above is generic for user of this script
Yes but "if it's not in our location, do the download" would seem a sensible thing to do (and the fewer things we download during the process the better and more secure we will be.

andrew-m-leonard: So openkeystore is not going to build on Solaris, so I think we're going to have to re-factor that code out of TemurinGenSBOM and into another signing app, called in a standalone job

andrew-m-leonard: @sxa in fact i'm thinking refactor the whole class into a separate invoked build-farm job, a bit like GPG signing. Trying to perform at the end of the build is troublsome

andrew-m-leonard:
Needs re-working with thought on jdk8 Solaris and possibly other issues..

PR #3300, opened 2023-03-27
Add configuration for jdk8u, jdk11u, and jdk17u on riscv64

luhenry:
The backports of Linux/RISC-V to JDK 8, 11, and 17 are in progress. To get a head's start on building and testing these backports, we want to special-case these repositories.

### DIFF
--- a/configureBuild.sh
+++ b/configureBuild.sh
@@ -77,6 +77,13 @@ doAnyBuildVariantOverrides() {
   if [[ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_SAP}" ]]; then
     local branch="sapmachine10"
     BUILD_CONFIG[BRANCH]=${branch:-${BUILD_CONFIG[BRANCH]}}
+  elif [[ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_HOTSPOT}" ]] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ]; then
+    if [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK8_VERSION}" ] \
+      || [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK11_VERSION}" ] \
+      || [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK17_VERSION}" ]; then
+      local branch="riscv-port"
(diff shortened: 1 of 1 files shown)


karianna (approved):

sxa (changes requested):

configureBuild.sh
```diff
@@ -179,6 +186,14 @@ setRepository() {
     suffix="openjdk/aarch32-port-jdk8u";
   elif [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK8_CORE_VERSION}" ] && [[ "${BUILD_CONFIG[OS_FULL_VERSION]}" == *"Alpine"* ]] && [[ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_TEMURIN}" ]]; then
     suffix="adoptium/alpine-jdk8u";
+  elif [[ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_HOTSPOT}" ]] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ]; then
+    if [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK8_VERSION}" ] \
+      || [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK11_VERSION}" ] \
+      || [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" == "${JDK17_VERSION}" ]; then
+      suffix="openjdk/riscv-port-${BUILD_CONFIG[OPENJDK_CORE_VERSION]}"
```
sxa, line 193: Similarly - this will need to be `OPENJDK_FOREST_NAME` instead of `OPENJDK_CORE_VERSION` similar to the lines underneath this to ensure it has the `u` suffix (Or adding an explicit `u` after the variable would also work)

sxa:
As an additional note, sice we currently don't have a suitable bootstrap JDK for RISC-V for JDK8+11 we will need to look at how to run 8 using this script (We have Bisheng or OpenJ9 builds for 11 that we can use)

luhenry:
For bootstraping, can we provide a one-off for the first ever build?

sxa:
> For bootstraping, can we provide a one-off for the first ever build?

Yep, that's the right thing to do IMHO

luhenry:
@sxa are the changes good to you?

sxa (approved):
Done a quick test with 17 and it seems to be ok, so approving :-)

Issue #3245, opened 2023-02-08
Add Reproducible Build Comparisons to Jenkins Pipelines 


andrew-m-leonard:
Add the ability in the ci-jenkins-pipelines to support dual build with integrated reproducible build comparison.
Only supported for certain versions: jdk-17+

The following existing toolset: https://github.com/adoptium/ci-jenkins-pipelines/tree/master/tools/reproduce_comparison
provides some tools, these need moving to temurin-build repo.
These are currently being updated/extended with some in-progress work here: https://github.com/andrew-m-leonard/openjdk-build/tree/repro_validate/tooling
note renaming of compareLinux.sh to [linux_repro_build_compare.sh](https://github.com/andrew-m-leonard/openjdk-build/blob/repro_validate/tooling/linux_repro_build_compare.sh)

**Suggested solution:**
- Utilise aqa-test framework to create a new "reproducible-test" job
- test job performs a second build by invoking the standard build job as a sub-job, but passing build_args of --build-reproducible-date "ISO8601 date", where the date is the same date used by the original build
- then compare the two builds, eg: https://github.com/andrew-m-leonard/openjdk-build/blob/4871528098f006f9680d3f3579351fd1ff823a37/tooling/linux_repro_build_compare.sh#L105
- produce a test result which includes the reproducible percentage, eg see: https://github.com/andrew-m-leonard/openjdk-build/blob/4871528098f006f9680d3f3579351fd1ff823a37/tooling/windows_repro_compare.sh#L45

zdtsw:
related to https://github.com/adoptium/ci-jenkins-pipelines/issues/397

sophia-guo:
What is the expected results? Looking at the console output  https://ci.adoptium.net/job/jdk17u-linux-x64-temurin_reproduce_compare/21/console, what issues or other steps should be taken care of ? @andrew-m-leonard

andrew-m-leonard:
@sophia-guo So the scripts need a JDK in the PATH so they can run "jimage" and "jmod". I suggest simply unpacking "original" into a 2nd folder that you then use as a JDK in PATH, so you have 3 folders:
1. originalJdkHome
2. originalHome
3. reproducedHome
and:
export PATH=<path to originalJdkHome>/bin:$PATH

The linux_repro_compare script takes the java home folder as the path, so I think you need to call it as:
[...]

PR #3313, opened 2023-04-05
Ensure build timestamp is consistent and set in SBOM

andrew-m-leonard:
To enable reproducible build testing:
- Set consistent nightly build timestamp and filename, consistent with timestamp used by the openjdk make file, thus enabling reproducibility...
- Set build timestamp into SBOM, so it's easily available to re-build reproducibly
- For reproducible "release" builds, get build timestamp generated by openjdk make file --with-source-date=version logic, so it can be set in SBOM


smlambert (approved):

sbin/build.sh
```diff
@@ -104,28 +104,23 @@ configureReproducibleBuildParameter() {
...
           CONFIGURE_ARGS="${CONFIGURE_ARGS//--enable-ccache/}"
       else
-          if [[ -n "${BUILD_CONFIG[BUILD_REPRODUCIBLE_DATE]}" ]]; then
-              # Use supplied date
-              addConfigureArg "--with-source-date=" "${BUILD_CONFIG[BUILD_REPRODUCIBLE_DATE]}"
-
-              # Specify --with-hotspot-build-time to ensure dual pass builds like MacOS use same time
-              # Use supplied date
-              addConfigureArg "--with-hotspot-build-time=" "${BUILD_CONFIG[BUILD_REPRODUCIBLE_DATE]}"
+          # Use BUILD_TIMESTAMP date
```
zdtsw, line 107: just thinking, since we will not build jdk18 any more, e.g line 96 can be simplify to` if [[ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -ge 17 ]]`

sbin/build.sh
```diff
@@ -104,28 +104,23 @@ configureReproducibleBuildParameter() {
...
+
+          # Convert BUILD_TIMESTAMP to seconds since Epoch
+          local buildTimestampSeconds
```
karianna, line 110: extract to util function?

sbin/build.sh
andrew-m-leonard: I think i'd rather leave it, as these scripts are for any user to use, and they may want for some reason to build 18

sbin/build.sh
andrew-m-leonard: yeah, good idea

andrew-m-leonard: added

zdtsw (approved):

karianna (approved):

sophia-guo:
Related #3245

Issue #3270, opened 2023-03-02
Should temurin build versions use --with-version-opt=LTS for LTS versions?


andrew-m-leonard:
Some vendors (in fact i've not found one that doesn't!) use --with-version-opt=LTS which results in version strings being extended with for example "jdk-17.0.6+10-LTS"
See: https://github.com/adoptium/jdk17u/blob/648f3f6a8fe9ea2e00b60710afa44208cfbc2c49/src/java.base/share/classes/java/lang/VersionProps.java.template#L69
```
private static final boolean isLTS =
        "@@VERSION_OPT@@".startsWith("LTS");
```

JEP [322](https://openjdk.org/jeps/322) states:
﻿﻿**If a release is part of a series of releases for which an implementor offers long-term support then the value of $OPT should start with "LTS", e.g., 11.0.2+13-LTS. This will cause "LTS" to be displayed prominently in the output of java --version, etc., more on which below.**

eg.:
**Zulu:**
```
openjdk version "17.0.6" 2023-01-17 LTS
OpenJDK Runtime Environment Zulu17.40+19-CA (build 17.0.6+10-LTS)
OpenJDK 64-Bit Server VM Zulu17.40+19-CA (build 17.0.6+10-LTS, mixed mode, sharing)
```
**Red Hat:**
```
openjdk version "17.0.6" 2023-01-17 LTS
OpenJDK Runtime Environment (Red_Hat-17.0.6.0+10-1) (build 17.0.6+10-LTS)
OpenJDK 64-Bit Server VM (Red_Hat-17.0.6.0+10-1) (build 17.0.6+10-LTS, mixed mode, sharing)
```
**Oracle:**
```
java version "17.0.6" 2023-01-17 LTS
Java(TM) SE Runtime Environment (build 17.0.6+9-LTS-190)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.6+9-LTS-190, mixed mode, sharing)
```

zdtsw:
I guess the question is for official releases if Temurin should have '-LTS' appended in the version?

karianna:
Will add to PMC as we know changing version strings can impact subscribers

jerboaa:
https://openjdk.org/jeps/223 introduces API, `Runtime.Version`, for version parsing. Is version output still a concern? If it was, wouldn't the same be a concern for cross-vendor switching?

jerboaa:
There are also jlink plugins which allow switching those:

```
[...]

PR #3340, opened 2023-05-11
Set --with-version-opt=LTS for LTS versions from jdk-21

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3270

Sets --with-version-opt for LTS release builds jdk-21+

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -389,8 +389,8 @@ configureVersionStringParameter() {
       addConfigureArg "--with-version-opt=" "${dateSuffix}"
       addConfigureArg "--with-version-pre=" "beta"
     else
-      # "LTS" builds (every 2 years) from jdk-21 will use "LTS" version opt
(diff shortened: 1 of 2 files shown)


sbin/build.sh
```diff
@@ -389,7 +389,12 @@ configureVersionStringParameter() {
       addConfigureArg "--with-version-opt=" "${dateSuffix}"
       addConfigureArg "--with-version-pre=" "beta"
     else
-      addConfigureArg "--without-version-opt" ""
+      # "LTS" builds (every 2 years) from jdk-21 will use "LTS" version opt
+      if [[ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -ge 21 ]] && [[ $(((BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]-21) % 4)) == 0 ]]; then
```
karianna, line 393: Probably nitpicking, but perhaps that logic could be extracted to a named function in case it changes

karianna (approved):

sbin/build.sh
andrew-m-leonard: Yeah, good idea

steelhead31 (approved):
LGTM

Issue #3341, opened 2023-05-11
Default embedded HOTSPOT_BUILD_USER creates non-comparable jvm.dll/.so


andrew-m-leonard:
For reproducible/comparable builds, the default hotspot build user name that ends up in the internal vm version info string, eg:

OpenJDK 64-Bit Server VM (19+23) for bsd-aarch64 JRE (19+23), built on 2022-09-20T00:00:00Z by **"adoptium"** with clang Apple LLVM 12.0.0 (clang-1200.0.32.29)

is different between Variant's which means if you try and compare same builds between variant's jvm.dll/.so differs.
The default build user name is arbitrary, but we should use an arbitrary unique default value, eg."admin", rather than using the Variant name.

PR #3342, opened 2023-05-11
Make embedded jvm internal build user name unique

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3341
by making the internal binary text for the "build user" unique.


steelhead31 (approved):
LGTM

zdtsw (approved):

sxa:
I'd probably have gone with `adoptium` rather than `admin` (which isn't a default username on any system AFAIK) since it's the result of building from the adoptium scripts (although another neutral ID would also be ok), but I won't insist on a change now :-)

Issue #3343, opened 2023-05-15
temurin-build scripts don't allow the complete set of user vendor strings


andrew-m-leonard:
The temurin build-scripts currently only support --vendor, to be consistent we should also support vendor-url, vendor-bug-url, vendor-vm-bug-url.

PR #3344, opened 2023-05-15
Enable build scripts support for user vendor strings

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3343

Change the ERROR exit policing of the vendor name, to just a WARNING.
And allow setting the vendor-url, vendor-bug-url, vendor-vm-bug-url.

### DIFF
--- a/makejdk-any-platform.1
+++ b/makejdk-any-platform.1
@@ -176,6 +176,18 @@ If true (default), a custom cacerts file will be generated based on Mozilla's
 list of CA certificates (see folder security/). If false, the file shipped by
 OpenJDK will be used. 
 .TP
+.BR \-\-vendor " " \fI<vendor>\fR
+specify the vendor name
+.TP
+.BR \-\-vendor-url " " \fI<vendor url>\fR
+specify the vendor url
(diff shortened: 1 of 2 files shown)


karianna (approved):

sxa (changes requested):
Also https://github.com/adoptium/temurin-build/blob/master/makejdk-any-platform.1 should be updated to document the new parameters.

andrew-m-leonard:
> Is there any reason why someone using their own value (through customer configure args) wouldn't be using a VARIANT of `hotspot`?
> 
> I'm curious as to what use case would cause this to be a problem, and by removing this check we're not going to be trapping typos on the valid VARIANT values.

@sxa Hmm, you make an interesting point, I need to go back and try this again, as the problems I was seeing related to vendor vs variant interchange in our scripts...

andrew-m-leonard:
Yep @sxa , my misstake I had accidently changed VARIANT in my env script. It's just the vendor- build args that are added now, for review please...

sxa (approved):

PR #3346, opened 2023-05-15
Fixed erroring issue with security/mk-cacerts.sh

jmjaffe37:
This allows us to run `security/mk-cacerts.sh` from the curl project without needing to modify `mk-ca-bundle.pl` manually to avoid error


jmjaffe37:
nvm: realized that there is a version of 'mk-ca-bundle.pl' within the adopt project which works fine

Issue #3096, opened 2022-09-08
Investigate how to make build more secure when download/install tools during build


zdtsw:
We have `checkFingerprint()` which uses checksum to verify certain tools we download onto build machine (e.g freemarker, alsa)

But we also have code e.g in `build-farm/platform-specific-configurations/*.sh`  directly download jdk from API server.
It is difficult to use checksum to verify these binaries due to nightly builds. 
Might be more places we are using "curl / wget "  to do similar things.

This ticket is to investigate from security perspective how can we make our builds more stable and secured

jiekang:
Just noting for easier reference, the download from API server is here atm:
https://github.com/adoptium/temurin-build/blob/master/build-farm/platform-specific-configurations/linux.sh#L155

zdtsw:
ref: https://blog.adoptium.net/2022/07/gpg-signed-releases/

sxa:
Looks like only alsa is remaining, and that is covered by a checksum to detect tampering, however https://www.alsa-project.org/files/pub/lib/ provides GPG signatures so we should use those (key ID `A6E59C91`, subject to fixing the GPG problems on Windows and [macos](https://github.com/adoptium/infrastructure/issues/3049).

sxa:
Also has been done in https://github.com/adoptium/temurin-build/pull/3351 and removed (since it's not used) on AIX+Solaris at https://github.com/adoptium/temurin-build/pull/3358.
Closing.

PR #3350, opened 2023-05-16
linux.sh: Pull some parts into functions and do GPG signature checks on boot JDK downloads

sxa:
Part of https://github.com/adoptium/temurin-build/issues/3096 - we should be verifing the GPG signatures when downloading a boot JDK during the build.

I've also taken this opportunity to split out a couple of large sections into functions which I've wanted to do for some time as it'll make the script easier to read and remove a lot of the stuff that is mostly not important when reading the file from the main flow, however it will mess with the `blame` history of the individual lines in there...

- `locateDragonwell8BootJDK`
- `setCrossCompilerEnvironment` (For RISC-V cross compilation)
- `downloadBootJDK` (The function I'm modifying here)

The downloadBootJDK additions are all of the lines relating to `gpg` and I've split the `curl | tar xpfz` pipe up so it does the curl download, then the signature check, then extracts. Slower but in keeping with our goal of having a secure supply chain.

So far this has been tested on Linux/aarch64 JDK21 (head)

We need a better way of pulling the signature information down, since this is still a bit of a mess, but that can be addressed separately...

Initial diff of the download code if you want to use it to review separately is 
[diff.log.txt](https://github.com/adoptium/temurin-build/files/11490312/diff.log.txt)

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -26,6 +26,171 @@ NATIVE_API_ARCH=$(uname -m)
 if [ "${NATIVE_API_ARCH}" = "x86_64" ]; then NATIVE_API_ARCH=x64; fi
 if [ "${NATIVE_API_ARCH}" = "armv7l" ]; then NATIVE_API_ARCH=arm; fi
 
+function locateDragonwell8BootJDK()
+{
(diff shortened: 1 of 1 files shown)


karianna (approved):

steelhead31 (approved):
LGTM!

PR #3351, opened 2023-05-16
Use GPG instead of checksum to verify ALSA download

sxa:
Needs to be verified on MacOS and Windows since they are having problems with GPG signing but this will switch to using GPG signatures instead of checksums to verify the ALSA download.

Some notes for reviews:
- The downloadFile function is no longer used after this - we could remove it completely
- Alternatively we could use it but modify it to only take the GPG KEY ID instead of a checksum, but since it's only used for ALSA that feels like unnecessary complexity.


karianna (dismissed):
needs linter fixes

steelhead31 (approved):
LGTM!

PR #3356, opened 2023-05-19
Fix GPG checking of boot JDK for EA downloads and Alpine

sxa:
Two fixes for the code merged in https://github.com/adoptium/temurin-build/pull/3350 yesterday:
- The check for a valid download of the GA boot JDK was incorrect (Checking the curl return code is not a valid way to do it as curl will download a message with the error code and return success)
- On Alpine if the current directory used when running the check is more than 83 characters long then the receiving of the keys from the keyserver will fail, so I'm going to use the value of `WORKSPACE` instead of 

For future reference and search fodder:
Failure on s390x/arm (Platforms where we have no GA build for JDK20 so this shows up on JDK21 builds) because it tries to download the signature of something that isn't there.
```
09:23:15  gpg: can't open `bootjdk.tar.gz.sig': No such file or directory
09:23:15  gpg: verify signatures failed: No such file or directory
```
Failure on Alpine:
```
09:24:39  gpg: connecting dirmngr at '/home/jenkins/workspace/build-scripts/jobs/jdk/jdk-alpine-linux-x64-temurin/workspace/libs/.gpg-temp/S.dirmngr' failed: Filename too long
```

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -18,7 +18,6 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=sbin/common/constants.sh
 source "$SCRIPT_DIR/../../sbin/common/constants.sh"
-
 # Bundling our own freetype can cause problems, so we skip that on linux.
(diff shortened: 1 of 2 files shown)


andrew-m-leonard (approved):

sxa:
Also verified on an Alpine and arm32 system for JDK21.

steelhead31 (approved):
LGTM!

PR #3360, opened 2023-05-23
Update api after build release

adamfarley:
Once a new series of builds have been released, it is important to update the api to ensure that the data from the builds of the next new java major version can be accurately reported on.

This change documents when+how to carry out this change, similar to the example [here](https://github.com/adoptium/api.adoptium.net/pull/578/files).


sxa (approved):
Approved, noting that this is consistent with the API docs added in https://github.com/adoptium/api.adoptium.net/pull/578/files which also acts as a sample if anyone needs it (We can probably have a better sample in the future where only that line is modified)

Issue #3359, opened 2023-05-22
Alpine Linux builds failing due to issue with gnugpg agent.


steelhead31:
The most recent build of Alpine is failing with the following : 

14:31:18  Checking and downloading Alsa dependency
14:31:18  Checking for ALSA
14:31:19    % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
14:31:19                                   Dload  Upload   Total   Spent    Left  Speed
14:31:22  
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  4  966k    4 49152    0     0  44351      0  0:00:22  0:00:01  0:00:21 44361
 43  966k   43  416k    0     0   206k      0  0:00:04  0:00:02  0:00:02  206k
 89  966k   89  864k    0     0   281k      0  0:00:03  0:00:03 --:--:--  281k
100  966k  100  966k    0     0   300k      0  0:00:03  0:00:03 --:--:--  300k
14:31:22    % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
14:31:22                                   Dload  Upload   Total   Spent    Left  Speed
14:31:22  
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   833  100   833    0     0   3682      0 --:--:-- --:--:-- --:--:--  3685
14:31:22  gpg: keybox '/home/jenkins/workspace/build-scripts/jobs/jdk17u/jdk17u-alpine-linux-x64-temurin/.gpg-temp/pubring.kbx' created
14:31:22  gpg: /home/jenkins/workspace/build-scripts/jobs/jdk17u/jdk17u-alpine-linux-x64-temurin/.gpg-temp/trustdb.gpg: trustdb created
14:31:22  gpg: key 8380596DA6E59C91: public key "ALSA Release Team (Package Signing Key v1) <release@alsa-project.org>" imported
**14:31:29  gpg: can't connect to the agent: IPC connect call failed**
14:31:29  gpg: Total number processed: 1
14:31:29  gpg:               imported: 1

PR #3361, opened 2023-05-23
Fix GPG Verification of ALSA on Alpine Linux

steelhead31:
Fixes : #3359 

Amended prepareWorkspace.sh for ALSA GPG verification to use a shortened path ( ~/home/.gpg-temp ) for everything except alpine which when running within docker cannot create a .gpg-temp directory on the root directory. Sine $HOME does not resolve to anything on the build alpine docker image,  its been amended to use /tmp. Pathing has been made unique by using .$$ (PID) as part of the path name, to minimise impact.

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -304,6 +304,7 @@ checkingAndDownloadingAlsa() {
 
     if echo ${BUILD_CONFIG[OS_FULL_VERSION]} | grep -qi "alpine" ; then
       export GNUPGHOME="${WORKSPACE:-$PWD}/.gpg-temp"
+      echo GNUPGHOME=$GNUPGHOME
       # Should we clear this directory up after checking?
       # Would this risk removing anyone's existing dir with that name?
       # Erring on the side of caution for now
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -305,7 +305,7 @@ checkingAndDownloadingAlsa() {
     # Erring on the side of caution for now
     mkdir -p "$GNUPGHOME" && chmod og-rwx "$GNUPGHOME"
     gpg --keyserver keyserver.ubuntu.com --recv-keys "${ALSA_LIB_GPGKEYID}" || true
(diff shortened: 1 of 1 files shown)


sbin/prepareWorkspace.sh
```diff
@@ -295,18 +295,32 @@ checkingAndDownloadingAlsa() {
...
-    export GNUPGHOME="${WORKSPACE:-$PWD}/.gpg-temp"
+
+    ## Add Special Rules For Alpine Linx As This Doesnt Work In docker
+
+    if echo "${BUILD_CONFIG[OS_FULL_VERSION]}" | grep -qi "alpine" ; then
+      # Use /tmp for alpine in preference to $HOME as Alpine fails gpg operation if PWD > 83 characters
+      # Alpine also cannot create ~/.gpg-temp within a docker context
+      export GNUPGHOME="/tmp/.gpg-temp.$$"
+    else
+      export GNUPGHOME="$HOME/.gpg-temp.$$"
```
sxa, line 310: Perhaps leave as `${WORKSPACE:-$HOME}` or `${WORKSPACE:-$PWD}` which is what it was previously

sbin/prepareWorkspace.sh
```diff
@@ -295,18 +295,32 @@ checkingAndDownloadingAlsa() {
...
+    fi
+
+    echo "GNUPGHOME=$GNUPGHOME"
+    mkdir -p "$GNUPGHOME" && chmod og-rwx "$GNUPGHOME"
+    gpg --keyserver keyserver.ubuntu.com --recv-keys "${ALSA_LIB_GPGKEYID}"
     # Should we clear this directory up after checking?
     # Would this risk removing anyone's existing dir with that name?
     # Erring on the side of caution for now
-    mkdir -p "$GNUPGHOME" && chmod og-rwx "$GNUPGHOME"
+    ## gpg --homedir $GNUPGHOME --keyserver keyserver.ubuntu.com --recv-keys "${ALSA_LIB_GPGKEYID}"
```
sxa, line 319: Don't think there's a need to leave this comment in :-)

sxa (changes requested):
Couple of tweaks

sxa (approved):
One typographical change in a comment which hopefully you can just quickly accept, then I'm ok with this.

andrew-m-leonard (approved):

PR #3363, opened 2023-05-24
Fix TemurinSignSBOM build failures due to upstream pki jar changes

andrew-m-leonard:
openkeystore updates to change their jar name
and fix TemurinSignSBOM.java incorrect exception handling

Fixes https://github.com/adoptium/temurin-build/issues/3364

### DIFF
--- a/cyclonedx-lib/build.xml
+++ b/cyclonedx-lib/build.xml
@@ -21,6 +21,7 @@
 -->
 
         <!-- Dependency versions -->
+        <property name="openkeystore-version" value="1.0.0"/>
         <property name="cyclonedx-core-java-version" value="7.3.2"/>
         <property name="jackson-core-version" value="2.14.2"/>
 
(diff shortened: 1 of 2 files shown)


zdtsw (approved):

zdtsw:
maybe we can consider to use their release tag e.g https://github.com/cyberphone/openkeystore/releases/tag/1.0.0  instead only clone repo on default branch?

sxa (approved):
Approving, noting that this is SBOM upstream tooling that has given the need for the change, not OpenJDK.

andrew-m-leonard:
> maybe later, we can consider to use their release tag e.g https://github.com/cyberphone/openkeystore/releases/tag/1.0.0 instead only clone repo on default branch?

ah good spot, that's new, they weren't providing release previously

andrew-m-leonard:
@zdtsw So they have tagged a release 1.0.0 level, I have updated this PR to clone that specific tag which is far safer.

PR #3377, opened 2023-05-31
Update version retire post-release section

andrew-m-leonard:
Update the example PR for post-release version retiring, noting it includes the evaluation pipeline.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -282,7 +282,7 @@ Once the PRs to change those files have been merged, the [adoptium-packages-linu
 Once all the release binaries have been published the following tasks should be completed:
 If the latest version just released has come to the end of its non-LTS lifecycle (2 CPU updates, eg.jdk-15.0.2)
 
-- disable and retire that version form the Nightly and Evaluation pipeline builds: e.g <https://github.com/adoptium/ci-jenkins-pipelines/pull/628/files>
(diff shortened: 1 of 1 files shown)


zdtsw (approved):

steelhead31 (approved):
LGTM

sxa (commented):
I am extremely reluctant to approve this. If the pipelines are deleted then any 'locked' jobs will disappear. On two separate occasions I have lost tagged built levels (which were advertised externally) for RISC-V as a result of this process because the tagged builds are never pushed explicitly to GitHub for evaluation pipelines.

For JDK19 I ended up recreating the build jobs at https://ci.adoptium.net/job/jdk19u-linux-riscv64-temurin/ as a result of this, so we need to take care not to just blindly remove these jobs without having some alternate solution in place for people who wish to use the evaluation platforms.

andrew-m-leonard:
> I am extremely reluctant to approve this. If the pipelines are deleted then any 'locked' jobs will disappear. On two separate occasions I have lost tagged built levels (which were advertised externally) for RISC-V as a result of this process because the tagged builds are never pushed explicitly to GitHub for evaluation pipelines.
> 
> For JDK19 I ended up recreating the build jobs at https://ci.adoptium.net/job/jdk19u-linux-riscv64-temurin/ as a result of this, so we need to take care not to just blindly remove these jobs without having some alternate solution in place for people who wish to use the evaluation platforms.

Yes agree, i've remove the - delete piplines and jobs line

PR #3383, opened 2023-06-05
Remove typo in aix cflags argument

adamfarley:
Adding an equals sign that needs to be present to make this a valid argument.


sxa (approved):
This whole format of adding parameters that needs the equals feels a bit wierd :-) But there' precedent elsewhere in the file so I'm ok with approving.

We should verify that adding the `--with-extra-cflags` doesn't impact the configure arguments in the log as it was before and hasn't caused anything to be removed.

karianna (approved):
I'm assuming this does not harm as PR builds passed

adamfarley:
[Here's](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk/job/jdk-aix-ppc64-temurin/225/) the latest run against the latest form of this fix. The build completes successfully, and the aborted status is only there because a test run failed (investigating now).

I think we're good to merge.

Issue #3314, opened 2023-04-05
Populate SBOM with minimal properties to enable reproducible-build


andrew-m-leonard:
The SBOM needs to contain properties that enable a full reproducible re-build.

andrew-m-leonard:
From the previous issue https://github.com/adoptium/temurin-build/issues/3256#issuecomment-1516027059
We concluded we at least need to add
    gcc version
    bootjdk version
This PR needs to update the script around abouts here: https://github.com/adoptium/temurin-build/blob/914459c510d90c27e22ee628f699e39872598f90/sbin/build.sh#L775

chandrashritii:
Got it. I'd like to take this up @andrew-m-leonard Could you please assign this issue to me?

> From the previous issue [#3256 (comment)](https://github.com/adoptium/temurin-build/issues/3256#issuecomment-1516027059) We concluded we at least need to add gcc version bootjdk version This PR needs to update the script around abouts here:
> 
> https://github.com/adoptium/temurin-build/blob/914459c510d90c27e22ee628f699e39872598f90/sbin/build.sh#L775

sxa:
Note that as an absolutely minimum the glibc package version is critical. EVERY different CentOS7 patch level contains a glibc that makes reproducibility incompatible with any other (I tested with the levels included at each CentOS7 patch level on x64)

I suspect most other packages are less sensitive to patches than glibc, but it's always going to be a risk so we should perhaps aim to include the glibc level in the SBOM ASAP and add everything else that comes out of the research as potential dependencies that could cause differences.

andrew-m-leonard:
@chandrashritii So to start with GCC versioning, and additional parts like operating system GLIBC mentioned above. So you'll need to investigate a bit but I'm thinking where the gcc setup occurs here: https://github.com/adoptium/temurin-build/blob/8e18f18fbf427747417b932c78741b0bc8511eed/build-farm/platform-specific-configurations/linux.sh#L315
That we could add some code there to interrogate what the exact version of gcc (gcc --version), and at the same time query the glibc version being used on the build machine, I think "ldd --version".

sxa:
`rpm -q glibc` should be the best way to get the glibc version from any rpm-based LInux distribution. `dpkg -l libc-bin` on Debian/Ubuntu (arm32 and riscv64 are built on Ubuntu, the others are on rpm-based distributions)

PR #3385, opened 2023-06-07
Populate sbom with minimal properties | reproducible builds

chandrashritii:
Fixes [#3314](https://github.com/adoptium/temurin-build/issues/3314)

- GCC version
- GLIBC version
- BootJDK version

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -804,8 +804,9 @@ generateSBoM() {
   addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "make_command_args" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/makeCommandArg.txt"
 
   # Below add build tools into metadata tools
-  # Add GLIBC version
-  addSBOMMetadataTools "${javaHome}" "${classpath}" "${sbomJson}" "GLIBC" "rpm -q glibc"
+  addGLIBCversionforLinux
+  addGCC
+  addBootJDK
   # Add ALSA 3rd party
   addSBOMMetadataTools "${javaHome}" "${classpath}" "${sbomJson}" "ALSA" "$(cat ${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/dependency_version_alsa.txt)"
   # Add FreeType 3rd party (windows + macOS)
@@ -840,7 +841,7 @@ checkingToolSummary() {
 
 # Below add versions to sbom | Facilitate reproducible builds
 
-checkingGLIBCforLinux() {
+addGLIBCforLinux() {
    export CC=${grep "^CC :=" ${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/build/*/spec.gmk}
    export SYSROOT_CFLAGS=${grep "^SYSROOT_CFLAGS :=" ${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/build/*/spec.gmk}
    export GLIBC_MAJOR="$(echo "#include <features.h>" | $CC $SYSROOT_CFLAGS -dM -E - 2>&1 | tr -s " " | grep "#define __GLIBC__" | cut -d" " -f3)"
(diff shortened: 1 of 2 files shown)


sbin/build.sh
```diff
@@ -804,6 +804,10 @@ generateSBoM() {
   addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "make_command_args" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/makeCommandArg.txt"
 
   # Below add build tools into metadata tools
+  # Add GCC version
+  addSBOMMetadataTools "${javaHome}" "${classpath}" "${sbomJson}" "GCC" "$(cat ${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/gcc_version.txt)"
+  # Add GLIBC version
+  addSBOMMetadataTools "${javaHome}" "${classpath}" "${sbomJson}" "GLIBC" "rpm -q glibc"
```
andrew-m-leonard, line 810: I think this should be $("rpm -q glibc")

sbin/prepareWorkspace.sh
```diff
@@ -333,6 +333,42 @@ checkingAndDownloadingAlsa() {
   echo "${ALSA_BUILD_URL}" > "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/dependency_version_alsa.txt"
 }
 
+checkingGCCforLinux() {
```
andrew-m-leonard, line 336: So I think rather than assuming the gcc used from the environment variables, we should use what was determined in the actual build. We currently get the "summary" here: https://github.com/adoptium/temurin-build/blob/2af5238b07c38fea4e9cb60b256adc1b67b19afb/sbin/build.sh#L836

Use a similar method to get just the gcc and gc++ versions used, eg:
Here's the output in the configure log:
```
 Tools summary:
* Boot JDK:       openjdk version "17.0.5" 2022-10-18 OpenJDK Runtime Environment Temurin-17.0.5+8 (build 17.0.5+8) OpenJDK 64-Bit Server VM Temurin-17.0.5+8 (build 17.0.5+8, mixed mode, sharing) (at /usr/lib/jvm/jdk-17)
* Toolchain:      gcc (GNU Compiler Collection)
* C Compiler:     Version 10.3.0 (at /usr/local/gcc10/bin/gcc-10.3)
* C++ Compiler:   Version 10.3.0 (at /usr/local/gcc10/bin/g++-10.3)
```
eg. something like (i've not tested this!)
```
sed -n '/^Tools summary:$/,$p' "${inputConfigFile}" | grep "C Compiler:" | sed 's/^.*C Compiler:.*Version\s+([0-9\.]).*$/\1/'
```

andrew-m-leonard:
@chandrashritii We need to use tr and cut for this since the format is fixed, ie.
```
sed -n '/^Tools summary:$/,$p' "${inputConfigFile}" | grep "C Compiler:" | tr -s " " | cut -d" " -f5
```
You may need to alter -f5 to -f6, depends on whether there is a leading space?

chandrashritii:
> @chandrashritii We need to use tr and cut for this since the format is fixed, ie.
> 
> ```
> sed -n '/^Tools summary:$/,$p' "${inputConfigFile}" | grep "C Compiler:" | tr -s " " | cut -d" " -f5
> ```
> 
> You may need to alter -f5 to -f6, depends on whether there is a leading space?

Thanks! @andrew-m-leonard -f5 works. I've also added a function call for sourcing the bootjdk version from the build tools summary, using a similar method as the one used for gcc

sbin/build.sh
```diff
@@ -804,6 +804,8 @@ generateSBoM() {
   addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "make_command_args" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/makeCommandArg.txt"
 
   # Below add build tools into metadata tools
+  # Add GLIBC version
+  addSBOMMetadataTools "${javaHome}" "${classpath}" "${sbomJson}" "GLIBC" "rpm -q glibc"
```
sxa, line 808: This will only work on `rpm`-based distributions. For arm32 and RISC-V we build on Ubuntu so `dpkg` will need to be used to identify the libc version used for the build.

(Also noting that  if we ever start using sysroots to run the build then this will not be the correct way to detect what was used, but that can be addressed if and when we make such a change)

andrew-m-leonard: @sxa yeah, you're right, and your point on sysroot is also key, given that can have been specified as a build arg...
I think the following will work based on using the compiler, but a little fiddly to work out:
1. Get $CC and $SYSROOT_CFLAGS from the build's generated spec.gmk: "build/*/spec.gmk" file
2. Call:

andrew-m-leonard: @chandrashritii The spec.gmk file in build.sh can be located from variables:
```
grep "^CC :=" ${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/build/*/spec.gmk
grep grep "^SYSROOT_CFLAGS :=" ${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/build/*/spec.gmk
```

andrew-m-leonard: Similarly we can also get the Gcc version easily from that:
```
"^CC_VERSION_NUMBER :="
```

andrew-m-leonard:
@chandrashritii Looks good, i've approved the CI build checks to run...

karianna:
Tests fail addGLIBCversionforLinux can't be found, probably needs declaring higher up / importing

karianna (changes requested):
failed builds

sbin/build.sh
```diff
@@ -804,6 +804,9 @@ generateSBoM() {
   addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "make_command_args" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/makeCommandArg.txt"
 
   # Below add build tools into metadata tools
+  addGLIBCversionforLinux
```
andrew-m-leonard, line 807: This should be addGLIBCforLinux

andrew-m-leonard (changes requested):

sbin/build.sh
chandrashritii: Thanks Andrew! Fixed

karianna (approved):

andrew-m-leonard (approved):
looks good

PR #3386, opened 2023-06-10
Enable SCM_REF as SHA or most recent tag that is reachable from a commit

sophia-guo:


build-farm/make-adopt-build-farm.sh
```diff
@@ -200,7 +200,7 @@ if [ "${RELEASE}" == "true" ]; then
 fi
 
 if [ "${RELEASE}" == "true" ] && [ "${VARIANT}" != "openj9" ]; then
-    export TAG="${SCM_REF}"
+    export TAG="${SCM_REF}"
```
karianna, line 203:
```suggestion
    export TAG="${SCM_REF}"
```

karianna (approved):

sbin/prepareWorkspace.sh
```diff
@@ -165,6 +171,19 @@ checkoutRequiredCodeToBuild() {
       else
         echo "Failed cmd: git fetch origin \"refs/tags/${tag}:refs/tags/${tag}\""
       fi
+    elif [ "$sha" ]; then
+      echo "Checking out sha ${sha}"
+      git checkout "${sha}" || rc=$?
+      if [ $rc -eq 0 ]; then
```
sxa, line 177: Any reason not to use `if git checkout "${sha}"` here? The use of `$rc` here look a little complicated.

sophia-guo: Just to keep the same style as other checks.

Issue #3339, opened 2023-05-10
[JDK 21+] static-libs build will need an update after JDK-8307194


jerboaa:
As part of the OpenJDK build, various images are getting produced. One of which is the `static-libs-image` for JDKs > 8 as shown here:
https://github.com/adoptium/temurin-build/blob/1faf3ecc4903b2f562af15b28c15bff97f46c675/sbin/build.sh#L557

With [JDK-8307194](https://bugs.openjdk.org/browse/JDK-8307194) the `static-libs-image` target changed to also include a static version of `libjvm.a`. That's not needed for the primary use case of those static-libs: GraalVM/Mandrel builds.

Therefore, we should change the target to `static-libs-graal-image` instead for JDK 22+.

jerboaa:
https://bugs.openjdk.org/browse/JDK-8307858 is the redo bug. Resolved for JDK 22+. The new make target to use is `static-libs-graal-image`

PR #3390, opened 2023-06-12
Use updated static libs target

jerboaa:
Upstream OpenJDK changed what the static-libs image means. It includes a static version of libjvm.a with JDK 22+. Therefore, in order to get a version without libjvm.a, static-libs-graal-image target is being used instead.

Closes: #3339

### DIFF
--- a/configureBuild.sh
+++ b/configureBuild.sh
@@ -132,7 +132,7 @@ setVariablesForConfigure() {
 
   # JDK 22+ uses static-libs-graal-image target, using static-libs-graal
   # folder.
-  if [ "$openjdk_core_version" == "${JDKHEAD_CORE_VERSION}" ]; then
+  if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -ge 22 ]; then
     local static_libs_path="static-libs-graal"
   else
     local static_libs_path="static-libs"
--- a/configureBuild.sh
+++ b/configureBuild.sh
@@ -129,8 +129,14 @@ setVariablesForConfigure() {
   # test-image, debug-image and static-libs-image targets are optional - build scripts check whether the directories exist
   local openjdk_test_image_path="test"
   local openjdk_debug_image_path="debug-image"
-  local openjdk_static_libs_image_path="static-libs"
 
+  # JDK 22+ uses static-libs-graal-image target, using static-libs-graal
+  # folder.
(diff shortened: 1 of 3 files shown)


karianna (approved):

sxa (commented):
Hmmm this check will presumably not do the right thing when JDK23 is split off, since JDK22 will revert to the old target. Can we update the conditionals to have a specific cut-off post 21?

jerboaa:
> Hmmm this check will presumably not do the right thing when JDK23 is split off, since JDK22 will revert to the old target. Can we update the conditionals to have a specific cut-off post 21?

@sxa I take it you are referring to the `configureBuild.sh` changes? My understanding is that there we'd have do do explicit equals comparisons since `JDK8_CORE_VERSION` and friends refer to the repo name like `jdk8u` (string not a number). I'll probably change it to explicitly enumerate jdk 11, 17, 21 to use `static-libs` and use `static-libs-graal` in the else branch. What do you think?

sxa:
> @sxa I take it you are referring to the `configureBuild.sh` changes? My understanding is that there we'd have do do explicit equals comparisons since `JDK8_CORE_VERSION` and friends refer to the repo name like `jdk8u` (string not a number). I'll probably change it to explicitly enumerate jdk 11, 17, 21 to use `static-libs` and use `static-libs-graal` in the else branch. What do you think?

IMHO it would be a shame for us to actively "break" the ability to rebuild the STS releases even if they are out of support. In `configureBuild.sh` we also have precedent for constructs such as `if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -ge 12 ]; then` so perhaps an equivalent of that would be appropriate?

jerboaa:
> > @sxa I take it you are referring to the `configureBuild.sh` changes? My understanding is that there we'd have do do explicit equals comparisons since `JDK8_CORE_VERSION` and friends refer to the repo name like `jdk8u` (string not a number). I'll probably change it to explicitly enumerate jdk 11, 17, 21 to use `static-libs` and use `static-libs-graal` in the else branch. What do you think?
> 
> IMHO it would be a shame for us to actively "break" the ability to rebuild the STS releases even if they are out of support. In `configureBuild.sh` we also have precedent for constructs such as `if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -ge 12 ]; then` so perhaps an equivalent of that would be appropriate?

OK. Updated.

sxa (approved):
LGTM - thanks

Issue #3397, opened 2023-06-21
jdk-17/21 builds not always deterministic if built with different system root path


andrew-m-leonard:
If the gcc system root include & lib path is different between two builds then the debug symbols get that path embedded and are thus not identical. The build src include paths are already mapped using -fdebug-prefix-map, the system paths need to also be mapped.

PR #3398, opened 2023-06-21
Ensure debug files are deterministic by mapping system include paths

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3397

Update the build-scripts to add cflags for -fdebug-prefix-map of the gcc system include paths, so that an absolute path is not embedded in the debug files, which then makes the JDK libraries different due to the embedded CRC code.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -162,8 +162,8 @@ configureReproducibleBuildDebugMapping() {
       fdebug_flags+=" -fdebug-prefix-map=${gcc_include}/="
     fi
 
-    addConfigureArg "--with-extra-cflags=" "${fdebug_flags}"
-    addConfigureArg "--with-extra-cxxflags=" "${fdebug_flags}"
+    addConfigureArg "--with-extra-cflags=" "'${fdebug_flags}'"
+    addConfigureArg "--with-extra-cxxflags=" "'${fdebug_flags}'"
   fi
 }
 
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -126,6 +126,44 @@ configureReproducibleBuildParameter() {
          addConfigureArg "--with-extra-cflags=" "-qnotimestamps"
          addConfigureArg "--with-extra-cxxflags=" "-qnotimestamps"
       fi
+
+      configureReproducibleBuildDebugMapping
+  fi
+}
+
+# For reproducible builds  we need to add debug mappings for the system header paths,
+# so that debug symbol files (and thus libraries) are deterministic
+configureReproducibleBuildDebugMapping() {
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
A test pipeline: https://ci.adoptium.net/job/build-scripts/job/openjdk17-pipeline/716/

karianna (approved):

karianna (changes requested):
some linter failures

sxa:
Noting that the single failure in the checks is the one fixed by https://github.com/adoptium/temurin-build/pull/3401

sxa (commented):
I've done an initial look over this - I wish it could be less complex :-)
It looks like this is going to put a parameter like:
`-fdebug-prefix-map=/some/sysroot/include=`
Would it be preferable to have it set to the location where users are most likely to have the files such as `/usr/include` instead of leaving it blank to avoid debuggers being unable to find the files when needed?
If I've misinterpreted this let me know.

andrew-m-leonard:
> I've done an initial look over this - I wish it could be less complex :-) It looks like this is going to put a parameter like: `-fdebug-prefix-map=/some/sysroot/include=` Would it be preferable to have it set to the location where users are most likely to have the files such as `/usr/include` instead of leaving it blank to avoid debuggers being unable to find the files when needed? If I've misinterpreted this let me know.

@sxa The = blank is consistent with the what the openjdk make files add for the actual openjdk src folder, but I guess making to a likely common for these might make sense. The path gets "dirname'd", thus the path added are eg.
```
13:32:39  Adding -fdebug-prefix-map for root include: /usr/include
13:32:39  Adding -fdebug-prefix-map for gcc include: /usr/local/gcc11/lib/gcc/x86_64-pc-linux-gnu/11.2.0/include
13:32:39  Adding -fdebug-prefix-map for g++ include: /usr/local/gcc11/include/c++/11.2.0
```
So currently, prior to this the User's gcc header path would have to be the same path and prefix a our compiler...

andrew-m-leonard:
@sxa How about prefix mapping them to these, it then gives the user the chance to symlink appropriately if they wanted to debug with symbols? Different locations for gcc, as likely not installed by user in /usr/include : 
- root include : /usr/include
- gcc include : /usr/local/gcc_include
- g++ include : /usr/local/gxx_include

sxa (approved):
LGTM. Noting that previously this will have needed the user to have the same version of the compiler in the same location as us (e.g. by downloading the same compiler) but this will now need a symlink or similar from whichever compiler symbols the user wants to point it. There may be an argument for leaving this as "our" location since it may not work with any other compiler other than ours in our location anyway.

PR #3424, opened 2023-07-10
Check file integrity for sbom dependencies after download

adamfarley:
Currently, we check the integrity of the first file, and then ant uses the "yes this file matches the sha256 we have on file" boolean (checksum.match) for all subsequent files.

This results in all files being declared intact, even if they don't match the shas we have for them.

To fix this, I've declared checksum.match to be a strictly local property, so we actually check every file. I also did the same for "downloaded.file.checksum" just to be safe, and I updated the checksums for the files which no longer match their checksum.


adamfarley:
Running a build to test this now. It already works locally.

karianna (approved):

zdtsw (approved):
/lgtm

PR #3431, opened 2023-07-12
Improve Resilience Of Signing Checks For MacOS

### LINKED ISSUE #3428: MacOS Signing - Improve Resiliency In sign.sh script
Improve MacOS Signing by improving the resiliency of the signing script. Use similar logic and process to that added to the ci-jenkins-pipelines.

steelhead31:
Fixes #3428 

Include a 20 retry for each library, when it fails to sign at the first attempt.

### DIFF
--- a/sign.sh
+++ b/sign.sh
@@ -180,11 +180,17 @@ signRelease()
         JDK_DIR=$(ls -d "${TMP_DIR}"/jdk*)
         JDK=$(basename "${JDK_DIR}")
         cd "${TMP_DIR}"
-        zip -r "${TMP_DIR}/unsigned.zip" "${JDK}"
+        zip -q -r "${TMP_DIR}/unsigned.zip" "${JDK}"
         cd -
         curl --fail --silent --show-error -o "${TMP_DIR}/signed.zip" -F file="@${TMP_DIR}/unsigned.zip" https://cbi.eclipse.org/macos/codesign/sign
         echo "Debug 1"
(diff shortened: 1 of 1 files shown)


sign.sh
```diff
@@ -132,11 +136,49 @@ signRelease()
...
+                success=false
+                iteration=$((iteration+1))
+                errcount=$((errcount+1))
+              fi
+            done
+            if [ "$errcount" -gt 0 ]
+            then
+              echo "Errors Encountered During Signing"
+              echo "Error Count = $errcount"
+              exit 1
```
adamfarley, line 176: I'm not sure I understand. I thought this fix was intended to retry 20 times in order to *avoid* failures. This appears to say "If we needed to retry downloading this file even once, then fail".

How would this be better than the current situation?

adamfarley (changes requested):
Confused. Your changes appear to cancel themselves out. Thoughts please.

sign.sh
steelhead31: The multi level nested loops and if statements make this slightly difficult to read.

steelhead31: Looks like its working ok in a successful run, Ive also tested it in my local test script, and it works correctly,, with a failure followed by a success

adamfarley: Thanks Scott. Looks excellent, well done. :)

adamfarley (approved):
LGTM 👍

andrew-m-leonard (approved):
lgtm

PR #3432, opened 2023-07-12
Windows reproducible comparison update

sophia-guo:
Create self-signed certificate for windows
Update windows tools' windows paths calling from cygwin
Turn on echo and verbose information only if command fails


sophia-guo:
Part of https://github.com/adoptium/ci-jenkins-pipelines/issues/743

karianna (approved):

PR #3434, opened 2023-07-13
Fix Issue With MacOS Codesigning Find Command

### LINKED ISSUE #3433: Fix Issue With MacOS Find Command In Code Signing Picking Up Extended Attributes
As seen in this job..

https://ci.adoptium.net/job/build-scripts/job/release/job/sign_build/25727/console

The MacOS signing fails as it is picking up the file containing the macos extended attributes when running on a linux system.

The find command is currently finding a file called .\_jdkxx.xxx.xx in the root of the unzipped JDK, this file contains only extended attributes for macos, and should be omitted from the signing process.

steelhead31:
Fixes #3433 

Successful test run of this change here:
https://ci.adoptium.net/job/build-scripts/job/release/job/sign_build/25728/console

### DIFF
--- a/sign.sh
+++ b/sign.sh
@@ -122,7 +122,7 @@ signRelease()
 
       # Sign all files with the executable permission bit set.
 
-      FILES=$(find "${TMP_DIR}" -perm +111 -type f -o -name '*.dylib' -type f || find "${TMP_DIR}" -perm /111 -type f -o -name '*.dylib' -type f)
+      FILES=$(find "${TMP_DIR}" -perm +111 -type f -not -name '.*' -o -name '*.dylib' || find "${TMP_DIR}" -perm /111 -type f -not -name '.*' -o -name '*.dylib')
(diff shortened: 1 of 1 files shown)


sxa (approved):
I'm approving regardless, but would using `find ${TMP_DIR}/*` be a nicer way of excluding the dot directories? Or are there other hidden files that we explcitly want to avoid?

I'm not sure the extra blank lines are needed but 🤷🏻 it's not a deal breaker for me.

steelhead31:
> I'm approving regardless, but would using `find ${TMP_DIR}/*` be a nicer way of excluding the dot directories? Or are there other hidden files that we explcitly want to avoid?
> 
> I'm not sure the extra blank lines are needed but 🤷🏻 it's not a deal breaker for me.

I initially tried the TMP\_DIR/\* method, but it continued to pick up the file.. its the only hidden binary file in the archive so shouldn't cause an issue going forward.

PR #3462, opened 2023-08-23
Store sbom dependencies as jenkins artifacts

adamfarley:
We've had a number of sbom-creation failures related to corrupted downloads from maven, so this stores the jars we need on Jenkins to avoid the maven issue.

Also includes the build.xml changes needed to fetch the jars from their new location.

### DIFF
--- a/cyclonedx-lib/build.xml
+++ b/cyclonedx-lib/build.xml
@@ -20,18 +20,8 @@
 // jscpd:ignore-start
 -->
 
-        <!-- Dependency versions -->
+        <!-- All other dependency versions are now listed in the ./getDependencies file -->
         <property name="openkeystore-version" value="1.0.0"/>
-        <property name="cyclonedx-core-java-version" value="7.3.2"/>
-        <property name="jackson-core-version" value="2.14.2"/>
-
-        <property name="jackson-annotations-version" value="2.14.2"/>
-        <property name="jackson-databind-version" value="2.14.2"/>
-        <property name="json-schema-version" value="1.0.77"/>
-
-        <property name="commons-codec-version" value="1.15"/>
-        <property name="commons-io-version" value="2.11.0"/>
-        <property name="github-package-url-version" value="1.4.1"/>
 
         <!-- classpath for running application -->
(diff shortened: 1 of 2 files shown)


adamfarley:
New job location: https://ci.adoptium.net/view/all/job/build.getDependency/

Test run: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-mac-x64-temurin/348/console

adamfarley:
> how is this new job going to be created in jenkins? https://github.com/adoptium/temurin-build/pull/3462/files#diff-da8c4d473444fbe9429f2fd06e2ff65a5a788ed9f7272842b1cabc7301d24e0d is just a jenkinsfile
> 
> shouldn't it be put in [jenkins repo ](https://github.com/adoptium/ci-jenkins-pipelines/tree/master/tools)

I tried to follow the test.getDependency example, where the job was created manually and the jenkins file stored in aqa-tests repo.

The build.getDependency job has already been created and tested, see [here](https://github.com/adoptium/temurin-build/pull/3462#issuecomment-1690213544) for details. Note that all recent job/s will fail because I already updated it to fetch the jenkins file from the location it will be stored at once this PR is merged.

andrew-m-leonard (approved):
lgtm

steelhead31 (approved):
LGTM

sophia-guo:
https://github.com/adoptium/temurin-build/pull/3462#issuecomment-1690228385

I would also prefer the jenkinsfile should be put in [ci-jenkins-pipeline repo](https://github.com/adoptium/ci-jenkins-pipelines/tree/master/tools)

sophia-guo:
Probably also prefer those jar put in job https://ci.adoptium.net/view/Dependencies/job/dependency_pipeline/ instead of separate build.dependency job

adamfarley:
If we do put the jenkinsfile into ci-jenkins-pipeline and also merge it into the dependency_pipeline job (as @sophia-guo suggests) then we should also do the same with test.getDependency for consistency.

I don't have as preference either way, so long as the dependencies are available through jenkins (to prevent the maven download issues this PR was designed to prevent).

@smlambert - What do you think?

smlambert:
Organizationally, 
- test pipeline scripts are in aqa-tests 
  - test.getDependency groovy script lives in the aqa-tests repo along with all test pipeline code, and it is fine to stay there
- build pipeline scripts are in ci-jenkins-pipeline 
- scripts that support standalone build of Temurin are in temurin-build

smlambert:
Sorry, by what I shared, I meant to say I agree about putting CI scripts (Jenkinsfile / groovy scripts meant to be run in a CI server, not in a standalone build) into ci-jenkins-pipeline repository.

Issue #3464, opened 2023-08-29
Windows sbom artifact has incorrect filename


andrew-m-leonard:
eg.https://github.com/adoptium/temurin17-binaries/releases/tag/jdk17u-2023-08-28-03-35-beta
```
OpenJDK17U-sbom_x64_windows_hotspot_2023-08-28-03-35.zip-metadata.json
OpenJDK17U-sbom_x64_windows_hotspot_2023-08-28-03-35.zip.json
OpenJDK17U-sbom_x64_windows_hotspot_2023-08-28-03-35.zip.json.sig
OpenJDK17U-sbom_x86-32_windows_hotspot_2023-08-28-03-35.zip-metadata.json
OpenJDK17U-sbom_x86-32_windows_hotspot_2023-08-28-03-35.zip.json
OpenJDK17U-sbom_x86-32_windows_hotspot_2023-08-28-03-35.zip.json.sig 
```
The artifact name should not have the "**.zip**" in it !

Problem due to this line been incorrect for Windows .zip : https://github.com/adoptium/temurin-build/blob/8300eaa111e4d58f257d8cbfbec79dec3cf98620/sbin/build.sh#L1673

PR #3466, opened 2023-08-31
Correct Windows SBOM artifact filename

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3464

Ensure SBOM artifact filename is correct on Windows builds.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -1670,7 +1670,14 @@ createOpenJDKTarArchive() {
   if [ -d "${sbomTargetPath}" ]; then
     # SBOM archive artifact as a .json file
     local sbomTargetName=$(echo "${BUILD_CONFIG[TARGET_FILE_NAME]//-jdk/-sbom}.json")
-    sbomTargetName="${sbomTargetName//\.tar\.gz/}"
+
+    # Remove the tarball extension from the name to be used for the SBOM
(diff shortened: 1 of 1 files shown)


steelhead31 (approved):
LGTM

tellison (approved):

sxa (commented):
It seems a little strange to have this required on Windows only and including the tar.gz switch - presumable it's only needed for .zip on that platform.
Is there an underlying reason why we need to special case windows here - does it not indicate some issue earlier on?

sxa (approved):
I'd probably have just gone with this to squash them both and skip the `if` clause to make the code simpler, but I'm ok with it as-is too.
```
sbomTargetName="${sbomTargetName//\.zip/}"
sbomTargetName="${sbomTargetName//\.tar\.gz/}"
```

andrew-m-leonard:
> I'd probably have just gone with this to squash them both and skip the `if` clause to make the code simpler, but I'm ok with it as-is too if you'd rather not change it.
> 
> ```
> sbomTargetName="${sbomTargetName//\.zip/}"
> sbomTargetName="${sbomTargetName//\.tar\.gz/}"
> ```

Yeah, I did think of doing that to start with, but thought better to be more precise, i'll go with what I have, thanks

Issue #3469, opened 2023-09-04
Nightly/Beta builds are not always reproducible due to use of ccache


andrew-m-leonard:
As was already highlighted for release builds, use of ccache causes non-deterministic builds: https://github.com/adoptium/temurin-build/issues/2973
For reproducible nightly/beta builds, we need to enable the same for jdk-17+

andrew-m-leonard:
I have confirmed doing a nightly build on x64Linux with ccache disabled, the JDK does seem to be consistently identical.

PR #3470, opened 2023-09-04
Ensure CCache disabled for all reproducible jdk-17 & jdk-19+ builds

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3469

Ensure reproducible jdk-17 & jdk-19+ builds disable ccache, as it produces non-deterministic binaries: https://github.com/adoptium/temurin-build/issues/2973#issuecomment-1150295701

Also, in similar vane, for nightly builds the TZ needs to always be set to UTC: https://github.com/adoptium/temurin-build/issues/3075

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -100,9 +100,8 @@ configureReproducibleBuildParameter() {
       then
           # TZ issue: https://github.com/adoptium/temurin-build/issues/3075
           export TZ=UTC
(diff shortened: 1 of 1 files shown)


steelhead31 (approved):
LGTM

gdams (approved):

tellison (approved):

PR #3471, opened 2023-09-05
Some fixes for Windows comparable build script

andrew-m-leonard:
Some fixes for comparable build script on Windows
- signtool needs picking up from PATH as location depends on local MSVS dev environment
- expandJDK was leaving the original files after expanding
- SELF_CERT not needed as script creates its own now


adamfarley (approved):
LGTM

gdams (approved):

Issue #3493, opened 2023-10-03
Some sbom's missing FreeType version


andrew-m-leonard:
When temurin-build downloads FreeType because it is not on the System, then that gets logged in the SBOM.
However if there is a "System" freetype that is detected, then its details are not stored in the SBOM, ie.
https://github.com/adoptium/temurin-build/blob/751e5c8490e556e6c2db55018b048eb0e2f04eea/sbin/prepareWorkspace.sh#L455

sxa:
Definitely sounds like a bug then, so my checks are failing correctly ;-)

andrew-m-leonard:
@sxa
So our inclusion of FreeType is somewhat "mixed" and hard to understand, i've so far concluded:
- jdk8 will check for existence of installedfreetype directory on the build machine workspace and use that (but it's hard to know what version that lib is..?)
- jdk8 if installedfreetype does not exist, then it downloads into a folder "freetype" the [hard coded](https://github.com/adoptium/temurin-build/blob/751e5c8490e556e6c2db55018b048eb0e2f04eea/sbin/common/config_init.sh#L471) version, and builds it and installs it... then copies to directory installedfreetype. This scenario gets stored in SBOM
- jdk11+ skips checking/downloading "19:06:42  Skipping Freetype", and then default to whatever is the "System" freetype:
```
19:07:17  checking for FREETYPE... yes
19:07:17  checking for freetype... yes (using pkg-config)
19:07:17  Using freetype: system
```

andrew-m-leonard:
@sxa Is there any reason we can't just use "Bundled" everywhere?

sxa:
> @sxa Is there any reason we can't just use "Bundled" everywhere?

I don't know. I haven't really looked into it.

PR #3497, opened 2023-10-05
Ensure SBOM FreeType version is set correctly for system and bundled settings

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3493

Determines what type of FreeType build it was "system" or "bundled", then look for the freetype.h from that freetype installation/openjdk bundle, to determine the FREETYPE_MAJOR,MINOR,PATCH

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -929,11 +929,13 @@ addFreeTypeVersionInfo() {
        FREETYPE_TO_USE="$(grep "^FREETYPE_TO_USE[ ]*:=" ${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/build/*/spec.gmk | sed "s/^FREETYPE_TO_USE[ ]*:=[ ]*//")"
    fi
 
+   echo "FREETYPE_TO_USE=${FREETYPE_TO_USE}"
+
    local version="Unknown"
    local freetypeInclude=""
    if [ "${FREETYPE_TO_USE}" == "system" ]; then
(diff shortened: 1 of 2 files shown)


andrew-m-leonard:
> Will this still pick up things properly in the case where we are pulling a version from git? Other than checking that (and a personal hatred of using bash arrays) this LGTM (Not looked in explicit detail, but I trust that you've tested it, and I really want this in before the October release cycle!)

@sxa Yes, works for all.. the FREETYPE_CFLAGS gets pointed at by autoconf to either the cloned freetype includes, or the system installed one...

sbin/build.sh
```diff
@@ -920,6 +920,62 @@ checkingToolSummary() {
    sed -n '/^Tools summary:$/,$p' "${inputConfigFile}" > "${outputConfigFile}"
 }
 
+# Determine FreeType version being used in the build from either the system or bundled freetype.h definition
+addFreeTypeVersionInfo() {
+   # Default to "system" the jdk8 only value
+   local FREETYPE_TO_USE="system"
+   if [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" != "${JDK8_CORE_VERSION}" ]; then
```
karianna, line 927: should that be `==`? Seems different to the comment above

karianna (changes requested):

sbin/build.sh
andrew-m-leonard: != is correct, the comment is missleading, i've changed it
jdk8 can only be system
jdk-11+ supports bundled or system

karianna (approved):

sxa (approved):
Based on my earlier question and this:

> != is correct, the comment is missleading, i've changed it jdk8 can only be system jdk-11+ supports bundled or system

do we have builds  where we are building freetype from a downloaded source? Approving regardless ...

andrew-m-leonard:
> Based on my earlier question and this:
> 
> > != is correct, the comment is missleading, i've changed it jdk8 can only be system jdk-11+ supports bundled or system
> 
> do we have builds where we are building freetype from a downloaded source? Approving regardless ...

@sxa Yes jdk8u, Windows & Mac

Issue #3518, opened 2023-11-01
jdk11u Alpine linux build failure: gpg keyserver timeout


andrew-m-leonard:
https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-alpine-linux-x64-temurin/277/console
```
18:05:42  GNUPGHOME=/tmp/.gpg-temp.194
18:05:42  gpg: keybox '/tmp/.gpg-temp.194/pubring.kbx' created
18:06:51  gpg: keyserver receive failed: Operation timed out
```

adamfarley:
PR created [here](https://github.com/adoptium/temurin-build/pull/3520) to extend the timeout. If this failure is a simple reaction to a higher server load causing our requests to take too long, this could give us the necessary tolerance for that slowness.

sxa:
Hmm we're getting quite a bit of variance in the workspace prep phase in those build jobs overall:
```
[sxa@fedora shm]$ for A in `seq 275 280`; do curl -s https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-alpine-linux-x64-temurin/$A/consoleText | egrep 'Configuring workspace|Initiating build'; done
[2023-10-26T18:05:40.893Z] build.sh : 18:05:40 : Configuring workspace inc. clone and cacerts generation ...
[2023-10-26T18:06:46.113Z] build.sh : 18:06:45 : Initiating build ...
[2023-10-28T17:06:45.353Z] build.sh : 17:06:45 : Configuring workspace inc. clone and cacerts generation ...
[2023-10-28T17:09:51.271Z] build.sh : 17:09:50 : Initiating build ...
[2023-10-31T18:05:36.287Z] build.sh : 18:05:36 : Configuring workspace inc. clone and cacerts generation ...
[2023-11-02T18:05:45.878Z] build.sh : 18:05:45 : Configuring workspace inc. clone and cacerts generation ...
[2023-11-02T18:13:15.347Z] build.sh : 18:13:15 : Initiating build ...
[2023-11-03T14:02:26.710Z] build.sh : 14:02:26 : Configuring workspace inc. clone and cacerts generation ...
[2023-11-03T14:03:56.815Z] build.sh : 14:03:56 : Initiating build ...
[2023-11-03T14:39:08.143Z] build.sh : 14:39:07 : Configuring workspace inc. clone and cacerts generation ...
[...]

PR #3520, opened 2023-11-03
Extending keyserver timeout to make alpine build more forgiving

adamfarley:
Right now, about 10% of alpine linux x64 builds fail due to timeouts when requesting keys from the keyserver.

Example: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-alpine-linux-x64-temurin/277

The hope is that these timeouts are just temporary slowness due to server load, and that an extended timeout is the correct fix.

Test run: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-alpine-linux-x64-temurin/279/

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -334,7 +334,7 @@ checkingAndDownloadingAlsa() {
     # Should we clear this directory up after checking?
     # Would this risk removing anyone's existing dir with that name?
     # Erring on the side of caution for now
-    gpg --keyserver keyserver.ubuntu.com --recv-keys "${ALSA_LIB_GPGKEYID}"
+    gpg --keyserver keyserver.ubuntu.com --keyserver-options timeout=300 --recv-keys "${ALSA_LIB_GPGKEYID}"
     echo -e "5\ny\n" |  gpg --batch --command-fd 0 --expert --edit-key "${ALSA_LIB_GPGKEYID}" trust;
     gpg --verify alsa-lib.tar.bz2.sig alsa-lib.tar.bz2 || exit 1
 
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -334,7 +334,6 @@ checkingAndDownloadingAlsa() {
     # Would this risk removing anyone's existing dir with that name?
     # Erring on the side of caution for now
     gpg --keyserver keyserver.ubuntu.com --keyserver-options timeout=300 --recv-keys "${ALSA_LIB_GPGKEYID}"
-    echo "Proof of timeout change."
     echo -e "5\ny\n" |  gpg --batch --command-fd 0 --expert --edit-key "${ALSA_LIB_GPGKEYID}" trust;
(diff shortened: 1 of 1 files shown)


adamfarley:
This PR may fix this issue: https://github.com/adoptium/temurin-build/issues/3518

adamfarley:
I noticed that the relevant code is duplicated. I don't see why this should be needed, so I'm removing it and tagging the original comitter to make sure it's not duplicated for a reason.

Also, I added a temporary echo cmd "Proof of timeout change." to make absolutely sure I'm modifying the right code here. I think I am, because it's preceded by the only instance of .gpg-temp creation that's followed by a pid ($$), which the output indicates is the thing that gets created prior to failure.

adamfarley:
Ok, we're definitely modifying the right code. Removing the echo cmd now.

https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-alpine-linux-x64-temurin/280/consoleFull

sxa (commented):
For clarification and historic reference we're talking about an issue with the GPG signature check process on the ALSA download which is giving this error, right?
```
18:05:42  GNUPGHOME=/tmp/.gpg-temp.194
18:05:42  gpg: keybox '/tmp/.gpg-temp.194/pubring.kbx' created
18:06:51  gpg: keyserver receive failed: Operation timed out
```
I'm somewhat surprised that that's timing out and I worry it's part of a more severe problem - if it hasn't git something after 50 seconds I'm not sure it'll get it after a longer time. How frequently has this occurred?

sxa:
Based on https://github.com/adoptium/temurin-build/issues/3518#issuecomment-1792606345 could you add an `uptime` command into the script (probably before the gpg receive so the line before where you had the `echo`) so we can have a few runs of this that we can analyse later to see if the machine is getting overloaded at times (Probably worth putting in a comment before the `uptime` call pointing at this issue/PR with a reminder to remove it later :-)

sxa (commented):
> I noticed that the relevant code is duplicated. I don't see why this should be needed, so I'm removing it and tagging the original comitter to make sure it's not duplicated for a reason.

Sine you asked me directly elsewhere I conform that yeah the de-duplication of this SGTM 👍🏻

andrew-m-leonard (approved):
lgtm

adamfarley:
> ...which is giving this error, right?
> ```
> 18:06:51  gpg: keyserver receive failed: Operation timed out
> ```

Yup, that's the one.

> 
> How frequently has this occurred?

Around 10% of the time. No apparent pattern.

> ...could you add an uptime command into the script (probably before the gpg receive so the line before where you had the echo)

This is slightly unclear, as the line where I had the "echo" is *after* the gpg recieve (gpg ... --recv-keys ...). 

I'm putting the uptime command into the script before the recv command. Do let me know if I've read that wrong.

sxa:
> I'm putting the uptime command into the script before the recv command. Do let me know if I've read that wrong.

Yeah that's fine with me :-)

sxa (approved):

PR #3522, opened 2023-11-06
Tidy up release_download_test.sh

netomi:
This PR adds support for the following things:

- support for running the script on machines other than linux-aarch64
- add support for output coloring if ansi terminal is available
- add support for additional options to ease testing (e.g. keep workspace instead of downloading all files of a release all the time)
- adding helper functions for common things like printing errors

Some things I want to add in a further PR:

- validation of signatures for all files of a mac binary using: https://github.com/indygreg/apple-platform-rs
- validation of signatures for all files of a windows binary using: https://github.com/mtrojnar/osslsigncode
- support for all of the included validations for a single binary, e.g. OpenJDK21U-jdk_x64_linux_hotspot_21.0.1_12.tar.gz
- so that consumers of temurin can validate if the binary they want to use passes all validation and they dont have to validate all binaries of a release which is quite a lot 

@mbarbero @sxa


tooling/release_download_test.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
karianna, line 1: I'm not sure we always have bash on our systems @sxa ?

tooling/release_download_test.sh
```diff
@@ -8,66 +8,191 @@
...
+VERBOSE=false
+KEEP_STAGING=false
+SKIP_DOWNLOADING=false
+
+NORMAL=""
+BOLD=""
+RED=""
+YELLOW=""
+
+# check if stdout is a terminal...
```
karianna, line 33: Useful function, extract to a util script for all scripts perhaps?

tooling/release_download_test.sh
```diff
@@ -8,66 +8,191 @@
...
    exit 1
 fi
 
+##########################################################################################################################
+#
+# Download release information from GitHub for the specified major version.
+#
+##########################################################################################################################
+
 if ! curl -sS "https://api.github.com/repos/adoptium/temurin${MAJOR_VERSION}-binaries/releases" > "$WORKSPACE/jdk${MAJOR_VERSION}.txt"; then
```
karianna, line 148: could this be a function?

tooling/release_download_test.sh
```diff
@@ -8,66 +8,191 @@
...
 fi
 
-[ "$VERBOSE" = "true" ] && echo "$(date +%T) : IVT: Downloading files from release repository"
+##########################################################################################################################
+#
+# Download all files of the specified release if required.
+#
+##########################################################################################################################
+
+if [ "$SKIP_DOWNLOADING" = "false" ]; then
```
karianna, line 159: could this be a function?

karianna: same goes for several commented blocks below.

tooling/release_download_test.sh
netomi: various scripts in the tooling directory use bash, so I assumed that bash would be fine.
The main reason to use bash it to be able to specify `set -o pipefail` which I consider to be a must for complex shell scripts imho.

tooling/release_download_test.sh
netomi: indeed, I was thinking the same that the various blocks should be extracted into functions, will do so.

netomi: I extracted most of the code into separate functions.

tooling/release_download_test.sh
netomi: this has been extracted into a common_logging.sh script

karianna (approved):
Looks good (fairly shallow review from me though, would be good if others dig in more)

tooling/release_download_test.sh
sxa: @karianna I had this discussion with @netomi elsewhere too as I'd prefer not to use bash if possible (IMHO using bash constructs leads to less readable scripts ;-) ) but we do have bash on all our systems for other things, and certainly on the systems which the jenkins job that runs this will be targetted at.

tooling/release_download_test.sh
```diff
@@ -8,164 +8,470 @@
...
-fi
+WORKSPACE=${WORKSPACE:-"$PWD"}
+VERBOSE=false
+KEEP_STAGING=false
+SKIP_DOWNLOADING=false
+
+TAG=""
+MAJOR_VERSION=""
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
sxa, line 30: Curious - why `${BASH_SOURCE[0]}` here and `${0}` in the error message a few lines later?

tooling/release_download_test.sh
```diff
@@ -8,164 +8,470 @@
...
+
+# shellcheck source=sbin/common/config_init.sh
+source "$SCRIPT_DIR/common_logging.sh"
+
+usage() {
```
sxa, line 35: Noting that this will require a change to the calling jenkins job when merged as the script is currently parameterless

sxa (commented):
Just going to do this review as a comment just now as there have been so may changes to the structure and operations that it's not too easy to review the changes in the git UI :-)

sxa:
Not immediately obvious why but this is failing fairly early on for me when tested in jenkins:
```
+ TAG=jdk-11.0.21+9
+ '[' -z jdk-11.0.21+9 ']'
+ '[' -z jdk-11.0.21+9 ']'
+ print_verbose 'IVT : Verifying Tag '\''jdk-11.0.21+9'\'''
+ '[' false = true ']'
```

tooling/release_download_test.sh
netomi: I copied this line from other scripts, should the use of ${0} in the usage be updated accordingly?

netomi:
> Not immediately obvious why but this is failing fairly early on for me when tested in jenkins:
> 
> ```
> + TAG=jdk-11.0.21+9
> + '[' -z jdk-11.0.21+9 ']'
> + '[' -z jdk-11.0.21+9 ']'
> + print_verbose 'IVT : Verifying Tag '\''jdk-11.0.21+9'\'''
> + '[' false = true ']'
> ```
> 
> EDIT: Same failure from the command line EDIT2: It works if you're using `-v` so the bug is somehow tied into the use of that parameter

I mainly tested with -v, will do without and fix the problem.

sxa:
Also if I run with `-v` I'm getting this in my jenkins job:
```
12:26:10 : IVT : Validating OpenJDK11U-sbom_aarch64_linux_hotspot_11.0.21_9.json ...
BOM validated successfully.
bash: tooling/validateSBOMcontent.sh: No such file or directory
ERROR: Failed checks on OpenJDK11U-sbom_aarch64_linux_hotspot_11.0.21_9.json
```
which is presumably related to a `cd` operation somewhere that means it's not picking up the second script properly

netomi:
> Also if I run with `-v` I'm getting this in my jenkins job:
> 
> ```
> 12:26:10 : IVT : Validating OpenJDK11U-sbom_aarch64_linux_hotspot_11.0.21_9.json ...
> BOM validated successfully.
> bash: tooling/validateSBOMcontent.sh: No such file or directory
> ERROR: Failed checks on OpenJDK11U-sbom_aarch64_linux_hotspot_11.0.21_9.json
> ```
> 
> which is presumably related to a `cd` operation somewhere that means it's not picking up the second script properly

fixed with latest commit. The refactoring into functions create that problem.

tooling/release_download_test.sh
netomi: I can also revert to use /bin/sh but then I have to remove the set -o pipefail.

tooling/release_download_test.sh
netomi: I am not super happy with this include here, the common_logging.sh will require that a parameter VERBOSE is defined, do you have any suggestion here?

sxa:
> fixed with latest commit. The refactoring into functions created that problem.

Yep confirmed that seems to resolve it it. However i'm not sure if it's from some of the `set` options but I'm now getting this at the end which I hadn't been for the original script:
```
+ echo 'Note: As this is a warning message this will not cause a non-zero return code by itself'
Note: As this is a warning message this will not cause a non-zero return code by itself
+ '[' 0 '!=' 0 ']'
+ exit 0
13:07:52 : IVT : Finished. Return code = 6
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

netomi:
Return code = 6 should indicate an error when running the validateSBOMcontent.sh for some of the sbom files. The latest did not seem to have an error as the exit code was 0, but one of the invocations before that?

sxa:
> Return code = 6 should indicate an error when running the validateSBOMcontent.sh for some of the sbom files. The latest did not seem to have an error as the exit code was 0, but one of the invocations before that?

Looks like something in the compiler expectation setting for windows:
```
+ '[' 'microsoft (Microsoft Visual Studio 2019)' '!=' 'gcc (GNU Compiler Collection)' ']'
+ echo 'ERROR: Compiler version not gcc (GNU Compiler Collection) (SBOM has microsoft (Microsoft Visual Studio 2019))'
ERROR: Compiler version not gcc (GNU Compiler Collection) (SBOM has microsoft (Microsoft Visual Studio 2019))
```

netomi:
I added now a -a option to enable ansi output. So you should be able to include that more easily into jenkins.
By default, ansi output is disabled.

tooling/release_download_test.sh
sxa: I think we should be consistent one way or another any personally I'd stick with `$0` everywhere unless you've got a specific advantage from the BASH specific version.

tooling/release_download_test.sh
sxa: Not sure I understand your comment here - you could pass a verbose option into `common_logging.sh` if required, but I feel I've missed the point :-)

I have no problems with the `-v` flag as-is unless you see an issue with it, but my comment was related to the fact that the mandatory `TAG` parameter is new (again, not a problem, and it's a good change, just needs dealing with in the automation when we merge this)

netomi: before that, the script did have this line

TAG=${1:-$TAG}

which I did not fully understand, does it mean that TAG will be assigned the first parameter. If this is absent, it will use the TAG variable that is defined externally e.g. using export before running the script? If this is needed for running on jenkins, we can readd it ofc.

tooling/release_download_test.sh
netomi: ok

tooling/release_download_test.sh
netomi: I reverted to the original behavior such that the TAG can be passed in as argument and as env variable.

sxa: > does it mean that TAG will be assigned the first parameter. If this is absent, it will use the TAG variable that is defined externally e.g. using export before running the script?

Yes that's correct. I'm ok with either approach and either will work in jenkins, would just need a one line modification in the jenkins script, but sine you've reverted it I'm happy with that :-)

sxa:
> > Return code = 6 should indicate an error when running the validateSBOMcontent.sh for some of the sbom files. The latest did not seem to have an error as the exit code was 0, but one of the invocations before that?
> 
> Looks like something in the compiler expectation setting for windows:
> 
> ```
> + '[' 'microsoft (Microsoft Visual Studio 2019)' '!=' 'gcc (GNU Compiler Collection)' ']'
> + echo 'ERROR: Compiler version not gcc (GNU Compiler Collection) (SBOM has microsoft (Microsoft Visual Studio 2019))'
> ERROR: Compiler version not gcc (GNU Compiler Collection) (SBOM has microsoft (Microsoft Visual Studio 2019))
> ```

Not quite sure how this was working previously but it wasn't being flagged despite it being a reasonably obvious issue. Anyway, https://github.com/adoptium/temurin-build/pull/3526/files should fix it - feel free to pull that change directly into your PR if you'd prefer.

netomi:
I will rebase this PR once the other is merged into the main branch.

sxa:
> I will rebase this PR once the other is merged into the main branch.

Have requested some quick approvals in the build channel which will allow you to rebase.

sxa:
> > I will rebase this PR once the other is merged into the main branch.

> Have requested some quick approvals in the build channel which will allow you to rebase.

Now merged - feel free to rebase :-)

sxa (approved):
LGTM - running one final test against jdk-17 then I'll get it merged :-) Thanks for the cleanup!

sxa:
Hmm the last test I did on jdk-17.0.9+9 failed for the win32 zip files as though the download didn't complete, but it did produce a file of the right size. Trying again ...

netomi:
can you provide a link to the jienkins job for inspection?

sxa:
> can you provide a link to the jienkins job for inspection?

It's restricted, so not easily but it looks like it might be reproducible outside jenkins anyway when you run against `jdk-17.0.9+9` - trying to run a few further tests now

tooling/release_download_test.sh
```diff
@@ -8,164 +8,490 @@
...
-  cd "staging/$TAG" || exit 3
   # Early access versions are currently in a different format
-  if echo "$TAG" | grep ea-beta; then
-     FILTER="ea_${MAJOR_VERSION}"
+  if echo "${TAG}" | grep ea-beta; then
+    filter="ea_${MAJOR_VERSION}"
   else
-     FILTER=$(echo "$TAG" | sed 's/+/%2B/g')
+    # shellcheck disable=SC2001
+    filter=$(echo "${TAG}" | sed 's/+/%2B/g')
```
sxa, line 155: This is needed to avoid problems when we produce point releases
```suggestion
    filter=$(echo "/${TAG}/" | sed 's/+/%2B/g')
```

sxa (changes requested):

sxa (approved):

Issue #3509, opened 2023-10-20
Investigate non-functioning of `--with-png=no` in freetype builds


sxa:
**What are you trying to do?**
Build freetype from source without libpng even when libpng is installed

**Expected behaviour:**
freetype is built without a dynamic link to libpng

**Observed behaviour:**
freetype is built with a dynamic link to libpng (Seen on macos jdk8u builds when cross-compiled on an aarch64 build machines)

**Any other comments:**
This line is in the build process but doesn't seem to take effect: https://github.com/adoptium/temurin-build/blob/698313c1bf2d036c334492a387586e2c0544aa6b/sbin/prepareWorkspace.sh#L482

It looks like the configure command in the freetype build invokes make which does `cd build/make && ./configure` when there is no configure script in that directory:
```
12:51:23  HEAD is now at 86bc8a950 * Version 2.9.1 released. =========================
12:51:23  make: error: couldn't stat Xcode's Info.plist (errno=No such file or directory)
12:51:23  /bin/sh: ./configure: No such file or directory
12:51:23  make: *** [setup] Error 127
```
(Noting that the build continues regardless, but the parameter to get rid of the libpng support does not take effect)

This problem was introduced via https://github.com/adoptium/infrastructure/issues/2536#issuecomment-1708716478 where libpng was explicitly installed in the x64 brew environment on the aarch64 build machines.

Two possible mitigations:
- Build jdk8u on the x64 machines again
- Force the detection of libpng to fail by removing the symlinks /usr/local/bin/libpng-config and /usr/local/lib/pkgconfig/libpng.pc (NOTE: This has been done on build-macstadium-macos11-arm64-2 and should be fixed properly after the release - potentially by removing the brew libpng package unless it's required for something else) The removed links are shown below
```
% ls -ld  /usr/local/bin/libpng-config /usr/local/lib/pkgconfig/libpng.pc
lrwxr-xr-x  1 administrator  admin  41 11 Sep 12:59 /usr/local/bin/libpng-config -> ../Cellar/libpng/1.6.40/bin/libpng-config
lrwxr-xr-x  1 administrator  admin  50 11 Sep 12:59 /usr/local/lib/pkgconfig/libpng.pc -> ../../Cellar/libpng/1.6.40/lib/pkgconfig/libpng.pc
```

sxa:
Note that https://ci.adoptium.net/computer/build-macstadium-macos11-arm64-2/ is the machine where the two links to libpng were removed.

[...]

PR #3523, opened 2023-11-06
Freetype: run bash autogen.sh before any ./configure commands

Haroon-Khel:
ref https://github.com/adoptium/temurin-build/issues/3509#issuecomment-1795762701

if `./configure --help` is run before `bash autogen.sh` then we get
```
* module: psnames   (Postscript & Unicode Glyph name handling)
cd builds/unix; \
	        ./configure 
/bin/sh: ./configure: No such file or directory
make: *** [setup] Error 127
```
and therefore `pngArg="--with-png=no"` does not get set

Running it after the autogen.sh script works
```
bash-3.2$ arch -x86_64 ./configure --help | grep with-png
  --with-png=[yes|no|auto]
```

Needs testing in jenkins

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -477,19 +477,21 @@ checkingAndDownloadingFreeType() {
       return
     fi
 
-    local pngArg=""
-    if ./configure --help | grep "with-png"; then
-      pngArg="--with-png=no"
-    fi
-
     local freetypeEnv=""
     if [[ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "i686" ]] || [[ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "i386" ]]; then
       freetypeEnv="export CC=\"gcc -m32\""
     fi
 
+    eval "${freetypeEnv}" && bash ./autogen.sh 
+
+    local pngArg=""
+    if ./configure --help | grep "with-png"; then
+      pngArg="--with-png=no"
+    fi
+
     # We get the files we need at $WORKING_DIR/installedfreetype
     # shellcheck disable=SC2046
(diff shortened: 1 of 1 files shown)


Haroon-Khel:
https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-mac-x64-temurin/427/ is queued up, link will be active when 426 finishes

karianna (approved):

andrew-m-leonard (approved):
lgtm

Haroon-Khel:
From the jenkins link above, the `--with-png` value gets set correctly 
`21:30:14  pngArg value: --with-png=no`

sbin/prepareWorkspace.sh
```diff
@@ -477,19 +477,21 @@ checkingAndDownloadingFreeType() {
...
-    if ./configure --help | grep "with-png"; then
-      pngArg="--with-png=no"
-    fi
-
     local freetypeEnv=""
     if [[ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "i686" ]] || [[ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "i386" ]]; then
       freetypeEnv="export CC=\"gcc -m32\""
     fi
 
+    eval "${freetypeEnv}" && bash ./autogen.sh
```
sxa, line 485: Should we check the return code of this? My decoupling it from the configure (which was run with a `&&` to ensure it wouldn't continue if the autogen failed) we've lost the exit code from this step which may lead to confusion in the logs.

sxa (commented):
LGTM and well spotted! Presumably we can remove https://github.com/adoptium/infrastructure/blob/ee8ecd8ddb812dd14641f557d3c3bb76f8a33097/ansible/playbooks/AdoptOpenJDK_Unix_Playbook/roles/Common/tasks/MacOSX.yml#L183 when this goes in?

sbin/prepareWorkspace.sh
Haroon-Khel: Done, now the script will exit if the autogen.sh script fails

sxa (approved):
May be better to exit with an explicit code of `1` if the autogen.sh fails, but otherwise this is good to go.

PR #3529, opened 2023-11-13
Add subject info the sbom

### LINKED ISSUE #3528: [SBOM] Attestation is missing a subject with digest
While looking in the sbom files for further validation, I noticed that currently there is no subject with digest stored in the sbom. See https://slsa.dev/attestation-model for terminology.

Without such a subject you will not be able to associate an sbom with an artifact and validate that a certain artifact (identified by its digest) is attested by the provenance that is provided (the sbom in the case of temurin).

Currently the sbom contains that information:

```
    "component" : {
      "name" : "Eclipse Temurin",
      "version" : "21.0.1+12-LTS",
      "description" : "Temurin JDK Component",
      "type" : "framework"
    },
```

we should add there the sha265 hash and also the filename of the artifact as generated during the build.

I can create a PR to add this and also update the validate script accordingly.

netomi:
This fixes #3528 .

This is a draft PR to gather feedback on the used approach.
Right now I split up the sbom generation into 2 steps:

 - generation of the overall sbom as before
 - finish the sbom by adding information for each generated artifact with their hashes

this has to be done after all archives have been created and moved to their final destination. We could also generate the sbom completely at the end, though I am not sure if no files should be modified in the target directory at all.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -848,11 +848,16 @@ generateSBoM() {
   local classpath="$(getCyclonedxClasspath)"
 
   local sbomTargetName=$(echo "${BUILD_CONFIG[TARGET_FILE_NAME]//-jdk/-sbom}.json")
+  # if the replacement did not work as the target file name does not contain the pattern '-jdk',
+  # construct the sbom name by just appending '-sbom.json' to the target file name.
+  if [[ ${sbomTargetName} != *"-sbom"* ]]; then
+    sbomTargetName=$(echo "${BUILD_CONFIG[TARGET_FILE_NAME]}-sbom.json")
+  fi
   # Remove the tarball extension from the name to be used for the SBOM
   if [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]]; then
-      sbomTargetName="${sbomTargetName//\.zip/}"
+    sbomTargetName="${sbomTargetName//\.zip/}"
   else
-      sbomTargetName="${sbomTargetName//\.tar\.gz/}"
+    sbomTargetName="${sbomTargetName//\.tar\.gz/}"
   fi
 
   local sbomJson="$(joinPath ${BUILD_CONFIG[WORKSPACE_DIR]} ${BUILD_CONFIG[TARGET_DIR]} ${sbomTargetName})"
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -877,39 +877,6 @@ generateSBoM() {
   addSBOMMetadataProperty "${javaHome}" "${classpath}" "${sbomJson}" "OS version" "${BUILD_CONFIG[OS_FULL_VERSION]^}"
   addSBOMMetadataProperty "${javaHome}" "${classpath}" "${sbomJson}" "OS architecture" "${BUILD_CONFIG[OS_ARCHITECTURE]^}"
 
-  # Create JDK Component
-  addSBOMComponent "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "${fullVer}" "${BUILD_CONFIG[BUILD_VARIANT]^} JDK Component"
-
-  # Below add different properties to JDK component
-  # Add variant as JDK Component Property
-  addSBOMComponentProperty "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "JDK Variant" "${BUILD_CONFIG[BUILD_VARIANT]^}"
-  # Add scmRef as JDK Component Property
-  addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "SCM Ref" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/scmref.txt"
-  # Add OpenJDK source ref commit as JDK Component Property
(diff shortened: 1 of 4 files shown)


netomi:
This is an example sbom when building locally:

[OpenJDK-sbom.json](https://github.com/adoptium/temurin-build/files/13332773/OpenJDK-sbom.json)

sxa:
> This is a draft PR to gather feedback on the used approach. Right now I split up the sbom generation into 2 steps:
 
> this has to be done after all archives have been created and moved to their final destination. We could also generate the sbom completely at the end, though I am not sure if no files should be modified in the target directory at all.

Good question ...
For our normal release builds I don't think there is a problem, however the nightly and EA [*] builds do get renamed during the process of publishing to GitHub. I think it's a good idea to get this merged now though with the caveat that the name information may not be correct for "non-production" builds (or we could include the checksum without the filename in those cases for now) but we should look at resolving that in a future update so we can make it fully consistent.

[*] - Noting that the process for EA (Early Access) builds is still very much a work-in-progres and therefore we have plenty of opportunities to fix this. The current way of doing things was an MVP to make it work from the perspective of customers - there are plenty of reasons to fix that, and being able to make the SBOM more useful as per this PR is definitely another one!

What do you think @andrew-m-leonard ?

sxa:
Also, I see you've changed it here but as a style note I personally prefer to have the comments all lined up in the file as it's more pleasing to the eye, but since this isn't code I wrote I won't require it to be changed ;-)

netomi:
As the artifacts are getting moved and renamed, I needed to add the final sbom generation at the very end to create the provenance with the final artifacts rather some intermediate state. Right now I use the final filename of the artifact as name of the component, but that is arbitrary, we could also other names that describe more conceptually and set the filename as property. So then the component name would be the same for EA and release builds, just the filename would be different which would also be reasonable imho.

Regarding the formatting: I started to work on that as there were some typos, but realized I should not change the formatting in general, so please ignore that changes. The idea of the PR is to collect feedback about the way the sbom gets generated and the output. The code to achieve is not clean at all atm.

sxa (commented):
@andrew-m-leonard Is there a reason why the main SBoM creation is before the creation of the final tar archives here?

sxa (commented):
The approach SGTM personally 👍🏻

andrew-m-leonard:
> @andrew-m-leonard Is there a reason why the main SBoM creation is before the creation of the final tar archives here?

Im think when this was originally added we were breaking new ground.. and my thoughts were on the JDK itself, and not the subsequent tar archive.

This change seems reasonable to me.

sxa:
@netomi Based on the above is there anything else you want to do before taking this out of draft status?

netomi:
I will need some more cleanups and the other components need some work (static-libs especially as the logic that I used so far does not work for them).

netomi:
I cleaned the PR.

Now there is only a single generateSBoM method at the end instead of splitting it up in two parts.

I changed the component names to "$COMPONENT Component" where component is one of JDK, JRE, STATIC-LIBS, SOURCES, TESTIMAGE, DEBUGIMAGE.

The filename of each component is now encoded as property with key Filename.

I did also some cleanups to avoid repeating the same snippets, e.g. when constructing filenames in an os independent way.

Something that could be changed is that the properties for each component are very much the same, and could be captured not on component level but on metadata level, or we omit it in same cases like for sources, but its also fine to keep it like that imho.

Attaching an example sbom generated locally using:

`
./makejdk-any-platform.sh -J /home/tn/.sdkman/candidates/java/11.0.18-tem/ --create-sbom --create-jre-image --create-source-archive jdk11u
`

[Uploading OpenJDK-sbom.json…]()

andrew-m-leonard:
> it looks like there is no sha256sum tool on MacOS:
> 
> what could be used otherwise?
> 
> I found this: https://unix.stackexchange.com/questions/426837/no-sha256sum-in-macos
> 
> Ah I found sha256File() in the prepareWorkspace.sh

On mac I use:
```
shasum -a 256 <file>
```

netomi:
On Macos the sha works now, but on windows the result looks like this:

          "content" : "\\605f4904f922500be134dff033d54926cc5bff70c3c2ec9bc0e07079e04409d7"

netomi:
Looks like cygwin sha256sum tool does not like windows paths. Making sure that files passed to the tool are not OS specific paths but rather unix like paths.

netomi:
Added a method to get the target file name of a component (e.g. jre or testimage) that also considers the case where the target file name does not contain a -jdk pattern in it like in the build workflow (there the target file name is OpenJDK.tar.gz), which means that the build script will always override that filename for all components as all components will end up with the same name. With this change, when no -jdk pattern is detected, just prepend the component name before the extension.

karianna (approved):

sbin/build.sh
```diff
@@ -826,64 +829,54 @@ generateSBoM() {
...
+
+  local javaHome="$(setupAntEnv)"
+
+  buildCyclonedxLib "${javaHome}"
+  # classpath to run java app TemurinGenSBOM
+  local classpath="$(getCyclonedxClasspath)"
+
+  local sbomTargetName=$(getTargetFileNameForComponent "sbom")
+  # Remove the tarball / zip extension from the name to be used for the SBOM
+  if [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]]; then
```
sxa, line 852: We could simplify this and not require the `if` clause by just stripping both e.g. `sbomTargetName=$(echo "${sbomTargetName}.json" | sed -e "s/\.zip//" -e "s/\.tar\.gz//")
But I'm ok either way.

sxa (approved):
I'm not an expert in this code (although more than I was a couple of weeks ago!) but this LGTM. We should look at the output (SBoM content) before and after this change (unless you already have a before & after pair you can add here?), but I'm ok with having this merged and see if it works, then fix up later if required.

Issue #3535, opened 2023-11-14
[SBOM] Capture all tools with their version / digest that are downloaded during a build


netomi:
The sbom should capture all external tools with their version / digests that are downloaded during the build. This mainly concerns the cyclonedx tools / dependencies.

Ideally, in the long run, there should not be a need to download these tools externally for each build, but they could be hosted in a tool repository that is checked out using a defined commit during the build to avoid relying on external services as much as possible.

sxa:
PR fo reht CycloneDX now merged, build.getDependencies re-run to pick up the new version, so #3538 can proceed

PR #3538, opened 2023-11-16
Capture CycloneDX JAR SHAs in SBoM

sxa:
**Note to reviewers: please allow me to merge this when approved.**

This is a test of including the CycloneDX jars (downloaded during the build process) into the SBoM.
There are several JAR files associated with the CycloneDX tool and this only covers one of them so is a "discussion starter" (hence raising this as a draft) as to how we should move forward if we want to capture this information in the SBoM. Is this the right place for it to be stored? Do we definitely need it? Would we exclude it if we downloaded this onto the machines via the playbooks? If we capture all of the JAR SHAs then what should the format be?
```
      {
        "name" : "FreeMarker",
        "version" : "N.A"
      },
      {
        "name" : "CycloneDX core java JAR SHA",
        "version" : "88193228f85a955127dc73e1c72efc9e08e18a01d227df47d0865dc20eceffd1"
      },
      {
        "name" : "Docker image SHA1",
        "version" : "adoptopenjdk/centos7_build_image@sha256:0dcc7d6c846a7ca4fa7814e0721c05e14a5a88f4bf67c43202586610
      }
```
Cc @andrew-m-leonard @netomi - this is part of satisfying https://github.com/adoptium/temurin-build/issues/3535

References to the new Formula stuff in CycloneDX 1.5 used in this PR:
- https://cyclonedx.org/docs/1.5/json/#formulation_items_bom-ref
- https://cyclonedx.github.io/cyclonedx-core-java/

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -869,6 +869,7 @@ generateSBoM() {
 
   # Set default SBOM formulation
   addSBOMFormulation "${javaHome}" "${classpath}" "${sbomJson}"
+  addSBOMFormulationComponent "${javaHome}" "${classpath}" "${sbomJson}"
 
   # Add Tool Summary section from configure.txt
   checkingToolSummary
@@ -988,9 +989,9 @@ addCycloneDXVersions() {
    else
        # This should probably cycle over all jars in the directory and include them
        # but this is an initial PoC for discussion ...
-       # Also - should we do somethign if the sha256sum fails? Currently SHA will show blank
+       # Also - should we do something if the sha256sum fails? Currently SHA will show blank
        local JarSha=$(sha256sum "${CYCLONEDB_DIR}/build/jar/cyclonedx-core-java.jar" | cut -d' ' -f1)
-       addSBOMFormulationProperty "${javaHome}" "${classpath}" "${sbomJson}" "CycloneDX core java JAR SHA" "${JarSha}"
+       addSBOMFormulationComponentProperty "${javaHome}" "${classpath}" "${sbomJson}" "CycloneDX SHAs" "CycloneDX core java JAR SHA" "${JarSha}"
    fi
 }
 
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -877,38 +877,9 @@ generateSBoM() {
   addSBOMMetadataProperty "${javaHome}" "${classpath}" "${sbomJson}" "OS version" "${BUILD_CONFIG[OS_FULL_VERSION]^}"
   addSBOMMetadataProperty "${javaHome}" "${classpath}" "${sbomJson}" "OS architecture" "${BUILD_CONFIG[OS_ARCHITECTURE]^}"
 
-  # Create JDK Component
-  addSBOMComponent "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "${fullVer}" "${BUILD_CONFIG[BUILD_VARIANT]^} JDK Component"
-
-  # Below add different properties to JDK component
-  # Add variant as JDK Component Property
-  addSBOMComponentProperty "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "JDK Variant" "${BUILD_CONFIG[BUILD_VARIANT]^}"
-  # Add scmRef as JDK Component Property
-  addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "SCM Ref" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/scmref.txt"
-  # Add OpenJDK source ref commit as JDK Component Property
-  addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "OpenJDK Source Commit" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/openjdkSource.txt"
-  # Add buildRef as JDK Component Property
-  addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "Temurin Build Ref" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/buildSource.txt"
-
-  # Add build timestamp
-  addSBOMComponentProperty "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "Build Timestamp" "${BUILD_CONFIG[BUILD_TIMESTAMP]}"
-
   # Set default SBOM formulation
   addSBOMFormulation "${javaHome}" "${classpath}" "${sbomJson}"
-  addSBOMFormulationComp "${javaHome}" "${classpath}" "${sbomJson}" "CycloneDX SHAs"
-
-  # Add Tool Summary section from configure.txt
-  checkingToolSummary
-  addSBOMComponentPropertyFromFile "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "Build Tools Summary" "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/dependency_tool_sum.txt"
-  # Add builtConfig JDK Component Property, load as Json string
-  built_config=$(createConfigToJsonString)
-  addSBOMComponentProperty "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "Build Config" "${built_config}"
-  # Add full_version_output JDK Component Property
-  addSBOMComponentProperty "${javaHome}" "${classpath}" "${sbomJson}" "Eclipse Temurin" "full_version_output" "${fullVerOutput}"
-  # Add makejdk_any_platform_args JDK Component Property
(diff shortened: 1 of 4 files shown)


andrew-m-leonard:
@sxa One thought would be do we need the other dependencies as well that cyclonedx pulls in jackson...
Then one method maybe to add an extrat ant target to the build.xml to dump the checksum's that are already in that file?
https://github.com/adoptium/temurin-build/blob/c7f1ad11a3c7a6d37622be2a689503142e473309/cyclonedx-lib/build.xml#L46

sxa:
> @sxa One thought would be do we need the other dependencies as well that cyclonedx pulls in jackson... Then one method maybe to add an extrat ant target to the build.xml to dump the checksum's that are already in that file?

Yeah I reckon I'd be ok with that as another potential option for retrieving the details.

Any thoughts on the format of the entries in the SBoM file?

andrew-m-leonard:
> 
> Any thoughts on the format of the entries in the SBoM file?

I asked on the CycloneDX slack, I think this is where we should have these:
https://cyclonedx.org/docs/1.5/json/#metadata_tools_oneOf_i0_components

sxa:
> I asked on the CycloneDX slack, I think this is where we should have these: https://cyclonedx.org/docs/1.5/json/#formulation_items_components

Looks like the formulation classes aren't available in our cyclonedx-core-java.jar, so I'm not sure how to do this.

stevespringett:
@sxa Formulation support was added to v1.5 of the CycloneDX specification, which is supported in cyclonedx-core-java v8.0 and higher. If your version of the jar does not support formulation, its likely an older version. The recommendation would be to update the jar to the current release.

See: https://github.com/CycloneDX/cyclonedx-core-java/tree/master/src/main/java/org/cyclonedx/model/formulation

sxa:
> @sxa Formulation support was added to v1.5 of the CycloneDX specification, which is supported in cyclonedx-core-java v8.0 and higher.

Yeah looks like we're on 1.4 just now so will need to upgrade if we're going to use that field for this.

sxa:
OK The current version of the code, run against 8.0.3, will add this into the SBoM:
```
  "formulation" : [
    {
      "components" : [
        {
          "name" : "CycloneDX SHAs",
          "properties" : [
            {
              "name" : "CycloneDX core java JAR SHA",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            }
          ],
          "type" : "framework"
        }
      ]
    }
  ]
```
There will be a bit of tidy up work to do but any comments on the format and/or names of the seconds are welcome. We probably ought to record all of the jar files that are pulled down during the build (unless, of course, we switch to installing these in the playbooks :-) )

sxa:
Updated to include all jars ... I stripped out the `.jar` suffix but I suspect that would be best left in:
```
  "formulation" : [
    {
      "components" : [
        {
          "name" : "CycloneDX SHAs",
          "properties" : [
            {
              "name" : "commons-codec",
              "value" : "b3e9f6d63a790109bf0d056611fbed1cf69055826defeb9894a71369d246ed63"
            },
            {
              "name" : "commons-io",
              "value" : "961b2f6d87dbacc5d54abf45ab7a6e2495f89b75598962d8c723cea9bc210908"
            },
            {
              "name" : "cyclonedx-core-java",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "github-package-url",
              "value" : "8e23280221afd1e6561d433dfb133252cd287167acb0eca5a991667118ff10a2"
            },
            {
              "name" : "jackson-annotations",
              "value" : "2c6869d505cf60dc066734b7d50339f975bd3adc635e26a78abb71acb4473c0d"
            },
            {
              "name" : "jackson-core",
              "value" : "b5d37a77c88277b97e3593c8740925216c06df8e4172bbde058528df04ad3e7a"
            },
            {
              "name" : "jackson-databind",
              "value" : "501d3abce4d18dcc381058ec593c5b94477906bba6efbac14dae40a642f77424"
            },
            {
              "name" : "jackson-dataformat-xml",
              "value" : "edbda6c775a36049cf0088b111ab958cca0dc70cb9326918d6cf153cb3fa426b"
            },
            {
              "name" : "json-schema",
              "value" : "968991e5718520cdd7b224770f790cf2c241cddf64d10a36c21f9f8b4a15e79c"
            },
            {
              "name" : "temurin-gen-sbom",
              "value" : "d092abca1da533019f02c6a707a67b14f4e38da581ab55c087ff7ad4a5031031"
            }
          ],
          "type" : "framework"
        }
      ]
    }
  ]
```

sxa:
Need to consider whether to switch this to using "Hash" objects on the component instead of just name/value properties.

sxa:
@netomi @andrew-m-leonard I think this is close to being ready now so I'll tentatively mark it as ready for review. The tests pass and it produces viable output for the CycloneDX SHAs. Since the last comment I've reinstated the `.jar` extension. I have a few follow-on concerns about the processes around this which I'll raise in another issue.

I've used a short name of `Comp` and `CompProp` for component[Properties] in the names - if anyone objects to that they can be changed but the names were getting quite long.

There are some notes about the lists in there too ... At the moment there is an assumption that there will only be one entry in there, so I'm always accessing the first element. Similarly there may be a debate about [whether we use hash objects](https://cyclonedx.github.io/cyclonedx-core-java/index.html?org/cyclonedx/model/Component.html) for this instead of a simple name/value pair with the SHA. I'm also currently generating them [instead of using the values in the build.xml](https://github.com/adoptium/temurin-build/pull/3538#issuecomment-1821243749) because I feel that's preferable in case they get adjusted somehow.

sbin/build.sh
```diff
@@ -877,6 +877,39 @@ generateSBoM() {
   addSBOMMetadataProperty "${javaHome}" "${classpath}" "${sbomJson}" "OS version" "${BUILD_CONFIG[OS_FULL_VERSION]^}"
   addSBOMMetadataProperty "${javaHome}" "${classpath}" "${sbomJson}" "OS architecture" "${BUILD_CONFIG[OS_ARCHITECTURE]^}"
 
+  # Create JDK Component
```
netomi, line 880: this block looks like wrong. This was the original block which created a single component: JDK
However, later on for each component that is generated during the build, a separate SBOM component will be generated. So this block should be removed and the formulation should be added in the for loop for each component imho.

sxa: Agreed - a side effect of attempting a rebase after your change went in - I'll need to do some manual fixup by the look of it. Thanks.

sxa: OK should be resolved now.

sxa:
Running a pipeline test at https://ci.adoptium.net/job/build-scripts/job/openjdk21-pipeline/228

sxa:
Note to reviewers: please allow me to merge this when approved.

netomi:
cant access the pipeline test build.

sxa:
> cant access the pipeline test build.

The github check for JDK8u has equivalent output and prints out the SBoM at the end: https://github.com/adoptium/temurin-build/actions/runs/7117734755/job/19379150291?pr=3538

@andrew-m-leonard Do we need the non-release pipelines to be restricted access now that we've split out the release ones?

sxa:
> Unless reviews explicitly object I'd like to merge this and then sort out the naming of the formulation entry (BomRef) in a subsequent PR to avoid accessing entry `(0)` in the lists.

Latest push has resolved this and is using a BomRef of `CycloneDX` to identify the correct entry in the list
```
 "formulation" : [
    {
      "components" : [
        {
          "name" : "CycloneDX jar SHAs",
          "properties" : [
            {
              "name" : "commons-codec.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "commons-io.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "cyclonedx-core-java.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "github-package-url.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "jackson-annotations.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "jackson-core.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "jackson-databind.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "jackson-dataformat-xml.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "json-schema.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            },
            {
              "name" : "temurin-gen-sbom.jar",
              "value" : "ecc371d12808dfe76047f87f8235665d74dd6cf8ec12c41d052715a3fd79e0b5"
            }
          ],
          "type" : "framework"
        }
      ],
      "bom-ref" : "CycloneDX"
    }
  ]
```
A couple of quesions (Maybe for @stevespringett ?)
- Is a `type` of `framework` suitable for the list here, or would `application` be better?
- Is using bom-ref like this as an identifier for the Component list entry a valid thing to do, and is `CyloneDX` a good identifier for this purpose (Storing the CycloneDX jar details) Maybe `CycloneDX formula` to help ensure uniqueness?

andrew-m-leonard (approved):
lgtm

karianna (approved):

PR #3543, opened 2023-11-23
Add workaround for brew update failure when running aqa smoke tests

netomi:
The tests for macos-13 fail due to an error when setting up the smoke tests.

This seems to be a pretty common problem, e.g. as documented here:

https://github.com/actions/runner-images/issues/6817

Workaround seems to be to either ignore the error or remove the files before running brew if they are present.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,7 +123,7 @@ jobs:
     steps:
     - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
-      # https://github.com/actions/runner-images/issues/6817
+    # https://github.com/actions/runner-images/issues/6817
     - name: (Mac) Workaround for homebrew
       shell: bash
       if: runner.os == 'macOS'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -123,6 +123,22 @@ jobs:
     steps:
     - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
+      # https://github.com/actions/runner-images/issues/6817
```
karianna, line 126:
```suggestion
    # https://github.com/actions/runner-images/issues/6817
```

karianna:
https://github.com/actions/setup-python/issues/577#issuecomment-1776328661 suggests forcing an overwrite instead?

netomi:
ideally that should be fixed in the runaqa action that calls brew install, this workaround is there to make the builds green as the macos13 build always seems to fail because of that.

karianna (approved):

andrew-m-leonard (approved):
lgtm

PR #3544, opened 2023-11-27
Rerun gpg recv-keys command multiple times

adamfarley:
That way, if a specific keyserver is slow and causes a timeout, we simply retry on a different server.

Testing here: https://ci.adoptium.net/job/build-scripts-pr-tester/job/build-test/job/jobs/job/jdk8u/job/jdk8u-alpine-linux-x64-temurin/141/

Note: that test is a rebuild if [this timeout failure](https://ci.adoptium.net/job/build-scripts-pr-tester/job/build-test/job/jobs/job/jdk8u/job/jdk8u-alpine-linux-x64-temurin/139/console), plus this PR's changeset.

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -336,7 +336,19 @@ checkingAndDownloadingAlsa() {
     # Note: the uptime command below is to aid diagnostics for this issue:
     # https://github.com/adoptium/temurin-build/issues/3518#issuecomment-1792606345
     uptime
-    gpg --keyserver keyserver.ubuntu.com --keyserver-options timeout=300 --recv-keys "${ALSA_LIB_GPGKEYID}"
+    # Will retry command below until it passes or we've failed 10 times.
+    # Note that the hkp://keys.gnupg.net keyserver uses round robin DNS to give a different keyserver each 
+    # time we run the command, so this rerun loop allows us to ignore problematic keyservers (timeouts etc).
+    for i in {1..10}; do
+      gpg --keyserver hkp://keys.gnupg.net --keyserver-options timeout=300 --recv-keys "${ALSA_LIB_GPGKEYID}"
+      if [[ $? == 0 ]]; then
+        break
+      elif [[ ${i} -lt 10 ]]; then
+        echo "gpg recv-keys attempt ${i} has failed. Retrying..."
+      else
+        echo "ERROR: gpg recv-keys attempt 10 has failed. Will not try again."
+      fi
+    done
     echo -e "5\ny\n" |  gpg --batch --command-fd 0 --expert --edit-key "${ALSA_LIB_GPGKEYID}" trust;
     gpg --verify alsa-lib.tar.bz2.sig alsa-lib.tar.bz2 || exit 1
 


adamfarley:
Ok, build's still running, but it's passed the point where we might see a timeout, so at least we know this change doesn't break anything. Seeking reviews and approval.

adamfarley:
Hi Jie. Yes, the issue is a timeout, mentioned [here](https://github.com/adoptium/temurin-build/issues/3518).

Also, I explained the change in keyserver via a comment in the change. This new keyserver is actually something that provides a different keyserver ip address each time you access it. That way, if we access a slow server, or a down server, we just loop back around and use a different server.

In theory anyway. For so0me reason this seems to work just fine on Alpine, mac, and Windows, but fails on all the other Linux flavours.

adamfarley:
Hi Jie,

> Does it make sense to link the PR to the issue via one of the methods described here: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

I considered doing this, but elected not to do this because I didn't want to automatically close the issue before Stewart had a chance to review the fix (on the off chance it got approved and merged yesterday). Plus, the last fix didn't solve the problem like we thought it would, so auto-close seemed unduly optimistic.

> In the future, it would be good to reference the issue if it exists in your initial PR comment so everyone can see it.

Good point. Will do. :)

> What was your source for your URL choice? When doing some preliminary searches online around GPG keyservers, there isn't too much 'official' looking documentation on active (and safe) keyservers to use...

[This](https://stackoverflow.com/questions/66217436/error-gpg-keyserver-receive-failed-no-name/68132500#68132500) was my source, which isn't even slightly trustworthy, but I didn't think that mattered in this case. [My understanding](https://askubuntu.com/questions/819002/how-can-gnupg-trust-keyserver-ubuntu-com) is that keyservers are not trustworthy sources of keys in and of themselves. This is why we specify a fingerprint when we use the gpg command, and we also verify the key upon receipt (see the gpg --verify command a few lines later).

> When making a change like that, it would be good to provide justification for it.

That's fair. I will remember this for future PRs.

sbin/prepareWorkspace.sh
```diff
@@ -336,7 +336,21 @@ checkingAndDownloadingAlsa() {
     # Note: the uptime command below is to aid diagnostics for this issue:
     # https://github.com/adoptium/temurin-build/issues/3518#issuecomment-1792606345
     uptime
-    gpg --keyserver keyserver.ubuntu.com --keyserver-options timeout=300 --recv-keys "${ALSA_LIB_GPGKEYID}"
+    # Will retry command below until it passes or we've failed 10 times.
+    # Note that the hkp://keys.gnupg.net keyserver uses round robin DNS to give a different keyserver each 
+    # time we run the command, so this rerun loop allows us to ignore problematic keyservers (timeouts etc).
+    keyserverHostArray=("keyserver.ubuntu.com" "keys.openpgp.org" "pgp.mit.edu")
+    for i in {0..2}; do
+      echo "Attempting to recieve keys from keyserver ${keyserverHostArray[$i]}"
```
netomi, line 344: typo receive

sbin/prepareWorkspace.sh
```diff
@@ -336,7 +336,21 @@ checkingAndDownloadingAlsa() {
     # Note: the uptime command below is to aid diagnostics for this issue:
     # https://github.com/adoptium/temurin-build/issues/3518#issuecomment-1792606345
     uptime
-    gpg --keyserver keyserver.ubuntu.com --keyserver-options timeout=300 --recv-keys "${ALSA_LIB_GPGKEYID}"
+    # Will retry command below until it passes or we've failed 10 times.
```
netomi, line 339: the comment does not seem to be aligned with the code. Here its written that we try 10 times, but the code below seem to try 3 different keyservers once.

sbin/prepareWorkspace.sh
adamfarley: Good catch, will fix. :)

sbin/prepareWorkspace.sh
adamfarley: We were trying to access one server 10 times, but this comment was not updated when I changed that. Will fix. Thank you. :)

andrew-m-leonard (approved):
lgtm

steelhead31 (approved):
LGTM

PR #3548, opened 2023-11-29
Add required library for Configure to work on Ubuntu 22 onwards.

### LINKED ISSUE #3547: Error whilst building via makejdk-any-platform.sh on Ubuntu22.04 x64

Attempting to build any ( 8,11,17,21 ) JDK on Ubuntu 22.04 fails with an error during the Autoconf/configure step.

```
 "error: C compiler cannot create executables"
```

**Expected behaviour:**

The build should work.

**Observed behaviour:**

The build doesn't work.

**Any other comments:**

After investigation, it appears this is related to a missing .o file

When running Autoconf/Configure for a non related C file, the same error occurs.. and after a little bit of debugging, the conftest.c that configure runs to check the c compiler works, fails to compile due to a missing "crti.o" file.

[This SO article details  a similar issue](https://stackoverflow.com/questions/6329887/how-to-fix-linker-error-cannot-find-crt1-o):

The library is present on an Ubuntu 22.04 installation in this path: /usr/lib/x86_64-linux-gnu

steelhead31:
Fixes #3547 

On Ubuntu 22.04 onwards, the object files required for autoconf/configure are in a new path, due to a change relating to multiarch implementations in Ubuntu.

The new path required to be added to LIBRARY_PATH is : /usr/lib/x86_64-linux-gnu

In order to get configure to work, the new path is appended to the LIBRARY_PATH.

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -194,6 +194,21 @@ function downloadBootJDK()
 if [ "${ARCHITECTURE}" == "x64" ]
 then
   export PATH=/opt/rh/devtoolset-2/root/usr/bin:$PATH
+  ## Fix For Issue https://github.com/adoptium/temurin-build/issues/3547
+  ## Add Missing Library Path For Ubuntu 22+
+  if [ -e /etc/os-release ]; then
+    ## source the os -release file if present
+    . /etc/os-release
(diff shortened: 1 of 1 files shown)


andrew-m-leonard (approved):
lgtm

sxa (approved):

adamfarley (approved):
LGTM

PR #3551, opened 2023-11-29
Add required library for Configure to work on Ubuntu 22 onwards. ( for both Arm64 & x64 )

steelhead31:
Extends the fix for https://github.com/adoptium/temurin-build/issues/3547

On Ubuntu 22.04 onwards, the object files required for autoconf/configure are in a new path, due to a change relating to multiarch implementations in Ubuntu.

The new path required to be added to LIBRARY_PATH is : /usr/lib/x86_64-linux-gnu for x64
The new path required to be added to LIBRARY_PATH is : /usr/lib/aarch64-linux-gnu for arm64

In order to get configure to work, the new path is appended to the LIBRARY_PATH.

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -201,9 +201,9 @@ fi
 if [ -e /etc/os-release ]; then
   ID=$(grep "^ID=" /etc/os-release | awk -F'=' '{print $2}')
   INT_VERSION_ID=$(grep "^VERSION_ID=" /etc/os-release | awk -F'"' '{print $2}' | awk -F'.' '{print $1}')
-  LIB_ARCH=$(uname -m)
+  LIB_ARCH=$(uname -m)-linux-gnu
   if [ "$ID" == "ubuntu" ] && [ "$INT_VERSION_ID" -ge "22" ]; then
-      export LIBRARY_PATH=/usr/lib/$(uname -m)-linux-gnu:$LIBRARY_PATH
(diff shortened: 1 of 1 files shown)


karianna (approved):

karianna:
Mac OS X failing GH is unrelated.

andrew-m-leonard (approved):
yep, looks good

sxa (changes requested):
Hmmm not sure we want to be so restrictive here - we have multiple other architectures - can we use `${ARCHITECTURE}-linux-gnu` on the non-x64 ones? (Needs checking to see if the names match for the others - arm32 (probably not), ppc64le and s390x. RISC-V not currently relevant as we don't use our own compiler yet but I suspect that would be ok .

steelhead31:
> Hmmm not sure we want to be so restrictive here - we have multiple other architectures - can we use `${ARCHITECTURE}-linux-gnu` on the non-x64 ones? (Needs checking to see if the names match for the others - arm32 (probably not), ppc64le and s390x. RISC-V not currently relevant as we don't use our own compiler yet but I suspect that would be ok .

> ${ARCHITECTURE}-

@sxa , architecture doesnt resolve quite right for x64, as the x86\_64 value is over-ridden with x64, I can use "uname -m" which should work and match across the many architectures, and simplify that piece of logic, I'll make an appropriate change and test.

steelhead31:
VPC has completed ok with this change: https://ci.adoptium.net/job/VagrantPlaybookCheck/OS=Ubuntu2204,label=vagrant/1812/console

sxa (approved):

Issue #3554, opened 2023-11-30
Build Auto Triage failure: ${PWD}/tooling/build_autotriage/build_autotriage.sh: No such file or directory


andrew-m-leonard:
https://github.com/andrew-m-leonard/openjdk-build/actions/runs/7044337061/job/19171821242

Looks like PWD is not defined?

adamfarley:
3 bugs have been found thus far:

1. That PWD may not be correctly specifying the location of the triage script, maybe.
2. That this action is triggering even when none of the triggers have been, well, triggered.
3. That even if the above 2 issues are resolved, we then see a failure that could be caused by the github action secrets.GITHUB_TOKEN environment variable lacking permissions to generate an issue in a repository.

Current working theories:
1. That PWD does not correctly expand when specified in a github action's "env" section. 
    Fix: Moving the PWD to the "Run Build Triage" section appears to have resolved the issue.
2. If a commit modifies one of the three files associated with the build-autotriage.yml github action, then pushing a branch containing that commit will trigger this action *even if* the user has not modified those files in their own commits specific to that branch. Current theory is that this is because github identifies "new" commits by comparing the branch to the default branch in the fork. If the user has not updated their master branch to contain the latest autotriage files, the action will trigger every time they commit a branch.
    Fix?: The latest fix for this was to only run the action when the commit is in a pull request. 
3. That we get the github token from the old repository, but we try to create the issue in the new repository. As a token's scope is the repository in which it was generated, this is a problem.
    Fix?: Currently working on this. One option is to print the script's output to the console if it's in a PR, but to perform as normal if it's triggered by the workflow. We could just make two options, but I'd rather not have code duplication if possible.

adamfarley:
Ok, I've removed the pull-request logic entirely. I can always add it into a copy of this action that is exclusive to my repository (and ignored) for testing changes.

Please review: https://github.com/adoptium/temurin-build/pull/3556

PR #3556, opened 2023-11-30
Make autotriage only run in pull requests

adamfarley:
This should prevent accidental triggering when someone fetches a commit containing a change to these files and pushes it to their fork.

Also moving a pwd envvar as its current location may cause problems in some forks.

Fixes https://github.com/adoptium/temurin-build/issues/3554


steelhead31 (approved):
LGTM

Issue #3504, opened 2023-10-17
FreeType version used across Temurin versions is inconsistent


andrew-m-leonard:
Issue: https://github.com/adoptium/temurin-build/issues/3493
highlighted the various Freetype versions used for Temurin versions, and the various different versions Temurin is built and linked with.
Should we look to provide a more consistent set of versions, and possibly use "Bundled" where possible?

sxa:
Running the tooling produced as part of https://github.com/adoptium/temurin-build/issues/3484 against JD11.0.21+9 the following was noted (using data from the SBOMs:

Platform | version
---|---
Linux/arm32 | 2.6.1
Linux/ppc64le | 2.8.0
Linux/s390x | 2.8.0
Linux/x64 | 2.3.11
Mac/aarch64 | 2.13.0
Mac/x64 | 2.13.0
Windows/x64 | 2.13.0
Windows/x86-32 | 2.13.0
AIX/ppc64 | 2.8.0
Alpine/x64 | 2.11.1
Linux/aarch64 | 2.8.0

andrew-m-leonard:
Trying to test whether using a different FreeType "system" version seems not easy. When using the build image and trying to reference a different FreeType other than the glibc 2.17 compatible one in the build container image, results in linking issues due to glibc linking. Similarly trying to build outside of docker it’s quite hard to replace a System FreeType due to it’s dependency links to other system components.

From a Reproducible Build perspective, using a different system FreeType would probably not alter the binary. It is also going to be not relevant when moving to using a specific DevKit, where the system FreeType is in the DevKit.

sxa:
> it’s quite hard to replace a System FreeType

Not sure why we'd want to do that - the goal ought to be to use the bundled freetype and ship that library with the JDK. We absolutely should not be interfering with OS-installed libraries.

> From a Reproducible Build perspective, using a different system FreeType would probably not alter the binary.

I'd dispute that - the headers are #included in the code so there's definitely a chance that different versions could make a difference to the binaries.

andrew-m-leonard:
> > it’s quite hard to replace a System FreeType
> 
> Not sure why we'd want to do that - the goal ought to be to use the bundled freetype and ship that library with the JDK. We absolutely should not be interfering with OS-installed libraries.
> 

I am trying to simply "prove" if using a different system FreeType would produce different binaries... so far I have not been able to do that..I can only use FreeType 2.8.0 which is the latest. It does not seem possible to go backwards... any ideas?

[...]

PR #3557, opened 2023-12-01
Default Temurin jdk-21+ to use bundled FreeType

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3504

PMC confirmed moving Temurin jdk21u to using "bundled" FreeType, to provide "consistent" and "reproducible" builds with FreeType.

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -19,8 +19,8 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=sbin/common/constants.sh
 source "$SCRIPT_DIR/../../sbin/common/constants.sh"
 
-if [[ ("${VARIANT}" == "${BUILD_VARIANT_TEMURIN}" || "${VARIANT}" == "${BUILD_VARIANT_HOTSPOT}") && "$JAVA_FEATURE_VERSION" -ge 21 ]]; then
-  # Temurin/hotspot jdk-21+ uses "bundled" FreeType
+if [[ "$JAVA_FEATURE_VERSION" -ge 21 ]]; then
+  # jdk-21+ uses "bundled" FreeType
   export BUILD_ARGS="${BUILD_ARGS} --freetype-dir bundled"
 else
   # Bundling our own freetype can cause problems, so we skip that on linux.
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -19,8 +19,8 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=sbin/common/constants.sh
 source "$SCRIPT_DIR/../../sbin/common/constants.sh"
 
-if [ "${VARIANT}" == "${BUILD_VARIANT_TEMURIN}" ] && [ "$JAVA_FEATURE_VERSION" -ge 21 ]; then
-  # Temurin jdk-21+ uses "bundled" FreeType
+if [ "${VARIANT}" == "${BUILD_VARIANT_TEMURIN}" || "${VARIANT}" == "${BUILD_VARIANT_HOTSPOT}" ] && [ "$JAVA_FEATURE_VERSION" -ge 21 ]; then
+  # Temurin/hotspot jdk-21+ uses "bundled" FreeType
   export BUILD_ARGS="${BUILD_ARGS} --freetype-dir bundled"
 else
   # Bundling our own freetype can cause problems, so we skip that on linux.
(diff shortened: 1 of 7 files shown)


sxa:
Are the AIX and Alpine changes here correct i.e. is freetype already included in those platforms or this this going to start adding support for it?

andrew-m-leonard:
Test jdk21u pipeline: https://ci.adoptium.net/job/build-scripts/job/openjdk21-pipeline/224/console

andrew-m-leonard:
Test pipeline with "tests" enabled:
https://ci.adoptium.net/job/build-scripts/job/openjdk21-pipeline/225/

build-farm/platform-specific-configurations/linux.sh
```diff
@@ -18,8 +18,14 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 # shellcheck source=sbin/common/constants.sh
 source "$SCRIPT_DIR/../../sbin/common/constants.sh"
-# Bundling our own freetype can cause problems, so we skip that on linux.
-export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
+
+if [[ ("${VARIANT}" == "${BUILD_VARIANT_TEMURIN}" || "${VARIANT}" == "${BUILD_VARIANT_HOTSPOT}") && "$JAVA_FEATURE_VERSION" -ge 21 ]]; then
```
sxa, line 22: I'm not convinced this syntax is working properly for the negative case ...
I wonder if there's really any reason to restrict this based on variant though?

Any issues with only using the `JAVA_FEATURE_VERSION` check?

(This comment applies to the same change on other platforms too of course)

andrew-m-leonard: OpenJ9 already makes a conscious decision for Windows: https://github.com/adoptium/temurin-build/blob/da2408e4ea988090835f15f29cb170873cced045/build-farm/platform-specific-configurations/windows.sh#L159

build-farm/platform-specific-configurations/mac.sh
```diff
@@ -68,6 +68,11 @@ else
   fi
 fi
 
+if [[ ("${VARIANT}" == "${BUILD_VARIANT_TEMURIN}" || "${VARIANT}" == "${BUILD_VARIANT_HOTSPOT}") && "$JAVA_FEATURE_VERSION" -ge 21 ]]; then
+  # Temurin/hotspot jdk-21+ uses "bundled" FreeType
+  export BUILD_ARGS="${BUILD_ARGS} --freetype-dir bundled"
```
adamfarley, line 73: If we're doing this for mac, can I suggest limiting the use of "--skip-freetype" [here](https://github.com/adoptium/temurin-build/blob/da2408e4ea988090835f15f29cb170873cced045/.github/workflows/build.yml#L170) to jdk21+?

I mean, this code will override that code, so this isn't a functional improvement. It just prevents us having a "skip" followed by a "bundle", which could confuse developers reading the command.

andrew-m-leonard: thanks Adam, i've updated that yml, it does not actually need to specify --skip-freetype as that determination is already done in prepareWorkspace.sh/build.sh

build-farm/platform-specific-configurations/linux.sh
pshipton: No concerns. Afaik for OpenJ9 AIX, Win, Mac already bundle freetype by default for all versions. This would start bundling on linux platforms for jdk21+.

andrew-m-leonard: Thanks @pshipton @AdamBrousseau 
Since I can't reliably test building jdk21u openj9, it's probably best if that's done in a separate PR.
So I'll get this merged, and the VARIANT check can simply be updated if tests your side prove ok?

pshipton: On the other hand, if it works for Temurin it should also work for OpenJ9. I'm willing to accept the change, and if it does end up causing a problem then deal with it.

andrew-m-leonard: ok sounds like a good plan, thanks @pshipton I shall update

andrew-m-leonard: updated PR to remove VARIANT check

sxa (approved):

pshipton (approved):

build-farm/platform-specific-configurations/linux.sh
pshipton: fwiw I built most of the openj9 linux platforms with `--with-freetype=bundled` without issue. The xlinux build is still pending but shouldn't be an issue since the other platforms succeeded.

pshipton: OpenJ9 has changed the configure defaults so freetype is bundled on all platforms but linux. With this change it should be bundled on linux too for jdk21+. Once the change is accepted we can see if it's working.

PR #3569, opened 2023-12-09
Add amount of disk space used by build to the log

sxa:
This will help log information that we can give to end users to tell how much space is required for builds, and also help with capacity planning on machines

### DIFF
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -255,5 +255,6 @@ CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM//\"/temporary
 bash -c "$MAC_ROSETTA_PREFIX $PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh --clean-git-repo --jdk-boot-dir ${JDK_BOOT_DIR} --configure-args \"${CONFIGURE_ARGS_FOR_ANY_PLATFORM}\" --target-file-name ${FILENAME} ${TAG_OPTION} ${OPTIONS} ${BUILD_ARGS} ${VARIANT_ARG} ${JAVA_TO_BUILD}"
 
 if [ -d "${WORKSPACE}" ]; then
-  echo "Total disk space in Kb consumed by build process:" "$(du -sk "$WORKSPACE")"
+  SPACEUSED=$(du -sk "$WORKSPACE")
+  echo "Total disk space in Kb consumed by build process: $SPACEUSED"
 fi
--- a/build-farm/make-adopt-build-farm.sh
+++ b/build-farm/make-adopt-build-farm.sh
@@ -254,4 +254,6 @@ CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM//\"/temporary
 # shellcheck disable=SC2086
 bash -c "$MAC_ROSETTA_PREFIX $PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh --clean-git-repo --jdk-boot-dir ${JDK_BOOT_DIR} --configure-args \"${CONFIGURE_ARGS_FOR_ANY_PLATFORM}\" --target-file-name ${FILENAME} ${TAG_OPTION} ${OPTIONS} ${BUILD_ARGS} ${VARIANT_ARG} ${JAVA_TO_BUILD}"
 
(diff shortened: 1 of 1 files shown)


sxa:
run tests

karianna (approved):

karianna:
Looks like TKG is broken - unrelated

steelhead31 (approved):
LGTM

sxa:
run tests

sxa:
FYI to reviewers the new push is to fix quoting which was a mess so hopefully doesn't affect you're review :-)

karianna (approved):

karianna:
@sxa not sure why the checks are all failing..

sxa:
See separate new issue referenced above for the GitHub check problems.

sxa:
Switching this to draft just now becuase it's not working. When I run `make-adopt-build-farm` it gets to the end of `build.sh` and then stops after `build.sh : 09:56:13 : All done!` so it's not falling back to this script - maybe the result of an exec somewhere? Another commit will be required for this to work as expected (Or it could potentially be moved into `build.sh` however it's more for our infrastructure's benefit so this feels like the right place for it)

sxa:
Rebasing to see if any of the actions failures fix themselves ...

karianna:
only the gpg failure on centos7

build-farm/make-adopt-build-farm.sh
```diff
@@ -252,4 +252,9 @@ echo "$MAC_ROSETTA_PREFIX $PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh --clea
 CONFIGURE_ARGS_FOR_ANY_PLATFORM="${CONFIGURE_ARGS_FOR_ANY_PLATFORM//\"/temporary_speech_mark_placeholder}"
 
 # shellcheck disable=SC2086
-bash -c "$MAC_ROSETTA_PREFIX $PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh --clean-git-repo --jdk-boot-dir ${JDK_BOOT_DIR} --configure-args \"${CONFIGURE_ARGS_FOR_ANY_PLATFORM}\" --target-file-name ${FILENAME} ${TAG_OPTION} ${OPTIONS} ${BUILD_ARGS} ${VARIANT_ARG} ${JAVA_TO_BUILD}"
\ No newline at end of file
+bash -c "$MAC_ROSETTA_PREFIX $PLATFORM_SCRIPT_DIR/../makejdk-any-platform.sh --clean-git-repo --jdk-boot-dir ${JDK_BOOT_DIR} --configure-args \"${CONFIGURE_ARGS_FOR_ANY_PLATFORM}\" --target-file-name ${FILENAME} ${TAG_OPTION} ${OPTIONS} ${BUILD_ARGS} ${VARIANT_ARG} ${JAVA_TO_BUILD}"
+
+if [ -d "${WORKSPACE}" ]; then
+  SPACEUSED=$(du -sk "$WORKSPACE")
+  echo "Total disk space in Kb consumed by build process: $SPACEUSED"
```
andrew-m-leonard, line 259: Could we also dump the current volume's free disk space? This could help figure the github runner issue?

andrew-m-leonard: Could we also print this info at the start of the build as well?

sxa: I considered that but I'd rather not do that in this PR because the before and after data is potentially misleading for docker containers which use shared space, plus we have Nagios monitoring to detect problems in that area
2. For the github runner case, presumably you could do a subsequent PR with the change you want - that would allow you to test it in the GHA checks even if you don't end up merging it.

sxa:
> only the gpg failure on centos7

There are still a few issues in this area which I believe others are covering at the moment. On the basis that this PR is (hopefully clearly) not the cause of them I'm happy to merge this with those checks failing.

sxa:
> Switching this to draft just now becuase it's not working. When I run `make-adopt-build-farm` it gets to the end of `build.sh` and then stops after `build.sh : 09:56:13 : All done!` so it's not falling back to this script - maybe the result of an exec somewhere?

This was user error on my part as I hadn't set the `WORKSPACE` directory before running my local test. It is working as expected. IMHO this is good to merge (@andrew-m-leonard if you're ok to approve given [my earlier response to you](https://github.com/adoptium/temurin-build/pull/3569#discussion_r1482821362) that would be appreciated.

sxa:
@andrew-m-leonard I'm going to merge this now since it has two approvers and your suggestions do not conflict with this. We can do a subsequent PR if you want to add the `df` somewhere.

build-farm/make-adopt-build-farm.sh
andrew-m-leonard: yes makes sense

PR #3572, opened 2023-12-11
Fix joinPathOS method that fails on solaris

netomi:
Previously printf was used which behaves differently on solaris it seems.

Using a method that should work anywhere.

Use IFS='/' to separate path elements and remove duplicate slashes afterwards using tr.

### DIFF
--- a/sbin/common/common.sh
+++ b/sbin/common/common.sh
@@ -98,7 +98,7 @@ function setDockerVolumeSuffix() {
 
 # Joins multiple parts to a valid file path for the current OS
 function joinPathOS() {
-  local path=$(printf '/%s' "${@}" | sed 's|/\+|/|g')
+  local path=$(IFS=/; echo "/$*" | tr -s /)
   if [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]]; then
       path=$(cygpath -w "${path}")
   fi


karianna (commented):
Looks OK but could do with a comment saying why we are using IFS

netomi:
> Looks OK but could do with a comment saying why we are using IFS

added some comment

sxa (commented):
I'm a bit unclear as to why we need `IFS` in this case - is the outcome of the joinPath any different from the somewhat simpler:
```
echo "${@}" | tr ' ' /
```
Or are there some edge cases that would be missed with that? I'm probably missing something obvious but thought it was worth checking :-)
(EDIT: I've missed out the `tr -s` on there that's likely needed too)

netomi:
I think that would also work:

```
echo "/${@}" | tr ' ' / | tr -s /
```

I wanted to have a reliable way to construct a valid path from components without having to worry about slashes and remove duplicate ones. While this is not necessary according to the POSIX standard, it still looks weird in logs.

netomi:
applied the change and removed the comment again. The comment was requested  to explain the use of IFS, I think right now its pretty straight forward.

karianna (approved):

sxa:
Test build: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-temurin/428/console

sxa:
LGTM - `OpenJDK SBOM will be /export/home/jenkins/workspace/build-scripts/jobs/jdk8u/jdk8u-solaris-sparcv9-temurin/workspace/target/OpenJDK8U-sbom_sparcv9_solaris_hotspot_2023-12-14-09-42.json.`

PR #3590, opened 2023-12-20
Enable GCC 10 for RISC-V

luhenry:
Given the rapid evolution of RISC-V, the newer the GCC toolchain, the better.

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -331,6 +331,10 @@ if [ "${VARIANT}" == "${BUILD_VARIANT_DRAGONWELL}" ] && [ "$JAVA_FEATURE_VERSION
   export CXX=/usr/local/gcc9/bin/g++-9.3
   # Enable GCC 10 for Java 17+ for repeatable builds, but not for our supported releases
   # Ref https://github.com/adoptium/temurin-build/issues/2787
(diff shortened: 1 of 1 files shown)


build-farm/platform-specific-configurations/linux.sh
```diff
@@ -331,6 +331,10 @@ if [ "${VARIANT}" == "${BUILD_VARIANT_DRAGONWELL}" ] && [ "$JAVA_FEATURE_VERSION
   export CXX=/usr/local/gcc9/bin/g++-9.3
   # Enable GCC 10 for Java 17+ for repeatable builds, but not for our supported releases
   # Ref https://github.com/adoptium/temurin-build/issues/2787
+if [ "${ARCHITECTURE}" == "ricv64" ] && [ "${VARIANT}" == "${BUILD_VARIANT_HOTSPOT}" ] && [ -r /usr/bin/gcc-10 ]; then
```
sxa, line 334: Typo in `riscv64` and it should be an `elif` statement. I'm also suggesting removing the variant check as our normal builds are with a `VARIANT` of `temurin` distinct from `hotpot` and the only potential exception of dragonwell is covered in the previous clause. Perhaps we should check for either of those two but given that gcc-9 seemed generally quite broken in earlier tests I'm inclined to just stick with this.
```suggestion
elif [ "${ARCHITECTURE}" == "riscv64" ] && [ -r /usr/bin/gcc-10 ]; then
```

sxa (changes requested):

sxa (approved):

tellison (approved):

PR #3601, opened 2024-01-08
fix labeler

karianna:
WARN: Because of https://github.com/actions/labeler?tab=readme-ov-file#notes-regarding-pull_request_target-event this PR will appear to have failed the GH Action check.  We won't know its worked until a subsequent PR is created.  This is why we can't have nice things :-)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -8,30 +8,38 @@
 #   - tests/*
 
 ghActions:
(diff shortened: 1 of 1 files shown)


gdams (approved):

sxa (approved):
Approving as it's equivalent to what we did in the pipeines repository

PR #3629, opened 2024-02-03
Special case boot version for JDK11u on riscv64

luhenry:


gdams (approved):

sxa (approved):
Well need to have this for the Temurin variant too but this will be ok for now

PR #3640, opened 2024-02-08
Add df comments to make-adopt-build-farm.sh

sxa:
For TESTING ONLY to verify what's happening with disk space in github actions.


sxa:
FYI @andrew-m-leonard this has some `df -k` checks which may help debugging the jdk11 actions failures.

sxa:
Looks like there is 10G available:
```
overlay         76026616 65962520  10047712  87% /
```

PR #3651, opened 2024-02-15
Correct check for alpine when setting up GNUPGHOME

andrew-m-leonard:
When build using makejdk-any-platform.sh, OPERATING_SYSTEM is not set:
```
build/sbin/prepareWorkspace.sh: line 323: OPERATING_SYSTEM: unbound variable
```
Need to check for Alpine like it does in https://github.com/adoptium/temurin-build/blob/c78dc6cf2d1f1ddbe689220afd80eb957ac6d2a5/build-farm/make-adopt-build-farm.sh#L46

Also GNUPGHOME setup for Alsa not correct if run directly from makejdk-any-platform.sh


sxa (approved):
I'd have through we'd have another variable for this, but I'm ok with this fix - thanks.

andrew-m-leonard:
> I'd have through we'd have another variable for this, but I'm ok with this fix - thanks.

thanks, not when not called through Adopt farm

adamfarley (approved):
LGTM

Issue #3659, opened 2024-02-19
GPG path lengths too long on RISC-V evaluation builds


sxa:
Ref: https://ci.adoptium.net/job/build-scripts/job/jobs/job/evaluation/job/jobs/job/jdk21u/job/jdk21u-evaluation-linux-riscv64-temurin/138/console (A re-run of the previous build to ensure it was running from the latest fixes in temurin-build):
```
10:31:54  GNUPGHOME=/mnt/disk-1/jenkins/workspace/build-scripts/jobs/evaluation/jobs/jdk21u/jdk21u-evaluation-linux-riscv64-temurin/workspace/.gpg-temp
10:31:54   10:31:53 up 51 days, 11:15,  0 users,  load average: 0.27, 0.08, 0.02
10:31:54  gpg: keybox '/mnt/disk-1/jenkins/workspace/build-scripts/jobs/evaluation/jobs/jdk21u/jdk21u-evaluation-linux-riscv64-temurin/workspace/.gpg-temp/pubring.kbx' created
10:32:00  gpg: connecting dirmngr at '/mnt/disk-1/jenkins/workspace/build-scripts/jobs/evaluation/jobs/jdk21u/jdk21u-evaluation-linux-riscv64-temurin/workspace/.gpg-temp/S.dirmngr' failed: File name too long
10:32:00  gpg: keyserver receive failed: No dirmngr
10:32:00  gpg recv-keys attempt has failed. Retrying after 10 second pause...
10:32:18  gpg: connecting dirmngr at '/mnt/disk-1/jenkins/workspace/build-scripts/jobs/evaluation/jobs/jdk21u/jdk21u-evaluation-linux-riscv64-temurin/workspace/.gpg-temp/S.dirmngr' failed: File name too long
10:32:18  gpg: keyserver receive failed: No dirmngr
```
FYI @luhenry @andrew-m-leonard since you've both been doing fixes in this area recently.

Related: https://github.com/adoptium/temurin-build/pull/3643  https://github.com/adoptium/temurin-build/pull/3630

luhenry:
It is https://github.com/adoptium/temurin-build/commit/88fc1e54aed1f11c205236a364fd7453b0cea412 which will override `GNUPGHOME` when not on alpine. @andrew-m-leonard could you please fix/revert?

I also don't understand why it's not enough to set it in `build-farm/platform-specific-configurations/alpine-linux.sh` given it's loaded before `sbin/prepareWorkspace.sh`?

[...]

PR #3663, opened 2024-02-20
Ensure GPG temp dir can be created for long build folder paths on linux riscv64

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3659

linux riscv64 GPG has folder path length limitations, similar to Alpine.

Successful GPG creation test: https://ci.adoptium.net/job/build-scripts/job/jobs/job/evaluation/job/jobs/job/jdk21u/job/jdk21u-evaluation-linux-riscv64-temurin/142/

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -324,7 +324,7 @@ checkingAndDownloadingAlsa() {
         # Use /tmp for alpine in preference to $HOME as Alpine fails gpg operation if PWD > 83 characters
         # Alpine also cannot create ~/.gpg-temp within a docker context
         GNUPGHOME="$(mktemp -d /tmp/.gpg-temp.XXXXXX)"
-    elif [ "${BUILD_CONFIG[TARGET_OS]}" == "linux" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ] && [ "$(pwd | wc -c)" -gt 83 ]; then
+    elif [[ "${BUILD_CONFIG[TARGET_OS]}" == "linux" ]] && [[ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ]] && [[ "$(pwd | wc -c)" -gt 83 ]]; then
         # linux riscv64 also has gpg pathlength issue
         GNUPGHOME="$(mktemp -d /tmp/.gpg-temp.XXXXXX)"
     else
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -324,6 +324,9 @@ checkingAndDownloadingAlsa() {
         # Use /tmp for alpine in preference to $HOME as Alpine fails gpg operation if PWD > 83 characters
         # Alpine also cannot create ~/.gpg-temp within a docker context
         GNUPGHOME="$(mktemp -d /tmp/.gpg-temp.XXXXXX)"
+    elif [ "${BUILD_CONFIG[TARGET_OS]}" == "linux" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ] && [ "$(pwd | wc -c)" -gt 83 ]; then
+        # linux riscv64 also has gpg pathlength issue
+        GNUPGHOME="$(mktemp -d /tmp/.gpg-temp.XXXXXX)"
     else
         GNUPGHOME="${BUILD_CONFIG[WORKSPACE_DIR]:-$PWD}/.gpg-temp"
     fi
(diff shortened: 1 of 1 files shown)


sxa (commented):
Is there a reason this needs to be an `elif` instead of adding it the section that detects Alpine? This feels like unnecessary duplication

sbin/prepareWorkspace.sh
```diff
@@ -324,6 +324,9 @@ checkingAndDownloadingAlsa() {
         # Use /tmp for alpine in preference to $HOME as Alpine fails gpg operation if PWD > 83 characters
         # Alpine also cannot create ~/.gpg-temp within a docker context
         GNUPGHOME="$(mktemp -d /tmp/.gpg-temp.XXXXXX)"
+    elif [ "${BUILD_CONFIG[OS_KERNEL_NAME]}" == "linux" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ] && [ "$(pwd | wc -c)" -gt 83 ]; then
```
luhenry, line 327: Should we instead take this path if `GNUPGHOME` is not defined?

andrew-m-leonard: not sure, we don't do in the build-farm path either ?

andrew-m-leonard: I think the intention is to create it in the WORKSPACE_DIR

luhenry: > not sure, we don't do in the build-farm path either ?

We should add it there as well then. Should we move that specific piece of code to a separate bash script that we include from all relevant places? That sounds easier than copy/pasting it all around.

andrew-m-leonard: @sxa I think you're more experienced with GPG than me, what's your view on honouring GNUPGHOME if it's already set in the environment? eg.by a "user"

sxa: As long as we're ok with a user potentially shooting themselves in the foot by setting it something >83 characters and having us honour that. I thin it's a subtle enough case that I don't mind either way (I'd probably err on the side of not doing the check as it's less code but 🤷🏻 - I'll leave it to you. It's an edge case that won't affect our build process unless we chose to define it on our machines in jenkins or something similar.

sxa:
> > Is there a reason this needs to be an `elif` instead of adding it the section that detects Alpine? This feels like unnecessary duplication
> 
> Not sure this complex if works in bash would it? hence I took the less elegant but works approach
> 
> ```
> if [[ -r /etc/alpine-release ] || [[ "${BUILD_CONFIG[OS_KERNEL_NAME]}" == "linux" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ]] ] && [ "$(pwd | wc -c)" -gt 83 ]; then
> ```

This is the syntax you're looking for if we're aiming for always on alpine, but also when the path is >83 characters on Linux/riscv64:

`if [ -r /etc/alpine-release ] || ( [ "${BUILD_CONFIG[OS_KERNEL_NAME]}" == "linux" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ] && [ "$(pwd | wc -c)" -gt 83 ] ); then echo true; fi`

sbin/prepareWorkspace.sh
andrew-m-leonard: I'm also keen on less code, so will leave that as is for the moment

andrew-m-leonard:
> > > Is there a reason this needs to be an `elif` instead of adding it the section that detects Alpine? This feels like unnecessary duplication
> > 
> > 
> > Not sure this complex if works in bash would it? hence I took the less elegant but works approach
> > ```
> > if [[ -r /etc/alpine-release ] || [[ "${BUILD_CONFIG[OS_KERNEL_NAME]}" == "linux" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ]] ] && [ "$(pwd | wc -c)" -gt 83 ]; then
> > ```
> 
> This is the syntax you're looking for if we're aiming for always on alpine, but also when the path is >83 characters on Linux/riscv64:
> 
> `if [ -r /etc/alpine-release ] || ( [ "${BUILD_CONFIG[OS_KERNEL_NAME]}" == "linux" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" == "riscv64" ] && [ "$(pwd | wc -c)" -gt 83 ] ); then ...`

Thanks @sxa updated

sxa (approved):
We could abstract out the length count from the expression, but sine the linter is happy I'm ok with this and the logic looks sound :-)

steelhead31 (approved):
LGTM

PR #3680, opened 2024-03-01
Fix GH action permissions

gdams:
Now that GH has changed the default permissions to read we need to explicitly define the permissions


smlambert (approved):

steelhead31 (approved):
LGTM

PR #3691, opened 2024-03-06
Reproducible jdk21u linux build fixes and updates for debug symbols

andrew-m-leonard:
Updates to the comparable build scripts to support Linux comparisons, and also recent changes due to backports for debug symbol mapping:
- Removal of linux debug symbol mapping for https://bugs.openjdk.org/browse/JDK-8323667 which is now backported to jdk21u and can be removed
- Fixes to the comparable build scripts to support Linux comparisons

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,12 +134,16 @@ jobs:
           rm /usr/local/bin/2to3-3.12 || true
           rm /usr/local/bin/idle3 || true
           rm /usr/local/bin/idle3.11 || true
+          rm /usr/local/bin/idle3.12 || true
           rm /usr/local/bin/pydoc3 || true
           rm /usr/local/bin/pydoc3.11 || true
+          rm /usr/local/bin/pydoc3.12 || true
           rm /usr/local/bin/python3 || true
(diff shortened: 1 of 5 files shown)


karianna (approved):

steelhead31 (approved):
LGTM

sxa (commented):
While this has already been merged I do want to have a further discussion on this (not in this PR though!) If devkits are to be our standard way of running for JDK21+ I think the scripts should use a pre-installed devkit on the machine **in place of** the gcc11 builds we are currently using.

IMHO the plan here ought to be to remove the need for those Adoptium GCC11 builds to be on the machine and replace them with the corresponding devkit in the build images which will get picked up by default by the configurations here.

We should also consider, in light of the devkits, whether to change the distributions we build on.

Also, it would be worth considering having devkits for some of the earlier releases too.

andrew-m-leonard:
Yes, the next progression when enabling DevKit via temurin-build (PR Draft https://github.com/adoptium/temurin-build/pull/3706/files), initially downloading..., but we can progress that to check if the requested DevKit eg.gcc-11.3.0-Centos7.9.2009-b01, is already installed in /usr/local/gcc11 ...

PR #3696, opened 2024-03-11
Support building in a user supplied OpenJDK build directory

andrew-m-leonard:
To allow users to specify an alternate directory for the openjdk build output, rather than the current default OpenJDK git source folder.
Some users prefer to keep build output in a separate directory to the git source, this option will allow that to be possible.
This change also provides a workaround for jdk-21.0.3 for upstream reproducible build bug: https://bugs.openjdk.org/browse/JDK-8326685

A new temurin-builds scripts build arg:
```
--user-openjdk-build-root-directory <UserDirectory>
```
The UserDirectory must be either "empty" or not exist (in which case it is created)

The temurin-build/sbin/build.sh unfortunately had numerous hardcoded openjdk build folder locations (build/src/build/*), this PR corrects that to accurately use the location being specified, or the default existing location. It was "stepping into" the openjdk src location based on the assumption that was the openjdk build root location.


karianna (approved):

steelhead31 (approved):
LGTM

PR #3706, opened 2024-03-14
Provide a temurin-build build arg --use-adoptium-devkit <devkit_tag>

andrew-m-leonard:
Enable users to build using an Adoptium built DevKit from https://github.com/adoptium/devkit-binaries
Adding a new build arg:
```
--use-adoptium-devkit <devkit_tag>
```
eg.:
```
--use-adoptium-devkit gcc-11.3.0-Centos7.6.1810-b01
```
This will download and verify the DevKit for the given architecture from the devkit-binaries release specified.
Then build using that devkit


andrew-m-leonard:
Currently based on PR https://github.com/adoptium/temurin-build/pull/3696/files
Will need rebasing once that is merged.

sxa:
Time for a controversial discussion starter ;-)

This is a follow-on to my recent comment in https://github.com/adoptium/temurin-build/pull/3691#pullrequestreview-1954991321 since it's especially relevant here but I strongly believe it would be preferable to have these devkits installed into the Linux container build image as a default on the system - replacing the Adoptium/GCC11 installs we currently put on - for use going forward so they can be used by anyone using our images who are typically not using the `make-adopt-build-farm.sh`

We should make the end-user process as slick as possible for being to reproduce the Temurin builds. I know there has been some concern expressed relating to the provenance of these things if we don't download but that would also be true of anything else in these images such as `make` and anything else that is on these images. Plus the docker images are checksummed and therefore verifiable.

So TL;DR I'd really like to modify the playbooks to add the devkit in instead of downloading for each build and keep those on the images in parallel for a while with the Adoptium/GCC11 ones, then look at removing the Adoptium/GCC11  later, especially since the Adoptium/GCC11 ones will be redundant and so potentially lead to confusion if they are picked up.

Also there may be justification for a new type of build image that contains the minimal amount of packages to build Temurin 20+ plus the devkit (This could be really quite small based on the tests I've done - the devkit tarballs seem no larger then the previous gcc-only ones)

sxa (commented):
Time for a controversial discussion starter ;-)

This is a follow-on to my recent comment in https://github.com/adoptium/temurin-build/pull/3691#pullrequestreview-1954991321 since it's especially relevant here but I strongly believe it would be preferable to have these devkits installed into the Linux container build image as a default on the system - replacing the Adoptium/GCC11 installs we currently put on - for use going forward so they can be used by anyone using our images who are typically not using the make-adopt-build-farm.sh

We should make the end-user process as slick as possible for being to reproduce the Temurin builds. I know there has been some concern expressed relating to the provenance of these things if we don't download but that would also be true of anything else in these images such as make and anything else that is on these images. Plus the docker images are checksummed and therefore verifiable.

So TL;DR I'd really like to modify the playbooks to add the devkit in instead of downloading for each build and keep those on the images in parallel for a while with the Adoptium/GCC11 ones, then look at removing the Adoptium/GCC11 later, especially since the Adoptium/GCC11 ones will be redundant and so potentially lead to confusion if they are picked up.

Also there may be justification for a new type of build image that contains the minimal amount of packages to build Temurin 20+ plus the devkit (This could be really quite small based on the tests I've done - the devkit tarballs seem no larger then the previous gcc-only ones)

andrew-m-leonard:
Yes agree, a follow on from this PR, would be to check if the requested devkit_tag (eg.gcc-11.3.0-Centos7.9.2009-b01) is already located in /usr/local/gcc11 (or whereever...), hence why I have added the tag to the devkit.info:
https://github.com/adoptium/ci-jenkins-pipelines/pull/955/files#diff-6f0823534c91127817b92b81710a0e3001e18d23eda57f34f4423a27daaece5dR45

Given the short runway for April, I want to get it functional downloading, before that next step.

sbin/prepareWorkspace.sh
```diff
@@ -577,6 +564,67 @@ prepareMozillaCacerts() {
...
+    mkdir -p "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/devkit"
+
+    devkit_tar="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/devkit/devkit.tar.xz"
+
+    setupGpg
+
+    # Determine DevKit tarball to download for this arch and release
+    devkitUrl="https://github.com/adoptium/devkit-binaries/releases/download/${BUILD_CONFIG[USE_ADOPTIUM_DEVKIT]}"
+    devkit_target="${BUILD_CONFIG[OS_ARCHITECTURE]}-linux-gnu"
+    devkit="devkit-$(echo "${BUILD_CONFIG[USE_ADOPTIUM_DEVKIT]}" | cut -d'-' -f1-3)-${devkit_target}-$(echo "${BUILD_CONFIG[USE_ADOPTIUM_DEVKIT]}" | cut -d'-' -f4)"
```
sxa, line 599: Is there a reason why we haven't made the filename a real substring of the release name? This manuipulation seems to add additional unnecessary complexity for retrieving the appropriate one with a parameter.

andrew-m-leonard: so I assume you mean the "reverse" ? as the filename contains the "arch", and the release contains DevKit's for all the arch's ?
Maybe easier to view an example : https://github.com/adoptium/devkit-binaries/releases/tag/gcc-11.3.0-Centos7.9.2009-b01
The filename is a "superset" of the release tag

andrew-m-leonard: so the above code needs to get the correct "architecture" tarball for the given release

sxa: > so I assume you mean the "reverse"

Correct 😳 

>  The filename is a "superset" of the release tag

🤔 
**Release tag:** gcc-11.3.0-Centos7.9.2009-b01
**Filename:** devkit-gcc-11.3.0-Centos7.9.2009-aarch64-linux-gnu-b01.tar.xz

The `b01` is not in the same place, so you have to insert another bit in the middle of the release tag to generate the filename instead of it just being `devkit-${tag}-${architecture}-linux-gnu.tar.xz

andrew-m-leonard: yeah, i'm just used to putting "build numbers" at the end of things :-)
I can switch it around as you suggest, sounds sensible, cheers

andrew-m-leonard: done

sbin/prepareWorkspace.sh
```diff
@@ -577,6 +564,67 @@ prepareMozillaCacerts() {
...
+    # GPG verify
+    gpg --keyserver keyserver.ubuntu.com --recv-keys 3B04D753C9050D9A5D343F39843C48A565F8F04B
+    echo -e "5\ny\n" |  gpg --batch --command-fd 0 --expert --edit-key 3B04D753C9050D9A5D343F39843C48A565F8F04B trust;
+    gpg --verify "${devkit_tar}.sig" "${devkit_tar}" || exit 1
+
+    tar --xz -xpf "${devkit_tar}" -C "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/devkit"
+    rm "${devkit_tar}"
+    rm "${devkit_tar}.sig"
+
+    # Validate devkit.info matches requirements
```
sxa, line 615:
```suggestion
    # Validate devkit.info matches value passed in and currnet architecture
```

sxa (approved):
Looks ok, but we perhaps need to have a bit of documentation somewhere on using the devkit and make it clear / throw a warning if the --with-devkit and --with-adoptium-devkit are both specified, since the latter will set the --with-devkit option to the appropriate location after download.

sbin/prepareWorkspace.sh
```diff
@@ -577,6 +564,67 @@ prepareMozillaCacerts() {
...
+    echo "Downloading DevKit : ${devkitUrl}/${devkit}.tar.xz"
+    curl -L --fail --silent --show-error -o "${devkit_tar}" "${devkitUrl}/${devkit}.tar.xz"
+    curl -L --fail --silent --show-error -o "${devkit_tar}.sig" "${devkitUrl}/${devkit}.tar.xz.sig"
+
+    # GPG verify
+    gpg --keyserver keyserver.ubuntu.com --recv-keys 3B04D753C9050D9A5D343F39843C48A565F8F04B
+    echo -e "5\ny\n" |  gpg --batch --command-fd 0 --expert --edit-key 3B04D753C9050D9A5D343F39843C48A565F8F04B trust;
+    gpg --verify "${devkit_tar}.sig" "${devkit_tar}" || exit 1
+
+    tar --xz -xpf "${devkit_tar}" -C "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/devkit"
```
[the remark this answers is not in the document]
andrew-m-leonard: thanks

andrew-m-leonard:
> Looks ok, but we perhaps need to have a bit of documentation somewhere on using the devkit and make it clear / throw a warning if the --with-devkit and --with-adoptium-devkit are both specified, since the latter will set the --with-devkit option to the appropriate location after download.

thanks, i for got to update the README.md, done...
I will do a blog about DevKits at some point

steelhead31 (approved):
LGTM

Issue #3711, opened 2024-03-18
Add reproducible comparison tests on windows


sophia-guo:
Issues:

- [x] https://github.com/adoptium/infrastructure/issues/3453
- [x] Auto trigger by upstream( beta-trigger) , figure out the jdk and sbom file - adding reproducible.mk
- [x] workdir doesn't work with jenkins user,
- [x] install apt-cygwin - no permission @steelhead31 if apt-cygwin is must we will have to ask infra to install it on the jenkins agents?
- [x] Invalid build.sh option: --with-version-opt 
- [x] configure: error: Your base path is too long. It is 170 characters long, but only 100 is supported. 
- [ ] Error: Cannot locate spec.gmk, given by SPEC=win_x86-64. https://github.com/adoptium/temurin-build/issues/3711#issuecomment-2025880943

steelhead31:
@sophia-guo the requirement for apt-cygwin is no longer relevant, as the playbooks have been updated to install jq , so there is no need for this... I'll put a PR in for this update.

sophia-guo:
Invalid build.sh option: --with-version-opt


```
14:11:15  Building JDK...
14:11:15  Starting ./makejdk-any-platform.sh to configure, build (Adoptium)OpenJDK binary
14:11:16  Parsing opt: --clean-git-repo
14:11:16  Possible opt arg: --jdk-boot-dir
14:11:16  Parsing opt: --jdk-boot-dir
14:11:16  Possible opt arg: /cygdrive/c/jenkins/workspace/Grinder/aqa-tests/TKG/output_17110446512142/Rebuild_Same_JDK_Reproducibility_Test_win_0/cmp20240321181053/jdk-20.0.2+9
14:11:16  Parsing opt: --with-version-opt
14:11:16  Possible opt arg: ea--target-file-name
14:11:16  Invalid build.sh option: --with-version-opt
14:11:16  cp: cannot stat '/cygdrive/c/jenkins/workspace/Grinder/aqa-tests/TKG/output_17110446512142/Rebuild_Same_JDK_Reproducibility_Test_win_0/cmp20240321181053/temurin-
```

https://ci.adoptium.net/view/Test_grinder/job/Grinder/9240/console

sophia-guo:
configure: error: Your base path is too long. It is 170 characters long, but only 100 is supported

Top dir 

```

[...]

PR #3718, opened 2024-03-19
Remove apt-cyg installation and check for jq.

steelhead31:
Solves part of #3711 

Remove installation of apt-cyg and jq, as these are installed via the playbooks.

The script now checks if jq is available, and errors if not present and in the path.

### DIFF
--- a/tooling/reproducible/windows_repro_build_compare.sh
+++ b/tooling/reproducible/windows_repro_build_compare.sh
@@ -137,38 +137,11 @@ Check_Parameters() {
   fi
 }
 
-Install_PreReqs() {
-  # Install Cygwin Package Manager If Not Present
-  echo "Checking If Apt-Cyg Is Already Installed"
-  if [ -f /usr/local/bin/apt-cyg ]; then
-    echo "Skipping apt-cyg Install"
-    APTCYG_INSTALLED="True"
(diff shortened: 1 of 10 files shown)


sophia-guo:
cygwin and jq are being available on adoptium jenkins. I think it’s still good to check and install if they aren't exist for local run.

sophia-guo (approved):
LGTM

karianna (approved):

steelhead31:
Local tests work ok, so once code freeze is over, this is good to merge.

tooling/reproducible/windows_repro_build_compare.sh
```diff
@@ -138,37 +139,43 @@ Check_Parameters() {
...
-      chmod +x apt-cyg
-      mv apt-cyg /usr/local/bin
-    else
-      echo "Checksum Is Not OK - Exiting"
-      exit 1
-    fi
+    echo "JQ Is Already Installed"
   fi
 
   # Install JQ Where Not Already Installed
```
sophia-guo, line 177: This line can be removed.

steelhead31: Done!, just testing...

PR #3722, opened 2024-03-21
Set build_full_name correctly when user build root directory specified

andrew-m-leonard:
configureBuild.sh specifies extra make args for the autoconf configuration for s390x, it unecessarily specifies
the CONF= parameter, which is not necessary as there is only one openjdk autoconf configuration, and
was incorrect when a user build root directory is specified.

build_full_name is the OpenJDK autoconf configuration name which is set to:
- When building within the OpenJDK src root:
```
$openjdk src root> bash ./configure ...    => build output to => build/linux-x86_64-server-release
```
- When building within a user specified build output directory:
```
$/user_directory> bash /path_to_openjdk_src/configure ...  => build output to => /user_directory
```

jdk8u s390x test build: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-linux-s390x-temurin/296/
jdk11u s390x test build: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-s390x-temurin/329/
jdk21u s390x test build: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-linux-s390x-temurin/105/


steelhead31 (approved):
LGTM.

sxa (approved):
This is a follow on to https://github.com/adoptium/temurin-build/pull/3696 which I haven't fully reviewed.

My initial concern was that this could change th4e defaults to squash the `variant` part of the build output directory in some cases, but I can't see that it's being included anyway so I don't think this will have any side effects. On that basis I'm approving.

Since I looked for the source, the reason for these additions being in there appear to have been lost in the depths of time as it was in a [giant rewrite of the codebase](https://github.com/adoptium/temurin-build/pull/393/files#diff-070a0531721fb82ff7f28fa30a025aa94f9de4db08619d0ee22c3325093cd4ec and the switch for JDK versions >=12 was added in [this commit](https://github.com/adoptium/temurin-build/commit/70b49dea2158ee6f1fb15912e443b5d214c317ab))

andrew-m-leonard:
> This is a follow on to #3696 which I haven't fully reviewed.
> 
> My initial concern was that this could change th4e defaults to squash the `variant` part of the build output directory in some cases, but I can't see that it's being included anyway so I don't think this will have any side effects. On that basis I'm approving although I'd say "cautiously" sine I don't know why the `CONF=` was added in the first place.
> 
> Since I looked for the source, the reason for these additions being in there appear to have been lost in the depths of time as it was in a [giant rewrite of the codebase](https://github.com/adoptium/temurin-build/pull/393/files#diff-070a0531721fb82ff7f28fa30a025aa94f9de4db08619d0ee22c3325093cd4ec and the switch for JDK versions >=12 was added in [this commit](https://github.com/adoptium/temurin-build/commit/70b49dea2158ee6f1fb15912e443b5d214c317ab))

@sxa yep, I search back through git blame history to no avail... but it was only for s390x, I am wondering if it was added when maybe building both a "zero" and a "server" variant within the same build...?
We could keep the logic and switch from either CONF=X or SPEC=<userbuilddirectory>, but I really couldn't see why we need it....

karianna:
I don't have recollection of this variable, so let's merge for now.

PR #3725, opened 2024-03-23
Added feature to download boot jdk on demand

judovana:
currenlty only dne for linux, shoudl be easy to extend for windows, not sure what whith other platfroms.

Is done via magic parameter of `download` for -J: -J download


judovana:
The windows label is wrong. I can not change t on my own. Please fix it to (linux) if possible.
I think I will add windows support once I get some cygwn machine,  but it will take time.

judovana:
I guess on windows, I will extract https://github.com/adoptium/temurin-build/blob/95385f4abc6cfe01e574f55cbde5a4c925bc031a/build-farm/platform-specific-configurations/windows.sh#L47-L99 to the `downloader.sh` as `downloadWindowsBootJDK` aybe with some adjustments... thoughts?

judovana:
ps: I had noticed the bash-lint issues, mos tof them are valid. Will elaborate as the review progresses here.

judovana:
Now works for windows and linuxes. I do not have  more paltforms to try on. Please consider this.

sbin/prepareWorkspace.sh
```diff
@@ -577,15 +577,33 @@ prepareMozillaCacerts() {
     fi
 }
 
-# Download all of the dependencies for OpenJDK (Alsa, FreeType etc.)
+downloadBootJdkIfNeeded () {
+  if [[ "${BUILD_CONFIG[JDK_BOOT_DIR]}" == "download" ]]; then
+    # the bootDir is used by refactored downloaders.sh methods; it would benice to change in future
```
karianna, line 582:
```suggestion
    # The bootDir is used by downloaders.sh methods; it would be nice to change in future
```

karianna: Change to what?

makejdk-any-platform.1
```diff
@@ -117,7 +117,9 @@ ignore the existing docker container if you have one already.
 specify the JDK boot dir.
 For reference, OpenJDK needs the previous version of a JDK in order to build
 itself. You should select the path to a JDK install that is N-1 versions below
-the one you are trying to build.
+the one you are trying to build. On some platforms, you can use magic keyword
+of `download`, and boot jdk will be downloaded for you in best attempt. It will
```
karianna, line 121: I think you need to update line 116 to show that `download` example?

build-farm/platform-specific-configurations/downloaders.sh
```diff
@@ -0,0 +1,147 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+####################################################################################
+# This file is gathering all download boot jdk functions, so they can b reused later
+# On long run, the methods - due theirs simialrity - should converge to one
+####################################################################################
+
+# this one should be swithing per (supported) oses in future
```
karianna, line 23:
```suggestion
# This function switches logic based on the (supported) os
```

build-farm/platform-specific-configurations/downloaders.sh
```diff
@@ -0,0 +1,147 @@
...
+function downloadBootJDK() {
```
karianna, line 24: Why are we not using this function now?

build-farm/platform-specific-configurations/downloaders.sh
```diff
@@ -0,0 +1,147 @@
...
+    echo "Downloading GA release of boot JDK version ${VER} failed."
+    # shellcheck disable=SC2034
+    releaseType="ea"
+    # shellcheck disable=SC2034
+    vendor="adoptium"
+    apiURL=$(eval echo ${apiUrlTemplate})
+    echo "Attempting to download EA release of boot JDK version ${VER} from ${apiURL}"
+    set +e
+    curl -L -o bootjdk.tar.gz "${apiURL}"
+    if ! grep "No releases match the request" bootjdk.tar.gz; then
```
karianna, line 73: seems like duplicated logic from above, could extract?

karianna (changes requested):

sbin/prepareWorkspace.sh
judovana: `    # The bootDir is used by downloaders.sh methods; it would be nice to change it in future so the download method is unified`

Sounds better?

makejdk-any-platform.1
judovana: Thanx. Adjusted by `;` not sure how is the expected man page format for such in-list exmaple

build-farm/platform-specific-configurations/downloaders.sh
judovana: I see it used  on https://github.com/adoptium/temurin-build/pull/3725/files#diff-1968af21c421fa9d7903dcd8950e640bc8ae257b7d7c5768c0fdd64d95fe3d7aR589

I had extracted two `downloadBootJDK` from `linux.sh` and `windows.sh` as `downloadLinuxBootJDK` and downloadWindowsBootJDK`. 
The original `linux.sh` is now calling `downloadLinuxBootJDK` and original `windows.sh` is calling `downloadWindowsBootJDK`

The "new" `downloadBootJDK` (maybe I shoud have change the name) is now called in `prepareWorkspace.sh` in `downloadBootJdkIfNeeded` 

`downloadBootJdkIfNeeded ` is decideing whether download should happenIf so, it is callin `downloader.sh`s downloadBootJDK` and that is decideding what platfrom to use.

build-farm/platform-specific-configurations/downloaders.sh
judovana: It is utterly duplicated. Most of body of the two extracted methods is very similar. I had really only re-factored linux.sh and windows.sh  boot jdk download to downloaders.sh.

It would be worthy to revisit those methods as separate PR, to unify as much logic as necessary.  AFAICJ all download methods should be processed like this, before the single replacement is crated.

If you wish, I can do some minor fixes (As the one you suggest here) but that may be contra productive in future.

judovana:
Thanx a lot for eyball!

karianna:
@judovana Needs a rebase

judovana:
> @judovana Needs a rebase

yup. Thats exactrlythe reasonwhy I do not want to change the extracted download code in this PR.

karianna (approved):
I'm broadly OK with the implementation.  Requesting other reviewers

sxa:
> The windows label is wrong. I can not change t on my own. Please fix it to (linux) if possible. I think I will add windows support once I get some cygwn machine, but it will take time.

Tried but I don't trust the labeller not to add windows back in again ;-)

sbin/prepareWorkspace.sh
```diff
@@ -623,15 +623,32 @@ downloadDevkit() {
...
 
-# Download all of the dependencies for OpenJDK (Alsa, FreeType etc.)
+downloadBootJdkIfNeeded () {
+  if [[ "${BUILD_CONFIG[JDK_BOOT_DIR]}" == "download" ]]; then
+    # The bootDir is used by downloaders.sh methods; it would be nice to change it in future so the download method is unified
+    bootDir="${BUILD_CONFIG[WORKSPACE_DIR]}/downloaded-boot-jdk-${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}"
+    if  [ -e "$bootDir" ] ; then
+      echo "Reusing $bootDir"
+    else
+      source "$SCRIPT_DIR/../build-farm/platform-specific-configurations/downloaders.sh"
```
sxa, line 633: Architecturally I'm not a fan of this. The stuff in the root of this repository and `sbin` is supposed to be fully independent. The bits in `build-farm` is for the adoptium build processes and configures the environment for our setup and wraps around the other stuff. Calling out from `sbin` to `platform-specific-configurations` is therefore not ideal.

I like the idea of the option to specify a boot JDK to download, but I think this implementation will need some refinement first.

judovana: It should be quite strightforward to move downloaders.sh next to common.sh and refactor them so the `$bootDir` is not missused.

I will elaborate on the `bootDir` as I really do not like it. Do you have any prefferences on `downloaders.sh` location?

sxa: I'm undecided but marginally leaning towards `sbin` or `sbin/common`. WDYT @andrew-m-leonard ?

build-farm/platform-specific-configurations/downloaders.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
+# shellcheck disable=SC1091
+
+################################################################################
```
gdams, line 4: replace with the correct header:

```output
# ********************************************************************************
# Copyright (c) 2018 Contributors to the Eclipse Foundation
#
# See the NOTICE file(s) with this work for additional
# information regarding copyright ownership.
#
# This program and the accompanying materials are made
# available under the terms of the Apache Software License 2.0
# which is available at https://www.apache.org/licenses/LICENSE-2.0.
#
# SPDX-License-Identifier: Apache-2.0
# ********************************************************************************
```

judovana: ty!

sbin/prepareWorkspace.sh
judovana: > Architecturally I'm not a fan of this. The stuff in the root of this repository and `sbin` is supposed to be fully independent. The bits in `build-farm` is for the adoptium build processes and configures the environment for our 

Actually there is an misunderstanding, caused by `bootJdk` variable name clash. I will rework it so it is no longer there.

The  `  if  [ -e "$bootDir" ] ; then` is independent  local check, if boot jdk was already downloaded, so it is not downloaded again.

judovana: done https://github.com/adoptium/temurin-build/pull/3725/commits/00b599c20d3fee1e035a3f23b1f497b51eb2ebcb

judovana:
ping please?

andrew-m-leonard (approved):
looks good

judovana:
@andrew-m-leonard  TY!

@sxa its upt o you now

> Architecturally I'm not a fan of this. The stuff in the root of this repository and `sbin` is supposed to be fully independent. The bits in `build-farm` is for the adoptium build processes and configures the environment for our setup and wraps around the other stuff. Calling out from `sbin` to `platform-specific-configurations` is therefore not ideal.
> 
> I like the idea of the option to specify a boot JDK to download, but I think this implementation will need some refinement first.

I had resoled the major issue you had (the global variable) but the shared file remained in `build-farm/platform-specific-configurations/downloaders.sh` . I left it here, as all the code is originally from  `build-farm/platform-specific-configurations/` files.  Maybe I would vote for leaving it there, with future work on other platforms, and move it only once it have been finished. But that can be very long time, thus the `build-farm/platform-specific-configurations/downloaders.sh` will be breaking the ' The stuff in the root of this repository and `sbin` is supposed to be fully independent." which I like, and do not want to be cause of its fade away.

sbin/prepareWorkspace.sh
andrew-m-leonard: > I'm undecided but marginally leaning towards `sbin` or `sbin/common`. WDYT @andrew-m-leonard ?

@sxa yes I agree with you, looking at this again I feel uncomfortable that "sbin" code is calling out to "build-farm", as this could lead code-creep in the future architecturally in the wrong direction.
@judovana We should move the downloaders.sh to sbin/common, which is the common place that both sbin and build-farm pull "common" things from please.

andrew-m-leonard (changes requested):

sbin/prepareWorkspace.sh
judovana: Sure, Thanx  a lot, on it!

judovana: done.  Tested on win x86, el8 aarch64 and f38 x64 aganst jdk8,17 and 21.

judovana:
done. Tested on win x86, el8 aarch64 and f38 x64 aganst jdk8,17 and 21.

sbin/prepareWorkspace.sh
```diff
@@ -652,15 +652,31 @@ downloadDevkit() {
...
+}
+
+# Download all of the dependencies for OpenJDK (Alsa, FreeType, boot-jdk etc.)
 downloadingRequiredDependencies() {
   if [[ "${BUILD_CONFIG[CLEAN_LIBS]}" == "true" ]]; then
     rm -rf "${BUILD_CONFIG[WORKSPACE_DIR]}/libs/freetype" || true
-
     rm -rf "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/installedalsa" || true
     rm -rf "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/installedfreetype" || true
+    rm -rf "${BUILD_CONFIG[WORKSPACE_DIR]}/downloaded-boot-jdk-${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" || true
```
sxa, line 675: I'm inclined to suggest leaving this in place as a cache for anyone using our scripts since the JDK is quite a big download, but I'm also ok with this.

I was going to comment on the fact that `|| true` is unnecessary with `rm -f` but sine that's also in the previous three lines I'll ignore it ;-)

judovana: I agree, and I'm keeping it.  It will be removed only if ` if [[ "${BUILD_CONFIG[CLEAN_LIBS]}" == "true" ]]; then`  
oook? Yah, I was wondering  on the double passig rf || true, but left in the style!

sxa (approved):
LGTM and I've very happy to see all of this logic moved out of the `platform-specific-configurations` directory :-)

I've done a few tests on Linux but not Windows.

In the future we should look at doing the same for AIX, Solaris and macos so I suggest creating an issue for that since I'm not going to mandate that we do it now.

Also, one other question for anyone else reading this - should we default to `download` if we can't find a JDK in any other location?

judovana:
> LGTM and I've very happy to see all of this logic moved out of the `platform-specific-configurations` directory :-)
> 
> I've done a few tests on Linux but not Windows.
> 
> In the future we should look at doing the same for AIX, Solaris and macos so I suggest creating an issue for that since I'm not going to mandate that we do it now.
> 
> Also, one other question for anyone else reading this - should we default to `download` if we can't find a JDK in any other location?


that would be NACK by me.  It is not small dependence, and one should have boot jdk around. The -J download is highly  comfortable option

andrew-m-leonard (approved):
looks good

PR #3727, opened 2024-03-25
Removing also dash "-" and underscore "_" from repository name

judovana:
this is partially bound to https://github.com/adoptium/temurin-build/pull/3724, where dash and underscore are one of major delimiters in directory names jdk11u-my_feature

May, or may have not sense without the 3724

### DIFF
--- a/sbin/common/common.sh
+++ b/sbin/common/common.sh
@@ -33,7 +33,7 @@ function setOpenJdkVersion() {
     BUILD_CONFIG[OPENJDK_CORE_VERSION]=${BUILD_CONFIG[OPENJDK_FOREST_NAME]%?}
   fi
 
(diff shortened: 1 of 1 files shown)


judovana:
Please let me know, if this have sense standalone, or only as part of  https://github.com/adoptium/temurin-build/pull/3724

karianna:
Let's make it part of 3724, easier to track the discussion there.

judovana:
Sure! ty!

judovana:
Just for record. This PR was very wrong. The final argumenr, which is santised on this line, is strictly `jdk[XY[u]]`  nothing else. In theory,  the  sed shold not be enhanced, but make more strict.  and remove jsut ` [jdku]` letters

I will do a PR, which will be asserting that the mandatory  argument of jdkXY is correct.

judovana:
Surpassed by https://github.com/adoptium/temurin-build/pull/3728

Issue #3730, opened 2024-03-26
comparable_patch.sh should skip debuginfo CRC differences


fitzsim:
comparable_patch.sh should not report .debuginfo CRC differences in .gnu_debuglink sections.  I will send a pull request that implements this feature.

andrew-m-leonard:
Fixed by https://github.com/adoptium/temurin-build/pull/3736

PR #3732, opened 2024-03-26
comparable_patch.sh: Neutralise debuginfo CRC differences (#3730)

fitzsim:
This patch removes .gnu_debuglink sections to eliminate CRC differences that occur when .debuginfo files differ, for example when they are built in the presence of different system headers.


fitzsim:
Closing in favour of https://github.com/adoptium/temurin-build/pull/3736.

PR #3734, opened 2024-03-27
Windows Reproducible Script - Fix Path Length

steelhead31:
Amend the working directory to be a significantly shorter path, and copy the output diff file to a more usable location.

### DIFF
--- a/tooling/reproducible/windows_repro_build_compare.sh
+++ b/tooling/reproducible/windows_repro_build_compare.sh
@@ -137,38 +137,11 @@ Check_Parameters() {
   fi
 }
 
-Install_PreReqs() {
-  # Install Cygwin Package Manager If Not Present
-  echo "Checking If Apt-Cyg Is Already Installed"
-  if [ -f /usr/local/bin/apt-cyg ]; then
-    echo "Skipping apt-cyg Install"
-    APTCYG_INSTALLED="True"
-  else
-    echo "Installing apt-cyg"
-    APTCYG_INSTALLED="False"
-    wget -q -O "./apt-cyg" "https://raw.githubusercontent.com/transcode-open/apt-cyg/master/apt-cyg"
-    ACTSHASUM=$(sha256sum "apt-cyg" | awk '{print $1}')
-    EXPSHASUM="d020050e2cb56fec990f16fd10695e153afd064cb0839ba935247b5a9e4c29a0"
-
-    if [ "$ACTSHASUM" == "$EXPSHASUM" ]; then
-      chmod +x apt-cyg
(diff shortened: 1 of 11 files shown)


steelhead31:
See #3711

andrew-m-leonard (approved):

tooling/reproducible/windows_repro_build_compare.sh
```diff
@@ -39,8 +39,8 @@ TARBALL_URL="$2"
 # These Values Should Be Updated To Reflect The Build Environment
 # The Defaults Below Are Suitable For An Adoptium Windows Build Environment
 # Which Has Been Created Via The Ansible Infrastructure Playbooks
-CURR_DIR=$(pwd)
-WORK_DIR="$CURR_DIR/cmp$(date +%Y%m%d%H%M%S)"
+
+WORK_DIR="C:/workspace/comp-jdk-build/"
```
sophia-guo, line 43: Would this dir be available for dev's local environment?

steelhead31: Yes, the script creates it, if it doesn't exist...

steelhead31: Ah, but I can probably make this better...

steelhead31: Let me make a quick fix..

steelhead31: should be ok now.

sophia-guo:
Can you also add move or add the compare results (.diff ) to current dir as the result dir, which can be accessed by aqa-tests?

steelhead31:
> Can you also add move or add the compare results (.diff ) to current dir as the result dir, which can be accessed by aqa-tests?

It currently puts it in  WORK\_DIR as seen below,  I'll see if I can move it to the directory the script was launched from..   
  
`mv "$WORK_DIR/compare/repro_diff.out" "$WORK_DIR"`

steelhead31:
@sophia-guo it will now record the directory win\_repro.sh script and copy the repro\_diff.out file to the same location, ie..   
  
`ScriptPath="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"`  
`cp "$WORK_DIR/compare/repro_diff.out" "$ScriptPath"`  
 

Please let me know if that will work for you?

steelhead31:
Found another bug, the script wasnt returning from its call to the repro\_compare.sh script, so I've added a fix for that, which Im just testing.

sophia-guo (approved):

PR #3737, opened 2024-03-28
enabled build from source dir/tarball via  to --openjdk-source-location/-l

judovana:
Surpassing https://github.com/adoptium/temurin-build/pull/3724

* not using per-jdk-major build dirs, using src, or whatever user set up
* always cleaning src (or whatever was set up) dir
* do not remove config.status, but instead cleaning, and not copying/unpacking ./build
*  do no tmodify getFirstTagFromOpenJDKGitRepo, modified getOpenJdkVersion instead
* not enforcing the `--tag` but warning if it is not present. Reusing default tag if not set (which works pretty well)
* kept the relative paths, as it is what all the functions in build epxects. Will change to absolute if it will be insisted on
*   It do not yet support build in container (although I would like to add)
  * should be jsut change in genrated docker file  from clone to cp in?
* works above both dirs and tarballs

Wdyt?


sbin/build.sh
```diff
@@ -1802,29 +1820,28 @@ getFirstTagFromOpenJDKGitRepo() {
     TAG_SEARCH="aarch64-shenandoah-jdk8u*-b*"
   fi
 
-  # If openj9 and the closed/openjdk-tag.gmk file exists which specifies what level the openj9 jdk code is based upon,
-  # read OPENJDK_TAG value from that file.
-  local openj9_openjdk_tag_file="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/closed/openjdk-tag.gmk"
-  if [[ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_OPENJ9}" ]] && [[ -f "${openj9_openjdk_tag_file}" ]]; then
-    firstMatchingNameFromRepo=$(grep OPENJDK_TAG ${openj9_openjdk_tag_file} | awk 'BEGIN {FS = "[ :=]+"} {print $2}')
+  if [ "${BUILD_CONFIG[OPENJDK_FOREST_DIR]}" == "true" ]; then
```
andrew-m-leonard, line 1823: don't need this, you've already checked and exit'd further up ^^

sbin/common/config_init.sh
```diff
@@ -410,6 +415,23 @@ function setBranch() {
   BUILD_CONFIG[BRANCH]=${BUILD_CONFIG[BRANCH]:-$branch}
 }
 
+# Set the local dir if used
+function setLocalDir() {
```
andrew-m-leonard, line 419: I'd call this setOpenjdkSourceDir() ?

sbin/common/config_init.sh
```diff
@@ -83,6 +83,8 @@ OPENJDK_BUILD_NUMBER
 OPENJDK_CORE_VERSION
 OPENJDK_FEATURE_NUMBER
 OPENJDK_FOREST_NAME
+OPENJDK_FOREST_DIR
+OPENJDK_FOREST_DIR_ABSPATH
```
andrew-m-leonard, line 87: Do we need both these? Just have OPENJDK_FOREST_DIR the defaults to "", to mean not specified

sbin/common/config_init.sh
```diff
@@ -410,6 +415,23 @@ function setBranch() {
...
 
+# Set the local dir if used
+function setLocalDir() {
+  if [ ! -e "${1}" ] ; then
+    echo "you have specified -o/--openjdk-source, but '${1}' do not exists"
+    exit 1
+  fi
+  BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]=$(readlink -f "$1");
+  if [ ! -e "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}" ] ; then
+    echo "you have specified -o/--openjdk-source, but '${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}' do not exists"
```
andrew-m-leonard, line 426: "does not exist"

sbin/prepareWorkspace.sh
```diff
@@ -37,14 +36,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
 ALSA_LIB_GPGKEYID=${ALSA_LIB_GPGKEYID:-A6E59C91}
 FREETYPE_FONT_SHARED_OBJECT_FILENAME="libfreetype.so*"
 
+
+copyFromDir() {
+  echo "Copyng existing ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} to `pwd`/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to be built"
```
andrew-m-leonard, line 41: "Copying OpenJDK source from "

sbin/prepareWorkspace.sh
```diff
@@ -37,14 +36,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+  # we really do not want to use .git for dirs, as we expect user have them set up, ignoring them
+  local files=$(find ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} -maxdepth 1 -mindepth 1 | grep -v -e "/workspace$" -e "/build$" -e "/.git" -e -"/build/")
+  cp -rf $files "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/"
+}
+
+unpackFromArchive() {
+  echo "Extracting source tarball ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to be built"
```
andrew-m-leonard, line 48: "Extracting OpenJDK source tarball "

sbin/prepareWorkspace.sh
```diff
@@ -37,14 +36,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
 ALSA_LIB_GPGKEYID=${ALSA_LIB_GPGKEYID:-A6E59C91}
 FREETYPE_FONT_SHARED_OBJECT_FILENAME="libfreetype.so*"
 
+
+copyFromDir() {
+  echo "Copyng existing ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} to `pwd`/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to be built"
+  # we really do not want to use .git for dirs, as we expect user have them set up, ignoring them
+  local files=$(find ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} -maxdepth 1 -mindepth 1 | grep -v -e "/workspace$" -e "/build$" -e "/.git" -e -"/build/")
```
andrew-m-leonard, line 43: Not sure about the ignores for workspace, build, .git, ...., this list could grow or become wrong...
I think we should assume the source tarball is exactly that, if someone puts something in their that is "wrong" then we are not really judging if it is right or wrong....?

sbin/prepareWorkspace.sh
```diff
@@ -37,14 +36,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+}
+
+unpackFromArchive() {
+  echo "Extracting source tarball ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to be built"
+  # if the tarball contains .git files, they should be ignored later withut shame
+  # todo, support also zips?
+  pushd "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+    local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}" | wc -l)
+    if [ "$topLevelItems" -eq "1" ] ; then
+      echo "Source tarball contans exaclty one directory, using"
```
andrew-m-leonard, line 54: "contains"

sbin/prepareWorkspace.sh
```diff
@@ -37,14 +36,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+      tar --strip-components 1 -xf "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}"
+    else
+      echo "Source tarball do not contains top level directory, using"
```
andrew-m-leonard, line 57: "does not contain a"

sbin/prepareWorkspace.sh
```diff
@@ -37,14 +36,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+      tar --strip-components 0 -xf "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}"
+    fi
+    rm -rf "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}/build"
```
andrew-m-leonard, line 60: We shouldn't be touching the source location...

sbin/common/config_init.sh
judovana: Sure. ty!

sbin/common/config_init.sh
judovana: No, you are right, and one should be enough. To split the flag and value seemed  to be much more readable to me. Do you insists?

sbin/common/config_init.sh
judovana: ty. fixed!

sbin/prepareWorkspace.sh
judovana: sure!

sbin/prepareWorkspace.sh
judovana: Gosh. that is nasty bug. Thanx for catching that. It was supposed to happen in `BUILD_CONFIG[OPENJDK_SOURCE_DIR]`. Will fix

judovana:
Thanx a lot for reading it out. I had fixed all what coudl be fixed without asking.  Two topic remained above.

sbin/prepareWorkspace.sh
judovana: Teh source tarball is most likely ok. But in case of local directory? That will always contain at least .git. And in many cases also the workspace and build:(

karianna:
@judovana Some linter issues

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+
+copyFromDir() {
+  echo "Copying OpenJDK source from  ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to be built"
+  # we really do not want to use .git for dirs, as we expect user have them set up, ignoring them
+  local files=$(find "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}" -maxdepth 1 -mindepth 1 | grep -v -e "/workspace$" -e "/build$" -e "/.git" -e -"/build/")
+  cp -rf "$files" "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/"
+}
+
+unpackFromArchive() {
+  echo "Extracting OpenJDK source tarbal ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to be built"
```
karianna, line 46:
```suggestion
  echo "Extracting OpenJDK source tarball ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to build the binary"
```

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+  # if the tarball contains .git files, they should be ignored later withut shame
```
karianna, line 47:
```suggestion
  # If the tarball contains .git files, they should be ignored later
```

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+  # todo, support also zips?
+  pushd "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+    local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}" | wc -l)
+    if [ "$topLevelItems" -eq "1" ] ; then
+      echo "Source tarball contains exaclty one directory, using"
```
karianna, line 52:
```suggestion
      echo "Source tarball contains exactly one directory"
```

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+      tar --strip-components 1 -xf "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}"
+    else
+      echo "Source tarball does not contain a top level directory, using"
```
karianna, line 55:
```suggestion
      echo "Source tarball does not contain a top level directory"
```

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+      tar --strip-components 0 -xf "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}"
+    fi
+    rm -rf "build"
+  popd
```
karianna, line 59:
```suggestion
    popd
```

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+}
+
+copyFromDirOrUnpackFromArchive() {
+  echo "Removing copy of openjdk source repository of $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} in 10 seconds..."
```
karianna, line 63:
```suggestion
  echo "Cleaning the copy of OpenJDK source repository from $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} and replacing with a fresh copy in 10 seconds..."
```

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+  sleep 10
```
karianna, line 64: Why have this delay?

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+  rm -rf "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+  mkdir  "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+  # note, that we are not persisting build directory
```
karianna, line 67:
```suggestion
  # Note that we are not persisting the build directory
```

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+  if [ -d "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}" ] ; then
+    copyFromDir
+  elif [ -f "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}" ] ; then
+    unpackFromArchive
+  else
+    echo "Not directory nor file ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}"
```
karianna, line 73:
```suggestion
    echo "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} is not a directory or a file "
```

sbin/prepareWorkspace.sh
```diff
@@ -104,8 +146,9 @@ checkoutAndCloneOpenJDKGitRepo() {
     fi
   fi
 
-  git clean -ffdx
-
+  if [ ! "${BUILD_CONFIG[OPENJDK_FOREST_DIR]}" == "true" ]; then
```
karianna, line 149:
```suggestion
  if [ "${BUILD_CONFIG[OPENJDK_FOREST_DIR]}" == "false" ]; then
```

sbin/prepareWorkspace.sh
```diff
@@ -116,6 +159,17 @@ checkoutAndCloneOpenJDKGitRepo() {
 # Checkout the required code to build from the given cached git repo
 # Set checkoutRc to result so we can retry
 checkoutRequiredCodeToBuild() {
+
+  if [ "${BUILD_CONFIG[OPENJDK_FOREST_DIR]}" == "true" ]; then
+    echo "skipping checkoutRequiredCodeToBuild - local directory under processing:"
```
karianna, line 164:
```suggestion
    echo "Skipping checkoutRequiredCodeToBuild - local directory under processing:"
```

karianna (changes requested):

makejdk-any-platform.1
```diff
@@ -128,6 +132,11 @@ specify any custom user make arguments.
 .BR \-\-no\-adopt\-patches
 This indicates to the build script that the build does not have Temurin's additional patches applied.
 .TP
+.BR  \-o ", " \-\-openjdk\-source " " \fI<dir/archive>\fR
+If set, must point to directory or archive with sources.
+Those will be used instead of cloning without any .git magic.
```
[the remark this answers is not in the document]
judovana: It is not jsut clonning. There is git reset or deduction of tags, and a bit of more. The git is called aprox 90times in build scripts.
So what about: `This local source code will be used instead of cloning from a git repository and without any git tag or git reset` ?

makejdk-any-platform.1
```diff
@@ -128,6 +132,11 @@ specify any custom user make arguments.
...
+You have to set main argument though, and you should use \-\-tag too.
```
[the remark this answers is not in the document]
judovana: Without manual tag, an default tag,  which is not exactly descriptive - eg `jdk-21.0.0+0` - is used. It is most visible in eg `java -version`. You yourslef knows the best what are you bulding, and can set a proper descriptive tag of targeted version and a description and iteration of your build purpose.

sbin/prepareWorkspace.sh
judovana: The `git reset` have same delay: https://github.com/adoptium/temurin-build/blob/646a79f2668378151985bc9517392ad955555ea6/sbin/prepareWorkspace.sh#L68

It may be good idea to wait a bit, before the finished build is accidently lost.

sbin/prepareWorkspace.sh
judovana: proepr boolean is so missing inbash...

makejdk-any-platform.1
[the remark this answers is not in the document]
sxa: Maybe this as slightly better wording?
```suggestion
You have to set main argument though, and it is recommended that you add \-\-tag too.
```

sbin/prepareWorkspace.sh
```diff
@@ -35,14 +34,55 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+
+unpackFromArchive() {
+  echo "Extracting OpenJDK source tarbal ${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to be built"
+  # if the tarball contains .git files, they should be ignored later withut shame
+  # todo, support also zips?
+  pushd "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+    local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}" | wc -l)
+    if [ "$topLevelItems" -eq "1" ] ; then
+      echo "Source tarball contains exaclty one directory, using"
+      tar --strip-components 1 -xf "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}"
```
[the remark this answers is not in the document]
judovana: True. But it is already used for eg alsa: 
`  tar -xf alsa-lib.tar --strip-components=1 -C "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/installedalsa/"`

I will remove  the `--strip-components 0` to help a bit with this. 

Speaking of alsa - can the download of alsa be omitted and pointed to lcoal files via the `./makejdk-any-platform.sh` ? If so, it is missing in man.

sbin/build.sh
```diff
@@ -1777,10 +1780,25 @@ getLatestTagJDK11plus() {
   fi
 }
 
+createDefaultTag() {
```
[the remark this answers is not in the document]
judovana: It is jsut extracted from `getFirstTagFromOpenJDKGitRepo` so not that much good change :(  But the method is very dummy, and I bet it do not work for many repos (eg that shenandoah-jdk8u or similar)

sbin/common/config_init.sh
karianna: Agree with @sxa on the rename here

karianna (changes requested):
Some suggested commits to be merged and Stewart's rename suggestion

sbin/common/config_init.sh
judovana: done!

sbin/prepareWorkspace.sh
judovana: Hello! Not following this one unless you insists.   If you look carefully the block in pushd/popd is indented:
```
unpackFromArchive() {
  echo "Extracting OpenJDK source tarbal ${BUILD_CONFIG[OPENJDK_FOREST_SOURCE_ARCHIVE_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to be built"
  # If the tarball contains .git files, they should be ignored later
  # todo, support also zips?
  pushd "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
    local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_FOREST_SOURCE_ARCHIVE_ABSPATH]}" | wc -l)
    if [ "$topLevelItems" -eq "1" ] ; then
      echo "Source tarball contains exaclty one directory, using"
      tar --strip-components 1 -xf "${BUILD_CONFIG[OPENJDK_FOREST_SOURCE_ARCHIVE_ABSPATH]}"
    else
      echo "Source tarball does not contain a top level directory, using"
      tar -xf "${BUILD_CONFIG[OPENJDK_FOREST_SOURCE_ARCHIVE_ABSPATH]}"
    fi
    rm -rf "build"
  popd
}
```

I don't know how global this habit is,  but its usefulness proved over years

sbin/prepareWorkspace.sh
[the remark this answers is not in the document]
judovana: moved to https://github.com/adoptium/temurin-build/issues/3752

sbin/prepareWorkspace.sh
judovana: resolved manually 519c52897d8a9794a207c69e110fed2ce2768dff

sbin/common/config_init.sh
```diff
@@ -412,6 +417,23 @@ function setBranch() {
...
 
+# Set the local dir if used
+function setOpenjdkSourceDir() {
+  if [ ! -e "${1}" ] ; then
+    echo "you have specified -o/--openjdk-source, but '${1}' do not exists"
+    exit 1
+  fi
+  BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]=$(readlink -f "$1");
+  if [ ! -e "${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}" ] ; then
+    echo "you have specified -o/--openjdk-source, but '${BUILD_CONFIG[OPENJDK_FOREST_DIR_ABSPATH]}' does not exists"
```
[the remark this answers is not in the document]
judovana: resolved manually: 275acb889b5ed10f208cba810c027ff1b7439242

judovana:
Hello! mostly  resolved/followed in best intentions.

Sorry to missing those 3d old... I'm not sure how I coud overlook them. Thanx a lot!

judovana:
ping please?

sbin/build.sh
```diff
@@ -183,8 +183,9 @@ configureMacOSCodesignParameter() {
 # Get the OpenJDK update version and build version
 getOpenJDKUpdateAndBuildVersion() {
   cd "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}"
-
-  if [ -d "${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/.git" ]; then
+  if [ "${BUILD_CONFIG[OPENJDK_FOREST_SOURCE_ARCHIVE]}" == "true" ]; then
+    echo "Version: local dir; OPENJDK_BUILD_NUMBER set as  ${BUILD_CONFIG[OPENJDK_BUILD_NUMBER]}"
```
karianna, line 187:
```suggestion
    echo "Version: local dir; OPENJDK_BUILD_NUMBER set as ${BUILD_CONFIG[OPENJDK_BUILD_NUMBER]}"
```

sbin/common/config_init.sh
```diff
@@ -81,6 +81,8 @@ OPENJDK_BUILD_NUMBER
 OPENJDK_CORE_VERSION
 OPENJDK_FEATURE_NUMBER
 OPENJDK_FOREST_NAME
+OPENJDK_FOREST_SOURCE_ARCHIVE
```
karianna, line 84: Archive to me usually means .zip or .tar etc.  Is OPENJDK_FOREST_SOURCE or something OPENJDK_LOCAL_SOURCE a better name?

karianna (changes requested):

sbin/common/config_init.sh
judovana: It can be archive archive.   You can build local directory and tarball  (both with sources) in same way.
As i replied to @sxa https://github.com/adoptium/temurin-build/pull/3737#issuecomment-2034530869 the  proepr name aka `OPENJDK_LOCAL_SOURCE_DIR_OR_ARCHIVE` is terrible, but why not.
```
sh makejdk-any-platform.sh  -o ~/git/java-21-openjdk-portable/openjdk-21.0.2+13.tar.xz  jdk21u
```
and
```
 sh makejdk-any-platform.sh -o  ~/git/jdk21u/ jdk21u
```

both do same job.

judovana: actually, `OPENJDK_LOCAL_SOURCE` is generic enough. @sxa  will use that.

karianna (approved):
Broadly LGTM - over to @sxa

judovana:
> Broadly LGTM - over to @sxa

TY!

andrew-m-leonard:
macOS jdk11u CI failure unrelated

andrew-m-leonard (approved):
looks good

sbin/prepareWorkspace.sh
```diff
@@ -79,6 +122,8 @@ checkoutAndCloneOpenJDKGitRepo() {
       echo "If this is inside a docker you can purge the existing source by passing --clean-docker-build"
       exit 1
     fi
+  elif [ "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE]}" == "true" ]; then
+    copyFromDirOrUnpackFromArchive
```
sxa, line 126: It feels a little odd to be calling this from within `checkoutAndCloneOpenJDKGitRepo` - perhaps we could invoke it from the higher level configureWorkspace` instead as an alternative. I guess it depends how much of the extra logic in this function we'd need to duplicate or abstract out.

sxa (commented):
A few comments/suggestions/discussion points before approving but overall this LGTM and I have no major objections.
I'd normally be more strict on it but if you want to leave the GNU-isms in place I'd be ok in this situation since it's not functionality that the temurin pipelines will use, and the platforms that will be affected are, how should I put it, "niche" ;-)

makejdk-any-platform.1
```diff
@@ -137,6 +141,11 @@ specify any custom user make arguments.
 .BR \-\-no\-adopt\-patches
 This indicates to the build script that the build does not have Temurin's additional patches applied.
 .TP
+.BR  \-o ", " \-\-openjdk\-source " " \fI<dir/archive>\fR
```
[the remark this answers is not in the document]
judovana: Its not minor. Its quite  major think.  if -l would be used, than it would be first short hand, which is letter, which is not used in the long one. The shortcut is critical. Just image writing the long one more then once per day. You would need to do a wrapper for that.  So to make you happy,  I would go with  \-l ", " \-\-openjdk\-source-location "

How much do you want to chage this? Switch is api, Unused by pipelnes, but still api.

sbin/prepareWorkspace.sh
```diff
@@ -38,14 +37,58 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+
+unpackFromArchive() {
+  echo "Extracting OpenJDK source tarball ${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to build the binary"
+  # If the tarball contains .git files, they should be ignored later
+  # todo, support also zips?
+  pushd "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+    local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}" | wc -l)
+    if [ "$topLevelItems" -eq "1" ] ; then
+      echo "Source tarball contains exactly one directory"
+      tar --strip-components 1 -xf "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"
```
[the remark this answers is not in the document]
judovana: But what else to do about it?-( I saw --strip-components on many places over all the repos in adoptium.

sbin/prepareWorkspace.sh
```diff
@@ -38,14 +37,58 @@ ALSA_LIB_CHECKSUM=${ALSA_LIB_CHECKSUM:-5f2cd274b272cae0d0d111e8a9e363f0878332915
...
+  # shellcheck disable=SC2086
+  cp -rf $files "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/"
+}
+
+unpackFromArchive() {
+  echo "Extracting OpenJDK source tarball ${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to build the binary"
+  # If the tarball contains .git files, they should be ignored later
+  # todo, support also zips?
+  pushd "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+    local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}" | wc -l)
```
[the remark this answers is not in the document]
judovana: Taht one I can do at least non  fatal.

sbin/prepareWorkspace.sh
judovana: The indentation or popd itself? The indentation of insides of pushd/popd block is quite good habbit.

sbin/prepareWorkspace.sh
judovana: ok. let me try.

makejdk-any-platform.1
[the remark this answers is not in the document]
sxa: > How much do you want to chage this? Switch is api, Unused by pipelnes, but still api.

I'd say 20-30% - certainly not something I'd block on if there are strong opposing feelings.

sbin/prepareWorkspace.sh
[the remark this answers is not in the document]
sxa: I've stripped it out from a few places in our code the past - if it's something that only runs on Linux/Windows (cygwin) install then it's ok since it's a reasonably safe assumption they we'll be using GNU tarbut this could be used elsewhere. I've worked around it with something along the lines of this in the past:
```
mkdir tmp
cd tmp && tar --strip-components 1 -xf "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"
cd .. && mv tmp/*/* .
rmdir tmp
```

makejdk-any-platform.1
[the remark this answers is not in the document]
judovana: I agree wit -o being confused by "output. Will try to  repalce it by ` "-l ", " --openjdk-source-location"`

sbin/prepareWorkspace.sh
judovana: Oh I see.  There is
```
  cd "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}"

  # Check that we have a git repo, we assume that it is a repo that contains openjdk source
  if [ -d "${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/.git" ] && [ "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE]}" == "false" ]; then
```

So the change of control dir, which is pretty important for any future steps would need to be handled.
And the main if is still using the `[ "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE]}" == "false" ]` 

becasue of the change of the dir, I would rather not chnage the location of the call, unless yo really insists.

In that case it will go :
```
function configureWorkspace() {
  if [[ "${BUILD_CONFIG[ASSEMBLE_EXPLODED_IMAGE]}" != "true" ]]; then
    createWorkspace
    downloadingRequiredDependencies
    downloadDevkit
    relocateToTmpIfNeeded
+if [ "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE]}" == "true" ]; then
+    copyFromDirOrUnpackFromArchiveWrapper
+else 
    checkoutAndCloneOpenJDKGitRepo
+fi
    applyPatches
    if [ "${BUILD_CONFIG[CUSTOM_CACERTS]}" = "true" ] ; then
      prepareMozillaCacerts
    fi
  fi
  writeDockerImageSHA
}
```
where the  wrapper:
```
function   copyFromDirOrUnpackFromArchiveWrapper() {
   cd "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}"
   copyFromDirOrUnpackFromArchive
    cd "${BUILD_CONFIG[WORKSPACE_DIR]}"
}
```

IN addition the method of `checkoutAndCloneOpenJDKGitRepo` is doing a bit more before `  cd "${BUILD_CONFIG[WORKSPACE_DIR]}"`   Although  it seems those lines looks oook to omit (temurin exit, update openj9 and createSourceTagFile ) I'm not sure If I feel safe to omit the,

sbin/prepareWorkspace.sh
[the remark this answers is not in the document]
judovana: > I've stripped it out from a few places in our code the past - if it's something that only runs on Linux/Windows (cygwin) install then it's ok since it's a reasonably safe assumption they we'll be using GNU tarbut this could be used elsewhere. I've worked around it with something along the lines of this in the past:
> 
> ```
> mkdir tmp
> cd tmp && tar --strip-components 1 -xf "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"
> cd .. && mv tmp/*/* .
> rmdir tmp
> ```

You mean 
```
...
 cd tmp && tar  -xf "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"
...
 ```

thas nice hack.   It would deserve method in commons.

judovana:
@sxa  I had elaborated on all your comments I had found. Please let me know if you need anything more. Thanx a lto for suggestions!

sbin/prepareWorkspace.sh
sxa: Yeah agree that the safest option is to leave it as-is although putting it in the `configureWorkspace` function as per your example does look a lot nicer. But I won't insist on it :-)

sxa (approved):
On the basis that all of my comments have been addressed I'll approve this.
I'll let you make the decision on whether to do the extra change to move things into configureWorkspace - I suggest we hold of merging this until Monday (Putting something like this on a Friday always seems like a bad idea!)

judovana:
sure. TY!

PR #3743, opened 2024-04-03
Enable checking for /usr/local/devkit matching required Adoptium DevKit

andrew-m-leonard:
Check and verify /usr/local/devkit for the required matching Adoptium DevKit
If not found then it will defer to downloading from devkit-runtimes

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -603,6 +603,7 @@ downloadDevkit() {
         echo "Attempting to download the required DevKit instead"
       else
         # Found a matching DevKit
+        echo "Using matching DevKit from location ${USR_LOCAL_DEVKIT}"
         BUILD_CONFIG[ADOPTIUM_DEVKIT_LOCATION]="${USR_LOCAL_DEVKIT}"
       fi
     fi
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -589,6 +589,8 @@ downloadDevkit() {
 
     BUILD_CONFIG[ADOPTIUM_DEVKIT_LOCATION]=""
 
+    local devkit_target="${BUILD_CONFIG[OS_ARCHITECTURE]}-linux-gnu"
+
     local USR_LOCAL_DEVKIT="/usr/local/devkit"
     if [[ -d "${USR_LOCAL_DEVKIT}" ]]; then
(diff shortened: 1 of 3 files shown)


sbin/prepareWorkspace.sh
```diff
@@ -581,44 +581,69 @@ setupGpg() {
...
     local devkit_target="${BUILD_CONFIG[OS_ARCHITECTURE]}-linux-gnu"
-    local devkit="devkit-${BUILD_CONFIG[USE_ADOPTIUM_DEVKIT]}-${devkit_target}"
 
-    # Download tarball and GPG sig
-    echo "Downloading DevKit : ${devkitUrl}/${devkit}.tar.xz"
-    curl -L --fail --silent --show-error -o "${devkit_tar}" "${devkitUrl}/${devkit}.tar.xz"
-    curl -L --fail --silent --show-error -o "${devkit_tar}.sig" "${devkitUrl}/${devkit}.tar.xz.sig"
+    local USR_LOCAL_DEVKIT="/usr/local/devkit"
+    if [[ -d "${USR_LOCAL_DEVKIT}" ]]; then
+      local usrLocalDevkitInfo="${USR_LOCAL_DEVKIT}/devkit.info"
```
sxa, line 596: I'm currently set up to install into a subdirectory of /usr/local/devkit with the name of the particular devkit in order to allow different ones to be installed, so this will need to be customizable, or otherwise defined somewhere. I don't think we should be expecting a single devkit on the machien in that location, which looks like what this PR is expecting.
At the moment I have /usr/local/devkit/s390x-on-s390x for s390x and on the others it's a directory under /usr/local/devkit [matching the name of the release](https://github.com/adoptium/infrastructure/pull/3488/files#diff-fc137c7ee8ffb4d34a53f0f2484bcb64340a2b18fd885a1384bac9d06419eab8)

andrew-m-leonard: ah ok, i'll add an extra build arg for "local devkit directory"

andrew-m-leonard: I don't really want to have to add hard coded directory paths to the pipeline config, how about I search /usr/local/devkit/<devkit version tag>/

andrew-m-leonard: I have updated local USR_LOCAL_DEVKIT="/usr/local/devkit" to include the DevKit release tag

sxa (approved):
LGTM

smlambert (approved):
👍  thanks @andrew-m-leonard !

PR #3744, opened 2024-04-03
Using proper laud countdown before cleaning of repository

judovana:
There is couple of
```
echo  "doing osmething terrible in Xs"
sleep X
```

here in log only the `doing osmething terrible in Xs` appear. 

I think it may be better to print out proper dynamic countdown instead of just dark screen. WDYT?

I had put it to commn.sh, but some other place may be better. Also I had used it as example only on two places in prepareWorkspace, where there definitely is more applicable places over scripts.

WDYT?

### DIFF
--- a/sbin/common/common.sh
+++ b/sbin/common/common.sh
@@ -250,6 +250,6 @@ function verboseSleep() {
   else
     local i="${1}"
   fi
-  while [ $i -gt 0 ] ; do echo -n " $i " && sleep 1 && i=$(($i-1)) ; done && echo " $i"
+  while [ "$i" -gt 0 ] ; do echo -n " $i " && sleep 1s && i=$((i-1)) ; done && echo " $i"
 }
 
--- a/sbin/common/common.sh
+++ b/sbin/common/common.sh
@@ -250,6 +250,6 @@ function verboseSleep() {
   else
     local i="${1}"
   fi
-  while [ $i -gt 0 ] ; do echo -n " $i " && sleep 1s && i=$(($i-1)) ; done && echo " $i"
+  while [ $i -gt 0 ] ; do echo -n " $i " && sleep 1 && i=$(($i-1)) ; done && echo " $i"
 }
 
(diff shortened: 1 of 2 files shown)


sbin/common/common.sh
```diff
@@ -243,3 +243,13 @@ function isFromJdk21LTS() {
...
 }
 
+# waits N second (10 by default), printing countdown every second.
+function verboseSleep() {
+  if [[ -z "${1}" ]] ; then
+    local i=10
+  else
+    local i="${1}"
+  fi
+  while [ $i -gt 0 ] ; do echo -n " $i " && sleep 1s && i=$(($i-1)) ; done && echo " $i"
```
sxa, line 253: `sleep 1s` won't work on Solaris so remove the `s`.
```suggestion
  while [ $i -gt 0 ] ; do echo -n " $i " && sleep 1 && i=$(($i-1)) ; done && echo " $i"
```

sxa (commented):
My initial reaction is that this is someone unneeded but I won't block it ;-)

sbin/common/common.sh
judovana: I actually took the `s` from one of temurin-build script, I was sure that defaul it is second for eternity... Thanx for confirming it :)

sbin/common/common.sh
```diff
@@ -243,3 +243,13 @@ function isFromJdk21LTS() {
   [[ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -ge 21 ]] && [[ $(((BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]-21) % 4)) == 0 ]]
 }
 
+# waits N second (10 by default), printing countdown every second.
```
karianna, line 246:
```suggestion
# Waits N seconds (10 by default), printing a countdown every second.
```

karianna (changes requested):

karianna (approved):

sbin/common/common.sh
sxa: Interesting - I might stick a PR in to change it everywhere else then ;-)

andrew-m-leonard (approved):

judovana:
ok. two approvers done. What now? Can anybody merge please?

judovana:
ty!

PR #3751, opened 2024-04-05
Ensure LD_LIBRARY_PATH set when building using a linux devkit

andrew-m-leonard:
When building using a linux devkit, the LD_LIBRARY_PATH needs setting prior to the build to use the devkit sysroot lib64:lib

### DIFF
--- a/build-farm/platform-specific-configurations/linux.sh
+++ b/build-farm/platform-specific-configurations/linux.sh
@@ -329,8 +329,16 @@ if [ $executedJavaVersion -ne 0 ]; then
     exit 1
 fi
 
-if [[ "${CONFIGURE_ARGS}" =~ .*"--with-devkit=".* ]]; then
-  echo "Using gcc from DevKit toolchain specified in configure args"
+devkit_regex="--with-devkit=([^ ]+)"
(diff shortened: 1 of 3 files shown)


sxa (approved):
I never liked the `REMATCH` syntax, but if it works I'm ok with it :-)
Do these need to be in `prepareWorkspace.sh` as well as `linux.sh`?
I'll approve regardless for this cycle but it would be good to avoid the dupliation in the future if possible.

andrew-m-leonard:
> I never liked the `REMATCH` syntax, but if it works I'm ok with it :-) Do these need to be in `prepareWorkspace.sh` as well as `linux.sh`? I'll approve regardless for this cycle but it would be good to avoid the dupliation in the future if possible.

I could do that better yes

andrew-m-leonard:
> I never liked the `REMATCH` syntax, but if it works I'm ok with it :-) Do these need to be in `prepareWorkspace.sh` as well as `linux.sh`? I'll approve regardless for this cycle but it would be good to avoid the dupliation in the future if possible.

yep, i've resolved it into one place

smlambert (approved):
LGTM

Issue #3752, opened 2024-04-06
downloaded alsa is not used,local files are used via  ./makejdk-any-platform.sh 


judovana:
alsa - can the download of alsa be omitted and pointed to lcoal files via the ./makejdk-any-platform.sh ? If so, it is missing in man.

judovana:
ok. I t was investiageted that alsa indeed cna not be currenlty set on cmdline. Will draft a pr.
Pleae anyubody remove documentation label...

judovana:
quotation from senior development:
```
 I would suggest adding a --user-alsa-installation <dir> and either ensure it gets copied to
${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/installedalsa
around abouts here: https://github.com/adoptium/temurin-build/blob/d889cc1223d106cc3a43e1fe9fa76a9559bfc10d/sbin/prepareWorkspace.sh#L655
or create a new BUILD_CONFIG[ALSA_LOCATION] which defaults to
${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/installedalsa
and can be overridden by --user-alsa-installation <dir> (edited) 
```

judovana:
The https://github.com/adoptium/temurin-build/blob/d889cc1223d106cc3a43e1fe9fa76a9559bfc10d/sbin/build.sh#L482 seems to be jdk8u only

judovana:
```
checking for alsa/asoundlib.h... no
configure: error: Could not find alsa!
configure exiting with result code 1
```
withou alsa-devel installed, the configure fails, and dwnloded alsa is not used

PR #3756, opened 2024-04-09
making local alsa properly optional

judovana:
This pr is dong two things (afaik so connected that should be together)
 - first commit  https://github.com/judovana/temurin-build/commit/5f0b38ff5726554480cbf9f179575176f88ee016 is fixing issue, when alsa was donwloaded and unpacked, but it was not used. The usage was for jdk8 only. If it was intentional then I will rework this.
 - second commit https://github.com/judovana/temurin-build/commit/f2afb6e360aa8eaa5d0b82d6ce6d8db91f864985 is making the download of alsa optional. If not downloaded, then the system alsa will be used, or user will specify theirs.

The second commit + my old https://github.com/adoptium/temurin-build/pull/3737 should be enough for offline build.

related to https://github.com/adoptium/temurin-build/issues/3752


judovana:
I admit I had made an fundamental build change, but Still I was not expecting build failures. In addition, they do not look related.... /me confused  ..even windows...

judovana:
ping please?

sbin/common/config_init.sh
```diff
@@ -31,6 +31,7 @@
 CONFIG_PARAMS=(
 ADOPTIUM_DEVKIT_LOCATION
 ADOPT_PATCHES
+ALSA
```
karianna, line 34: Hmm, is this label to indicate that we skip Alsa or something else?

sbin/prepareWorkspace.sh
```diff
@@ -667,7 +668,13 @@ downloadingRequiredDependencies() {
     echo "Non-Linux-based environment detected, skipping download of dependency Alsa."
   else
     echo "Checking and downloading Alsa dependency because OSTYPE=\"${OSTYPE}\""
-    checkingAndDownloadingAlsa
+    if [[ "${BUILD_CONFIG[ALSA]}" == "true" ]]; then
+      checkingAndDownloadingAlsa
+    else
+      echo ""
+      echo "---> Skipping the process of checking and downloading the Alsa dependency, a pre-built version provided via -C/--configure-args <---"
```
karianna, line 675:
```suggestion
      echo "---> Skipping the process of checking and downloading the Alsa dependency, a pre-built version should be provided via -C/--configure-args <---"
```

sbin/build.sh
```diff
@@ -505,6 +504,14 @@ configureDebugParameters() {
   fi
 }
 
+configureAlsaLocation() {
+  if [[ ! "${CONFIGURE_ARGS}" =~ "--with-alsa" ]]; then
```
karianna, line 508: Could merge those two if clauses.

makejdk-any-platform.1
```diff
@@ -45,6 +45,12 @@ Build JDK (tip), defaults to https://github.com/adoptium/jdk
 
 .SH OPTIONS
 .TP
+.BR \-A ", " \-\-skip-alsa
+skip downloading of alsa automatically.
```
karianna, line 49:
```suggestion
Skip downloading of alsa automatically.
```

karianna (changes requested):

sbin/common/config_init.sh
judovana: I got inspired by `FREETYPE`  whic have the same name style and is doinf practically the same.

sbin/build.sh
judovana: Right. Could.  In  `configureFreetypeLocation`  and in `configureZlibLocation`  are similar structures, although in first they can not be merged, and in second it would not be nice.
It seemed better to me to keep it like this for readability. Bash conditions can be terrible.

I'm voting to keep it as it is. If you insist, will redo asap.

karianna (approved):

judovana:
tyvm!

judovana:
@andrew-m-leonard  @sxa  Thoughts?

andrew-m-leonard (approved):
looks good, thanks Jiri

judovana:
tyvm!

/integrate

PR #3772, opened 2024-04-24
dependency_data files do not need licenses

adamfarley:
...on the basis that they only contain version strings, jar names, and SHAs that are obtained from external sources.

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -46,5 +46,6 @@ header:
     - 'sbin/*.template'
     - '.github/linters/*'
     - 'cyclonedx-lib/getDependencies'
+    - 'cyclonedx-lib/dependency_data/**'
     - 'makejdk-any-platform.1'
     - 'serverTimestamp.properties'


tellison (approved):

tellison:
Do you need to remove some incorrectly added too?

karianna (approved):

karianna:
@adamfarley Feel free to merge once you've addressed Tim's Q.

adamfarley:
> Do you need to remove some incorrectly added too?

I don't think so, no. I'm adding just one entry, and it exclusively covers a directory that contains sha256 files and version string files.

I did review the rest of the list. Maybe you're suggesting that getDependencies should have a license at the top of that groovy file, and its entry be excluded here? I didn't think it did, as other groovy files in our repos don't have licenses ([example](https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/aqaTestPipeline.groovy)).

adamfarley:
Thanks @gdams

Issue #3807, opened 2024-05-10
Need to check version of ALSA being used for jdk21+ DevKit Linux builds?


andrew-m-leonard:
A reproducible build validation check of jdk-23+21 has highlighted a miss-match in alsa version:
```
Binary files openjdk_build/images/jdk/lib/libjsound.so and jdk-23+21/lib/libjsound.so differ
```

A few recent changes have highlighted a question over version of ALSA being used?
1. https://github.com/adoptium/temurin-build/pull/3756 recently made jdk-11+ builds specify --with-alsa=/workspace/installedalsa, whereas before it was finding the OS system ALSA
2. Change to DevKits means a "sysroot" is being used, which means the version of ALSA is found from the DevKit sysroot and not the OS ALSA

andrew-m-leonard:
PR https://github.com/adoptium/temurin-build/pull/3756 probably needs correcting to only apply to jdk-8 as it was previously

Using ALSA from the sysroot of the DevKit is probably the right thing for DevKit builds, but the SBOM ALSA version is probably wrong in that case...

judovana:
Hi. noted. Will look into it (unless it will block you in meantime).

judovana:
Do I understand correctly, that for jdk8, you need to persists current behaviour, where the alsa tarabll si downloaded and used, but for any newr  jdk, you need  it to not download alsa and use system one?

In that case, I would like to invert `BR \-A ", " \-\-skip-alsa` behavior so it will enable download of alsa on demand. Except jdk8, where it will behave like now. wdyt?

judovana:
To have the option different for various jdks sounds like a bad idea.  So to invert it? By default system alsa will be used and only if -A is specified, then one willbe downloaded.

For jdk11+ it is as per this issue necessary, and for jdk8 it shoudl work to.

andrew-m-leonard:
So https://github.com/adoptium/temurin-build/pull/3756 actually fixed an issue whereby we are downloading ALSA-1.1.6 and then storing in the SBOM that ALSA-1.1.6 was used, however for jdk-11+ we were actually using the "System" ALSA which is typically 1.1.8. Now with this PR we are using the downloaded ALSA-1.1.6 that is specified in the SBOM.

The problem now is when a "DevKit" is specified, which should aid reproducible builds, but the ALSA being used is now specified --with-alsa=ALSA-1.1.6 and not the DevKit sysroot alsa, thus a DevKit does not predicate a reproducible build.

[...]

PR #3809, opened 2024-05-14
Correctly obtain ALSA version for SBOM

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3807 issue(2.)

Correctly obtain the ALSA version from the built spec.gmk

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -952,7 +952,7 @@ generateSBoM() {
   addBootJDK
 
   # Add ALSA 3rd party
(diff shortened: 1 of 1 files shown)


judovana:
How does this behave with alsa download enabled?

gdams (approved):

andrew-m-leonard:
> How does this behave with alsa download enabled?

It retreives the built libsoundj path to the alsa used, which for download will be the path in --with-alsa=...

Issue #3746, opened 2024-04-04
sbom strace analysis is filtering some downloaded tooling


andrew-m-leonard:
From enabling DevKit, I can see the downloaded Devkit into workspace/build/devkit is not being analysed and thus the gcc 11.3.0 is not listed as a dependency.
Possibly other dependencies under workspace/build might be missing too, like ALSA? (although that might be listed manually elsewhere)

PR #3824, opened 2024-05-27
Update parsing TEMURIN_BUILD_ARGS

sophia-guo:
Fix https://github.com/adoptium/aqa-tests/issues/5345

### DIFF
--- a/tooling/reproducible/linux_repro_build_compare.sh
+++ b/tooling/reproducible/linux_repro_build_compare.sh
@@ -84,6 +84,114 @@ setEnvironment() {
   ls -ld "$CC" "$CXX" "/usr/lib/jvm/jdk-${BOOTJDK_VERSION}/bin/javac" || exit 1
 }
 
+containsElement() {
+  local e
+  for e in "${@:2}"; do
+    if [ "$e" == "$1" ]; then
+      return 0  # Match found
+    fi
+  done
+  return 1  # No match found
+}
+
+setBuildArgs() {
+  local CONFIG_ARGS=("--disable-warnings-as-errors" "--enable-dtrace" "--without-version-pre" "--without-version-opt" "--with-version-opt")
+  local NOTUSE_ARGS=("--configure-args")
+  export BOOTJDK_HOME="/usr/lib/jvm/jdk-${BOOTJDK_VERSION}"
+  echo "Parsing Make JDK Any Platform ARGS For Build"
+  # Split the string into an array of words
+  IFS=' ' read -ra words <<< "$TEMURIN_BUILD_ARGS"
+
+  # Add The Build Time Stamp In Case It Wasnt In The SBOM ARGS
+  words+=("--build-reproducible-date")
+  words+=("\"$BUILDSTAMP\"")
+
+  # Initialize variables
+  param=""
+  value=""
+  params=()
+
+  # Loop through the words
+  for word in "${words[@]}"; do
+    # Check if the word starts with '--'
+    if [[ $word == --* ]] || [[ $word == -b* ]]; then
+      # If a parameter already exists, store it in the params array
+      if [[ -n $param ]]; then
+        params+=("$param=$value")
+      fi
+      # Reset variables for the new parameter
+      param="$word"
+      value=""
+    else
+      value+="$word "
+    fi
+  done
+  
+    # Add the last parameter to the array
+  params+=("$param=$value")
+
+  # Read the separated parameters and values into a new array
+  export fixed_param=""
+  export fixed_value=""
+  export fixed_params=()
+  export new_params=""
+  CONFIG_ARRAY=()
+  BUILD_ARRAY=()
+  IGNORED_ARRAY=()
+
+  for p in "${params[@]}"; do
+    IFS='=' read -ra parts <<< "$p"
+    prefixed_param=${parts[0]}
+    fixed_param="${prefixed_param%%[[:space:]]}"
+    prepped_value=${parts[1]}
+    fixed_value=$(echo "$prepped_value" | awk '{$1=$1};1')
+    # Handle Special parameters
+    if [ "$fixed_param" == "--jdk-boot-dir" ]; then fixed_value="$BOOTJDK_HOME" ; fi
+    
+    # Fix Build Variant Parameter To Strip JDK Version
+    if [ "$fixed_param" == "--build-variant" ] ; then
+      # Remove Leading White Space
+      trimmed_value=$(echo "$prepped_value" | awk '{$1=$1};1')
+      IFS=' ' read -r variant jdk <<< "$trimmed_value"
+      if [[ $jdk == jdk* ]]; then
+        variant="$variant "
+      else
+        temp="$variant "
+        variant="$jdk"
+        jdk="$temp"
+      fi
+      fixed_value=$variant
+    fi
+
+    # Check if fixed_param is in CONFIG_ARGS
+    if containsElement "$fixed_param" "${CONFIG_ARGS[@]}"; then
+      # Add Config Arg To New Array
+      if [ "$fixed_param" == "--with-version-opt" ] ; then
+        STRINGTOADD="$fixed_param=$fixed_value"
+        CONFIG_ARRAY+=("$STRINGTOADD")
+      else
+        STRINGTOADD="$fixed_param"
+        CONFIG_ARRAY+=("$STRINGTOADD")
+      fi
+    elif containsElement "$fixed_param" "${NOTUSE_ARGS[@]}"; then
+      # Strip Parameters To Be Ignored
+      STRINGTOADD="$fixed_param $fixed_value"
+      IGNORED_ARRAY+=("$STRINGTOADD")
+    else
+      # Not A Config Param Nor Should Be Ignored, So Add To Build Array
+      STRINGTOADD="$fixed_param $fixed_value"
+      BUILD_ARRAY+=("$STRINGTOADD")
+    fi
+  done
+
+  IFS=' ' build_string="${BUILD_ARRAY[*]}"
+  IFS=' ' config_string=$"${CONFIG_ARRAY[*]}"
+  final_params="$build_string --configure-args \"$config_string\" $jdk"
+
+  echo "Make JDK Any Platform Argument List = "
+  echo "$final_params"
+}
+
 cleanBuildInfo() {
   # shellcheck disable=SC3043
   local DIR="$1"
@@ -126,17 +234,17 @@ BOOTJDK_VERSION=$(jq -r '.metadata.tools[] | select(.name == "BOOTJDK") | .versi
 GCCVERSION=$(jq -r '.metadata.tools[] | select(.name == "GCC") | .version' "$SBOM" | sed 's/.0$//')
 LOCALGCCDIR=/usr/local/gcc$(echo "$GCCVERSION" | cut -d. -f1)
 TEMURIN_BUILD_SHA=$(jq -r '.components[0] | .properties[] | select (.name == "Temurin Build Ref") | .value' "$SBOM" | awk -F/ '{print $NF}')
-TEMURIN_BUILD_ARGS=$(jq -r '.components[0] | .properties[] | select (.name == "makejdk_any_platform_args") | .value' "$SBOM" | cut -d\" -f4 | sed -e "s/--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt/'--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt'/" -e "s/ --disable-warnings-as-errors --enable-dtrace/ '--disable-warnings-as-errors --enable-dtrace'/" -e 's/\\n//g' -e "s,--jdk-boot-dir [^ ]*,--jdk-boot-dir /usr/lib/jvm/jdk-$BOOTJDK_VERSION,g")
+TEMURIN_BUILD_ARGS=$(jq -r '.components[0] | .properties[] | select (.name == "makejdk_any_platform_args") | .value' "$SBOM")
 TEMURIN_VERSION=$(jq -r '.metadata.component.version' "$SBOM" | sed 's/-beta//' | cut -f1 -d"-")
-
+BUILDSTAMP=$(jq -r '.components[0].properties[] | select(.name == "Build Timestamp") | .value' "$SBOM")
 NATIVE_API_ARCH=$(uname -m)
 if [ "${NATIVE_API_ARCH}" = "x86_64" ]; then NATIVE_API_ARCH=x64; fi
 if [ "${NATIVE_API_ARCH}" = "armv7l" ]; then NATIVE_API_ARCH=arm; fi
 
 checkAllVariablesSet
-
 downloadTooling
 setEnvironment
+setBuildArgs
 
 if [ -z "$JDK_PARAM" ] && [ ! -d "jdk-${TEMURIN_VERSION}" ] ; then
     JDK_PARAM="https://api.adoptium.net/v3/binary/version/jdk-${TEMURIN_VERSION}/linux/${NATIVE_API_ARCH}/jdk/hotspot/normal/eclipse?project=jdk"
@@ -158,12 +266,7 @@ if [ "${isJdkDir}" = true ]; then
   comparedDir=$JDK_PARAM
 fi
 
-rc=0
-echo " cd temurin-build && ./makejdk-any-platform.sh $TEMURIN_BUILD_ARGS 2>&1 | tee build.$$.log; echo \$?" | sh > temp.log
-rc=$(tail -n 1 temp.log)
-if [ $rc -ne 0 ]; then
-  echo "JDK rebuild failed. Stop and Exit..."
-fi
+echo " cd temurin-build && ./makejdk-any-platform.sh $final_params 2>&1 | tee build.$$.log" | sh
 
 echo Comparing ...
 mkdir compare.$$
@@ -173,7 +276,7 @@ cp "$SBOM" SBOM.json
 
 cleanBuildInfo "${comparedDir}"
 cleanBuildInfo "compare.$$/jdk-$TEMURIN_VERSION"
-
+rc=0
 # shellcheck disable=SC2069
 diff -r "${comparedDir}" "compare.$$/jdk-$TEMURIN_VERSION" 2>&1 > "reprotest.diff" || rc=$?
 
--- a/tooling/reproducible/linux_repro_build_compare.sh
+++ b/tooling/reproducible/linux_repro_build_compare.sh
@@ -158,7 +158,7 @@ if echo "$TEMURIN_BUILD_ARGS" | grep -v configure-args\ \\\"; then
         sed -e "s/--with-jobs=[0-9]*//g" \
          -e "s/--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt/'--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt'/" \
          -e "s/--disable-warnings-as-errors --enable-dtrace *--with-version-opt=ea/'--disable-warnings-as-errors --enable-dtrace --with-version-opt=ea'/" \
-         -e "s/--disable-warnings-as-errors --enable-dtrace/'--disable-warnings-as-errors --enable-dtrace'/" \
+         -e "s/ --disable-warnings-as-errors --enable-dtrace/ '--disable-warnings-as-errors --enable-dtrace'/" \
          -e 's/\\n//g')
 fi
 
--- a/tooling/reproducible/linux_repro_build_compare.sh
+++ b/tooling/reproducible/linux_repro_build_compare.sh
@@ -25,7 +25,7 @@ ANT_SHA=9028e2fc64491cca0f991acc09b06ee7fe644afe41d1d6caf72702ca25c4613c
 ANT_CONTRIB_VERSION=1.0b3
 ANT_CONTRIB_SHA=4d93e07ae6479049bb28071b069b7107322adaee5b70016674a0bffd4aac47f9
 isJdkDir=false
-
+USING_DEVKIT="true"
 installPrereqs() {
   if test -r /etc/redhat-release; then
     yum install -y gcc gcc-c++ make autoconf unzip zip alsa-lib-devel cups-devel libXtst-devel libXt-devel libXrender-devel libXrandr-devel libXi-devel
@@ -75,12 +75,15 @@ downloadAnt() {
   fi
 }
 
-setEnvironment() {
+setGccEnvironment() {
   export CC="${LOCALGCCDIR}/bin/gcc-${GCCVERSION}"
   export CXX="${LOCALGCCDIR}/bin/g++-${GCCVERSION}"
   export LD_LIBRARY_PATH="${LOCALGCCDIR}/lib64"
-  # /usr/local/bin required to pick up the new autoconf if required
+}
+
+setAntEnvironment() {
   export PATH="${LOCALGCCDIR}/bin:/usr/local/bin:/usr/bin:$PATH:/usr/local/apache-ant-${ANT_VERSION}/bin"
+  # /usr/local/bin required to pick up the new autoconf if required
   ls -ld "$CC" "$CXX" "/usr/lib/jvm/jdk-${BOOTJDK_VERSION}/bin/javac" || exit 1
 }
 
@@ -90,12 +93,29 @@ cleanBuildInfo() {
   sed -i '/^BUILD_INFO=.*$/d' "${DIR}/release"
 }
 
+setTemurinBuildArgs() {
+  local buildArgs="$1"
+  local bootJdk="$2"
+  local ignoreOptions=("--enable-sbom-strace ")
+  for ignoreOption in "${ignoreOptions[@]}"; do
+    buildArgs="${buildArgs/${ignoreOption}//}"
+  done
+  #reset --jdk-boot-dir
+  # shellcheck disable=SC2001
+  buildArgs="$(echo "$buildArgs" | sed -e "s|--jdk-boot-dir [^ ]*|--jdk-boot-dir /usr/lib/jvm/jdk-${bootJdk}|")"
+  echo "${buildArgs}"
+}
+
 downloadTooling() {
+  local using_DEVKIT=$1
+
   if [ ! -r "/usr/lib/jvm/jdk-${BOOTJDK_VERSION}/bin/javac" ]; then
     echo "Retrieving boot JDK $BOOTJDK_VERSION" && mkdir -p /usr/lib/jvm && curl -L "https://api.adoptium.net/v3/binary/version/jdk-${BOOTJDK_VERSION}/linux/${NATIVE_API_ARCH}/jdk/hotspot/normal/eclipse?project=jdk" | (cd /usr/lib/jvm && tar xpzf -)
   fi
-  if [ ! -r "${LOCALGCCDIR}/bin/g++-${GCCVERSION}" ]; then
-    echo "Retrieving gcc $GCCVERSION" && curl "https://ci.adoptium.net/userContent/gcc/gcc$(echo "$GCCVERSION" | tr -d .).$(uname -m).tar.xz" | (cd /usr/local && tar xJpf -) || exit 1
+  if [[ "${using_DEVKIT}" == "false" ]]; then
+    if [ ! -r "${LOCALGCCDIR}/bin/g++-${GCCVERSION}" ]; then
+      echo "Retrieving gcc $GCCVERSION" && curl "https://ci.adoptium.net/userContent/gcc/gcc$(echo "$GCCVERSION" | tr -d .).$(uname -m).tar.xz" | (cd /usr/local && tar xJpf -) || exit 1
+    fi
   fi
   if [ ! -r temurin-build ]; then
     git clone https://github.com/adoptium/temurin-build || exit 1
@@ -125,16 +145,24 @@ GCCVERSION=$(jq -r '.metadata.tools[] | select(.name == "GCC") | .version' "$SBO
 LOCALGCCDIR=/usr/local/gcc$(echo "$GCCVERSION" | cut -d. -f1)
 TEMURIN_BUILD_SHA=$(jq -r '.components[0] | .properties[] | select (.name == "Temurin Build Ref") | .value' "$SBOM" | awk -F/ '{print $NF}')
 TEMURIN_BUILD_ARGS=$(jq -r '.components[0] | .properties[] | select (.name == "makejdk_any_platform_args") | .value' "$SBOM")
-TEMURIN_BUILD_ARGS="${TEMURIN_BUILD_ARGS//--enable-sbom-strace /}"
 TEMURIN_VERSION=$(jq -r '.metadata.component.version' "$SBOM" | sed 's/-beta//' | cut -f1 -d"-")
 #BUILDSTAMP=$(jq -r '.components[0].properties[] | select(.name == "Build Timestamp") | .value' "$SBOM")
 NATIVE_API_ARCH=$(uname -m)
 if [ "${NATIVE_API_ARCH}" = "x86_64" ]; then NATIVE_API_ARCH=x64; fi
 if [ "${NATIVE_API_ARCH}" = "armv7l" ]; then NATIVE_API_ARCH=arm; fi
+if [[ "$TEMURIN_BUILD_ARGS" =~ .*"--use-adoptium-devkit".* ]]; then
+  USING_DEVKIT="false"
+fi
 
 checkAllVariablesSet
-downloadTooling
-setEnvironment
+downloadTooling "$USING_DEVKIT"
+if [[ "${USING_DEVKIT}" == "false" ]]; then
+  setGccEnvironment
+fi
+setAntEnvironment
+echo "original temurin build args is ${TEMURIN_BUILD_ARGS}"
+TEMURIN_BUILD_ARGS=$(setTemurinBuildArgs "$TEMURIN_BUILD_ARGS" "$BOOTJDK_VERSION")
+
 
 if [ -z "$JDK_PARAM" ] && [ ! -d "jdk-${TEMURIN_VERSION}" ] ; then
     JDK_PARAM="https://api.adoptium.net/v3/binary/version/jdk-${TEMURIN_VERSION}/linux/${NATIVE_API_ARCH}/jdk/hotspot/normal/eclipse?project=jdk"
@@ -155,6 +183,7 @@ if [ "${isJdkDir}" = true ]; then
   comparedDir=$JDK_PARAM
 fi
 
+echo "Rebuild Temurin build args is $TEMURIN_BUILD_ARGS"
 echo " cd temurin-build && ./makejdk-any-platform.sh $TEMURIN_BUILD_ARGS 2>&1 | tee build.$$.log" | sh
 
 echo Comparing ...
--- a/tooling/reproducible/linux_repro_build_compare.sh
+++ b/tooling/reproducible/linux_repro_build_compare.sh
@@ -25,7 +25,7 @@ ANT_SHA=9028e2fc64491cca0f991acc09b06ee7fe644afe41d1d6caf72702ca25c4613c
 ANT_CONTRIB_VERSION=1.0b3
 ANT_CONTRIB_SHA=4d93e07ae6479049bb28071b069b7107322adaee5b70016674a0bffd4aac47f9
 isJdkDir=false
-USING_DEVKIT="true"
+USING_DEVKIT="false"
 installPrereqs() {
   if test -r /etc/redhat-release; then
     yum install -y gcc gcc-c++ make autoconf unzip zip alsa-lib-devel cups-devel libXtst-devel libXt-devel libXrender-devel libXrandr-devel libXi-devel
@@ -83,8 +83,6 @@ setGccEnvironment() {
 
 setAntEnvironment() {
   export PATH="${LOCALGCCDIR}/bin:/usr/local/bin:/usr/bin:$PATH:/usr/local/apache-ant-${ANT_VERSION}/bin"
-  # /usr/local/bin required to pick up the new autoconf if required
-  ls -ld "$CC" "$CXX" "/usr/lib/jvm/jdk-${BOOTJDK_VERSION}/bin/javac" || exit 1
 }
 
 cleanBuildInfo() {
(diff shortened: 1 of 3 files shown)


sophia-guo:
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10109/

sophia-guo:
Grinder runs extremely long. Seems like doing some unexpected steps

```
13:52:53  
13:53:06  find "/home/jenkins/temurin-build/workspace/./build//straceOutput" -type f -name 'outputFile.*' | xargs -n100 grep -v ENOENT | cut -d'"' -f2 | grep "^/" | eval "grep -Ev '(\.java$|\.d$|\.o$|\.d\.targets$|^/home/jenkins/temurin-build|\+\+\+|\-\-\-|^/dev/|^/proc/|^/sys/|^/tmp/)'" | sort | uniq
13:53:06  /bin/sh
13:53:06  /etc/ld.so.cache
13:53:06  /etc/nsswitch.conf
13:53:06  /etc/passwd
13:53:06  /home/jenkins/workspace/build-scripts/jobs/jdk21u/jdk21u-linux-x64-temurin/workspace/build/openjdkbuild/build.log
13:53:06  /home/jenkins/workspace/build-scripts/jobs/jdk21u/jdk21u-linux-x64-temurin/workspace/build/openjdkbuild/buildtools/classlist_classes
13:53:06  /home/jenkins/workspace/build-scripts/jobs/jdk21u/jdk21u-linux-x64-temurin/workspace/build/openjdkbuild/buildtools/classlist_classes/_the.CLASSLIST_JAR-javacserver.conf
13:53:06  /home/jenkins/workspace/build-scripts/jobs/jdk21u/jdk21u-linux-x64-temurin/workspace/build/openjdkbuild/buildtools/classlist_classes/_the.CLASSLIST_JAR.config_vardeps
13:53:06  /home/jenkins/workspace/build-scripts/jobs/jdk21u/jdk21u-linux-x64-temurin/workspace/build/openjdkbuild/buildtools/classlist_classes/_the.CLASSLIST_JAR.vardeps
13:53:06  /home/jenkins/workspace/build-scripts/jobs/jdk21u/jdk21u-linux-x64-temurin/workspace/build/openjdkbuild/buildtools/classlist_classes/_the.CLASSLIST_JAR_batch
13:53:06  /home/jenkins/workspace/build-scripts/jobs/jdk21u/jdk21u-linux-x64-temurin/workspace/build/openjdkbuild/buildtools/classlist_classes/_the.CLASSLIST_JAR_batch.cmdline
1
```

Is this related with https://github.com/adoptium/temurin-build/issues/3746? @andrew-m-leonard

karianna:
Linter also unhappy.

sophia-guo:
Turn off strace for reproducing build
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10140/

sophia-guo:
Lots of binary diff. @andrew-m-leonard would it be related with turn off strace? 

PR is ready.

andrew-m-leonard:
> Lots of binary diff. @andrew-m-leonard would it be related with turn off strace?
> 
> PR is ready.

no strace should have no affect.

I've done some analysis, and it looks like the Grinder test build is using the wrong gcc compiler:
```
17:14:20  configure: Using gcc C++ compiler version 11.3.0 [g++-11.3 (Adoptium/SXA 11.3.0) 11.3.0]
```
vs the [original build](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-linux-x64-temurin/200/consoleFull):
```
20:02:16  configure: Using gcc C++ compiler version 11.3.0 [g++ (GCC) 11.3.0]
```

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -84,8 +84,116 @@ setEnvironment() {
...
+  for e in "${@:2}"; do
+    if [ "$e" = "$1" ]; then
+      return 0  # Match found
+    fi
+  done
+  return 1  # No match found
+}
+
+setBuildArgs() {
+  local CONFIG_ARGS=("--disable-warnings-as-errors" "--enable-dtrace" "--without-version-pre" "--without-version-opt" "--with-version-opt")
```
andrew-m-leonard, line 99: We should be using the original build configure_args, not specifying these...?

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -84,8 +84,116 @@ setEnvironment() {
...
+  local NOTUSE_ARGS=("--configure-args" "--enable-sbom-strace")
```
andrew-m-leonard, line 100: --configure-args should not be being ignored, only --enable-sbom-strace

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -84,8 +84,116 @@ setEnvironment() {
...
+        variant="$jdk"
+        jdk="$temp"
+      fi
+      fixed_value=$variant
+    fi
+
+    # Check if fixed_param is in CONFIG_ARGS
+    if containsElement "$fixed_param" "${CONFIG_ARGS[@]}"; then
+      # Add Config Arg To New Array
+      if [ "$fixed_param" == "--with-version-opt" ] ; then
```
andrew-m-leonard, line 170: shouldn't be needed, use original --configure-args

tooling/reproducible/linux_repro_build_compare.sh
andrew-m-leonard: we need to fix the SBOM makejdk_any_platform_args which is losing the --configure-args "quotes"...

replace the code here: https://github.com/adoptium/temurin-build/blob/39a0b0094099f3aad0661211859c944fa0610c7d/makejdk-any-platform.sh#L54
with this, I "think" should fix that:
```
# Store params to this script as "buildinfo"
makeJdkArgs=""
for arg in "$@"; do
  # Quote the argument if it contains spaces
  if [[ "${arg}" =~ .*" ".* ]]; then
    makeJdkArgs="${makeJdkArgs} \"${arg}\""
  else
    makeJdkArgs="${makeJdkArgs} ${arg}"
  fi
done
echo "${makeJdkArgs}" > ./workspace/config/makejdk-any-platform.args
```

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -84,8 +84,116 @@ setEnvironment() {
   ls -ld "$CC" "$CXX" "/usr/lib/jvm/jdk-${BOOTJDK_VERSION}/bin/javac" || exit 1
```
andrew-m-leonard, line 84: split this method into separate:
setGccEnvironment()
setAntEnvironment()
Also remove check on javac, as it might not be located in /usr/lib/jvm/jdk-

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -126,23 +233,22 @@ BOOTJDK_VERSION=$(jq -r '.metadata.tools[] | select(.name == "BOOTJDK") | .versi
...
-
+BUILDSTAMP=$(jq -r '.components[0].properties[] | select(.name == "Build Timestamp") | .value' "$SBOM")
 NATIVE_API_ARCH=$(uname -m)
 if [ "${NATIVE_API_ARCH}" = "x86_64" ]; then NATIVE_API_ARCH=x64; fi
 if [ "${NATIVE_API_ARCH}" = "armv7l" ]; then NATIVE_API_ARCH=arm; fi
 
 checkAllVariablesSet
-
 downloadTooling
 setEnvironment
```
andrew-m-leonard, line 245: split this into two:
```
setAntEnvironment()

if [[ NOT_USING_DEVKIT ]]; then
  setGccEnvironment()
fi
```

andrew-m-leonard: in downloadTooling() add DEVKIT check... :
```
if [[ NOT_USING_DEVKIT ]]; then
    if [ ! -r "${LOCALGCCDIR}/bin/g++-${GCCVERSION}" ]; then
    echo "Retrieving gcc $GCCVERSION" && curl "https://ci.adoptium.net/userContent/gcc/gcc$(echo "$GCCVERSION" | tr -d .).$(uname -m).tar.xz" | (cd /usr/local && tar xJpf -) || exit 1
  fi
fi
```

andrew-m-leonard (changes requested):

tooling/reproducible/linux_repro_build_compare.sh
andrew-m-leonard: yes just tested it, seems to do the right thing, you can then consume configure_args with this fix:
```
        {
          "name" : "makejdk_any_platform_args",
          "value" : "--clean-git-repo --jdk-boot-dir /home/jenkins/temurin-build/jdk-22 --configure-args \" --disable-warnings-as-errors --with-version-opt=LTS --with-version-pre=ea\" --target-file-name jdk-hotspot.tar.gz --release --clean-libs --tag jdk-23+24_adopt --create-sbom --user-openjdk-build-root-directory /home/jenkins/build2 --use-adoptium-devkit gcc-11.3.0-Centos7.6.1810-b03 --freetype-dir bundled --use-jep319-certs --create-debug-image --build-variant temurin jdk"
        },
```

tooling/reproducible/windows_repro_build_compare.sh
```diff
@@ -57,13 +57,14 @@ SIGNTOOL_BASE="C:/Program Files (x86)/Windows Kits/10"
 # Define What Are Configure Args & Redundant Args
 # This MAY Need Updating If Additional Configure Args Are Passed
```
andrew-m-leonard, line 58: ditto, not sure this all of this right, we should use original configure-args, and update args that need to point to local path equivalents, eg.--with-ucrt-dll-dir needs to point at local Windows dir path

andrew-m-leonard (changes requested):

sophia-guo:
> https://github.com/adoptium/temurin-build/pull/3824#issuecomment-2141514799

gcc is downlaod from [adoptium](https://ci.adoptium.net/userContent/gcc/)

https://github.com/adoptium/temurin-build/blame/master/tooling/reproducible/linux_repro_build_compare.sh#L98-L100

tooling/reproducible/linux_repro_build_compare.sh
sophia-guo: --configure-args  is not ignored. Its basically using the same logic as window repro_build code. it was added back `final_params="$build_string --configure-args \"$config_string\" $jdk"`. The idea is do with --configure-args differently as it need the double quoted.

tooling/reproducible/linux_repro_build_compare.sh
sophia-guo: I'm not sure about this --configure-args. will this work for other cases? Like the value of it may be variable as `--disable-warnings-as-errors --with-version-opt=LTS --enable-dtrace --without-version-pre`?

tooling/reproducible/windows_repro_build_compare.sh
sophia-guo: No idea maybe confirm with @steelhead31

tooling/reproducible/linux_repro_build_compare.sh
andrew-m-leonard: Yes, you should not need this if we fix the SBOM --configure-args BUILD_ARG, which is currently missing the double-quotes
The problem here is we are adding assumptions about what configure args are expected, we should just pass them through.
The Windows logic is wrong too.

tooling/reproducible/linux_repro_build_compare.sh
andrew-m-leonard: @sophia-guo i've created a PR to fix this, so we can get some builds you can test this with...
https://github.com/adoptium/temurin-build/pull/3835

sophia-guo:
Using temurin master with https://github.com/adoptium/temurin-build/pull/3835 in

https://ci.adoptium.net/view/Test_grinder/job/Grinder/10155/

tooling/reproducible/linux_repro_build_compare.sh
sophia-guo: @andrew-m-leonard how is `NOT_USING_DEVKIT` defined? Is it defined by following options?

```

if [[ "${CONFIGURE_ARGS}" =~ .*"--with-devkit=".* ]]; then
  echo "Using gcc from DevKit toolchain specified in configure args"
elif [[ "${BUILD_ARGS}" =~ .*"--use-adoptium-devkit".* ]]; then
  echo "Using gcc from Adoptium DevKit toolchain specified in --use-adoptium-devkit build args"
```

sophia-guo: SBOM has the args `--use-adoptium-devkit`

andrew-m-leonard: > @andrew-m-leonard how is `NOT_USING_DEVKIT` defined? Is it defined by following options?
> 
> ```
> 
> if [[ "${CONFIGURE_ARGS}" =~ .*"--with-devkit=".* ]]; then
>   echo "Using gcc from DevKit toolchain specified in configure args"
> elif [[ "${BUILD_ARGS}" =~ .*"--use-adoptium-devkit".* ]]; then
>   echo "Using gcc from Adoptium DevKit toolchain specified in --use-adoptium-devkit build args"
> ```

Correct, --use-adoptium-devkit will download the required devkit

sophia-guo:
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10233/

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
karianna, line 1: Do we have to move to bash?

sophia-guo:
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10233/

@andrew-m-leonard  still binaries differ. Original SBOM has option `--use-adoptium-devkit`. So we did setGccEnvironment, which download gcc from [adoptium](https://ci.adoptium.net/userContent/gcc/)

https://github.com/adoptium/temurin-build/blame/master/tooling/reproducible/linux_repro_build_compare.sh#L98-L100, is it correct?

Also diff from release

```
diff -r /home/jenkins/jdkbinary/release compare.1/jdk-21.0.4+5/release
2,3c2,3
< IMPLEMENTOR_VERSION="Temurin-21.0.4+5-202406031931"
< JAVA_RUNTIME_VERSION="21.0.4-beta+5-202406031931"
---
> IMPLEMENTOR_VERSION="Temurin-21.0.4+5-202406052324"
> JAVA_RUNTIME_VERSION="21.0.4-beta+5-202406052324"
11,14c11
< BUILD_SOURCE="git:2ad2f7c73bb23e23cd4f20208c01576bec75f1f3"
< BUILD_SOURCE_REPO="https://github.com/adoptium/temurin-build.git"
< SOURCE_REPO="https://github.com/adoptium/jdk21u.git"
< FULL_VERSION="21.0.4-beta+5-202406031931"
---
> FULL_VERSION="21.0.4-beta+5-202406052324"
17c14
< JVM_VERSION="21.0.4-beta+5-202406031931"
---
> JVM_VERSION="21.0.4-beta+5-202406052324"
```

Is this related with value of timestamp, which need to be reset? ( if that's the case, why we didn't have this issue before?)

tooling/reproducible/linux_repro_build_compare.sh
sophia-guo: Using /bin/sh linter isn't happy and I'm not sure why /bin/sh was initially used. I've searched all shell scripts in all adoptium repos and found all shell scripts are using /bin/bash. So I did the change.

andrew-m-leonard:
@sophia-guo you need to add BUILD_ARG of "--build-reproducible-date ${buildTimeStamp}"
where buildTimeStamp comes from the SBOM
like was done before: https://github.com/adoptium/ci-jenkins-pipelines/blob/49d73e26f7ff74a7872d00b856af003745ef2dbb/tools/reproduce_comparison/Jenkinsfile#L67

sophia-guo:
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10270/

sophia-guo:
@andrew-m-leonard tests is running in the contain with basic docker centos:7 . I also double checked in the container there is no local /usr/local/gcc11 /usr/local/gcc11 
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10352/console

C Compiler:     Version 11.3.0 (at /usr/local/gcc11/bin/gcc-11.3) is set by 

https://github.com/adoptium/temurin-build/pull/3824/files#diff-65a42dfc13c7c802493b4d5ddf3040f762df150947ea544841d387a4b82da73fR120-R123

and https://github.com/adoptium/temurin-build/pull/3824/files#diff-65a42dfc13c7c802493b4d5ddf3040f762df150947ea544841d387a4b82da73fR163-R166.

So  CC and CXX environment was setup by download and export if original sbom says '--use-adoptium-devkit'.

sophia-guo:
This PR is fixing the parameter parsing, I think we can get this done first. Any related reproducible issues or underlying temurin-build or sbom issues we can open specific issues correspondingly. @andrew-m-leonard opinion?

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -126,23 +149,30 @@ BOOTJDK_VERSION=$(jq -r '.metadata.tools[] | select(.name == "BOOTJDK") | .versi
...
-TEMURIN_BUILD_ARGS=$(jq -r '.components[0] | .properties[] | select (.name == "makejdk_any_platform_args") | .value' "$SBOM" | cut -d\" -f4 | sed -e "s/--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt/'--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt'/" -e "s/ --disable-warnings-as-errors --enable-dtrace/ '--disable-warnings-as-errors --enable-dtrace'/" -e 's/\\n//g' -e "s,--jdk-boot-dir [^ ]*,--jdk-boot-dir /usr/lib/jvm/jdk-$BOOTJDK_VERSION,g")
+TEMURIN_BUILD_ARGS=$(jq -r '.components[0] | .properties[] | select (.name == "makejdk_any_platform_args") | .value' "$SBOM")
 TEMURIN_VERSION=$(jq -r '.metadata.component.version' "$SBOM" | sed 's/-beta//' | cut -f1 -d"-")
-
+BUILDSTAMP=$(jq -r '.components[0].properties[] | select(.name == "Build Timestamp") | .value' "$SBOM")
 NATIVE_API_ARCH=$(uname -m)
 if [ "${NATIVE_API_ARCH}" = "x86_64" ]; then NATIVE_API_ARCH=x64; fi
 if [ "${NATIVE_API_ARCH}" = "armv7l" ]; then NATIVE_API_ARCH=arm; fi
+if [[ "$TEMURIN_BUILD_ARGS" =~ .*"--use-adoptium-devkit".* ]]; then
+  USING_DEVKIT="false"
```
andrew-m-leonard, line 159: This should be:
```
  USING_DEVKIT="true"
```

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -25,7 +25,7 @@ ANT_SHA=9028e2fc64491cca0f991acc09b06ee7fe644afe41d1d6caf72702ca25c4613c
 ANT_CONTRIB_VERSION=1.0b3
 ANT_CONTRIB_SHA=4d93e07ae6479049bb28071b069b7107322adaee5b70016674a0bffd4aac47f9
 isJdkDir=false
-
+USING_DEVKIT="true"
```
andrew-m-leonard, line 28: Should assume no DevKit default
```
USING_DEVKIT="false"
```

andrew-m-leonard (changes requested):

andrew-m-leonard:
> @andrew-m-leonard tests is running in the contain with basic docker centos:7 . I also double checked in the container there is no local /usr/local/gcc11 /usr/local/gcc11 https://ci.adoptium.net/view/Test_grinder/job/Grinder/10352/console
> 
> C Compiler: Version 11.3.0 (at /usr/local/gcc11/bin/gcc-11.3) is set by
> 
> https://github.com/adoptium/temurin-build/pull/3824/files#diff-65a42dfc13c7c802493b4d5ddf3040f762df150947ea544841d387a4b82da73fR120-R123
> 
> and https://github.com/adoptium/temurin-build/pull/3824/files#diff-65a42dfc13c7c802493b4d5ddf3040f762df150947ea544841d387a4b82da73fR163-R166.
> 
> So CC and CXX environment was setup by download adoptium gcc tar and export if original sbom says '--use-adoptium-devkit'.

@sophia-guo The USING_DEVKIT boolean is the wrong way around...

sophia-guo:
https://ci.adoptium.net/job/Grinder/10386/console

@andrew-m-leonard

andrew-m-leonard:
That's great, only these differ now:
```
17:27:46  < BUILD_SOURCE="git:2ad2f7c73bb23e23cd4f20208c01576bec75f1f3"
17:27:46  < BUILD_SOURCE_REPO="https://github.com/adoptium/temurin-build.git"
17:27:46  < SOURCE_REPO="https://github.com/adoptium/jdk21u.git"
```
the reason for that is the failures further up, eg:
```
17:25:03  Unable to fetch build SHA, does a work tree exist?...
17:25:03  Unable to fetch OpenJDK Source SHA, does a work tree exist?...
17:25:03  fatal: attempt to fetch/clone from a shallow repository
17:25:03  fatal: Could not read from remote repository.
17:25:03  
17:25:03  Please make sure you have the correct access rights
17:25:03  and the repository exists.
17:25:03  fatal: attempt to fetch/clone from a shallow repository
17:25:03  fatal: Could not read from remote repository.
17:25:03  
17:25:03  Please make sure you have the correct access rights
17:25:03  and the repository exists.
17:25:03  fatal: attempt to fetch/clone from a shallow repository
17:25:03  fatal: Could not read from remote repository.
```

You need to work out why those are failing.... not sure why?

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -126,23 +147,30 @@ BOOTJDK_VERSION=$(jq -r '.metadata.tools[] | select(.name == "BOOTJDK") | .versi
 GCCVERSION=$(jq -r '.metadata.tools[] | select(.name == "GCC") | .version' "$SBOM" | sed 's/.0$//')
 LOCALGCCDIR=/usr/local/gcc$(echo "$GCCVERSION" | cut -d. -f1)
 TEMURIN_BUILD_SHA=$(jq -r '.components[0] | .properties[] | select (.name == "Temurin Build Ref") | .value' "$SBOM" | awk -F/ '{print $NF}')
-TEMURIN_BUILD_ARGS=$(jq -r '.components[0] | .properties[] | select (.name == "makejdk_any_platform_args") | .value' "$SBOM" | cut -d\" -f4 | sed -e "s/--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt/'--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt'/" -e "s/ --disable-warnings-as-errors --enable-dtrace/ '--disable-warnings-as-errors --enable-dtrace'/" -e 's/\\n//g' -e "s,--jdk-boot-dir [^ ]*,--jdk-boot-dir /usr/lib/jvm/jdk-$BOOTJDK_VERSION,g")
+TEMURIN_BUILD_ARGS=$(jq -r '.components[0] | .properties[] | select (.name == "makejdk_any_platform_args") | .value' "$SBOM")
```
sxa, line 150: I don't think we should remove these bits of sed that quote the configure args completely at the moment.

While the format of the SBOM [changed to ensure the configure args quoted in the SBOM](https://github.com/adoptium/temurin-build/pull/3835) this is still needed if we want to rebuild the last GA with these scripts (which require the devkit adjustments too, so we can't just rely on using an old version of the temurin-build repo to rebuild the latest GA)

Ref: [Old GA SBOM without quotes](https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.3%2B9/OpenJDK21U-sbom_x64_linux_hotspot_21.0.3_9.json) - [New SBOM with quotes](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-linux-x64-temurin/216/artifact/workspace/target/OpenJDK21U-sbom_x64_linux_hotspot_2024-06-03-19-28.json)

Can we (at least until we've done a few more GA levels) do something to detect if the `makejdk_any_platform_args` is already quoted, and if not apply this `sed` command?

sxa (changes requested):
Looks great overall and great to see this script getting a few updates! I've added a few comments, mostly just suggestions, but I've marked this as "Request Changes" because we should definitely retain the configure_args parsing for the last GA release to keep the "old" quoting working for that version to make sure we have a viable story for end users to rebuild the GA version.

One other note: When this is merged we should make the commit message a bit clearer to indicate that it is the parsing for the Linux reproducible build script e.g. `Update build args parsing in Linux reproducible build script`

andrew-m-leonard:
@sophia-guo yeah, I wasn't quite right in my previous discussion on the release.txt, we need to discover why addBuildSourceRepo is failing:
https://github.com/adoptium/temurin-build/blob/8ae8d2885735baa5fd95552542c1490d90e071fa/sbin/build.sh#L2184

I have seen this issue before, and I think it maybe the "git -C " option, if I remember correctly the "-C" option is only valid on certain versions of git, I suspect we should do this differently, maybe "cd repo && git config ..." ?

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -75,28 +75,50 @@ downloadAnt() {
...
   local DIR="$1"
   # BUILD_INFO name of OS level build was built on will likely differ
   sed -i '/^BUILD_INFO=.*$/d' "${DIR}/release"
 }
 
+setTemurinBuildArgs() {
+  local buildArgs="$1"
+  local bootJdk="$2"
+  local timeStamp="$3"
+  local ignoreOptions=("--enable-sbom-strace ")
```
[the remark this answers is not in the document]
sophia-guo: It's not a code standard issue. I personally feel the array is simpler if extra options need to be filtered out. Using multiple -e make it less readable and more error-prone for me.

tooling/reproducible/linux_repro_build_compare.sh
sophia-guo: My question would be it's currently used by anyone outside of Adoptium for former GAs? It's more like under development instead of production, which should allow the failures? For now if the PR is not in the issue with the devkit still exists. That is the issue with using the an old version of the temurin-build repo to rebuild without devkit adjustments will be an known issue. If necessary I would prefer to update some related documents to address this.

andrew-m-leonard:
@sophia-guo yes, it looks like the test job container is using centos:7 with git package: Package git.x86_64 0:1.8.3.1-25.el7_9 will be installed
I believe git 1.8 does not support "git -C"
So we can either upgrade git used in the test container, or fix the build.sh to use "cd <dir> && git config ..."

tooling/reproducible/linux_repro_build_compare.sh
sxa: Since we are actively pushing the story that we're reproducible it would be a mistake not to be able to do this with the scripts easily. I've done videos in the past and if we now have to saw "Well we can't do that just now" it's not a good story. I strongly believe we should support that scenario and that these scripts should never be considered only for our internal use.

It should be fairly simple to support this. Going forward we should be in the situation where we can point people at the GA SHA of temurin-build but in the meantime I feel it's very important to allow end users to be able to  prove reproducibility easily given how much we're pushing this in our marketing at the moment.

sxa: > If necessary I would prefer to update some related documents to address this.

Anything like that would be very time limited, and I'm not sure how we'd publish that in a way that people would be likely to find it. The reproducible build script has already been publicised in the past, so I feel it's important that it can be seen to work as demonstrated previously with the most recent GA level.

tooling/reproducible/linux_repro_build_compare.sh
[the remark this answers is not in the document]
sxa: I'm not convinced this logic is working. When I tried it out it converted:

```
--create-sbom --enable-sbom-strace --use-adoptium-devkit
```
in the string to:
```
--create-sbom /--use-adoptium-devkit
```
so if you want to leave it with the bash-specific construct, I think you'll need to switch `//` to `/` in the translation expression.
```

File: tooling/reproducible/linux_repro_build_compare.sh [hunk 12]
```diff
@@ -126,23 +147,30 @@ BOOTJDK_VERSION=$(jq -r '.metadata.tools[] | select(.name == "BOOTJDK") | .versi
...
+fi
 
 checkAllVariablesSet
+downloadTooling "$USING_DEVKIT"
+if [[ "${USING_DEVKIT}" == "false" ]]; then
+  setGccEnvironment
+fi
+setAntEnvironment
+echo "original temurin build args is ${TEMURIN_BUILD_ARGS}"
+TEMURIN_BUILD_ARGS=$(setTemurinBuildArgs "$TEMURIN_BUILD_ARGS" "$BOOTJDK_VERSION" "$BUILDSTAMP")
```
sxa, line 168: Extra blank line here should be removed

tooling/reproducible/linux_repro_build_compare.sh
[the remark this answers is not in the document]
sophia-guo: Confused. @sxa which logic did you try? Seems you were trying buildArgs="$(echo buildArgs | sed -e 's/--enable-sbom-strace//g')"? As I said I would prefer using the array, similar to the window reproducible comparing script.

tooling/reproducible/linux_repro_build_compare.sh
sxa: The following after the `jq` line that sets `TEMURIN_BUILD_ARGS` should be ok for all situations:
```
if echo "$TEMURIN_BUILD_ARGS" | grep -v configure-args\ \\\"; then
  echo "Using old method to escape configure-args in SBOM (build PR 3835)"
  TEMURIN_BUILD_ARGS=$(echo $TEMURIN_BUILD_ARGS | \
        cut -d\" -f4 | \
        sed -e "s/--with-jobs=[0-9]*//g" \
         -e "s/--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt/'--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt'/" \
         -e "s/--disable-warnings-as-errors --enable-dtrace *--with-version-opt=ea/'--disable-warnings-as-errors --enable-dtrace --with-version-opt=ea'/" \
         -e "s/--disable-warnings-as-errors --enable-dtrace/'--disable-warnings-as-errors --enable-dtrace'/" \
         -e 's/\\n//g' \
         -e "s,--jdk-boot-dir [^ ]*,--jdk-boot-dir /usr/lib/jvm/jdk-$BOOTJDK_VERSION,g")
fi
```
Noting that the removal of `--with-jobs=` is because it's not necessary for reproducing the builds, and was added to work around an issue when `enable-sbom-strace` is enabled (Ref: https://github.com/adoptium/ci-jenkins-pipelines/commit/68ffc9e3836d0690c9d4981c1b9d033cad6d8e1a). This could potentially be done in the same place as the removal of `enable-sbom-strace` but it needs to be done before the other translations in the above sed are done.

I've verified the above changes with the following and it looks good:
- Latest in jenkins: `./linux_repro_build_compare.sh https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-linux-aarch64-temurin/195/artifact/workspace/target/OpenJDK21U-sbom_aarch64_linux_hotspot_2024-06-13-10-35.json https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-linux-aarch64-temurin/195/artifact/workspace/target/OpenJDK21U-jdk_aarch64_linux_hotspot_2024-06-13-10-35.tar.gz` (Now deleted so can't be re-built again)
- Latest EA: `./linux_repro_build_compare.sh https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.4%2B5-ea-beta/OpenJDK21U-sbom_aarch64_linux_hotspot_21.0.4_5-ea.json https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.4%2B5-ea-beta/OpenJDK21U-jdk_aarch64_linux_hotspot_21.0.4_5-ea.tar.gz`
- Latest GA: `./linux_repro_build_compare.sh https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.3%2B9/OpenJDK21U-sbom_aarch64_linux_hotspot_21.0.3_9.json https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.3%2B9/OpenJDK21U-jdk_aarch64_linux_hotspot_21.0.3_9.tar.gz`

Note to self: The location specified by `--user-openjdk-build-root-directory` should be cleared before each subsequent attempt.
Second note to self: This script does not work if a fork of the adoptium/temurin-build is used as that repository is hard coded in the script, so - for example - jdk21u-linux-aarch64-temurin/197 does not work without a change to the script.

tooling/reproducible/linux_repro_build_compare.sh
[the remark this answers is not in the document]
sxa: > Seems you were trying buildArgs="$(echo buildArgs | sed -e 's/--enable-sbom-strace//g')"

No, the version with sed works ok - the current version with the array adds in an extra `/` unless you remove one of them:

```
sxa@fedora:~$ buildArgs="abc --enable-sbom-strace def"
sxa@fedora:~$ echo $buildArgs | sed -e 's/--enable-sbom-strace//g'
abc  def
sxa@fedora:~$ ignoreOptions=("--enable-sbom-strace ")
sxa@fedora:~$   for ignoreOption in "${ignoreOptions[@]}"; do
    buildArgs="${buildArgs/${ignoreOption}//}"
  done
sxa@fedora:~$ echo $buildArgs
abc /def
sxa@fedora:~$

File: tooling/reproducible/linux_repro_build_compare.sh (hunk 10 above)
Comment (sophia-guo) [reply tooling/reproducible/linux_repro_build_compare.sh:150]: Assume those covers all possible combinations of configure-args of former GAs.

File: tooling/reproducible/linux_repro_build_compare.sh (hunk 11 above)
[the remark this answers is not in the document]
Comment (sophia-guo) [reply tooling/reproducible/linux_repro_build_compare.sh:98]: I see, updated.

File: tooling/reproducible/linux_repro_build_compare.sh (hunk 10 above)
Comment (sophia-guo) [reply tooling/reproducible/linux_repro_build_compare.sh]: Updated. @sxa

File: tooling/reproducible/linux_repro_build_compare.sh [hunk 13]
```diff
@@ -126,23 +147,41 @@ BOOTJDK_VERSION=$(jq -r '.metadata.tools[] | select(.name == "BOOTJDK") | .versi
...
+TEMURIN_BUILD_ARGS=$(jq -r '.components[0] | .properties[] | select (.name == "makejdk_any_platform_args") | .value' "$SBOM")
+
+# Using old method to escape configure-args in SBOM (build PR 3835)", can be removed later.
+if echo "$TEMURIN_BUILD_ARGS" | grep -v configure-args\ \\\"; then
+  TEMURIN_BUILD_ARGS=$(echo "$TEMURIN_BUILD_ARGS" | \
+        cut -d\" -f4 | \
+        sed -e "s/--with-jobs=[0-9]*//g" \
+         -e "s/--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt/'--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt'/" \
+         -e "s/--disable-warnings-as-errors --enable-dtrace *--with-version-opt=ea/'--disable-warnings-as-errors --enable-dtrace --with-version-opt=ea'/" \
+         -e "s/--disable-warnings-as-errors --enable-dtrace/'--disable-warnings-as-errors --enable-dtrace'/" \
```
sxa, line 161: Apologies - copy error on my part, this needs a space at the beginning otherwise part of the parameters get escaped twice:
```suggestion
         -e "s/ --disable-warnings-as-errors --enable-dtrace/'--disable-warnings-as-errors --enable-dtrace'/" \
```

sxa:
Noting that with the changes to support older GA builds, both 21.0.3 and 21.0.2 GA builds are reproducible with this script. 21.0.1 does not work due to a checksum failure on `cyclonedx-core-java.jar` (which is likely because it's using the "latest" version in https://github.com/adoptium/temurin-build/blob/16ac032d56eb0820393ca7752c13a4bab98eaf92/cyclonedx-lib/build.xml#L47 and there has been a subsequent update)

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -126,23 +147,41 @@ BOOTJDK_VERSION=$(jq -r '.metadata.tools[] | select(.name == "BOOTJDK") | .versi
...
+TEMURIN_BUILD_ARGS=$(jq -r '.components[0] | .properties[] | select (.name == "makejdk_any_platform_args") | .value' "$SBOM")
+
+# Using old method to escape configure-args in SBOM (build PR 3835)", can be removed later.
+if echo "$TEMURIN_BUILD_ARGS" | grep -v configure-args\ \\\"; then
+  TEMURIN_BUILD_ARGS=$(echo "$TEMURIN_BUILD_ARGS" | \
+        cut -d\" -f4 | \
+        sed -e "s/--with-jobs=[0-9]*//g" \
+         -e "s/--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt/'--disable-warnings-as-errors --enable-dtrace --without-version-pre --without-version-opt'/" \
+         -e "s/--disable-warnings-as-errors --enable-dtrace *--with-version-opt=ea/'--disable-warnings-as-errors --enable-dtrace --with-version-opt=ea'/" \
+         -e "s/ --disable-warnings-as-errors --enable-dtrace/ '--disable-warnings-as-errors --enable-dtrace'/" \
```
sxa, line 161:
```suggestion
         -e "s/ --disable-warnings-as-errors --enable-dtrace/' --disable-warnings-as-errors --enable-dtrace'/" \
```

sxa (changes requested):
Apologies - space in the wrong place in the last change you did for me so hopefully you can just accept this change.
Other than this if you can give me a response to https://github.com/adoptium/temurin-build/pull/3824#discussion_r1639823161 that would be good (WDYT @andrew-m-leonard) and then we can get this in :-)

sxa (approved):
LGTM Although it would be good to get a response to https://github.com/adoptium/temurin-build/pull/3824#discussion_r1639823161 that would be good (WDYT @andrew-m-leonard since I think that name was your suggestion?)

Ignore my change request from a few minutes ago if you saw it - I've been starying at my screen for too long today and my suggestion was wrong ;-)

Also reiterating [my earlier comment](https://github.com/adoptium/temurin-build/pull/3824#pullrequestreview-2118361698) in case it gets lost when this is merged that it would be good to make the commit message a bit more descriptive when this is merged e.g. `Update build args parsing in Linux reproducible build script`

tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -75,28 +75,50 @@ downloadAnt() {
   fi
 }
 
-setEnvironment() {
+setGccEnvironment() {
```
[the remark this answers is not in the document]
andrew-m-leonard: yes agree, that is what was going through my head when I suggested it, ie.set the Gcc environment when not using a DevKit...

andrew-m-leonard:
@sophia-guo just change the setGccEnvironment() method namae as per Stewart's suggestion, then we're good to go i think

andrew-m-leonard (approved):
lgtm

Issue #3745, opened 2024-04-04
sbom strace analysis error: printNumberOfAllProcessedFiles: command not found


andrew-m-leonard:
https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-linux-x64-temurin/168/console
```
09:37:58  /home/jenkins/workspace/build-scripts/jobs/jdk21u/jdk21u-linux-x64-temurin/sbin/../tooling/strace_analysis.sh: line 263: printNumberOfAllProcessedFiles: command not found
```

Issue #3689, opened 2024-03-06
optimize strace_analysis.sh to not use set +eu


andrew-m-leonard:
The strace analysis script relies on set +eu, and checking for shell rc using $?, we should change this to set -eu and do shell commands in if checks.

PR #3826, opened 2024-05-28
Strace fixes

andrew-m-leonard:
Fix to various strace_analysis issues:
- Correct syntax errors
- Make work with set -eu
- Detect and log DevKit Adoptium version
- Correctly filter openjdk cloned source and openjdk build output that should not be analysed

Fixes https://github.com/adoptium/temurin-build/issues/3746
Fixes https://github.com/adoptium/temurin-build/issues/3745
Fixes https://github.com/adoptium/temurin-build/issues/3689

### DIFF
--- a/tooling/strace_analysis.sh
+++ b/tooling/strace_analysis.sh
@@ -22,19 +22,27 @@
 
 # Before executing this script, strace output files need to be generated
 # $1 is path of strace output folder
-# $2 is path of temurin-build folder, for exmaple: /home/user/Documents/temurin-build"
+# $2 is path of temurin-build folder, for example: /home/user/Documents/temurin-build"
 # $3 is javaHome
 # $4 is classpath
 # $5 is sbomJson
+# $6 is path of openjdk build output folder
+# $7 is path of cloned openjdk folder
+# $8 is Optional, path of devkit
 
-set +eu
+set -eu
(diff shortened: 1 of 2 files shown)


tooling/strace_analysis.sh
```diff
@@ -22,19 +22,27 @@
...
+# $2 is path of temurin-build folder, for example: /home/user/Documents/temurin-build"
 # $3 is javaHome
 # $4 is classpath
 # $5 is sbomJson
+# $6 is path of openjdk build output folder
+# $7 is path of cloned openjdk folder
+# $8 is Optional, path of devkit
 
-set +eu
+set -eu
```
karianna, line 33: Was the deliberate?

karianna (changes requested):

tooling/strace_analysis.sh
andrew-m-leonard: @karianna yes see https://github.com/adoptium/temurin-build/issues/3689, "set +eu" meant it would swallow failures!, rather than check for successful shell command completion. set -eu means we properly manage failures

karianna (approved):

tellison (approved):

sophia-guo (approved):

PR #3839, opened 2024-06-05
Exit from setupAntEnv is not propagated to its only usecase

judovana:
I had hit this when I forget to run all the build chain with JAVA_HOME set.  It died, but without any trace. This PR is adding the trace of what went wrong.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -871,10 +871,10 @@ generateSBoM() {
     return
   fi
 
-  # exit from $(setupAntEnv) is not propagated. We have to ensure, the exit propagates, and is fatal for the script
-  # co calling it one more time, rather sooner then later, without subshell.
-  setupAntEnv
-  local javaHome="$(setupAntEnv)"
+  # exit from local var=$(setupAntEnv) is not propagated. We have to ensure, the exit propagates, and is fatal for the script
+  # So the declaration is split. In that case the bug do not occur and thus the `exit 2` from setupAntEnv is correctly propagated
+  local javaHome
+  javaHome="$(setupAntEnv)"
 
   buildCyclonedxLib "${javaHome}"
   # classpath to run java app TemurinGenSBOM
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -871,7 +871,8 @@ generateSBoM() {
     return
   fi
 
-  # exit from $(setupAntEnv) is not propagated. We have to ensure it runs, so calling it one more time
+  # exit from $(setupAntEnv) is not propagated. We have to ensure, the exit propagates, and is fatal for the script
+  # co calling it one more time, rather sooner then later, without subshell.
   setupAntEnv
   local javaHome="$(setupAntEnv)"
 
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -52,6 +52,8 @@ export CONFIGURE_ARGS=""
 export ADDITIONAL_MAKE_TARGETS=""
 export GIT_CLONE_ARGUMENTS=()
 
+JAVA_HOME_MISSING_ERROR="Unable to find a suitable JAVA_HOME to build the cyclonedx-lib"
+
 # Parse the CL arguments, defers to the shared function in common-functions.sh
 function parseArguments() {
   parseConfigurationArguments "$@"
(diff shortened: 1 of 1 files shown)


judovana:
I do not have preffernce whether it is https://github.com/adoptium/temurin-build/pull/3839/commits/895dd6a0851ba8b3c0a3287eb6623f1dcb0d1668 or  https://github.com/adoptium/temurin-build/pull/3839/commits/8b866a55e1c10291b09675ad916d944c2c0a6a1f.

I like https://github.com/adoptium/temurin-build/pull/3839/commits/895dd6a0851ba8b3c0a3287eb6623f1dcb0d1668 more, but it may be to vague.

sbin/build.sh
```diff
@@ -860,6 +860,8 @@ generateSBoM() {
     return
   fi
 
+  # exit from $(setupAntEnv) is not propagated. We ahve to ensure it is fatal, co calling it one more times
```
karianna, line 863:
```suggestion
  # exit from $(setupAntEnv) is not propagated. We have to ensure it runs, so calling it one more time
```

judovana: This do not seem right. The patch is ensuring, that the `exit 2` is propagated, thus really making the call to `setupAntEnv` fatal, if it fails

karianna: Argh, sorry, I committed that change, please do override it with your explanation

judovana: ty, updated!

judovana:
spellcheck should have warned to this: https://www.shellcheck.net/wiki/SC2155#problematic-code-in-the-case-of-local

judovana:
Hard to say where it is intentional:
```
$  grep -ir  SC2155
build-farm/platform-specific-configurations/mac.sh:# shellcheck disable=SC2155
build-farm/platform-specific-configurations/alpine-linux.sh:# shellcheck disable=SC2155
build-farm/platform-specific-configurations/aix.sh:# shellcheck disable=SC2155
build-farm/platform-specific-configurations/windows.sh:# shellcheck disable=SC2155
build-farm/platform-specific-configurations/linux.sh:# shellcheck disable=SC2155
sbin/prepareWorkspace.sh:# shellcheck disable=SC2155,SC1091,SC2196,SC2235
sbin/common/config_init.sh:# shellcheck disable=SC2153,SC2155
sbin/common/common.sh:# shellcheck disable=SC2155
sbin/build.sh:# shellcheck disable=SC2155,SC2153,SC2038,SC1091,SC2116,SC2086
```

its usually some terrible `eval`

sbin/build.sh
```diff
@@ -871,7 +871,10 @@ generateSBoM() {
     return
   fi
 
-  local javaHome="$(setupAntEnv)"
+  # exit from local var=$(setupAntEnv) is not propagated. We have to ensure, the exit propagates, and is fatal for the script
```
karianna, line 874:
```suggestion
  # exit from local var=$(setupAntEnv) is not propagated. We have to ensure that the exit propagates, and is fatal for the script
```

sbin/build.sh
```diff
@@ -871,7 +871,10 @@ generateSBoM() {
...
+  # So the declaration is split. In that case the bug do not occur and thus the `exit 2` from setupAntEnv is correctly propagated
```
karianna, line 875:
```suggestion
  # So the declaration is split. In that case the bug does not occur and thus the `exit 2` from setupAntEnv is correctly propagated
```

karianna (changes requested):

judovana:
@karianna thanx.  The smallest I can do as a gratitude for your patience, is to sign myself up for some english class over summer:( do x does.. what a shame...

karianna (approved):

sxa (commented):
I'd like @andrew-m-leonard to review this one too since it's part of the setup for CycloneDX SBOM generation, so adding him as an approver. LGTM though. (Reviewing this as a "Comment" instead of "Approve" so that it doesn't get a second approval and merged until Andrew has approved!
Coincidentally this is related to a shellcheck exclusion that I queried in [another PR](https://github.com/adoptium/temurin-build/pull/3852#pullrequestreview-2122731100) today.

judovana:
It is clear bug. If it do not fail asap, it will die later, just in not debugg-able  exetption.

Thanx for eyball!

andrew-m-leonard (approved):
I think it's good, I didn't realise it did that...!

judovana:
ty!

PR #3842, opened 2024-06-10
Added support for zipped sources

judovana:
Sources tarballs and binaries are available both in zip, tar.gz and tar.xz.  "my" `-l/--openjdk-source-location` was counting only tarballs and was unpacking them with `tar -xf` with workaround around `--strip-components 1`. Now, if the file ends with `.zip` it is unpacked by `unzip` with same workaround for `--strip-components 1` as tar.

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -78,7 +78,7 @@ unpackFromArchive() {
         echo "Source zip unpacked as if it contains exactly one directory"
         unzipGnuAbsPathWithStripComponents1 "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"
     else
-      local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"  | grep "/$" -c) || local topLevelItems=1
+      local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}" | grep "/$" -c) || local topLevelItems=1
       if [ "$topLevelItems" -eq "1" ] ; then
         echo "Source tarball contains exactly one directory"
         untarGnuAbsPathWithStripComponents1 "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -75,7 +75,7 @@ unpackFromArchive() {
   # If the tarball contains .git files, they should be ignored later
   pushd "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
     if [ "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]: -4}" == ".zip" ] ; then
-        echo "Source zip unpacked as if ti contains exactly one directory"
+        echo "Source zip unpacked as if it contains exactly one directory"
         unzipGnuAbsPathWithStripComponents1 "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"
     else
       local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"  | grep "/$" -c) || local topLevelItems=1
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -51,7 +51,8 @@ copyFromDir() {
 # this is workarounding --strip-components 1 missing on gnu tar
 # it requires  absolute tar-filepath as it changes dir and is hardcoded to one
 # similar approach can be used also for zip in future
-unpackGnuAbsPathWithStripComponents1() {
+# warning! this method do not merge if (parts of!) destination exists.
+unpackGnuAbsPathWithStrip1Component() {
   local tmp=$(mktemp -d)
   pushd "$tmp" > /dev/null
     "$@"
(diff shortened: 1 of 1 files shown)


judovana:
@sxa  I had jsut realised one interesting fact. The magical method, workarounding  `--strip-components=1` with `mv` have one regressin against untar/unzip - it will not merge dirs

Long story short
```
mkdir -p a/lib/
mkdir -p b/lib/
touch  a/lib/file1
mv a/lib b/
touch  a/lib/file2
mv a/lib b/
mv: cannot move 'a/lib' to 'b/lib': Directory not empty
# however co would do
cp -r  a/lib b/
ls b/lib/
file1  file2
```

I hit it elsewhere, but nice expalnation is  https://askubuntu.com/a/269818 ; Worthy to fix?

sbin/prepareWorkspace.sh
```diff
@@ -51,29 +51,41 @@ copyFromDir() {
...
   echo "Extracting OpenJDK source tarball ${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]} to $(pwd)/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]} to build the binary"
   # If the tarball contains .git files, they should be ignored later
-  # todo, support also zips?
   pushd "./${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
-    local topLevelItems=$(tar --exclude='*/*' -tf  "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"  | grep "/$" -c) || local topLevelItems=1
-    if [ "$topLevelItems" -eq "1" ] ; then
-      echo "Source tarball contains exactly one directory"
-      untarGnuAbsPathWithStripComponents1 -xf "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]}"
+    if [ "${BUILD_CONFIG[OPENJDK_LOCAL_SOURCE_ARCHIVE_ABSPATH]: -4}" == ".zip" ] ; then
+        echo "Source zip unpacked as if ti contains exactly one directory"
```
karianna, line 78:
```suggestion
        echo "Source zip unpacked as if it contains exactly one directory"
```

karianna (changes requested):

sxa:
> @sxa I had jsut realised one interesting fact. The magical method, workarounding `--strip-components=1` with `mv` have one regressin against untar/unzip - it will not merge dirs
> 

> I hit it elsewhere, but nice expalnation is https://askubuntu.com/a/269818 ; Worthy to fix?

Interesting, although is that a situation that is likely to ever be seen? We would only expect one top level directory in such a tarball/zipball so this feels very much like an edge case (and perhaps an error if it does have multiple top level directories)

judovana:
Funnily  enough, I hit this issue, when I missues this function to merge Openjdk.tar.gz and Openjdk-staticlibs.tar.gz as I need them in one archive ((so unpack both to one dir, and then pack that one), and as I had temurin scripts already around, and I was more lazy then usually, I reused the method... I do nto think it is rare in common worl, but it shoudl be super rare in the scope those two functions are for.

karianna (approved):

sbin/prepareWorkspace.sh
```diff
@@ -51,29 +51,41 @@ copyFromDir() {
...
-untarGnuAbsPathWithStripComponents1() {
+unpackGnuAbsPathWithStripComponents1() {
   local tmp=$(mktemp -d)
   pushd "$tmp" > /dev/null
-    tar "$@"
+    "$@"
   popd  > /dev/null
   mv "$tmp"/*/* .
   mv "$tmp"/*/.* . || echo "no hidden files in tarball"
   rmdir "$tmp"/*
```
RadekCap, line 61: There are directories, this will fail at least on Windows.

```suggestion
  rm -rf "$tmp"/*
  rm -rf "$tmp"
```

judovana: I think this suggestion is wrong. It is intentional, that rmdir will fail if it is not empty.  If something remained here, then I think there was bug before. What directories remained here?  Was there something in console before?

RadekCap: You're right. I've forgotten to update the `previous` lines. We can't move the files because it leads to:

`mv /tmp/tmp.iwVmIoZLAI/jdk-21.0.0+0-static-libs/lib .
mv: cannot move '/tmp/tmp.iwVmIoZLAI/jdk-21.0.0+0-static-libs/lib' to './lib': Directory not empty`

The update was from 

`````
  mv "$tmp"/*/* .
  mv "$tmp"/*/.* . || echo "no hidden files in tarball"
`````

to

`````
  cp -r "$tmp"/*/* .
  cp -r "$tmp"/*/.* . || echo "no hidden files in tarball"
`````

Whole suggestion looks like this:

`````
  cp -r "$tmp"/*/* .
  cp -r "$tmp"/*/.* . || echo "no hidden files in tarball"
  rm -rf "$tmp"/*
  rm -rf  "$tmp"
`````

Comment (judovana) [reply sbin/prepareWorkspace.sh:61]: That is slightly different case. I do not want to mix those two changesets in this PR.

To change from `mv and rmdir` to `cp and rm -rf` really shoudl be separate change set. Maybe different function Aka  
```
// this method is very simialr to unpackGnuAbsPathWithStripComponents1
// however is able to mege archive even if targets exists, and rpevent
//mv /tmp/tmp.iwVmIoZLAI/jdk-21.0.0+0-static-libs/lib . mv: cannot move '/tmp/tmp.iwVmIoZLAI/jdk-21.0.0+0-static-libs/lib' to './lib': Directory not empty
// hoewer, is slightly unsafe. If you can, use unpackGnuAbsPathWithStripComponents1
unpackAndMErgeGnuAbsPathWithStripComponents1() {
  local tmp=$(mktemp -d)
  pushd "$tmp" > /dev/null
    "$@"
  popd  > /dev/null
  cp -ar "$tmp"/*/* .
  cp -ar "$tmp"/*/.* . || echo "no hidden files in tarball"
  rm -rf "$tmp"/*
  rm -rf "$tmp"
}
```

sxa (approved):
I'll make a couple of comments, but I'm also ok if they are not addressed so this isn't a blocker:
- `"$@"` on its own does look a bit odd ;-)
- While the format of the function was set in the previous PR It might be a bit nicer to rename `untarGnuAbsPathWithStripComponents1` to `untarGnuAbsPathWithStrip1Component` which I think makes it a bit clearer what the "1" means

Issue #3845, opened 2024-06-11
Windows & Mac SBOM contains wrong "Build Timestamp" value


andrew-m-leonard:
Due to the 2 pass openjdk make procedure for Signing Windows & Mac binaries, the SBOM "Build Timestamp" is set to the 2nd pass build arg value that is not used, as the binary was built in the 1st pass with the value generated during the 1st pass!
eg.
https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk22u/job/jdk22u-windows-x64-temurin/11/
1st make pass:
```
--with-version-opt=202406081216
```
2nd make pass:
```
--with-version-opt=202406081302
```
Actual JDK version opt used and encoded in "java -version":
```
14:46:48  OpenJDK Runtime Environment Temurin-22.0.1+8-202406081216 (build 22.0.1-beta+8-202406081216)
```
SBOM value:
```
name	"Build Timestamp"
value	"2024-06-08 13:02:30"
```

PR #3846, opened 2024-06-11
Ensure SBOM build timestamp is set to the ACTUAL timestamp compiled from spec.gmk

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/3845

Ensure SBOM Build Timestamp is derived from openjdk make spec.gmk configuration value, so we know it is exactly as is compiled into the binary.

Test build: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk22u/job/jdk22u-windows-x64-temurin/13/
Looks good


sophia-guo (approved):

steelhead31 (approved):
LGTM!

PR #3854, opened 2024-06-17
don't use git -C as it depends on later git version support

andrew-m-leonard:
To fix issue highlighted by https://github.com/adoptium/temurin-build/pull/3824#issuecomment-2168316106
whereby earlier versions of git doe not support "git -C dir".
Change to not use -C to provide better git tooling support.

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -2171,7 +2171,7 @@ addJVMVariant() {
 }
 
 addBuildSHA() { # git SHA of the build repository i.e. openjdk-build
-  local buildSHA=$(git -C "${BUILD_CONFIG[WORKSPACE_DIR]}" rev-parse HEAD 2>/dev/null)
+  local buildSHA=$(cd "${BUILD_CONFIG[WORKSPACE_DIR]}" && git rev-parse HEAD 2>/dev/null)
   if [[ $buildSHA ]]; then
     # shellcheck disable=SC2086
     echo -e BUILD_SOURCE=\"git:$buildSHA\" >>release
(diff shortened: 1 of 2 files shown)


sophia-guo (approved):

andrew-m-leonard:
@sxa for review please

sbin/build.sh
```diff
@@ -2221,7 +2221,7 @@ addJ9Tag() {
   # This code makes sure that a version number is always present in the release file i.e. openj9-0.21.0
   local j9Location="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}/openj9"
   # Pull the tag associated with the J9 commit being used
-  J9_TAG=$(git -C $j9Location describe --abbrev=0)
+  J9_TAG=$(cd "$j9Location describe" && git --abbrev=0)
```
sxa, line 2224: This looks wrong - the `describe` shouldn't be part of the path
```suggestion
  J9_TAG=$(cd "$j9Location" describe && git describe --abbrev=0)
```

sxa (changes requested):
One fix needed, otherwise if the linter is happy and the scripts still run then this looks good.

sbin/build.sh
andrew-m-leonard: @sxa oo, good spot! over zealous there

judovana:
I had spend really a lot fo time to get newer git to my older platfroms.. it is obvious I pushed my energy to wrong direction. Although I have suspicions, that -C is used also in aqa-tests @sophia-guo  do you know?

sxa:
> I had spend really a lot fo time to get newer git to my older platfroms.. it is obvious I pushed my energy to wrong direction. Although I have suspicions, that -C is used also in aqa-tests @sophia-guo do you know?

FYI We had some issues when the test code started to use `reference-if-able` which was new in some version of git that wasn't on the older CentOS systems so we ended up install git 2.15 from source on those platforms via a new playbook role: https://github.com/adoptium/infrastructure/blob/master/ansible/playbooks/AdoptOpenJDK_ITW_Playbook/roles/GIT_Source/tasks/main.yml but that's not being used in the situation that Sophia is testing here, which is on a relatively "clean" image.

git 2.15 seemed relatively easy to build if you need it for some older distribution version - sounds like you ended up with something similar.

Issue #3750, opened 2024-04-05
sbom strace enablement for Alpine linux needs to use apk info --who-owns instead of rpm -qf


andrew-m-leonard:
Alpine linux SBOM strace enablement fails as Alpine does not use rpm

ref: https://github.com/adoptium/infrastructure/issues/3452

Haroon-Khel:
Done by https://github.com/adoptium/temurin-build/pull/3864

PR #3864, opened 2024-06-25
Enable sbom strace for Alpine

Haroon-Khel:
ref https://github.com/adoptium/temurin-build/issues/3750

### DIFF
--- a/tooling/strace_analysis.sh
+++ b/tooling/strace_analysis.sh
@@ -173,9 +173,16 @@ processFiles() {
         filePath="$(readlink -f "$file")"
         non_pkg=false
 
+        package_query="rpm -qf"
+        # Alpine uses apk info --who-owns
+        cat /etc/os-release | grep "Alpine Linux" 2>&1>/dev/null
+        isAlpine=$(echo $?)
+        if [[ "$isAlpine" == "0" ]]; then
+            package_query="apk info --who-owns"
+        fi
         # Attempt to determine rpm pkg
         # shellcheck disable=SC2069
-        if ! rpm -qf "$filePath" 2>&1>/dev/null; then
+        if ! ${package_query} "$filePath" 2>&1>/dev/null; then
             # bin, lib, sbin pkgs may be installed under the root symlink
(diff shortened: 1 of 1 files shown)


tooling/strace_analysis.sh
```diff
@@ -173,9 +173,16 @@ processFiles() {
         filePath="$(readlink -f "$file")"
         non_pkg=false
 
+        package_query="rpm -qf"
+        # Alpine uses apk info --who-owns
+        cat /etc/os-release | grep "Alpine Linux" 2>&1>/dev/null
+        isAlpine=$(echo $?)
+        if [[ "$isAlpine" == "0" ]]; then
```
andrew-m-leonard, line 180: Not sure this will work as $(echo $?) might be run in a different shell,
can we use this instead?
```
if grep "Alpine Linux" /etc/os-release 2>&1>/dev/null; then
```

sxa: `2>&1>/dev/null` won't work. Use `>/dev/null 2>&1` or `>/dev/null 2>/dev/null`

tooling/strace_analysis.sh
```diff
@@ -173,9 +173,14 @@ processFiles() {
         filePath="$(readlink -f "$file")"
         non_pkg=false
 
+        package_query="rpm -qf"
+        # Alpine uses apk info --who-owns
+        if grep "Alpine Linux" /etc/os-release >/dev/null 2>&1; then
+            package_query="apk info --who-owns"
+        fi
```
sxa, line 180:
```suggestion
        # Alpine uses apk info --who-owns
        if grep "Alpine Linux" /etc/os-release >/dev/null 2>&1; then
            package_query="apk info --who-owns"
        else
            package_query="rpm -qf"
        fi
```

sxa (approved):
One suggestion to make the logic clearer. Noting also that neither of these will work on dpkg-based build platforms such as arm32 and riscv64.

andrew-m-leonard (approved):

Issue #3865, opened 2024-06-26
jdk8u win32 failing to build freetype


andrew-m-leonard:
https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x86-32-temurin/478/
From freetype.log : 
```
  c1 : fatal error C1083: Cannot open source file: '..\..\..\src\base\ftfntfmt.c': No such file or directory [C:\workspace\openjdk-build\workspace\libs\freetype\builds\windows\vc2010\freetype.vcxproj]
```

ftfntfmt.c does not exist in freetype VER-2-5-3 that we are trying to build !
I think the problem is appearing due to recent change: https://github.com/adoptium/temurin-build/commit/7c3fe8eb949d33674133b2ccbde6ed642d81517a
although not entirely sure why yet

Suspect it's due to: # Don't apply the patch for OpenJ9 (OpenJ9 doesn't need the patch and, technically, it should only be applied for version 2.8.1).

judovana:
So the guilty is I guess:
```
-  if ... [ "${ARCHITECTURE}" = "x64" ]...
+  if [ ... "${BUILD_CONFIG[OS_ARCHITECTURE]}" = "x86_64" ...]
```

Where indeed, it seems visible on various build combinations I'm running (eg https://github.com/adoptium/temurin-build/issues/3863 ) .  You had warrend me:
>  OS_ARCHITECTURE is the "BUILD arch", whereas ARCHITECTURE is the TARGET arch

So maybe that neeeds different arch?

judovana:
Also note, that the GH actions were running in all cases, and passed.

andrew-m-leonard:
> So the guilty is I guess:
> 
> ```
> -  if ... [ "${ARCHITECTURE}" = "x64" ]...
> +  if [ ... "${BUILD_CONFIG[OS_ARCHITECTURE]}" = "x86_64" ...]
> ```
> 
> Where indeed, it seems visible on various build combinations I'm running (eg #3863 ) . You had warrend me:
> 
> > OS_ARCHITECTURE is the "BUILD arch", whereas ARCHITECTURE is the TARGET arch
> 
> So maybe that neeeds different arch?

@judovana this check on build arch was a dubious way of checking for Freetype version 2.8.1 !

judovana:
Indeed. I'm reading your fix for the issue. Surprsingly complex. Tahnx a lot!

judovana:
[...]

PR #3830, opened 2024-05-30
replaced ${ARCHITECTURE}" by  "${BUILD_CONFIG[OS_ARCHITECTURE]}"

judovana:
This should fix windows issue I see:
```
Certificate was added to keystore
Number of certs processed: 147

real	2m26.134s
user	0m42.929s
sys	0m24.072s
build.sh : 20:48:54 : Initiating build ...
Version: local dir; OPENJDK_BUILD_NUMBER set as
/home/tester/temurinbuild-1717011006/temurin-build/sbin/build.sh: line 193: ARCHITECTURE: unbound variable
+ BUILD_RESULT=1
+ popd
```

I'm just testing it on widows VM. But it seems that the ARCHITETURE is indeed not set. Please eyball carefully, I'm unable to judge the impact

PR #3866, opened 2024-06-26
Update Windows jdk8u freetype version check to correctly update freetype.vcxproj for version 2.8.1

andrew-m-leonard:
Fixes: https://github.com/adoptium/temurin-build/issues/3865

Previous PR: https://github.com/adoptium/temurin-build/pull/3830 changed the ARCHITECTURE check to use the target OS arch rather than the build-farm build arch, at the time it wasn't obvious it was using the fact of the "build arch" being x64 to reason the Freetype was 2.8.1...

The patchFreetypeWindows() should be checking for actual freetype version 2.8.1 rather than assuming the version from the build ARCHITECTURE being built
jdk8u x86-32 uses Freetype 2.5.3
jdk8u x64 uses Freetype 2.8.1

Test builds:
- jdk8u win32 : https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x86-32-temurin/479/consoleFull
- jdk8u win64 : https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x64-temurin/505/consoleFull

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -190,10 +190,28 @@ getOpenJDKUpdateAndBuildVersion() {
 patchFreetypeWindows() {
   # Allow freetype 2.8.1 to be built for JDK8u with Visual Studio 2017 (see https://github.com/openjdk/jdk8u-dev/pull/3#issuecomment-1087677766).
   # Don't apply the patch for OpenJ9 (OpenJ9 doesn't need the patch and, technically, it should only be applied for version 2.8.1).
-  if [ "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" = "${JDK8_CORE_VERSION}" ] && [ "${BUILD_CONFIG[OS_ARCHITECTURE]}" = "x86_64" ] && [ "${BUILD_CONFIG[BUILD_VARIANT]}" != "${BUILD_VARIANT_OPENJ9}" ]; then
-    rm "${BUILD_CONFIG[WORKSPACE_DIR]}/libs/freetype/builds/windows/vc2010/freetype.vcxproj"
-    # Copy the replacement freetype.vcxproj file from the .github directory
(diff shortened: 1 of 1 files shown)


sbin/build.sh
```diff
@@ -190,10 +190,28 @@ getOpenJDKUpdateAndBuildVersion() {
...
+    echo "Checking cloned freetype source version for version 2.8.1, that needs updated builds/windows/vc2010/freetype.vcxproj ..."
+    # Determine cloned freetype version
+    local freetype_version=""
+    # Obtain FreeType version from freetype.h
+    local freetypeInclude="${BUILD_CONFIG[WORKSPACE_DIR]}/libs/freetype/include/freetype/freetype.h"
+    if [[ -f "${freetypeInclude}" ]]; then
+      local ver_major="$(grep "FREETYPE_MAJOR" "${freetypeInclude}" | grep "#define" | tr -s " " | cut -d" " -f3)"
+      local ver_minor="$(grep "FREETYPE_MINOR" "${freetypeInclude}" | grep "#define" | tr -s " " | cut -d" " -f3)"
+      local ver_patch="$(grep "FREETYPE_PATCH" "${freetypeInclude}" | grep "#define" | tr -s " " | cut -d" " -f3)"
+      local freetype_version="${ver_major}.${ver_minor}.${ver_patch}"
```
sxa, line 203: Could be done with a single `awk` comment which feels a little nicer/efficient than the multiple pipes but I'm ok either way
```suggestion
      local freetype_version=$(awk '/^#define *FREETYPE_MAJOR/{MAJOR=$3}/^#define *FREETYPE_MINOR/{MINOR=$3}/^#define *FREETYPE_PATCH/{PATCH=$3}END{print MAJOR "." MINOR "." PATCH}' ${freetypeInclude})
```

sxa (approved):
LGTM although I do wonder if we can do anything to mitigate the need for this patching going forward particularly with us using a very old compiler for this.

Issue #2922, opened 2022-05-12
Investigate changing jdk8u to build with VS2019 for Windows for July 2022 PSU


karianna:
Upstream change has been merged into jdk8u-dev

sxa:
@karianna This has not been implemented. Slightly worryingly the [64-bit JDK8 build logs](https://ci.adoptopenjdk.net/view/Failing%20Temurin%2jobs/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x64-temurin) currently show this:
```
19:08:02  Tools summary:
19:08:02  * Environment:    cygwin version 3.1.7(0.340/5/3) (root at /cygdrive/c/cygwin64)
19:08:02  * Boot JDK:       openjdk version "1.7.0_75"  OpenJDK Runtime Environment (build 1.7.0_75-b13)  OpenJDK Client VM (build 24.75-b04, mixed mode)   (at /cygdrive/c/openjdk/jdk-7)
19:08:02  * Toolchain:      microsoft (Microsoft Visual Studio 2017 - CURRENTLY NOT WORKING)
19:08:02  * C Compiler:     Version 19.16.27043 (at /cygdrive/c/progra~2/mib055~1/2017/commun~1/vc/tools/msvc/1416~1.270/bin/hostx86/x64/cl)
19:08:02  * C++ Compiler:   Version 19.16.27043 (at /cygdrive/c/progra~2/mib055~1/2017/commun~1/vc/tools/msvc/1416~1.270/bin/hostx86/x64/cl)
```
Are you concerned about this build not using 2019 or the "CURRENTLY NOT WORKING" message?

[32-bit JDK8 is still on VS2013:](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x86-32-temurin)
```
19:09:38  Tools summary:
19:09:38  * Environment:    cygwin version 3.1.4(0.340/5/3) (root at /cygdrive/c/cygwin64)
19:09:38  * Boot JDK:       openjdk version "1.7.0_75"  OpenJDK Runtime Environment (build 1.7.0_75-b13)  OpenJDK Client VM (build 24.75-b04, mixed mode)   (at /cygdrive/c/openjdk/jdk7)
19:09:38  * Toolchain:      microsoft (Microsoft Visual Studio 2013)
[...]

PR #3907, opened 2024-08-07
Use VS 2022 by default for all JDK versions + skip freetype patching

sxa:
Updating to use VS2022 for all Temurin builds. This change has been approved by the PMC today, but if it causes any unexpected issues it can be reverted.

freetype patching can (and should be, otherwise it tries to reference things that don't exist) be skipped when we're not building JDK8 with VS2017 otherwise [it tries to reference a file that isn't where it's expecting](https://github.com/adoptium/temurin-build/issues/2922#issuecomment-2273611105).

I'm currently testing with this branch on a machine which only has the VS2022 build tools installed and will report back once it's complete, but it's got to the middle of the build with jdk8u so I'm confident no other issues will show up, and standalone testing before creating this PR looked good.

Alternate solutions:
- I was tempted to just comment out the individual `TOOLCHAIN_VERSION` overrides instead of deleting them for now so that we knew where they all were.
- We could change all of the other versions explicitly to 2022 instead of providing a default at the start but this feels a lot cleaner given that we really do have a "standard" version that we're changing to.

This should leave OpenJ9 unaffected, but I'll tag in @AdamBrousseau too (LMK if you want this to apply to OpenJ9 as well)

Closes https://github.com/adoptium/temurin-build/issues/2922

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -225,7 +225,7 @@ jobs:
        fail-fast: false
        matrix:
          os: [windows-2022]
-         version: [jdk8u, jdk11u, jdk17u, jdk]
+         version: [jdk11u, jdk17u, jdk]
          variant: [temurin]
 
      env:
--- a/build-farm/platform-specific-configurations/windows.sh
+++ b/build-farm/platform-specific-configurations/windows.sh
@@ -113,21 +113,17 @@ then
       export INCLUDE="C:\Program Files\Debugging Tools for Windows (x64)\sdk\inc;$INCLUDE"
       export PATH="$PATH:/c/cygwin64/bin"
       export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
-      TOOLCHAIN_VERSION="2017"
     elif [ "${JAVA_TO_BUILD}" == "${JDK9_VERSION}" ]
     then
-      TOOLCHAIN_VERSION="2013"
       export BUILD_ARGS="${BUILD_ARGS} --freetype-version 2.5.3"
     elif [ "${JAVA_TO_BUILD}" == "${JDK10_VERSION}" ]
     then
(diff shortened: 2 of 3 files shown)


AdamBrousseau:
I believe Semeru has switched over to VS2022 for all jdk versions for some time now. We should do the same for the OpenJ9 builds at Adoptium.

sxa:
> I believe Semeru has switched over to VS2022 for all jdk versions for some time now. We should do the same for the OpenJ9 builds at Adoptium.

Thanks - PR updated. That makes it simpler :-)

sxa:
PR has failed the JDK8 check but it's attempting to build with compiler version 19.40.33807 instead of the 19.37.32822 which we use which I believe is the cause of the check failure and would impact running on our machines.

sxa:
Removing jdk8u/windows test for now (Ideally I'd make it conditionally set `CONFIGURE_ARGS=--with-toolchain=2019` to let it build on an older version for now, but I haven't got the syntax worked out yet)

gdams (approved):
LGTM (although as discussed in Slack, I'd like to switch over to using the docker image as a base once we've got one published)

sxa:
jdk8u test removed for now as the default win2022 image in GitHub Actions has a pre-installed VS2022 compiler which is newer than the one we want and fails to compiler jdk8u.

AdamBrousseau:
Let us know when you have a Windows build docker image. I'd be interested in using it too.

sxa:
> Let us know when you have a Windows build docker image. I'd be interested in using it too.

@AdamBrousseau There is a functional dockerfile for this in https://github.com/adoptium/infrastructure/pull/3702/files :-)

steelhead31 (approved):
Looks good!

karianna:
@sxa safe to merge?

sxa:
> @sxa safe to merge?

Yep I think so now - was leaving it a bit to ensure that anyone who wanted to comment could do so.

PR #3908, opened 2024-08-09
Use VS 2022 by default for all JDK versions + skip freetype patching

andrew-m-leonard:
Update to https://github.com/adoptium/temurin-build/pull/3907 to change github build actions to use VS2022 BuildTools


sxa:
I'm going to close this as the experiments being attempted in this PR were so far unsuccessful and we should merge https://github.com/adoptium/temurin-build/pull/3907 instead

PR #3923, opened 2024-08-22
only specify --openjdk-target when cross compiling

gdams:
This will partially fix https://github.com/adoptium/adoptium-support/issues/937 for macOS aarch64

### DIFF
--- a/build-farm/platform-specific-configurations/mac.sh
+++ b/build-farm/platform-specific-configurations/mac.sh
@@ -70,7 +70,7 @@ else
     if [ "${ARCHITECTURE}" == "x64" ]; then
       # We can only target 10.9 on intel macs
       export cxx_flags_bucket="${cxx_flags_bucket} -mmacosx-version-min=10.9"
-    elif [ "${ARCHITECTURE}" == "aarch64" ]; then
+    elif [[ "${MACHINEARCHITECTURE}" == "x64" ]] && [[ "${ARCHITECTURE}" == "aarch64" ]]; then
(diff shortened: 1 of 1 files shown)


gdams:
Actually this doesn't work as it tries to compile x64 on macOS regardless

sxa (approved):
I guess this was added in for cross-compiling for aarch64 on x64 hosts, which we're not doing so it makes sense to remove it in the aarch64->aarch64 case.
On that bases, I approve 👍🏻 Thanks for tracking it down.

karianna (approved):

build-farm/platform-specific-configurations/mac.sh
```diff
@@ -70,7 +70,7 @@ else
     if [ "${ARCHITECTURE}" == "x64" ]; then
       # We can only target 10.9 on intel macs
       export cxx_flags_bucket="${cxx_flags_bucket} -mmacosx-version-min=10.9"
-    elif [ "${ARCHITECTURE}" == "aarch64" ]; then
+    elif [[ "${MACHINEARCHITECTURE}" == "x64" ]] && [[ "${ARCHITECTURE}" == "aarch64" ]]; then
```
karianna, line 73: Will this elif trigger given the if above?

karianna (changes requested):

build-farm/platform-specific-configurations/mac.sh
gdams: yes it will trigger on an x64 machine that is cross-compiling aarch64

karianna: Ugh, sorry missed the variable name being different

karianna (approved):

PR #3924, opened 2024-08-22
Add execute permission to macos_repro_build_compare.sh

sophia-guo:
Update mac script permission same as all other shell script files.


sxa (commented):
LGTM but can you update the commit message so it's a bit clearer in the history please e.g.
`Add execute permission to macos_repro_build_compare.sh`

karianna (approved):

PR #3935, opened 2024-09-03
Update to openssl 3.0.15

keithc-ca:
See https://github.com/eclipse-openj9/openj9/pull/20098.

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -364,7 +364,7 @@ updateOpenj9Sources() {
   if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_OPENJ9}" ]; then
     cd "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}" || return
     # NOTE: fetched openssl will NOT be used in the RISC-V cross-compile situation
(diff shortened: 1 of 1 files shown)


sbin/prepareWorkspace.sh
```diff
@@ -364,7 +364,7 @@ updateOpenj9Sources() {
   if [ "${BUILD_CONFIG[BUILD_VARIANT]}" == "${BUILD_VARIANT_OPENJ9}" ]; then
     cd "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}" || return
     # NOTE: fetched openssl will NOT be used in the RISC-V cross-compile situation
-    bash get_source.sh --openssl-version=3.0.14
+    bash get_source.sh -openssl-branch=openssl-3.0.15
```
karianna, line 367: why the change to go after the branch?

keithc-ca: Option handling was unified in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/781 (and back-ports). We thought all uses of non-standard options (e.g. `--openssl-version`) had been removed so support for them was removed by https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/784. It turns out that places like this still existed so they were temporarily restored by https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/786. We would eventually like to remove the non-standard options (again): this is a step in that direction.

andrew-m-leonard (approved):

Issue #3970, opened 2024-10-04
Upgrade temurin-build/cyclonedx-lib build to use cyclonedx-core-java latest "1.6 spec" release


andrew-m-leonard:
To support CycloneDX CDXA document we need to move to the 1.6 spec
Latest cyclonedx-core-java release: https://github.com/CycloneDX/cyclonedx-core-java/releases/tag/cyclonedx-core-java-9.0.5

Requires dependency job issue fix: https://github.com/adoptium/temurin-build/issues/3969

PR #3985, opened 2024-10-10
Upgrade TemurinGenSBOM to use CycloneDX v9.0.5 spec 1.0.6

andrew-m-leonard:
- Upgrade to CycloneDX v9.0.5 spec 1.0.6 to support the new CycloneDX Attestations (CDXA) documents
- Resolve issue with SBOM dependency jar versioning with the https://ci.adoptium.net/view/all/job/build.getDependency/ job
- Allow local dependency cache dir, to allow jars to be pre-installed by ansible.... Added new --local-dependency-cache-dir build arg

Fixes https://github.com/adoptium/temurin-build/issues/3970


sbin/build.sh
```diff
@@ -887,23 +887,30 @@ buildCyclonedxLib() {
   else
     ANTBUILDFILE="${CYCLONEDB_DIR}/build.xml"
   fi
+
+  # Do we have a local cache for the dependency jars?
+  local localJarCacheOption=""
```
sxa, line 892: Can we use a default value for this since it's likely to be in a common location on all machines, and it would be good to use the cached version automatically if it is available instead of requiring it to use a build option?

sbin/build.sh
```diff
@@ -1211,21 +1218,20 @@ addCycloneDXVersions() {
...
        for JAR in "${CYCLONEDB_DIR}/build/jar"/*.jar; do
-         JarName=$(basename "$JAR")
+         JarName=$(basename "$JAR" | cut -d'.' -f1)
          if [ "$(uname)" = "Darwin" ]; then
             JarSha=$(shasum -a 256 "$JAR" | cut -d' ' -f1)
          else
             JarSha=$(sha256sum "$JAR" | cut -d' ' -f1)
          fi
-         addSBOMFormulationComponentProperty "${javaHome}" "${classpath}" "${sbomJson}" "CycloneDX" "CycloneDX jar SHAs" "${JarName}" "${JarSha}"
+         addSBOMFormulationComponentProperty "${javaHome}" "${classpath}" "${sbomJson}" "CycloneDX" "CycloneDX jar SHAs" "${JarName}.jar" "${JarSha}"
```
sxa, line 1227: What's the reason for stripping off the `.jar` extension a few lines above then adding it back in?

sbin/build.sh
andrew-m-leonard: @sxa bearing in mind "users" can use this script, do you see it a reasonable default for users in general?

sbin/build.sh
andrew-m-leonard: @sxa  because line 1230 index into the dependency_data.properties needs just the component name, so i'd need to strip it for that...

sbin/build.sh
andrew-m-leonard: also what should the default be, and i'd need to add platform specific differences here..?

andrew-m-leonard: my assumption was we would specify the adoptium ci defaults in the ci-jenkins-pipelines job configs

adamfarley: If we're not specifying a default location within the script itself, I advise that we add error checking for a blank value.

Example below.

sbin/build.sh
```diff
@@ -887,23 +887,30 @@ buildCyclonedxLib() {
   else
     ANTBUILDFILE="${CYCLONEDB_DIR}/build.xml"
   fi
+
+  # Do we have a local cache for the dependency jars?
+  local localJarCacheOption=""
+  if [[ -n "${BUILD_CONFIG[LOCAL_DEPENDENCY_CACHE_DIR]}" ]]; then
+    localJarCacheOption="-Dlocal.deps.cache.dir=${BUILD_CONFIG[LOCAL_DEPENDENCY_CACHE_DIR]}"
+  fi
```
adamfarley, line 895:
```suggestion
  else
    echo "ERROR: LOCAL_DEPENDENCY_CACHE_DIR was not set in job parameters, and is required."
    exit 1
  fi
```

adamfarley: LGTM pending this one requested change

andrew-m-leonard: the option is not mandatory

adamfarley (approved):
LGTM

sbin/build.sh
sxa: The defaults should be consistent with wherever we're going to install them in the playbooks.

So yes, potentially different for each platform.

The fact that users can use this is precisely why we should have a default - so the users can use the defaults on a system configured with the playbooks, such as the simplest option where they are using it container images.

andrew-m-leonard: I've added some defaults, we can correct them when we get to the infra issue if they don't match.

andrew-m-leonard:
Raised infrastructure issue for ansible update to download local cache of jars: https://github.com/adoptium/infrastructure/issues/3806

sbin/build.sh
sxa: SGTM. I definitely wouldn't want it to be under $HOME since that will typically be as location writable by the build processes, but I'm ok with merging this for now and changing later since we have these placeholders now 😀

sxa (approved):

PR #3992, opened 2024-10-11
Remove hyphen from bootjdk symlinks

Haroon-Khel:
Ref https://github.com/adoptium/infrastructure/issues/2912 and the infra pr https://github.com/adoptium/infrastructure/pull/3773

The pr removes the hyphen from the bootjdk symlinks, ie /usr/lib/jvm/jdk-11 becomes /usr/lib/jvm/jdk11. The change has been made to the linux, mac and windows playbooks, so the build scripts need to be updated to be able to detect the new symlinks.

Leaving this as a draft so it doesnt get merged before or during the release

### DIFF
--- a/build-farm/platform-specific-configurations/alpine-linux.sh
+++ b/build-farm/platform-specific-configurations/alpine-linux.sh
@@ -49,7 +49,7 @@ export GNUPGHOME
 
 BOOT_JDK_VARIABLE="JDK${JDK_BOOT_VERSION}_BOOT_DIR"
 if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
-  bootDir="$PWD/jdk-$JDK_BOOT_VERSION"
+  bootDir="$PWD/jdk$JDK_BOOT_VERSION"
   # Note we export $BOOT_JDK_VARIABLE (i.e. JDKXX_BOOT_DIR) here
   # instead of BOOT_JDK_VARIABLE (no '$').
   export "${BOOT_JDK_VARIABLE}"="$bootDir"
--- a/build-farm/platform-specific-configurations/windows.sh
+++ b/build-farm/platform-specific-configurations/windows.sh
@@ -33,16 +33,16 @@ if [ "$ARCHITECTURE" == "aarch64" ]; then
 fi
 BOOT_JDK_VARIABLE="JDK${JDK_BOOT_VERSION}_BOOT_DIR"
 if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
-  bootDir="$PWD/jdk$JDK_BOOT_VERSION"
+  bootDir="$PWD/jdk-$JDK_BOOT_VERSION"
   # Note we export $BOOT_JDK_VARIABLE (i.e. JDKXX_BOOT_DIR) here
   # instead of BOOT_JDK_VARIABLE (no '$').
   export "${BOOT_JDK_VARIABLE}"="$bootDir"
   if [ ! -x "$bootDir/bin/javac.exe" ]; then
     # Set to a default location as linked in the ansible playbooks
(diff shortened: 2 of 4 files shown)


gdams (approved):

Haroon-Khel:
Without the hyphens in the symlinks the upstream configure is able to find the bootjdks without a problem, see https://github.com/adoptium/infrastructure/issues/2912#issuecomment-2435792702

build-farm/platform-specific-configurations/alpine-linux.sh
```diff
@@ -49,7 +49,7 @@ export GNUPGHOME
 
 BOOT_JDK_VARIABLE="JDK${JDK_BOOT_VERSION}_BOOT_DIR"
 if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
-  bootDir="$PWD/jdk-$JDK_BOOT_VERSION"
```
sxa, line 52: Just noting that for these versions it doesn't matter whether they have the `-` or not as it's a temporary directory used to download the build anyway, but no harm in it being consistent with the "default` top level path names on the machines that we will be creating.

sxa (approved):
LGTM. Noting that make-adopt-build-farm.sh passes `JDK_BOOT_DIR` into makejdk-any-platform:

https://github.com/adoptium/temurin-build/blame/62ecfbaeb60cbe5f4072bd81b8d4a3546625dcb2/build-farm/make-adopt-build-farm.sh#L253

And https://github.com/adoptium/temurin-build/pull/393 was added to cover a lot of the `JDKXX_BOOT_DIR` variables but there is potential for stripping a lot of that code our in the ultimate goal of simplification later.

FYI @AdamBrousseau - this will probably go live soon along with the correspinding infrastructure PR. you were ok with this in https://github.com/adoptium/infrastructure/issues/2912#issuecomment-1412147028 but since that was a while ago now I'm letting you know in case you have any last minute objections :-)

karianna:
@Haroon-Khel - I know there's another PR related to this, will let you merge this if that's safe to do so independantly

Haroon-Khel:
To reiterate my comment here https://github.com/adoptium/infrastructure/pull/3773#issuecomment-2457381477

Im going to hold off on the windows changes. It looks like the upstream configure looks in [C:\Program Files\Java//cygdrive/c/Program Files/Java](https://github.com/openjdk/jdk/blob/c33a8f52b613e5eff02f572eda876cbbfc7c22cf/make/autoconf/boot-jdk.m4#L273) which is where we already unpack the JDK on our machines
https://github.com/adoptium/infrastructure/blob/822366e945b4ef1f540aca9a29180f7999cf6026/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Java_install/tasks/main.yml#L37 and https://github.com/adoptium/infrastructure/blob/822366e945b4ef1f540aca9a29180f7999cf6026/ansible/playbooks/AdoptOpenJDK_Windows_Playbook/roles/Java8/tasks/main.yml#L27

build-farm/platform-specific-configurations/windows.sh
```diff
@@ -33,16 +33,16 @@ if [ "$ARCHITECTURE" == "aarch64" ]; then
 fi
 BOOT_JDK_VARIABLE="JDK${JDK_BOOT_VERSION}_BOOT_DIR"
 if [ ! -d "$(eval echo "\$$BOOT_JDK_VARIABLE")" ]; then
-  bootDir="$PWD/jdk$JDK_BOOT_VERSION"
+  bootDir="$PWD/jdk-$JDK_BOOT_VERSION"
```
karianna, line 36: PR says to remove hyphen but this change is adding?

Haroon-Khel: Ive reversed the windows changes, see my comment here https://github.com/adoptium/temurin-build/pull/3992#issuecomment-2457395642

PR #3995, opened 2024-10-14
adapting comparable_patch.sh to new signature of removeExcludedFiles

judovana:
https://github.com/adoptium/temurin-build/commit/08621c72262ba260e4bb6451a9c75bf1c0ab365d#diff-65e4993f80b74f0a4078e853b798cbf6770215b4c31ae1377aa2704eec01ae16R254

### DIFF
--- a/tooling/reproducible/comparable_patch.sh
+++ b/tooling/reproducible/comparable_patch.sh
@@ -338,7 +338,8 @@ echo "Successfully removed all Signatures from ${JDK_DIR}"
 removeExcludedFiles
 
 # Needed due to vendor variation in jmod re-packing after signing, putting attributes in different order
-processModuleInfo
+# Comparable patch, as per read-me, requires java on path
+processModuleInfo "${JDK_DIR}" "${OS}" "$(dirname "$(dirname "$(readlink -f  "$(which java)")")")"
 
 # Patch Windows VS_VERSION_INFO[COMPANY_NAME]
 if [[ "$OS" =~ CYGWIN* ]] && [[ "$PATCH_VS_VERSION_INFO" = true ]]; then


judovana:
Thanx, I had found only one and half of small issue. hth

tooling/reproducible/comparable_patch.sh
```diff
@@ -338,7 +338,8 @@ echo "Successfully removed all Signatures from ${JDK_DIR}"
 removeExcludedFiles
 
 # Needed due to vendor variation in jmod re-packing after signing, putting attributes in different order
-processModuleInfo
+# Comparable patch, as per read-me, requires java on path
+processModuleInfo "${JDK_DIR}" "${OS}" "$(dirname "$(dirname "$(readlink -f  "$(which java)")")")"
```
andrew-m-leonard, line 342: Using "which java" maybe ok, but wondering if we ought to take a "backup" at the start of the JDK being compared, thus ensuring the class_version is good enough for the "javap" processing. This would be similar as was done here: https://github.com/adoptium/temurin-build/blob/5035fa3b6bfc1787c94d7b1e91e2979619b9e02a/tooling/reproducible/repro_compare.sh#L24

andrew-m-leonard (changes requested):

tooling/reproducible/comparable_patch.sh
judovana: To make backup of jdk ok. My redundant top level wrapper and also the (probably) existing one are doing backup of jdk/allowing to set up 3rd jdk . So that is ok. Unluckily I have not seen the  way to pass it into the comparable_patch.sh properly.

Comparable_patch.sh is clear, it needs  java/javac on path.  Whoever modifed processModuleInfo, have violated  that. 

Comparable patch should remain as it is, and should not be doing nay backups of jdk. Thats the task for its wrappers to deal with it as it is appropriate. 

There can be follow up on this set of patches, which will elaborate on usage of path by comaprable_patch, to add additional solution, but right now the  comaprable_patch.sh is broken, and this path  is fixing it.

karianna (approved):

andrew-m-leonard (approved):
yep lets get this in as is, thanks

karianna:
rerunning the failed job

Issue #3946, opened 2024-09-23
Implement a new Sign-JSON type job for secure signing of SBOM json using an EF PEM/signService


andrew-m-leonard:
EPIC https://github.com/adoptium/ci-jenkins-pipelines/issues/610 documents the parts required to implement SBOM json signing.
This issue is to implement the necessary temurin-build/ci-jenkins-pipelines parts to interface with an EF provided signing PEM or "json signing service".

Haroon-Khel:
Work that needs to be done:

https://ci.adoptium.net/job/build-scripts/job/release/job/sign_temurin_jsf/ has been created to replicate https://ci.adoptium.net/job/build-scripts/job/release/job/sign_temurin_gpg/ but to use the PEM format. The sign_temurin_jsf job needs to be updated to use https://github.com/adoptium/temurin-build/blob/master/cyclonedx-lib/sign_src/TemurinSignSBOM.java to sign the SBOM file

Because we dont want the jenkins worker node to build any of the cyclone dx jars, this needs to run in the build scripts during the jdk build, https://github.com/adoptium/temurin-build/blob/62ecfbaeb60cbe5f4072bd81b8d4a3546625dcb2/cyclonedx-lib/build.xml#L68, and then archived, and then pulled onto the jenkins worker node during the sign_temurin_jsf job

Haroon-Khel:
@sxa It looks like the SBOM signing jars can only be built with JDK17+
https://github.com/adoptium/temurin-build/actions/runs/11597199261/job/32290097938?pr=4017

This would make it difficult to build the jars on a build node for platforms like Solaris for example. It should be discussed the feasibility of building the jars on the eclipse signing node

andrew-m-leonard:
> Andrew's idea of [cacheing the cyclone dx](https://github.com/adoptium/infrastructure/issues/3806) dependency jars can be done on the eclipse signing node. Instead of them being downloaded during the ant build step

From chatting with Stewart and Thomas, ideally we don't want to build on the signing node if we can help it. We could add in a "build-sbom-jars-and-sign" job, which builds on any suitable node, and is then used as upstream to the "sign_temurin_jsf" job.

Haroon-Khel:
[...]

PR #4017, opened 2024-10-30
Build build-sign-sbom ant target

Haroon-Khel:
ref https://github.com/adoptium/temurin-build/issues/3946

Adding the `build-sign-sbom` target to the `build` target ensures that the `temurin-sign-sbom.jar` gets built during https://github.com/adoptium/temurin-build/blob/62ecfbaeb60cbe5f4072bd81b8d4a3546625dcb2/sbin/build.sh#L885 when the other cyclone dx libraries get built

### DIFF
--- a/cyclonedx-lib/build.xml
+++ b/cyclonedx-lib/build.xml
@@ -119,7 +119,7 @@
                 <download-file-with-default-options destfile="github-package-url.jar"/>
         </target>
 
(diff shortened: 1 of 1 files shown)


Haroon-Khel:
The java its using with ant needs to be jdk17 or higher to build the `temurin-sign-sbom.jar`

Haroon-Khel:
Building the SBOM libraries is done in this jenkins job, https://ci.adoptium.net/job/build_sign_sbom_libraries/. Doesnt need to be done in the build scripts

Issue #3860, opened 2024-06-20
Overhaul README.md in this repository


sxa:
There have been very few updates to the top level README.md in the temurin-build and ci-jenkins-pipelines repositories, and so we should look at them and see if they are "fit for purpose" and whether the information on here that we present to our user base is (a) useful (b) accurate and (c) fit for purpose.

I've started with an overhaul of the ci-jenkins-pipelines README.md in https://github.com/adoptium/ci-jenkins-pipelines/pull/1048 but here are some discussion points for the one in here:

- We have multiple ways of setting up a build environment - what do we want to recommend and promote?
  - Run the playbooks on your local machine
  - List the prereqs that are required (Maybe complex as it's different on each platform?)
  - Use the `-D` option on makejdk-any-platform.sh to automatically build in a new container image to avoid discouraging podman users (Also, why do we have this section yet also reference the dockerfile generator in the "build natively" section? Is this likely to be confusing to first time users?) NOTE: This is converted to `-D` in https://github.com/adoptium/temurin-build/pull/3796/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5
  - Duplicating the "man page" for makejdk-any-platform.sh in the README.md seems unnecessary - we should only have this information in one place ... But which location is right? Is the "man page" format still useful?
  - Is the metadata section useful? Should it be in this repository's README or the ci-jenkins-pipelines one? My gut feel is that it should probably be here, but maybe collapsed (as I did in my pipelines PR for now) or moved elsewhere as it's perhaps not "interesting" particularly now that we have the SBoM generation.
  - We should perhaps shout more about SBoM and reproducibility options in the README.md
  - Should SmokeTesting.md and Testing.md be moved somewhere else instead of the top level directory? Maybe Testing.md should be part of CONTRIBUTING.md and the smoke test doc to `test/functional/buildAndPackage/README.md`?

judovana:
One friendly note. The man page is good. The best of its existence  is the usage during --help for build: https://github.com/adoptium/temurin-build/blob/8646220187f9e64087249085e519a9d3d18ae97e/sbin/common/config_init.sh#L320  . Maybe the `.1` file can be linked  to its generated html version?

sxa:
>  Maybe the .1 file can be linked to its generated html version?

[...]

PR #4019, opened 2024-10-31
Document missing disable-adopt-branch-safety

judovana:


karianna (approved):

makejdk-any-platform.1
```diff
@@ -106,6 +106,9 @@ It is known issue: https://github.com/adoptium/temurin-build/issues/3862
 .BR \-\-debug-docker
 debug OpenJDK build script in a docker container. Only valid if \fB-D\fR is selected.
 .TP
+.BR \-\-disable-adopt-branch-safety
+disable the default check for tags/branches in case of \fBtemurin\fR as --build-variant.
```
karianna, line 110: What's the check?  More context will help

judovana: Tbh I'm not sute. i think itis checking whether you are in correct repo, in correcy branch with correct tags.I will try to diga bot more

karianna: Yes and the comment that it is checking if you are in the correct tag is good context to add.

karianna (changes requested):

makejdk-any-platform.1
judovana: @karianna  ok. In that case I have to verify that I understood it correctly. Tahnk you!

andrew-m-leonard: The more precise description is actually not related to "tags" but the detection of the TEMURIN_MARKER_FILE, which is the file README.JAVASE, which is only present in the Adoptium "mirror" : https://github.com/adoptium/jdk21u/blob/dev/README.JAVASE. This check is to ensure Temurin is building from the Adoptium mirror source, and not something else..., hence the use of the word "safety".
The "fake tag" bit is just an additional bit to avoid the missing _adopt tag associated with not using an Adoptium mirror.

judovana: Thanx a lo, updated.

karianna (approved):

judovana:
@andrew-m-leonard  ping please?

sxa (approved):

sxa:
Noting that there likely needs to be a load of checking of everything and deduplication (`.1` page and `README.md`) as part of https://github.com/adoptium/temurin-build/issues/3860

judovana:
thanx. I have completely forgot about readme.md ad the #3860

Issue #4021, opened 2024-11-01
Rebuild_Same_JDK_Reproducibility_Test_win fails with Redist devkit


andrew-m-leonard:
Change to using Windows Redist devkit broke reproducible test, as it is trying to verify the local machine UCRT dlls match the original build, when it won't because they are dynamically fetched from devkit-binaries during the build..

```
09:42:01  REQ UCRT = 10.0.26100.1 (WinBuild.160101.0800)
09:42:01  Check For UCRT DLL In : C:/Program Files (x86)/Windows Kits/10/Redist
09:42:03  ERROR - A Version Of ucrtbase.dll matching 10.0.26100.1 (WinBuild.160101.0800)
09:42:03   was not found on this system - Exiting
09:42:03  -----------------------------------
09:42:03  Rebuild_Same_JDK_Reproducibility_Test_win_0_FAILED
```

Simply don't check local system if --use-adoptium-devkit used.

PR #4022, opened 2024-11-01
Update windows_repro_build_compare.sh to allow for windows Redist devkit

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/4021

Don't check local Windows system UCRT if --use-adoptium-devkit used, as the correct versions will be downloaded.

Grinder test: https://ci.adoptium.net/job/Grinder/11287/

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -79,9 +79,10 @@ configureDevKitConfigureParameter() {
     if [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]]; then
       # Windows DevKit, currently only Redist DLLs
 
+
       # Default to build architecture unless target ARCHITECTURE variable is set
       def target_arch="${BUILD_CONFIG[OS_ARCHITECTURE]}"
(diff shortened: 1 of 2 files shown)


smlambert (approved):

sophia-guo:
This situation also happens to linux https://github.com/adoptium/temurin-build/blob/master/tooling/reproducible/linux_repro_build_compare.sh#L166, which suggests another opportunity for code refactoring and consolidating for different platforms. 

Creating a reusable function or module will make the logic easier to understand, test and maintain.

andrew-m-leonard:
> This situation also happens to linux https://github.com/adoptium/temurin-build/blob/master/tooling/reproducible/linux_repro_build_compare.sh#L166, which suggests another opportunity for code refactoring and consolidating for different platforms.
> 
> Creating a reusable function or module will make the logic easier to understand, test and maintain.

It is the same yes, although the apart from the check for the --use-adoptium-flag, the GCC vs VisualStudio logic is different

adamfarley (approved):
LGTM

PR #4024, opened 2024-11-04
Fix check if webpki library is already available

netomi:
This fixes the availability check for the webpki library.

It was checking the wrong path (build/webpki....) instead of (build/jar/webpki....) so when running repeatedly the target build-sign-sbom, the library was cloned and rebuilt everytime.


karianna (approved):

sxa (approved):
I've resolved the merge conflict - approving

PR #4039, opened 2024-11-14
Add --enable-linkable-runtime for JDK 24+

### LINKED ISSUE #4035: [JDK 24] Enable JEP 493 in Temurin 24 builds
Description
-----------------
[JEP 493](https://openjdk.org/jeps/493) has this summary:

```
Reduce the size of the JDK by approximately 25% by enabling the jlink tool
to create custom run-time images without using the JDK's JMOD files. This
feature must be enabled when the JDK is built; it will not be enabled by
default, and some JDK vendors may choose not to enable it.
```

Proposal
-----------------
Add `--enable-linkable-runtime` to Temurin 24 builds so as to enable this feature. It only affects users running `jlink` so as to produce custom runtimes. The nice property would be that the Temurin 24 JDK would ship without `jmods` directory by default. In order to cover cases where a link from the run-time image is not suitable I suggest to also package the jmods image as a separate artefact (like `testimage`, `static-libs` and friends) and make those available for each build via the API for users actually needing them.

Pros
-----------------
By default, the Temurin 24 JDK download shrinks in size without any restrictions in terms of run-time needs. `jlink` users should be able to use it to link custom runtimes.

Cons
-----------------
Certain use-cases might not be possible. For example, cross linking, linking from a modified runtime. See [JEP 493](https://openjdk.org/jeps/493) for details. For those users there would be the option to download the `jmods` artefact separately and get the same features as JDK 23+ had.

jerboaa:
For JDK's including JEP 493, enable the build option for producing a `jlink` that is capable of linking from the run-time image. A build with it no longer contains the `jmods` folder. Thus, the size savings. See below for some numbers.

Closes #4035

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -174,6 +174,17 @@ configureMacOSCodesignParameter() {
   fi
 }
 
+# JDK 24+ includes JEP 493 which allows for the JDK to enable
+# linking from the run-time image (instead of only from JMODs). Enable
+# this option. This has the effect, that no 'jmods' directory will be
+# produced in the resulting build. Thus, the tarball and, especially the
+# extracted tarball will be smaller in terms of disk space size.
+configureLinkableRuntimeParameter() {
+  if [[ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -ge 24 ]]; then
+    addConfigureArg "--enable-linkable-runtime" ""
+  fi
+}
+
 # Get the OpenJDK update version and build version
 getOpenJDKUpdateAndBuildVersion() {
   cd "${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}"
@@ -574,6 +585,7 @@ configureCommandParameters() {
   configureVersionStringParameter
   configureBootJDKConfigureParameter
   configureDevKitConfigureParameter
+  configureLinkableRuntimeParameter
   configureShenandoahBuildParameter
   configureMacOSCodesignParameter
   configureDebugParameters


jerboaa:
Test JDK 24 pipeline with this is:
https://ci.adoptium.net/blue/organizations/jenkins/build-scripts%2Fopenjdk24-pipeline/detail/openjdk24-pipeline/36/pipeline/121

A few numbers for the compressed JDK tarball/zip - according to Jenkins:

|Build | Before  | After |
|--------| ------------- | ------------- |
|Linux x64 (jdk)| 205.54 MiB | 133.03 MiB |
|Windows x64 (jdk) |205.00 MiB|133.40 MiB |
|MacOSX aarch64 (jdk)|199.04 MiB|129.29 MiB|
|AIX ppc64 (jdk)|185.14 MiB|110.57 MiB|

karianna:
I know a bunch of folks use Temurin + OpenJFX as a combination.  IIRC Gluon does publish jmods so I think we should be good there but it would be good to make a big push to get folks to test this.

karianna (approved):

jerboaa:
> I know a bunch of folks use Temurin + OpenJFX as a combination.

OK. It's not clear how this change affects that though. Could you clarify?

sxa (approved):

jerboaa:
> We'll need to raise a PR to ci-jenkins-pipelines to prevent building an exploded image and signing the jmods:
> 
> https://github.com/adoptium/ci-jenkins-pipelines/blob/5bd79eb1d95a033c4ee364a8f9fcc270ad653178/pipelines/build/common/openjdk_build_pipeline.groovy#L1878-L1913

OK.

jerboaa:
> We'll need to raise a PR to ci-jenkins-pipelines to prevent building an exploded image and signing the jmods:
> 
> https://github.com/adoptium/ci-jenkins-pipelines/blob/5bd79eb1d95a033c4ee364a8f9fcc270ad653178/pipelines/build/common/openjdk_build_pipeline.groovy#L1878-L1913

https://github.com/adoptium/ci-jenkins-pipelines/pull/1145 should do that I think.

PR #4072, opened 2024-12-02
Use standard sed syntax for translating filenames

sxa:
Solaris's `sed` doesn't support the `-r` parameter. Since I believe this was only required for the OR section on the extension I've split it out into two separeate `-e` parameters for sed. One for the `.tar.gz` case and the other for `.zip`. I've also escaped the brackets.

To test:
```
X=jdk8u-hotspot.tar.gz
echo $X | sed -e "s/\(.*\)\(\.tar\.gz\)/\1-$component\2/" -e "s/\(.*\)\(\.zip\)/\1-${component}\2/"
```
Noting that even on Linux the original version was likely not working:
```
X=jdk8u-hotspot.tar.gz
echo $X | sed -e "s/(.*)(\.tar.gz)/\1-component\2/"   
sed: -e expression #1, char 32: invalid reference \2 on `s' command's RHS
echo $X | sed -e "s/\(.*\)\(\.tar.gz\)/\1-component\2/"   
jdk8u-hotspot-component.tar.gz
```
Testing at https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-x64-temurin-simple/17/

Identified as part of https://github.com/adoptium/infrastructure/issues/3742


netomi (approved):
The mentioned reasoning is correct, the '|' operator is not available in basic regexp only extended. The workaround works well imho after doing a test.

sbin/build.sh
```diff
@@ -2094,7 +2094,9 @@ getTargetFileNameForComponent() {
     echo "${target_file_name}" | sed "s/-jdk/-${component}/"
   else
     # if no pattern is found, append the component name right before the extension.
-    echo "${target_file_name}" | sed -r "s/(.+)(\.tar\.gz|\.zip)/\1-${component}\2/"
+    echo "${target_file_name}" | sed \
```
karianna, line 2097: Worth adding a comment about using Solaris friendly sed?  I've seen a few cases where this sort of thing gets reverted by Linux focused folks :-)

karianna (approved):

sbin/build.sh
judovana: Isn't shell check able to to check posix? But I guess that is syntax only and not sed/grep and others...

sxa: > Worth adding a comment about using Solaris friendly sed? I've seen a few cases where this sort of thing gets reverted by Linux focused folks :-)

I have a view on that ... If we think we really need to tell people explictly "Don't change this to a non-standard syntax" I can add it (such things aren't just for Solaris)... however I feel adding such comments everywhere would be a job for life ;-)

> Isn't shell check able to to check posix? But I guess that is syntax only and not sed/grep and others...

I suspect so yeah. There have also been cases where we have used `if` blocks to explicitly check for and use non-standard versions, so they'd have to be skipped.

Issue #4074, opened 2024-12-04
Feature Proposal: A feature that prevents the build scripts from behaving insecurely


adamfarley:
**Summary**

An argument or flag that causes the build scripts to deliberately disable functionality considered to potentially risk build security.

**Purpose of this Issue**

1) To establish the mechanism for triggering the inhibition of said functionality.
2) To unify file downloads (in the build process) through a single part of the code base.
3) To make 1 inhibit 2.
4) To raise a separate issue to discuss further ways we could expand this effects of 1.

Issue #3741, opened 2024-04-01
introduce bash unit-tests


judovana:
hello! 
Was there any attempt to include tests at least to temurin-builds?
Looking to https://github.com/adoptium/temurin-build/pull/3736 or https://github.com/adoptium/temurin-build/pull/3728 are excellent candidates... And I belive in past thre were many simialar pathces introducing standalone single function.

I would vote to use my ancient and favourite:
  https://github.com/rh-openjdk/run-folder-as-tests
as it offers ultimate freedom, to to run individual sets, driven by:
  https://github.com/lehmannro/assert.sh
  https://github.com/kward/shunit2

wdyt?

karianna:
GPT-4 says:

There isn't a single de facto standard for unit testing shell scripts as the choice often depends on the specific needs of a project, the complexity of the shell scripts, and the preferences of the development team. However, several popular frameworks and tools are commonly used in the industry. Here are a few notable ones:

1. **shUnit2**: An xUnit based unit test framework for Bourne based shell scripts, `shUnit2` is one of the older and more established frameworks. It provides a straightforward way to write and run tests, making it a popular choice for many developers.

2. **Bats (Bash Automated Testing System)**: `Bats` is a TAP-compliant testing framework for Bash scripts. It provides a simple way to write tests right in Bash, using special `@test` syntax. `Bats` also supports test setup and teardown, making it suitable for more complex test scenarios.

3. **ShellSpec**: A full-featured BDD (Behavior Driven Development) unit testing framework for shell scripts. `ShellSpec` supports various shells (not just Bash) and offers a rich set of features, including mocking and coverage reporting. Its syntax is inspired by `RSpec`, making it familiar to Ruby developers.

4. **bash_unit**: A simple Bash script to automate testing of Bash scripts, applications, and shell libraries. It's less complex than some of the other options, making it easy to get started with minimal setup.

5. **Roundup**: Another simple framework for writing and running tests for shell scripts. `Roundup` provides an easy-to-understand syntax and is designed to be portable and easy to deploy.

Each of these tools has its strengths and is suitable for different kinds of projects and preferences. For simpler scripts or projects where Bash is not the primary focus, `shUnit2` or `bash_unit` might be sufficient. For projects requiring more sophisticated testing features, such as mocking or BDD-style tests, `ShellSpec` or `Bats` could be more appropriate.

The best approach is often to evaluate a few of these frameworks based on the specific requirements of your project and choose the one that best fits your needs.

[...]

PR #4076, opened 2024-12-04
Adding secure mode option to build scripting

adamfarley:
This mode will eventually be used to disable a range of build script functionality that raises the potential security risk level during the build process.

This PR is intended to cover the creation of the new mode, the relevant documentation, 

After this commit is merged, we will need to add the flag to the pipeline job configurations in order to pass it into build jobs.

This PR is connected to https://github.com/adoptium/temurin-build/issues/4074 but does not fully resolve it.

### DIFF
--- a/sbin/common/config_init.sh
+++ b/sbin/common/config_init.sh
@@ -63,6 +63,7 @@ DISABLE_ADOPT_BRANCH_SAFETY
 DOCKER_FILE_PATH
 DOCKER_SOURCE_VOLUME_NAME
 ENABLE_SBOM_STRACE
+ENABLE_SECURE_MODE
 FREETYPE
 FREETYPE_DIRECTORY
 FREETYPE_FONT_BUILD_TYPE_PARAM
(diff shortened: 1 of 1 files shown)


sbin/common/config_init.sh
```diff
@@ -558,6 +562,9 @@ function configDefaults() {
 
   BUILD_CONFIG[ENABLE_SBOM_STRACE]="false"
 
+  # Set default value to "false", for maximum user convenience. "false" enables potentially-insecure functionality, like the dynamic download of boot JDKs.
```
judovana, line 565: The boot jdk is one of few downloads, which you must manually force to happen.  Eg. alsa is on contrary download without warning.  Free type is downloaded as sources?  What else is/can be  downloaded?  /me just curious.  Thanx for the downlaod unification effort. Highly appreciated!

adamfarley: Mainly boot JDKs. AIX, Alpine Linux, and Mac download boot JDKs if they can't find one locally.

There are also 22 other uses of wget, and many more that are indirect.

e.g. buildDocker.sh and dockerfile-generator wget many things, like the list of available releases and the mkdocker.sh script from openj9.

judovana:
Just FYI, there is: https://github.com/adoptium/temurin-build/issues/3741 It seems that you have initiated the work :)

PR #4082, opened 2024-12-09
add initial python version of script to add metadata to sbom

### LINKED ISSUE #3917: configure args is in the metadata file but not the main SBoM file
I thought we had the configure args in the main SBoM now - I guess not 😢 
Ideally we'd just pull the SBoM from the API but since we need the metadata file as well I guess it makes sense to have the two lines look similar.
We typically don't want to encourage people going directly to our backend location on github.

_Originally posted by @sxa in https://github.com/adoptium/adoptium.net/pull/2949#discussion_r1707377946_

We should ensure that all information required for reproducibility is in the main SBoM file so the metadata file (Primarily for use by the adoptium API) is not required for reproducibility verification.

The goal here should be to ensure that out documentationg like [this for Linux/x64](https://adoptium.net/en-GB/docs/reproducible-verification-builds/reproduce-linux-x64/) does not include `curl -L -O https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.4%2B7/OpenJDK21U-sbom_x64_linux_hotspot_21.0.4_7-metadata.json` because all of the information has been added to the SBoM file.

Scanteianu:
resolves #3917


Scanteianu:
superseded by #4122

PR #4092, opened 2024-12-17
Amend Windows Reproducible Ant Check To Be Less Strict

### LINKED ISSUE #4062: Windows Reproducible Build Script - Ant Download And Install Fails

The windows reproducible build script fails when trying to download and install a version of Ant deduced from the SBOM ( see 

[**https://ci.adoptium.net/job/Test_openjdk21_hs_special.system_x86-64_windows/8/consoleFull**](https://ci.adoptium.net/job/Test_openjdk21_hs_special.system_x86-64_windows/8/consoleFull)

The reproducible build test deduced a requirement of Ant 1.10.5 from the SBOM, however newer machines (  test-azure-win2022-x64-4 )  only have the later Ant Version 1.10.15 installed. The jenkins automated process however has insufficient permissions to install an older version.

**Expected behaviour:**

**Test should run ok with any suitable version of Ant.**

**Observed behaviour:**

**Test fails, when it should either detect a suitable version (  either  1.10.XXXX is allowed or ANY version of Ant is allowed )**

**Any other comments:**

steelhead31:
Fixes #4062 

Update the ant validation logic in the windows reproducible build script to allow any minor version of the correct major version. Both are defined in the script header, 1.10.{X} will be treated as allowed, rather than pinning to the exact version in the SBOM.

### DIFF
--- a/tooling/reproducible/windows_repro_build_compare.sh
+++ b/tooling/reproducible/windows_repro_build_compare.sh
@@ -43,7 +43,8 @@ REPORT_DIR="$3"
 # The Defaults Below Are Suitable For An Adoptium Windows Build Environment
 # Which Has Been Created Via The Ansible Infrastructure Playbooks
 WORK_DIR="/cygdrive/c/comp-jdk-build"
-ANT_VERSION="1.10.5"
+ANT_VERSION_ALLOWED="1.10"
+ANT_VERSION_REQUIRED="1.10.15"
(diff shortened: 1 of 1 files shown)


steelhead31:
Tested OK locally.

sophia-guo (approved):

karianna (approved):

Issue #4096, opened 2024-12-19
Reproducible test failure on Windows


sxa:
I spotted this a few days ago but I don't think it's been raised (If so, feel free to close this as a duplicate)
This seems to have been in [all runs of special.system](https://ci.adoptium.net/job/Test_openjdk21_hs_special.system_x86-64_windows) since job 27 on the 5th December: 
```
18:09:15  Reading The SBOM Content & Validating The Structure..
18:09:15  
18:09:15  SBOM Is Structurally Sound.. Extracting Values:
18:09:15  
18:09:15  assertion "cb == jq_util_input_next_input_cb" failed: file "/cygdrive/d/a/scallywag/jq/jq-1.7-1.x86_64/src/jq-1.7/src/util.c", line 381, function: jq_util_input_get_position
18:09:16  -----------------------------------
18:09:16  Rebuild_Same_JDK_Reproducibility_Test_win_0_FAILED
18:09:16  -----------------------------------
```

adamfarley:
I haven't seen this, and I'm not aware of any dups.

For completeness, here's when you first noted this problem: https://github.com/adoptium/infrastructure/issues/3286#issuecomment-2507707091

adamfarley:
The [internet suggests](https://stackoverflow.com/questions/71662276/jq-not-processing-a-json-file-in-cygwin) an sbom formatting issue. Didn't we update cygwin recently? We should check out the sbom formatting.

sxa:
> Didn't we update cygwin recently? 

Not explicitly. The build machines were new installs from the playbooks but this failure is occurring on the test machines where we run the reproducible test.

andrew-m-leonard:
It's the update to CDX 1.6 spec where metadata.tools has been deprecated in favour of new tools information "components" format

PR #4097, opened 2024-12-20
Update reproducible tooling to select updated CDX 1.6 tools

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/4096

CDX 1.6 spec deprecates metadata.tools, replaced by metadata.tools.components ToolInformation
https://github.com/adoptium/temurin-build/blob/master/tooling/release_download_test.sh also needs updating to download latest cyclonedx-cli to support CDX 1.6

Grinder: https://ci.adoptium.net/job/Grinder/12138/console

### DIFF
--- a/tooling/release_download_test.sh
+++ b/tooling/release_download_test.sh
@@ -425,12 +425,12 @@ download_cyclonedx_tool() {
   esac
 
   case "${cyclonedx_os}-${cyclonedx_arch}" in
-      linux-x64)   cyclonedx_checksum="bd26ccba454cc9f12b6860136e1b14117b829a5f27e993607ff526262c5a7ff0";;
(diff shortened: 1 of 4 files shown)


sxa (approved):
Approving on the basis that you will have tested that this fixes the problem and you're far more of an expert in this area than I am ;-)

Is there an opportunity here for docs on "What to change when you update the SBoM?". As another example, [release_download_test.sh](https://github.com/adoptium/temurin-build/blob/master/tooling/release_download_test.sh) (called from the [download_and_sbom_validation](https://ci.adoptium.net/job/build-scripts/job/release/job/download_and_sbom_validation/) job) is another thing that might be impacted when people change the SBoM

steelhead31 (approved):
Looks ok.

PR #4123, opened 2025-01-31
warn on false negatives if repro_compare.sh is run after  comparable_patch.sh in  ReproducibleBuilds.md

judovana:
I may be misunderstanding, but it seems, that if you run `comparable_patch.sh` you can not use `repro_compare.sh` for final comparison. You would get false negatives. Run plain `diff jdk1 jdk2` Should be the only  thing to call. 
I would like to add this as warning to the readme.


andrew-m-leonard (approved):

karianna (approved):

judovana:
tyvm!

PR #4075, opened 2024-12-04
Generate XML SBOM

Haroon-Khel:
Changing the hardcoded `--jsonFile` flag to `--xmlFile` in `sbom.sh` should command TemurinGenSBOM.java to generate an XML file instead of a json. And then I just changed any json variable name to xml to clear any confusion

PR #4127, opened 2025-02-05
Generate SBOM in XML

Haroon-Khel:
Re do of https://github.com/adoptium/temurin-build/pull/4075/files


andrew-m-leonard (changes requested):
This is on hold after this week PMC, subject to getting user feedback as to whether this is an acceptable change...

PR #4128, opened 2025-02-05
Same results for similar and identical

judovana:
Hello!

When I was following the:  https://github.com/adoptium/temurin-build/commit/d1c15a677275e44c3fed8c287159743794e88f6a, I realised that after the `diff jdk1 jdk2` I'm moreover replicating what repro_compare.sh is doing.  Thus I thought it would be better to leave it all on repro_compare.sh
, and just allow it to skip the whole preprocessing. WDYT?


judovana:
Tyvm!

PR #4144, opened 2025-03-03
Redirect output to build.log and add all related info to testoutput

sophia-guo:
Reduce reproducible tests output to console. Redirect all output to build.log, which will be archived as part of testoutput.

Also add sbom.json to testoutput for mac and win. (linux already enabled.)

Related https://github.com/adoptium/ci-jenkins-pipelines/issues/1188


andrew-m-leonard (approved):
lgtm

karianna (approved):

PR #4162, opened 2025-04-17
cacerts: follow redirects in http requests

gdams:
The CACert updater job has been failing for some time: https://github.com/adoptium/temurin-build/actions/runs/14481848049/job/40620227207

It looks like Mozilla changed the http endpoints to redirect to an edge server which was breaking our code as we weren't following redirects


steelhead31 (approved):
Looks good.

karianna (approved):

PR #4218, opened 2025-08-13
Disable jdk24 and enable jdk25 in AutoTriage

adamfarley:
Plus an update to the RELEASING doc.


adamfarley:
Thanks @smlambert and @karianna 🤓

Issue #4159, opened 2025-04-10
Bundled freetype depends on system harfbuzz


zzambers:
We are facing issue with jdk8 build, with bundled freetype:
```
java.lang.UnsatisfiedLinkError:
/usr/lib/jvm/java-1.8.0-openjdk-jdk8u452.b07-0.ojdk8~u~upstream.hotspot.release.sdk.el7.x86_64.tarxz/jre/lib/amd64/libfontmanager.so:
/lib64/libharfbuzz.so.0: undefined symbol: FT_Get_Color_Glyph_Layer
```

It is actually not libfontmanager.so library (directly) that depends on harfbuzz (jdk8 does not depend on harfbuzz), but rather bundled freetype (in jre/lib/amd64) depends on it:
```
 objdump -x ./libfreetype.so.6 | grep NEEDED
  NEEDED               libz.so.1
  NEEDED               libharfbuzz.so.0
  NEEDED               libc.so.6
```
Harfbuzz in turn depens on freetype. This causes problem, if system has newer harfbuzz, which is not compatible with older freetype (bundled in jdk).

Seems like freetype links against system harfbuzz automatically, if found:
```
checking for HARFBUZZ... yes
checking for library containing clock_gettime... none required
configure: creating ./config.status
config.status: creating ftoption.h
config.status: creating unix-cc.mk
config.status: creating unix-def.mk
config.status: creating ftconfig.h
config.status: executing libtool commands
configure:

Library configuration:
  external zlib: yes (pkg-config)
  bzip2:         no
  libpng:        no
  harfbuzz:      yes (pkg-config) 
```

Maybe [configure command](https://github.com/zzambers/temurin-build/blob/d017e2c38ecf8117ba62471652e069b67376e19c/sbin/prepareWorkspace.sh#L597) could be modified to avoid dependence on harfbuzz in similar way as with png.

zzambers:
Actually I am now looking at build error on fedora 42, which looks like related problem. When reproducing steps of temurin-build I have run into this configure problem (JDK 8):
```
configure:45697: Found freetype include files at /home/tester/freetype-lib/include using --with-freetype
configure:45990: checking for freetype includes
configure:45992: result: /home/tester/freetype-lib/include
configure:45995: checking for freetype libraries
configure:45997: result: /home/tester/freetype-lib/lib
configure:49149: checking if we can compile and link with freetype
[...]

PR #4238, opened 2025-08-28
Avoid dependency of bundled freetype on harfbuzz

zzambers:
This avoids dependency of bundled freetype on system harfbuzz, as it can cause [issue with unresolved symbols](https://github.com/adoptium/temurin-build/issues/4159) depending on system.

Problem is, that harfbuzz in turn depends on freetype (also see [docs](https://github.com/freetype/freetype/blob/48f91b53311929deaf0a26140f06d9f657f1db13/docs/INSTALL.UNIX#L109)), but if version of system freetype is different (newer) than bundled freetype version, (system) harfbuzz fails on unresolved symbols.

Freetype automatically builds with harfbuzz support, if found. (linking against system harfbuzz). This explicitly disables harfbuzz support in freetype. It only affects builds (jdk < 11) with bundled freetype. Another option would be to also bundle harfbuzz.

Note that temurin builds on linux [no longer bundle freetype](https://github.com/adoptium/temurin-build/pull/2226).

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -588,11 +588,11 @@ checkingAndDownloadingFreeType() {
     eval "${freetypeEnv}" && bash ./autogen.sh || exit 1
 
     local pngArg=""
-    if ./configure --help | grep "with-png"; then
+    if bash ./configure --help | grep "with-png"; then
       pngArg="--with-png=no"
     fi
     local harfbuzzArg=""
-    if ./configure --help | grep "with-harfbuzz"; then
+    if bash ./configure --help | grep "with-harfbuzz"; then
       harfbuzzArg="--with-harfbuzz=no"
     fi
 
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -591,10 +591,14 @@ checkingAndDownloadingFreeType() {
     if ./configure --help | grep "with-png"; then
       pngArg="--with-png=no"
     fi
+    local harfbuzzArg=""
+    if ./configure --help | grep "with-harfbuzz"; then
+      harfbuzzArg="--with-harfbuzz=no"
+    fi
 
     # We get the files we need at $WORKING_DIR/installedfreetype
     # shellcheck disable=SC2046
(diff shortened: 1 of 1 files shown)


sbin/prepareWorkspace.sh
```diff
@@ -591,10 +591,14 @@ checkingAndDownloadingFreeType() {
     if ./configure --help | grep "with-png"; then
       pngArg="--with-png=no"
     fi
+    local harfbuzzArg=""
+    if ./configure --help | grep "with-harfbuzz"; then
```
judovana: I would definitley call it with interpreter - `sh ./configure` to be sure it passes also from various custom tarballs.

sbin/prepareWorkspace.sh
```diff
@@ -591,10 +591,14 @@ checkingAndDownloadingFreeType() {
...
+      harfbuzzArg="--with-harfbuzz=no"
```
judovana, line 596: The only need to have this if is to distinguish jdk8 from others? (jsut more clever, to count possible future jdk8 with harfbuz..) . Right?

judovana:
hi! Nice detective work! There is so much IFs and wrappers:(

judovana:
The system free type is new for me. Its 5 years old. Since that, the world had tunred a bit, and whole world had moved to in-tree bundeld freetype...

karianna:
Paging @jerboaa here as well as I don't trust my own opinions on linux build package changes for OpenJDK

sbin/prepareWorkspace.sh
jerboaa: If we do this we should do it on line 591 as well. So this seems prior art. Since the final `configure` call of the `freetype` sources is using `bash ./configure` I'd suggest to use that consistently when calling `--help` as well.

jerboaa (commented):
Please make the `configure --help` calls consistent. Looks good otherwise.

sbin/prepareWorkspace.sh
zzambers: No sure I understand the question. Whole function is [used for jdk 8 only](https://github.com/adoptium/temurin-build/blob/4ea25fa5e9ee5e7e77b66b6bf60bac3573c4c9f9/sbin/prepareWorkspace.sh#L853) (counting only alive jdks). Newer jdk's (11+) have in-tree freetype.

karianna (approved):

Issue #4230, opened 2025-08-21
Retro Action: FAQ update - list of compilers


adamfarley:
> Another thing noticed as part of some other work but relevant to the release. The [list of compilers in the temurin-build FAQ](https://github.com/adoptium/temurin-build/blob/master/FAQ#what-are-the-prerequisites-for-a-system-used-for-builds) does not mention that we use a devkit for JDK21+ 

 _Originally posted by @sxa in [#84](https://github.com/adoptium/temurin/issues/84#issuecomment-3163720838)_

sxa:
Also to note that we are using GCC14.2 for Linux on jdk25+ as of today
Ref: https://github.com/adoptium/infrastructure/issues/3742

PR #4241, opened 2025-08-29
Update Temurin's GCC versions in FAQ.md to match reality of what we build with

sxa:
Preview at https://github.com/sxa/temurin-build/blob/compiler_updates/FAQ.md#what-are-the-prerequisites-for-a-system-used-for-builds

Fixes https://github.com/adoptium/temurin-build/issues/4230


sxa:
@andrew-m-leonard I've only updated Linux and Windows here and have not updated the others. LMK if you think updates are needed to other platforms and I can add them here or we can cover that in a separate PR.

karianna (approved):

andrew-m-leonard (approved):
lgtm

Issue #4098, opened 2024-12-20
Create new jobs to handle running Solaris builds via a Linux proxy machine


sxa:
This issue covers the implementation of what was discussed in https://github.com/adoptium/infrastructure/issues/3742#issuecomment-2478948681

Current status: Prototype jobs have been created for x64 and SPARC at:
- https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-x64-temurin-simple
- https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-temurin-simple

These are currently connecting to the target machine as the `vagrant` user (even on SPARC where I've created a user with that name for consistency).

sxa:
SPARC job doesn't seem to be doing too well: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-temurin-simple/
```
Compiling /export/home/local/vagrant/temurin-build/build-farm/workspace/build/src/hotspot/src/share/vm/runtime/sweeper.cpp
"/export/home/local/vagrant/temurin-build/build-farm/workspace/build/src/hotspot/src/share/vm/interpreter/abstractInterpreter.hpp", line 36: Error: Could not open include file "interp_masm_x86.hpp".
"/export/home/local/vagrant/temurin-build/build-farm/workspace/build/src/hotspot/src/share/vm/interpreter/templateTable.hpp", line 34: Error: Could not open include file "interp_masm_x86.hpp".
"/export/home/local/vagrant/temurin-build/build-farm/workspace/build/src/hotspot/src/share/vm/interpreter/templateTable.hpp", line 361: Error: Could not open include file "templateTable_x86_64.hpp".
"/export/home/local/vagrant/temurin-build/build-farm/workspace/build/src/hotspot/src/share/vm/interpreter/templateInterpreter.hpp", line 191: Error: Could not open include file "templateInterpreter_x86.hpp".
"/export/home/local/vagrant/temurin-build/build-farm/workspace/build/src/hotspot/src/share/vm/interpreter/interpreter.hpp", line 122: Error: The type "InterpreterMacroAssembler" is incomplete.
"/export/home/local/vagrant/temurin-build/build-farm/workspace/build/src/hotspot/src/share/vm/interpreter/interpreter.hpp", line 128: Error: The type "InterpreterMacroAssembler" is incomplete.
[...]

Issue #4099, opened 2024-12-20
Create new jobs to handle running Solaris tests via a Linux proxy machine


sxa:
This covers the implementation of what has been discussed in https://github.com/adoptium/infrastructure/issues/3742#issuecomment-2529132798

Current status: Prototype jobs have been created for x64 and SPARC at:
- https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-x64-temurin-simpletest
- https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-temurin-simpletest

These are currently connecting to the target machine as the `vagrant` user (even on SPARC where I've created a user with that name for consistency).

Jobs are currently set up to run the full AQA suite of tests instead of running as individual jobs but that can be changed later if desired and archive the artefacts

sxa:
Prototype now working (although hard coded to a specific tag. There is a `dotests.x64.sh` script and a `dotests.sparcv9` script on the proxy host which is copied across to the target machine using `scp` as `dotests.sh` and that is executed.

It needs to be parameterised to be able to take the tag/URL as a parameter (for use when retrieving the artifact directly from jenkins, since we can't use `copyArtifact`) but otherwise it works.
It currently has a loop which can loop over each suite that is required and then copy the TAP output back to the proxy machine for archiving. The [proxy agent](https://ci.adoptium.net/computer/dockerhost-azure-solarisproxy/) is running as the `solaris` user on [dockerhost-azure-ubuntu2204-x64-1](https://ci.adoptium.net/computer/dockerhost-azure-ubuntu2204-x64-1/)

sxa:
## Verification (Solaris/x64)

Failures (based on looking at https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-x64-temurin-simpletest/50/tapResults:
- sanity.openjdk:
  - jdk/lambda/vm/InterfaceAccessFlagsTest.java
- sanity.system (All due to being unable to locate `mauve/mauve.jar`:
  - MauveSingleThrdLoad_HS_5m_0
  - MauveSingleInvocLoad_HS_5m_0
  - MauveMultiThrdLoad_5m_0
- extended.system (All except OAuthTest_0 are due to being unabel to locate `mauve/mauve.jar`:
  - MiniMix_5m_0
  - MiniMix_10m_0
[...]

PR #4253, opened 2025-09-09
update solarissimple dotests.sh to allow artifact copying in jenkins

sxa:
Comments at the end for the final `gmake` will be removed before merging and are purely while the script is being tested.

Part of the [Solaris EPIC](https://github.com/adoptium/infrastructure/issues/3742#issuecomment-3248772139)

Closes https://github.com/adoptium/temurin-build/issues/4098 https://github.com/adoptium/temurin-build/issues/4099 and probably the epic, but I'll do that manually once I'm happy as there are a couple of updates that will be done to switch branches in the jobs once this is merged.

### DIFF
--- a/sbin/solaris/dotests.sh
+++ b/sbin/solaris/dotests.sh
@@ -12,7 +12,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # ********************************************************************************
-
+# shellcheck disable=SC2006
 #
 # dotests.sh
 # Designed for running via a proxy machine for the Temurin Solaris tests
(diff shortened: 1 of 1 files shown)


sxa:
This seems to be running ok so I'm happy to get this merged and start using it. Noting that it will not affect the Jdk25 release this month as it is in solaris specific code which is only invoked for jdk8u

sxa:
> Can I ask you to add "# shellcheck disable=SC2006" to the top of the file rather than suggesting that the Linter go do one?

Done, but in reality I didn't because it didn't seem like the most productive use of my time ... and the comment made the reasoning clear ;-)

Also noting that it's a comment that I'd added previously, not specifically in this PR.

adamfarley:
> > Can I ask you to add "# shellcheck disable=SC2006" to the top of the file rather than suggesting that the Linter go do one?
> 
> Done, but in reality I didn't because it didn't seem like the most productive use of my time ... and the comment made the reasoning clear ;-)
> 
> Also noting that it's a comment that I'd added previously, not specifically in this PR.

Looks like there are other linter errors too. Hmm.

The script should work as intended as-is, so I'll approve. If I feel strongly about the linter failures, I can fix it in a separate, housekeeping PR.

sxa:
Thanks @adamfarley - feel free to raise an issue on the other linter fixes if you don't plan to do a PR immediately as it probably makes sense to look at some of them in the future

PR #4265, opened 2025-10-07
Allow getOpenJDKTag() to default to version of checked out BRANCH if no tags in repo

andrew-m-leonard:
Fixes https://github.com/adoptium/ci-jenkins-pipelines/issues/497

This should be the final fix to allow building from "branches" from openjdk. The problem was i think from various other changes to the build script handling of tags, shas, etc, whereby if no git tag could be resolved which would be the case for a personal branch (as it contains no tags), then it would not default to anything...and just fail!

With this PR, you should now be able to either:
1. Build from a personal fork branch of an adoptium mirror by specifying BUILD_ARGS, eg:
```
-r https://github.com/andrew-m-leonard/jdk-adopt -b aix_patch
```
2. Build from an upstream openjdk hotspot branch by specifying BUILD_ARGS, eg:
```
-r https://github.com/openjdk/jdk -b jdk25 --disable-adopt-branch-safety
```

### DIFF
--- a/sbin/prepareWorkspace.sh
+++ b/sbin/prepareWorkspace.sh
@@ -229,8 +229,8 @@ checkoutRequiredCodeToBuild() {
         tag="${BUILD_CONFIG[BRANCH]}"
         BUILD_CONFIG[TAG]="${tag}"
         BUILD_CONFIG[SHALLOW_CLONE_OPTION]=""
-      elif git cat-file commit "${BUILD_CONFIG[BRANCH]}"; then
-        echo "look like the BUILD_CONFIG[BRANCH] (${BUILD_CONFIG[BRANCH]}) scm ref given is a valid sha, so treat it as a sha"
+      elif git cat-file commit "${BUILD_CONFIG[BRANCH]}" 2>/dev/null; then
+        echo "looks like the BUILD_CONFIG[BRANCH] (${BUILD_CONFIG[BRANCH]}) scm ref given is a valid sha, so treat it as a sha"
         sha="${BUILD_CONFIG[BRANCH]}"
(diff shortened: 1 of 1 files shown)


karianna (approved):

sxa (commented):
I'm testing this, but noting that it does say
```
fatal: Not a valid object name <branchName>
BUILD_CONFIG[BRANCH} (<branchName>) is a valid branch, fetching and resetting to branch
```
It seems to have worked though :-)

andrew-m-leonard:
> I'm testing this, but noting that it does say
> 
> ```
> fatal: Not a valid object name <branchName>
> BUILD_CONFIG[BRANCH} (<branchName>) is a valid branch, fetching and resetting to branch
> ```
> 
> It seems to have worked though :-)

@sxa do you have a link please? i'd like to check that...?

andrew-m-leonard:
Standard build test on Windows https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-windows-x64-temurin-andrew/1/

Build from branch test: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk/job/jdk-linux-x64-temurin/602/

andrew-m-leonard:
@sxa ah yes, that's from this check for SHA: elif git cat-file commit "${BUILD_CONFIG[BRANCH]}"; then
I can add 2>/dev/null to quiet that message

andrew-m-leonard:
Fixed... "fatal: Not a valid object name <branchName>" should no longer be seen

sxa:
> @sxa do you have a link please? i'd like to check that...?

Nope - I was running locally at a command line when I saw that. But the fix looks good - thanks. I'll run one more build before approving

Issue #3967, opened 2024-10-02
Status of Reproducible comparing tests on linux


sophia-guo:
PPC64le and aarch64 was enabled on Sep 25th. https://github.com/adoptium/ci-jenkins-pipelines/pull/1115. There is no data yet as  no betaTrigger builds triggered with test enabled. 

Recent linux update was merged in [Sep 6th. ](https://github.com/adoptium/temurin-build/pull/3921).

Most recent ea pipeline jobs are triggered by test disabled (not sure why, for example https://ci.adoptium.net/job/build-scripts/job/openjdk21-pipeline/333/), which means most reproducible comparing tests are quite old and may not be worth looking at. FYI @smlambert @andrew-m-leonard 

Only one test build with jdk24 is with recent update. https://trss.adoptium.net/output/test?id=66f72b85d24e1b006ee0dd01 @andrew-m-leonard triggered by https://ci.adoptium.net/job/build-scripts/job/openjdk24-pipeline/25/, test jenkins job is https://ci.adoptium.net/job/Test_openjdk24_hs_special.system_x86-64_linux/11/.  Both job and rerun job have same results. 6 files differs.

```
17:51:56  Number of files: 105604
17:51:56  Comparing jdk-24+17 with tarJDK/jdk-24+17 ... output to file: reprotest.diff
17:52:00  Files jdk-24+17/jmods/expanded_java.base.jmod/lib/classlist and tarJDK/jdk-24+17/jmods/expanded_java.base.jmod/lib/classlist differ
17:52:00  Files jdk-24+17/jmods/expanded_java.base.jmod/lib/server/libjvm.so and tarJDK/jdk-24+17/jmods/expanded_java.base.jmod/lib/server/libjvm.so differ
17:52:00  Files jdk-24+17/lib/classlist and tarJDK/jdk-24+17/lib/classlist differ
17:52:00  Files jdk-24+17/lib/server/classes.jsa and tarJDK/jdk-24+17/lib/server/classes.jsa differ
17:52:00  Files jdk-24+17/lib/server/classes_nocoops.jsa and tarJDK/jdk-24+17/lib/server/classes_nocoops.jsa differ
17:52:00  Files jdk-24+17/lib/server/libjvm.so and tarJDK/jdk-24+17/lib/server/libjvm.so differ
17:52:00  Number of differences: 6
17:52:00  ReproduciblePercent = 99.99 %
17:52:00  Differences found..., logged in: reprotest.diff

```

[...]

PR #4268, opened 2025-10-10
Restrict detailed file diff to classlist files

adamfarley:
Otherwise we get many megabytes of output, which overloads the test output YAML handler and fails the job.

This is incorrect, as the detailed output should be limited to the contents of the classlist file ([as per this comment](https://github.com/adoptium/temurin-build/issues/3967#issuecomment-2391102560)).

Detailing the diffs for every file prevents further test execution and prevents us detecting other failures.

Tested here: https://ci.adoptium.net/job/Grinder/15011/

### DIFF
--- a/tooling/reproducible/repro_compare.sh
+++ b/tooling/reproducible/repro_compare.sh
@@ -83,7 +83,7 @@ echo "$(date +%T) : diff complete - rc=$rc. Output written to file: ${output}"
 
 cat "${output}"
 
-grep "Files .*/classlist differ" "${output}" | while read -r line; do
+grep "Files .*/classlist" "${output}" | while read -r line; do
   FILE1=$(echo "$line" | awk '{print $2}')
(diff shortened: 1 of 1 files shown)


adamfarley:
Also excluding PR build checks for changes to the PR folder, as I don't believe anything in "tooling" is used by the core build process.

tooling/reproducible/repro_compare.sh
```diff
@@ -83,7 +83,7 @@ echo "$(date +%T) : diff complete - rc=$rc. Output written to file: ${output}"
 
 cat "${output}"
 
-grep "Files .*" "${output}" | while read -r line; do
+grep "Files .*/classlist differ" "${output}" | while read -r line; do
```
andrew-m-leonard:
```suggestion
grep "Files .*/classlist" "${output}" | while read -r line; do
```

adamfarley:
Thanks @sxa and @andrew-m-leonard 😎

PR #4270, opened 2025-10-13
Updating jdk25 to jdk25u for autotriage

adamfarley:


adamfarley:
Thanks @Haroon-Khel and @karianna 😎

PR #4283, opened 2025-10-24
Update expected build gcc for linux platforms to 14.2.0

adamfarley:
Except for Alpine and risc, which stay the same.

Note: This will also be cherry-picked for the October 2025 release branch.

PR #4284, opened 2025-10-24
Update expected build gcc for linux platforms to 14.2.0

adamfarley:
Except for Alpine and risc, which stay the same.

Cherry-picked from https://github.com/adoptium/temurin-build/pull/4283


sxa (approved):
LGTM

PR #4287, opened 2025-10-28
adaptation of JDK-8 to build with in-tree freetype

judovana:
Hello!

There is https://github.com/openjdk/jdk8u-dev/pull/709 agreed to go to jdk8, so a change in temurin-build is necessary to continue tp support future jdk8/

This is minimalist change, which is allowing to build jdk8 as temurin, but I do not belive it is correct, enough, nor final.

What is supposed to be supported in this script? Older jdk8?  Jdk9+10? And so on. If none, maybe whole freetype download and build can be remove? Due to nature of tis commit, the jdk8 build in CI is most liekly going to fail before https://github.com/openjdk/jdk8u-dev/pull/709  is merged and used.

 Any  guidance or thoughts highly appreciated

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -548,12 +548,12 @@ configureAlsaLocation() {
   fi
 }
 
-setDefaultFreeType() {
+setBundledFreeType() {
   echo "Freetype set from bundled in jdk"
   freetypeDir=${BUILD_CONFIG[FREETYPE_DIRECTORY]:-bundled}
 }
 
-setFreeTypeFromSrcs() {
+setFreeTypeFromExternalSrcs() {
   echo "Freetype set from local sources"
   addConfigureArg "--with-freetype-src=" "${BUILD_CONFIG[WORKSPACE_DIR]}/libs/freetype"
 }
@@ -568,10 +568,10 @@ configureFreetypeLocation() {
     if [[ "${BUILD_CONFIG[FREETYPE]}" == "true" ]]; then
       local freetypeDir="${BUILD_CONFIG[FREETYPE_DIRECTORY]}"
       if isFreeTypeInSources ; then
-        setDefaultFreeType
+        setBundledFreeType
       else
         if [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]]; then
-          setFreeTypeFromSrcs
+          setFreeTypeFromExternalSrcs
         else
           setFreeTypeFromInstalled
         fi


karianna:
For completeness it would be good to detect jdk8u_<supported version>+

judovana:
> For completeness it would be good to detect jdk8u_+

You mean to distinguish the freshly released "jdk8 5b3"  as it went out 2014 or to distinguish the particular u472 and older x future ones?

karianna:
being able to still support older jdk8 would be useful for completeness.

judovana:
That may be easier then to fully support the custom build of free type.  If the tag will be present, which usually is, then it can be decided. If not, the future in-tree approach should take place. Or maybe add switch?

However to fully remove the custom freetype, would simplify the scripts a lot. I'm still just brainstorming. Am not sure which direction to take. Thanx a lot for inputs!

judovana:
> being able to still support older jdk8 would be useful for completeness.

@karianna  WDYT now? It is based on tag, so pretty limited, but had not found a better way.

This do not work wit `-l`  ... but not sure if it is worthy of fixing this case

judovana:
I would like to use https://github.com/adoptium/temurin-build/blob/f4bcbeb80b2dc0a624721f09a91c925e6df2bc90/sbin/prepareWorkspace.sh#L592 like approach, but in this moment, the sources do not exists. Thoughts?

judovana:
The https://github.com/openjdk/jdk8u-dev/pull/709  was integrated

sbin/build.sh
```diff
@@ -548,22 +548,77 @@ configureAlsaLocation() {
   fi
 }
 
+getFreeTypePerJdk8Tag() {
+  if [ -n "${BUILD_CONFIG[TAG]}" ]; then
+    local majorBuildVersion=$(echo "${BUILD_CONFIG[TAG]}" | sed "s/jdk8u//" | sed "s/-.*//")
+    if [ 0${majorBuildVersion} -lt 482 ] ; then
+      echo "old"
+    elif [ 0${majorBuildVersion} -gt 482 ] ; then
```
karianna: Should this be gt and equal to?

karianna: Can we just simplify the logic here to capture `8u<version>` and see if `<version>` is >=482 and then apply the freetype from source?  There seems to be excess logic here

karianna (changes requested):

sbin/build.sh
judovana: > Should this be gt and equal to?

Nope. The "eq" is handled in final else. More readable it is:
```
if  majorBuildVersion< 482  ; then  //always old
     return  "old"
 elif majorBuildVersion > 482] ; then    //always new
     return  "new"
else // that means ==
  decideByMinorBuildNumber()
fi
```

judovana: > Can we just simplify the logic here to capture `8u<version>` and see if `<version>` is >=482 and then apply the freetype from source? There seems to be excess logic here

Of course we can. I made it to the most precise granularity, and if "b" is unimportant (which I Actually doubt) then of course. But it was done in early  stage of  482, so maybe there can be done one exact match to  482-b01 and 482-b02 to avodi math check on b's XY.

I would like to use https://github.com/adoptium/temurin-build/blob/f4bcbeb80b2dc0a624721f09a91c925e6df2bc90/sbin/prepareWorkspace.sh#L592 like approach, but in this moment, the sources do not exists.   Does anybody  have any ideas? I will connect t some upcoming call due of this. @jiekang  any idea who to ping/when to join a call?

judovana: @karianna  Thanx for eyball, highly appreciated!

sbin/build.sh
```diff
@@ -548,22 +548,77 @@ configureAlsaLocation() {
...
+      echo "new"
+    else
+      # the change was introduces in 482, lets determine the build promotion, and decide
```
karianna:
```suggestion
      # The change was introduced in 482, lets determine the build number, and decide
```

karianna (changes requested):

judovana:
@andrew-m-leonard  @sxa  thoughts please?

sxa (commented):
Noting that I have managed to build the 8u482-b02 version with the original build scripts without any problem so it's not necessarily clear (unless I made a mistake in my build) [what error I should expect](https://github.com/adoptium/temurin-build/issues/4306#issue-3641662183).

Noting also that if we change the version of freetype in the build the [SBOM validation script](https://github.com/adoptium/temurin-build/blob/master/tooling/validateSBOMcontent.sh) will also need to be updated to match what gets produced.

sbin/build.sh
andrew-m-leonard: @judovana Is there a way to tell from the actual "source" what it is doing/supports? I don't know... but say if (some make file exists?!)

andrew-m-leonard:
fyi, jdk8u482-b02_adopt will be appearing shortly once i've resolved : https://github.com/adoptium/mirror-scripts/issues/76

adamfarley:
@sxa - Does removing/changing the "--with-freetype=/usr/local/" bit in solaris.sh fall under the scope of this PR?

https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-temurin-simple/120/console
```
configure: error: Valid values for --with-freetype are 'system' and 'bundled'
```

sbin/build.sh
andrew-m-leonard: although not ideal, you could gate this logic on the existance or not of .github/workflows/freetype.vcxproj ,
which has been deleted as the submit workflow does not support building freetype src with this update?

judovana:
> Noting that I have managed to build the 8u482-b02 version with the original build scripts (not from the temurin tags) without any problem so it's not necessarily clear (unless I made a mistake in my build) [what error I should expect](https://github.com/adoptium/temurin-build/issues/4306#issue-3641662183).
> 


Hi @sxa  thats exceptionally weird, it can not work via temurin scripts, unless you are setting up freetype manually, you will always get `configure: error: Valid values for --with-freetype are 'system' and 'bundled'`  bbecause its reading had chnaged:

```
$$ cd ~/git/temurin-build/
$$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$$ git status
On branch master
Your branch is up to date with 'origin/master'.
$$ git pull
remote: Enumerating objects: 4, done.
...
   a3adb91..c6c180f  noMoreFreeTypeFor8 -> origin/noMoreFreeTypeFor8
Already up to date.
$$ git log
commit 1159fa56119232e8588a174de80d51a1158d8691 (HEAD -> master, origin/master, origin/HEAD)
Author: Adam Farley <adfarley@ibm.com>
Date:   Fri Oct 24 17:36:30 2025 +0100

    Update expected build gcc for linux platforms to 14.2.0 (#4283)
    
    Except for Alpine and risc, which stay the same.
    
    Signed-off-by: Adam Farley <adfarley@ibm.com>
    
$$  sh  makejdk-any-platform.sh  --tag jdk8u482-b02  jdk8u
Starting makejdk-any-platform.sh to configure, build (Adoptium)OpenJDK binary
Parsing opt: --tag
...
checking for cups/ppd.h... yes
configure: error: Valid values for --with-freetype are 'system' and 'bundled'
No configurations found for /home/jvanek/git/temurin-build/workspace/build/src/! Please run configure to create a configuration.
Makefile:55: *** Cannot continue.  Stop.
OpenJDK make failed, archiving make failed logs
Archiving and compressing with pigz

real	0m0.001s
user	0m0.000s
sys	0m0.002s
Your archive was created as /home/jvanek/git/temurin-build/workspace/build/src/build/linux-x86_64-normal-server-release/OpenJDK.tar.gz
Moving the artifact to location /home/jvanek/git/temurin-build/workspace/target//OpenJDK-makefailurelogs.tar.gz
archive done.
Failed to make the JDK, exiting
```
After the fix:
```
$$ git checkout noMoreFreeTypeFor8
Switched to branch 'noMoreFreeTypeFor8'
Your branch is behind 'origin/noMoreFreeTypeFor8' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)
$$ git pull
Updating a3adb91..c6c180f
Fast-forward
 sbin/build.sh | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
$$ git log
commit c6c180f6a015113395a03ee2a2878ccb20b484ef (HEAD -> noMoreFreeTypeFor8, origin/noMoreFreeTypeFor8)
Author: judovana <judovana@email.cz>
Date:   Mon Nov 17 10:45:48 2025 +0100

    Update sbin/build.sh
    
    Co-authored-by: Martijn Verburg <martijnverburg@gmail.com>

commit a3adb91b195deb0039c8f1555ac0400132af9465
Author: Jiri Vanek <jvanek@redhat.com>
Date:   Thu Nov 13 11:57:56 2025 +0100

    The in-tree freetype will appear in 482-b02

$$  sh  makejdk-any-platform.sh  --tag jdk8u482-b02  jdk8u
Starting makejdk-any-platform.sh to configure, build (Adoptium)OpenJDK binary
Parsing opt: --tag
...
Using freetype: bundled
checking fontconfig/fontconfig.h usability... yes
...
## Starting langtools
Compiling 2 files for BUILD_TOOLS
```

judovana:
> @sxa - Does removing/changing the "--with-freetype=/usr/local/" bit in solaris.sh fall under the scope of this PR?
> 
> https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-temurin-simple/120/console
> 
> ```
> configure: error: Valid values for --with-freetype are 'system' and 'bundled'
> ```

Thats it. The patch is adjusting exactly this. Tahnx @adamfarley

judovana:
> Noting also that if we change the version of freetype in the build the [SBOM validation script](https://github.com/adoptium/temurin-build/blob/master/tooling/validateSBOMcontent.sh) will also need to be updated to match what gets produced.

Thats a lot of `EXPECTED_FREETYPE` versions. In theory, once this [effort](https://github.com/openjdk/jdk8u-dev/pull/720) is finished, all jdks (except 26) will have same, bundeld version. Isnt that correct?

sbin/build.sh
judovana: hi! I would love to, But jdk sources do not exists in this moment. Otherwise check like https://github.com/adoptium/temurin-build/blob/f4bcbeb80b2dc0a624721f09a91c925e6df2bc90/sbin/prepareWorkspace.sh#L592  would always work, and would be much much better.

The only other way is to deference all this after jdk sources are built, which may not be trivial, as in "legacy" mode, the certificates are prepared long before jdk. 

Tbh I would love to get rid of tag check, because the tag do not cover eg the build from tarball and maybe more.

I'm proposing to merge this, so jdk8 remain buildable, but would elaborate, on change of order. 
That woudl mean to get rid of tag parsing, and try to reorder cloning the jdk, and building of cacerts. I  already looked into it, and it seems doable. Only I got scared of such big change.
Or I will try to do so asap.

judovana:
> > Noting also that if we change the version of freetype in the build the [SBOM validation script](https://github.com/adoptium/temurin-build/blob/master/tooling/validateSBOMcontent.sh) will also need to be updated to match what gets produced.
> 
> Thats a lot of `EXPECTED_FREETYPE` versions. In theory, once this [effort](https://github.com/openjdk/jdk8u-dev/pull/720) is finished, all jdks (except 26) will have same, bundeld version. Isnt that correct?

I did some reading, and some experiments. Can I deffere this to separate PR?
How to test that file?

It do not seem correct to me. I thought, that everywhere except `8`, the bundled freetype is used - that means, single version, as all jdks except tip (and for now 8)  are on 2.13.3, no metter of OS. And jdk8 willb e on 2.9.0 now, 2.10.4 soon, and hopefully 2.13.3 before CPU...

judovana:
@andrew-m-leonard 

hi! I was partially wrong. Or better confused myself. Jdk exists in the moment of those checks, so I will switch to somehting like
```
isFreeTypeInSources() {
 if [ ! -e  build/src ] ; then
    echo "No jdk found to determine libfreetype/src presence"
    exit 1
 fi
 find build/src | grep  -q libfreetype/src
 return $?
}
```

That will simplify the whole code a lot. Howver,  waht I want to do, and what made me to write that "jdk sources do nto exists in that moment" is this
```
 function configureWorkspace() {
   if [[ "${BUILD_CONFIG[ASSEMBLE_EXPLODED_IMAGE]}" != "true" ]]; then
     createWorkspace
+    checkoutAndCloneOpenJDKGitRepo
     downloadingRequiredDependencies
     downloadDevkit
     relocateToTmpIfNeeded
-    checkoutAndCloneOpenJDKGitRepo
     applyPatches
     if [ "${BUILD_CONFIG[CUSTOM_CACERTS]}" = "true" ] ; then
       prepareMozillaCacerts
```

Without it, the freetype would be downloaded, and build, but later not used. Thats why I moved to the cryptic tag parsing.

I will now elaborate on this, and will stop building and downloading (and of course using)  the freetype if  `isFreeTypeInSources` will be true.

judovana:
```
+    checkoutAndCloneOpenJDKGitRepo
     downloadingRequiredDependencies
     downloadDevkit
     relocateToTmpIfNeeded
-    checkoutAndCloneOpenJDKGitRepo
```

Its now done. I t loosk much nicer (thanx!) but tbh, I'm not sure what all could be broken by that... Seems like nothing, but I do nto see to all edges of this build infra. 

This is exactly what I wanted to avoid, and called it "minimalistic". it is no longer minimalistic:)

Tested against:
`sh makejdk-any-platform.sh jdk8` (failing later in build)
`sh makejdk-any-platform.sh jdk9` (failing later in build)
`sh makejdk-any-platform.sh jdk11u`
`sh makejdk-any-platform.sh jdk8u`
`sh makejdk-any-platform.sh jdk21u`
`sh makejdk-any-platform.sh  --tag jdk8u482-b02  jdk8u`
`sh makejdk-any-platform.sh --tag jdk8u482-b01  jdk8u`
`sh  makejdk-any-platform.sh -l java-1.8.0-openjdk-jdk8u482.b01-0.ojdk8~u~upstream.src.tarxz  jdk8u`
`sh  makejdk-any-platform.sh -l java-1.8.0-openjdk-jdk8u482.b02-0.ojdk8~u~upstream.src.tarxz  jdk8u`

judovana:
The ECA check seems wrong: 
image

image

sbin/prepareWorkspace.sh
```diff
@@ -850,8 +867,7 @@ downloadingRequiredDependencies() {
   fi
 
   if [[ "${BUILD_CONFIG[FREETYPE]}" == "true" ]]; then
-    case "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" in
-      jdk8* | jdk9* | jdk10*)
+     if ! isFreeTypeInSources ".." ; then
```
karianna: I think the whitespace is out by one here (intend 3 vs 2)

karianna: Same with below else block

sbin/prepareWorkspace.sh
```diff
@@ -40,6 +40,23 @@ FREETYPE_FONT_SHARED_OBJECT_FILENAME="libfreetype.so*"
 # sha256 of https://github.com/adoptium/devkit-binaries/releases/tag/vs2022_redist_14.40.33807_10.0.26100.1742
 WINDOWS_REDIST_CHECKSUM="ac6060f5f8a952f59faef20e53d124c2c267264109f3f6fabeb2b7aefb3e3c62"
 
+isFreeTypeInSources() {
+  local libfreetypeid="libfreetype/src"
+  local location="$1/build/src"
+  if [ ! -e "$location" ] ; then
+    echo "No jdk found to determine $libfreetypeid presence"
```
karianna: Is that no JDK or no JDK source?

sbin/build.sh
```diff
@@ -548,22 +548,34 @@ configureAlsaLocation() {
...
 configureFreetypeLocation() {
   if [[ ! "${CONFIGURE_ARGS}" =~ "--with-freetype" ]]; then
     if [[ "${BUILD_CONFIG[FREETYPE]}" == "true" ]]; then
       local freetypeDir="${BUILD_CONFIG[FREETYPE_DIRECTORY]}"
-      if [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]]; then
-        case "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" in
-          jdk8* | jdk9* | jdk10*) addConfigureArg "--with-freetype-src=" "${BUILD_CONFIG[WORKSPACE_DIR]}/libs/freetype" ;;
-          *) freetypeDir=${BUILD_CONFIG[FREETYPE_DIRECTORY]:-bundled} ;;
-        esac
+      if isFreeTypeInSources "." ; then
```
karianna: Why do we need to pass in "." if it's a hard coded value, that function doesn't need to take $1 in that case?

judovana: Hi! Sorry. It was already late and I was not sure  which of the `${BUILD_CONFIG[WORKSPACE_DIR]}` `${BUILD_CONFIG[WORKING_DIR]}` `${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}` can be or cant  be absolute. Fixed.

sbin/prepareWorkspace.sh
judovana: rephrased

sbin/prepareWorkspace.sh
judovana: The file is mostly to two spaces. I had overlooked that there are few (eg this one) aligned to 3. I had truncated it to two in reasonable neighbourhood.  Am ok to rise it to three in the same scope, but thats going slightly against the file.

sxa:
> @sxa - Does removing/changing the "--with-freetype=/usr/local/" bit in solaris.sh fall under the scope of this PR?

Yeah that's broken too so will need to be changed. For [java version reasons](https://github.com/adoptium/temurin-build/blob/master/SOLARIS.md) Solaris is no longer built in the main pipelines but is built by https://github.com/adoptium/temurin-build/blob/master/sbin/solaris/build-simple.sh

We could either modify that script to force the new options into the configure args, or do it in the jenkins jobs themselves. Depends how messy it would get :-)

andrew-m-leonard:
@judovana Suspect some of your commits are not signed correctly for ECA??

judovana:
> The latest update looks good, if test builds are fine, then i'm happy. Thanks Jiri

Really?-) I'mn slightly nervouse about them... Thanx a lot for eyball! Will wait for @sxa and @karianna  now.

judovana:
> @judovana Suspect some of your commits are not signed correctly for ECA??

You are right @andrew-m-leonard ..
```
Author Eclipse account cannot be found.

    Email: ContBont@q* DOT com
```

What is that, I need to find...

judovana:
> > @judovana Suspect some of your commits are not signed correctly for ECA??
> 
> You are right @andrew-m-leonard ..
> 
> ```
> Author Eclipse account cannot be found.
> 
>     Email: ContBont@q* DOT com
> ```
> 
> What is that, I need to find...

Ok found it: https://github.com/adoptium/temurin-build/pull/4287/commits/1cbc980d5b85471ba66a613c30a8b08f96f47336

Its the  `conatinerQa bot authored and conatinerQa bot committed 18 hours ago `  What is it... But it seems like locla system missconfiguration
```
git config --list | grep -e user.email -e user.name
user.name=conatinerQa bot
user.email=ContBont@qa.com
```

Will fix

judovana:
> Will fix

Fixed. And also found a culprint. TY!

judovana:
> > @sxa - Does removing/changing the "--with-freetype=/usr/local/" bit in solaris.sh fall under the scope of this PR?
> 
> Yeah that's broken too so will need to be changed. For [java version reasons](https://github.com/adoptium/temurin-build/blob/master/SOLARIS.md) Solaris is no longer built in the main pipelines but is built by https://github.com/adoptium/temurin-build/blob/master/sbin/solaris/build-simple.sh
> 
> We could either modify that script to force the new options into the configure args, or do it in the jenkins jobs themselves. Depends how messy it would get :-)


@sxa I had probably never used `--with-freetype=/usr/local/` but I think soalris shold now use the default  bundled. 
That probably means, the `--with-freetype=/usr/local/` and the whole path `setFreeTypeFromInstalled` probably should be removed?

Guess I correctly it is 8 only?

judovana:
> > > @sxa - Does removing/changing the "--with-freetype=/usr/local/" bit in solaris.sh fall under the scope of this PR?
> > 
> > 
> > Yeah that's broken too so will need to be changed. For [java version reasons](https://github.com/adoptium/temurin-build/blob/master/SOLARIS.md) Solaris is no longer built in the main pipelines but is built by https://github.com/adoptium/temurin-build/blob/master/sbin/solaris/build-simple.sh
> > We could either modify that script to force the new options into the configure args, or do it in the jenkins jobs themselves. Depends how messy it would get :-)
> 
> @sxa I had probably never used `--with-freetype=/usr/local/` but I think soalris shold now use the default bundled. That probably means, the `--with-freetype=/usr/local/` and the whole path `setFreeTypeFromInstalled` probably should be removed?
> 
> Guess I correctly it is 8 only?

ok, is not...
```
  export CUPS="--with-cups=/opt/sfw/cups"
  export MEMORY=4000
elif [ "${ARCHITECTURE}" == "sparcv9" ]; then
  export CUPS="--with-cups=/opt/csw/lib/ --with-cups-include=/usr/local/cups-1.5.4-src"
  export FREETYPE="--with-freetype=/usr/local/"
  export MEMORY=16000
fi
```

No idea why it is there:(  Would jsut remove it...

sxa:
> Guess I correctly it is 8 only?

Yeah Solaris is only built on jdk8. Adoptium doesn't have a suitable compiler for building 11 which is the only other LTS version that works on Solaris from the upstream codebase.

sxa:
@andrew-m-leonard Are there any implications of us having `--skip-freetype` in the `BUILD_ARGS`?
(Which I think just involves us downloading and building our own from source before the build)

EDIT: It looks like when we run through our `make-adopt-build-farm.sh` (tested on my laptop with the centos6 build image) it's coming up with
```
Using freetype: system
```
instead of what I'm getting from `makejdk-any-platform.sh`:
```
Using freetype: bundled
```
which we might need to do some checking of before merging this

sxa:
Running checks in the following jobs (Solaris ones [needed this playbook fix](https://github.com/adoptium/infrastructure/pull/4156) too) so we can see what the build output looks like on our systems with the full build Adoptium scripts:
- [Windows Server 2022](https://ci.adoptium.net/job/VagrantPlaybookCheck/2213/OS=Win2022,label=vagrant/console): FAILED on the VS2013 install (Unrelated to this PR) [Re-running with FASTMODE](https://ci.adoptium.net/job/VagrantPlaybookCheck/OS=Win2022,label=vagrant/2214/)
- [CentOS6/x64](https://ci.adoptium.net/job/VagrantPlaybookCheck/2218/OS=CentOS6,label=vagrant/console): PASSED
- [Solaris](https://ci.adoptium.net/job/VagrantPlaybookCheck/2217/OS=Solaris10,label=vagrant/console): FAILED

Solaris build failed with:
```
ld: fatal: file /usr/sfw/lib/libfreetype.so: wrong ELF class: ELFCLASS32
ld: fatal: file /usr/sfw/lib/libfreetype.so: wrong ELF class: ELFCLASS32
ld: fatal: file processing errors. No output written to /export/home/local/vagrant/openjdk-build/workspace/build/src/build/solaris-x86_64-normal-server-release/jdk/lib/amd64/libfontmanager.so
gmake[2]: *** [lib/Awt2dLibraries.gmk:958: /export/home/local/vagrant/openjdk-build/workspace/build/src/build/solaris-x86_64-normal-server-release/jdk/lib/amd64/libfontmanager.so] Error 2
```

sxa:
Also noting that the for the r[egular temurin builds of `jdk8u482-b02_adopt`](https://ci.adoptium.net/job/build-scripts/job/openjdk8-pipeline/3040/console) before this fix is applied the Linux platforms ran through ok but [Win64](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x64-temurin/588/console), [Win32](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x86-32-temurin/529/console), [mac/x64](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-mac-x64-temurin/575/console) and [AIX](https://ci.adoptium.net/job/sbom_validator_job/155/console) all had issues relatign to Freetype.

Most of those failed the build step (not always for the same reason!) but AIX failed the SBOM validation step:
```
ERROR: FreeType version not 2.8.0 (SBOM has Unknown)
FREETYPE is Unknown
```

judovana:
> @andrew-m-leonard Are there any implications of us having `--skip-freetype` in the `BUILD_ARGS`? (Which I think just involves us downloading and building our own from source before the build)

It should still be valid for older jdk8 (and possibly 9 and 10, although they do not build anyway), or not? Am moreover trying to understand what all I could miss.

> 
> EDIT: It looks like when we run through our `make-adopt-build-farm.sh` (tested on my laptop with the centos6 build image) it's coming up with

That script do not seem to adjust freetype:( Unless ` ${CONFIGURE_ARGS}` brings it from elsewhere?
> 
> ```
> Using freetype: system
> ```
> 
> instead of what I'm getting from `makejdk-any-platform.sh`:
> 
> ```
> Using freetype: bundled
> ```
> 
> which we might need to do some checking of before merging this

THat would be major divergence which should not be there. Thanx for noticing it.

judovana:
> Also noting that the for the r[egular temurin builds of `jdk8u482-b02_adopt`](https://ci.adoptium.net/job/build-scripts/job/openjdk8-pipeline/3040/console) before this fix is applied the Linux platforms ran through ok but [Win64](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x64-temurin/588/console), [Win32](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-windows-x86-32-temurin/529/console), [mac/x64](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-mac-x64-temurin/575/console) and [AIX](https://ci.adoptium.net/job/sbom_validator_job/155/console) all had issues relatign to Freetype.
> 
> Most of those failed the build step (not always for the same reason!) but AIX failed the SBOM validation step:
> 
> ```
> ERROR: FreeType version not 2.8.0 (SBOM has Unknown)
> FREETYPE is Unknown
> ```

The logs throws 404. I would b ver interested in final arguments. The sbom failure is expected, as the build should pick up bundled 2.9.  But the fact that build passed keeps me wondering where is the catch.   Were there any built artifacts at all? I guess yes, but false positivum may be also the reason.  I'm still very nervous from swapping the two work-flow lines.

sxa:
> The logs throws 404.

Can you clarify which link is giving a 404 - all the ones in my previous comment should be publicly accessible 🤔 

> Were there any built artifacts at all?

From Linux? Yep! https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-linux-aarch64-temurin/521/artifact/workspace/target/ is an example of the job with artifacts from a successful jdk8u482-b02 on Linux/aarch64

I haven't gone through it in detail but noting that we set `BUILD_ARGS` to have `--skip-freetype` in the Temurin build scripts on Linux which _may_ be impacting the behaviour here
Ref: https://github.com/adoptium/temurin-build/blob/a44863d150ed2303b6833a02d313b7defb8ab15f/build-farm/platform-specific-configurations/linux.sh#L26

judovana:
> > The logs throws 404.
> 
> Can you clarify which link is giving a 404 - all the ones in my previous comment should be publicly accessible 🤔

https://ci.adoptium.net/job/build-scripts/job/openjdk8-pipeline/3040/console => 404 others works
> 
> > Were there any built artifacts at all?
> 
> From Linux? Yep! https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-linux-aarch64-temurin/521/artifact/workspace/target/ is an example of the job with artifacts from a successful jdk8u482-b02 on Linux/aarch64

Good :)
> 
> I haven't gone through it in detail but noting that we set `BUILD_ARGS` to have `--skip-freetype` in the Temurin build scripts on Linux which _may_ be impacting the behaviour here Ref:
> 
> https://github.com/adoptium/temurin-build/blob/a44863d150ed2303b6833a02d313b7defb8ab15f/build-farm/platform-specific-configurations/linux.sh#L26

Investigating. Thanx!

sxa:
> > > The logs throws 404.

> > Can you clarify which link is giving a 404 - all the ones in my previous comment should be publicly accessible 🤔

> https://ci.adoptium.net/job/build-scripts/job/openjdk8-pipeline/3040/console => 404 others works

Thanks - raised https://github.com/adoptium/ci-jenkins-pipelines/issues/1327 to investigate

judovana:
> Investigating. Thanx!

Ok, the conclusion is quite interesting.  When you use `--skip-freetype` it is changing the build to use defaults, thats why it passed on the systems which were using it. It would build both pre bundled_freetype and also post one.
In the scope of temurin build "configure" it is "jsut" `BUILD_CONFIG[FREETYPE]="true/false"` , however the output is much more interesting:
### no switch; old jdk:
configure run with `--with-freetype=/home/jvanek/git/temurin-build/workspace/./build//installedfreetype`
resulting to
```
checking for cups/ppd.h... yes
configure: Found freetype include files at /home/jvanek/git/temurin-build/workspace/./build//installedfreetype/include using --with-freetype
checking for freetype includes... /home/jvanek/git/temurin-build/workspace/build/installedfreetype/include
checking for freetype libraries... /home/jvanek/git/temurin-build/workspace/build/installedfreetype/lib
checking if we can compile and link with freetype... yes
checking if we should bundle freetype... yes
checking fontconfig/fontconfig.h usability... yes
```
### --skip-freetype switch; old jdk
configure run without any freetype, so run to defaults, which is
```
checking for cups/ppd.h... yes
checking for FREETYPE... yes
checking for freetype... yes (using pkg-config)
checking if we can compile and link with freetype... yes
checking if we should bundle freetype... no
checking fontconfig/fontconfig.h usability... yes
```
### no switch; new jdk
configure run with ` --with-freetype=bundled`
leading of course to
```
......
checking for cups/ppd.h... yes
Using freetype: bundled
checking fontconfig/fontconfig.h usability... yes
......
```
### --skip-freetype switch; new jdk
configure run without any freetype, so run to defaults, which is, however
```
checking for cups/ppd.h... yes
checking for FREETYPE... yes
checking for freetype... yes (using pkg-config)
Using freetype: system
checking fontconfig/fontconfig.h usability... yes
```

where new jdk refers to anything newer then 482-b01

So although it looks scary, when summed up:
 * external; built by temurin-build or cached - set for no switch; old jdk
 * system; for --skip-freetype switch; old jdk
 * bundled; for no switch; new jdk
 * system; for  --skip-freetype switch; new jdk
 (please review above)

It seems, that this behavior is correct, as for **jdk21 with** `--skip-freetype`  can see:
`BUILD_CONFIG[FREETYPE]="false"` 
`Using freetype: system`
 and  no "freetype" in configure . 

**Jdk21 without** it, I can see:
`BUILD_CONFIG[FREETYPE]="true"`
`Using freetype: bundled`
and with `--with-freetype=bundled`  in configure

So - in my opinion - no action is needed for this patch,  but every wrapper of build.sh (eventually ./makejdk-any-platform.sh/prepareWorkspace.sh) should drop the  `--skip-freetype` :

<pre>
<font color="#A347BA">build-farm/platform-specific-configurations/mac.sh</font><font color="#2AA1B3">:</font>    export BUILD_ARGS=&quot;${BUILD_ARGS} --<font color="#C01C28"><b>skip-freetype</b></font>&quot;
<font color="#A347BA">build-farm/platform-specific-configurations/alpine-linux.sh</font><font color="#2AA1B3">:</font>  export BUILD_ARGS=&quot;${BUILD_ARGS} --<font color="#C01C28"><b>skip-freetype</b></font>&quot;
<font color="#A347BA">build-farm/platform-specific-configurations/solaris.sh</font><font color="#2AA1B3">:</font>export BUILD_ARGS=&quot;${BUILD_ARGS} --<font color="#C01C28"><b>skip-freetype</b></font> --make-args SHELL=/bin/bash&quot;
<font color="#A347BA">build-farm/platform-specific-configurations/aix.sh</font><font color="#2AA1B3">:</font>  export BUILD_ARGS=&quot;${BUILD_ARGS} --<font color="#C01C28"><b>skip-freetype</b></font>&quot;
<font color="#A347BA">build-farm/platform-specific-configurations/windows.sh</font><font color="#2AA1B3">:</font>      export BUILD_ARGS=&quot;${BUILD_ARGS} --<font color="#C01C28"><b>skip-freetype</b></font>&quot;
<font color="#A347BA">build-farm/platform-specific-configurations/windows.sh</font><font color="#2AA1B3">:</font>      export BUILD_ARGS=&quot;${BUILD_ARGS} --<font color="#C01C28"><b>skip-freetype</b></font>&quot;
<font color="#A347BA">build-farm/platform-specific-configurations/windows.sh</font><font color="#2AA1B3">:</font>      export BUILD_ARGS=&quot;${BUILD_ARGS} --<font color="#C01C28"><b>skip-freetype</b></font>&quot;
<font color="#A347BA">build-farm/platform-specific-configurations/windows.sh</font><font color="#2AA1B3">:</font>      export BUILD_ARGS=&quot;${BUILD_ARGS} --<font color="#C01C28"><b>skip-freetype</b></font>&quot;
<font color="#A347BA">build-farm/platform-specific-configurations/linux.sh</font><font color="#2AA1B3">:</font>  export BUILD_ARGS=&quot;${BUILD_ARGS} --<font color="#C01C28"><b>skip-freetype</b></font>&quot;
<font color="#A347BA">CHANGELOG.md</font><font color="#2AA1B3">:</font>1. `-sf` changes to `-F`, (long form `--<font color="#C01C28"><b>skip-freetype</b></font>` stays the same).
<font color="#A347BA">makejdk-any-platform.1</font><font color="#2AA1B3">:</font>.BR \-F &quot;, &quot; \-\-<font color="#C01C28"><b>skip-freetype</b></font>
<font color="#A347BA">README.md</font><font color="#2AA1B3">:</font>-F, --<font color="#C01C28"><b>skip-freetype</b></font>
<font color="#A347BA">sbin/common/config_init.sh</font><font color="#2AA1B3">:</font>        &quot;--<font color="#C01C28"><b>skip-freetype</b></font>&quot; | &quot;-F&quot; )
<font color="#A347BA">cyclonedx-lib/build.xml</font><font color="#2AA1B3">:</font>                  &lt;arg value=&quot;--clean-git-repo --jdk-boot-dir /usr/lib/jvm/jdk-16 --configure-args  --disable-warnings-as-errors --enable-ccache --enable-dtrace --target-file-name OpenJDK17-jdk_x64_linux_hotspot_17_35.tar.gz --release --clean-libs --tag jdk-17+35_adopt --<font color="#C01C28"><b>skip-freetype</b></font> --use-jep319-certs --create-debug-image --build-variant hotspot jdk17&quot;/&gt;
<font color="#A347BA">cyclonedx-lib/build.xml</font><font color="#2AA1B3">:</font>                  &lt;arg value=&quot;--clean-git-repo --jdk-boot-dir /usr/lib/jvm/jdk-16 --configure-args  --disable-warnings-as-errors --enable-ccache --enable-dtrace --target-file-name OpenJDK17-jdk_x64_linux_hotspot_17_35.tar.gz --release --clean-libs --tag jdk-17+35_adopt --<font color="#C01C28"><b>skip-freetype</b></font> --use-jep319-certs --create-debug-image --build-variant hotspot jdk17&quot;/&gt;
</pre>

In addition the `--skip-freetype`  in **README.md** and   **makejdk-any-platform.1** should be adjusted, as it actually swap bundled/system freetype, and thus (probably) deprecated, as the image would not be portable.

WDYT?

judovana:
> > Investigating. Thanx!

While studying above, I had noticed one more possible issue. The `--freetype-dir` do not seem to work as expected.

 * On "old-jdk8" the docs man/readme lies -  as using the  `--skip-freetype` leads to use system freetype,  the --freetype-dir is ignored (which is in ooposite to docs, which says
```
       -f, --freetype-dir <path>
              specify the location of an existing FreeType  library.   This  is
              typically used in conjunction with <-F>.
...
       -F, --skip-freetype
              skip building Freetype automatically.  This is typically used  in
              conjunction with <-f>.
```

When `--skip-freetype` is omitted, the `--freetype-dir` works as expected.

 * On "new-jdk8" and on 11+the usage of `-freetype-dir` is leading to imminent configure  error
   * with added `--skip-freetype` its useless as before, as system is used.


My proposal is to terminate build  asap, once `--freetype-dir` in jdk with in-tree freetype is used, fix the docs (replace `  This is typically used  in conjunction with <-f>.`  by ` This will lead to dynamic linking of system freetype` and deprecate `-f`  as it is only for non-freetype-bundling jdks.
Similarly,  it if the "old jdk8" will detect usage of   `--skip-freetype` and  `--freetype-dir`  it will halt, with verbose explanation. And + the `--skip-freetype explaining system/bundled` as mentioned earlier. 

Unlike various `platform-specific-configurations`, `CHANGELOG.md` and  build.xml` changes,  where I hesitate, this should be part of the pr.. WDYT?

judovana:
> So - in my opinion - no action is needed for this patch, but every wrapper of build.sh (eventually ./makejdk-any-platform.sh/prepareWorkspace.sh) should drop the `--skip-freetype` :
> 
> build-farm/platform-specific-configurations/mac.sh:    export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
> build-farm/platform-specific-configurations/alpine-linux.sh:  export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
> build-farm/platform-specific-configurations/solaris.sh:export BUILD_ARGS="${BUILD_ARGS} --skip-freetype --make-args SHELL=/bin/bash"
> build-farm/platform-specific-configurations/aix.sh:  export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
> build-farm/platform-specific-configurations/windows.sh:      export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
> build-farm/platform-specific-configurations/windows.sh:      export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
> build-farm/platform-specific-configurations/windows.sh:      export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
> build-farm/platform-specific-configurations/windows.sh:      export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
> build-farm/platform-specific-configurations/linux.sh:  export BUILD_ARGS="${BUILD_ARGS} --skip-freetype"
> CHANGELOG.md:1. `-sf` changes to `-F`, (long form `--skip-freetype` stays the same).
> makejdk-any-platform.1:.BR \-F ", " \-\-skip-freetype
> README.md:-F, --skip-freetype
> sbin/common/config_init.sh:        "--skip-freetype" | "-F" )
> cyclonedx-lib/build.xml:                  <arg value="--clean-git-repo --jdk-boot-dir /usr/lib/jvm/jdk-16 --configure-args  --disable-warnings-as-errors --enable-ccache --enable-dtrace --target-file-name OpenJDK17-jdk_x64_linux_hotspot_17_35.tar.gz --release --clean-libs --tag jdk-17+35_adopt --skip-freetype --use-jep319-certs --create-debug-image --build-variant hotspot jdk17"/>
> cyclonedx-lib/build.xml:                  <arg value="--clean-git-repo --jdk-boot-dir /usr/lib/jvm/jdk-16 --configure-args  --disable-warnings-as-errors --enable-ccache --enable-dtrace --target-file-name OpenJDK17-jdk_x64_linux_hotspot_17_35.tar.gz --release --clean-libs --tag jdk-17+35_adopt --skip-freetype --use-jep319-certs --create-debug-image --build-variant hotspot jdk17"/>
> 
> In addition the `--skip-freetype` in **README.md** and **makejdk-any-platform.1** should be adjusted, as it actually swap bundled/system freetype, and thus (probably) deprecated, as the image would not be portable.
> 
For tuning, initiated: https://github.com/adoptium/temurin-build/pull/4320

judovana:
> docs (replace ` This is typically used in conjunction with <-f>.` by ` This will lead to dynamic linking of system freetype` and deprecate `-f` as it is only for non-freetype-bundling jdks. Similarly, it if the "old jdk8" will detect usage of `--skip-freetype` and `--freetype-dir` it will halt, with verbose explanation. And + the `--skip-freetype explaining system/bundled` as mentioned earlier.

hints welcomed! https://github.com/adoptium/temurin-build/pull/4321

sxa:
> In addition the --skip-freetype in README.md and makejdk-any-platform.1 should be adjusted, as it actually swap bundled/system freetype, and thus (probably) deprecated, as the image would not be portable.

I would tentatively agree with all of your analysis :-)

I note that it looks like we use `--skip-freetype` on all versions before 21. Would it be correct to make this change only on jdk8 or would we be looking at it for 8 and 11 too? I guess we'd need to be certain on why we initially added this option in. Looking back, @andrew-m-leonard did some analysis in https://github.com/adoptium/temurin-build/issues/3504 (and his [associated PR](https://github.com/adoptium/temurin-build/pull/3557) added the bits to make it act differently on jdk21+) so I suggest he takes another look at this PR in the context of the changes he made at that time. 

I feel we should definitely adjust the `build-farm/platform-specific-configurations` files under this PR too to ensure that Temurin is building for each version/platform with the options we want it it.

judovana:
> > In addition the --skip-freetype in README.md and makejdk-any-platform.1 should be adjusted, as it actually swap bundled/system freetype, and thus (probably) deprecated, as the image would not be portable.
> 
> I would tentatively agree with all of your analysis :-)
> 
> I note that it looks like we use `--skip-freetype` on all versions before 21. Would it be correct to make this change only on jdk8 or would we be looking at it for 8 and 11 too? I guess we'd need to be certain on why we initially added this option in. Looking back, @andrew-m-leonard did some analysis in #3504 (and his [associated PR](https://github.com/adoptium/temurin-build/pull/3557) added the bits to make it act differently on jdk21+) so I suggest he takes another look at this PR in the context of the changes he made at that time.
> 

Thanx for the info!
The https://bugs.openjdk.org/browse/JDK-8193017 is before jdk17 forked. You may see that eg https://bugs.openjdk.org/browse/JDK-8348596 is backoprted everywhere, except 8, were I had jsut not yet requested it (waiting for https://github.com/openjdk/jdk8u-dev/pull/720 to be merged before requesting it)

So I think this change should be global. But there may of course be a reasons I'm not awaere off:(

> I feel we should definitely adjust the `build-farm/platform-specific-configurations` files under this PR too to ensure that Temurin is building for each version/platform with the options we want it it.

I had opened it in separate PR (https://github.com/adoptium/temurin-build/pull/4320) , but i have no objections to includ eit here. I put it separately, because it is affecting also 11 and 17.

andrew-m-leonard:
The move to "bundled" I did back in 2023, was to support reproducible builds, so as to ensure deterministic freetype binary.

sxa:
> The move to "bundled" I did back in 2023, was to support reproducible builds, so as to ensure deterministic freetype binary.

Is there a particular reason that change wasn't done across all versions (even though they wouldn't have been reproducible)? That would have seemed the obvious thing to do which makes me nervous that there was some specific problem that was potentially going to be caused.

andrew-m-leonard:
> > The move to "bundled" I did back in 2023, was to support reproducible builds, so as to ensure deterministic freetype binary.
> 
> Is there a particular reason that change wasn't done across all versions (even though they wouldn't have been reproducible)? That would have seemed the obvious thing to do which makes me nervous that there was some specific problem that was potentially going to be caused.

From the PMC minutes discussing this back in Nov 22nd 2023:
For reproducible and consistency purposes we would prefer to go with bundling the freetype library. If a user is used to using fonts provided by their system freetype they may notice a difference with this change
Plan to start with JDK21 on moving to bundled fontlib before rolling out to earlier versions.
Don’t expect to see problems with OpenJDK support of the latest version of fontlib.
Question if this lack of reproducibility is known upstream, and if the fontlib is bundled by default. If not this should be raised there too.
DECISION: PMC agreed to start bundling fontlib as part of the build from JDK21 immediately, then backporting to earlier JDK version builds if this is successful.

andrew-m-leonard:
So the plan originally was to backport to earlier versions... but that was not done yet... So seems reasonable to do that now

judovana:
Thanx a lot for all that insight.  Highly appreciated!

judovana:
hi! Did run quite a few tests around the freetype switches on both  freetype bundled and external-needed jdks, and it seems it behave as it should now.

Ok t go by me ;)

Please, state your wish how to handle all related PRs:  https://github.com/adoptium/temurin-build/issues/4306#issuecomment-3584889700

I do not mind to include them in this, or as separate. Being separate have afaik many benefits, but it is up to you. W can even discuss them in separate PRs, and then move to this one.  Maybe even some logic (liek the checks may move elsewhere (eg to the man/reasme one, but  for that I'm more against, then for) . Note, that the configuration one, probably can not pass until this one is finished. Similarly the sbom one can not pass before 2.13.3 reaches jdk8u (already under review)

sxa:
> I do not mind to include them in this, or as separate. Being separate have afaik many benefits, but it is up to you.

I'm OK with the multi-PR approach :-) Thanks for all your work on this.

sbin/build.sh
```diff
@@ -548,22 +548,34 @@ configureAlsaLocation() {
...
 configureFreetypeLocation() {
   if [[ ! "${CONFIGURE_ARGS}" =~ "--with-freetype" ]]; then
     if [[ "${BUILD_CONFIG[FREETYPE]}" == "true" ]]; then
       local freetypeDir="${BUILD_CONFIG[FREETYPE_DIRECTORY]}"
-      if [[ "$OSTYPE" == "cygwin" ]] || [[ "$OSTYPE" == "msys" ]]; then
-        case "${BUILD_CONFIG[OPENJDK_CORE_VERSION]}" in
-          jdk8* | jdk9* | jdk10*) addConfigureArg "--with-freetype-src=" "${BUILD_CONFIG[WORKSPACE_DIR]}/libs/freetype" ;;
-          *) freetypeDir=${BUILD_CONFIG[FREETYPE_DIRECTORY]:-bundled} ;;
-        esac
+      if isFreeTypeInSources ; then
```
karianna, line 570: This reads oddly, if we detect the freetype lib in sources (e.g., `${BUILD_CONFIG[WORKSPACE_DIR]}/libs/freetype`), would we not use that?  I think the function may need renaming here.

sbin/prepareWorkspace.sh
```diff
@@ -850,19 +894,18 @@ downloadingRequiredDependencies() {
...
-          echo ""
-          echo "---> Skipping the process of checking and downloading the FreeType Font dependency, a pre-built version provided at ${BUILD_CONFIG[FREETYPE_DIRECTORY]} <---"
-          echo ""
-        fi
-      ;;
-      *) echo "Using bundled Freetype" ;;
-    esac
+    if [ "0${freeTypeInSources}" -ne 0 ]  ; then
+      if [ -z "${BUILD_CONFIG[FREETYPE_DIRECTORY]}" ]; then
+        echo "Checking and download FreeType Font dependency"
```
karianna:
```suggestion
        echo "Checking and downloading FreeType Font dependency"
```

sbin/prepareWorkspace.sh
```diff
@@ -850,19 +894,18 @@ downloadingRequiredDependencies() {
...
+        checkingAndDownloadingFreeType
+      else
+        echo ""
+        echo "---> Skipping the process of checking and downloading the FreeType Font dependency, a pre-built version provided at ${BUILD_CONFIG[FREETYPE_DIRECTORY]} <---"
```
karianna:
```suggestion
        echo "---> Skipping the process of checking and downloading the FreeType Font dependency, a pre-built version is provided at ${BUILD_CONFIG[FREETYPE_DIRECTORY]} <---"
```

sbin/prepareWorkspace.sh
```diff
@@ -947,10 +990,10 @@ createSourceTagFile(){
 function configureWorkspace() {
   if [[ "${BUILD_CONFIG[ASSEMBLE_EXPLODED_IMAGE]}" != "true" ]]; then
     createWorkspace
+    checkoutAndCloneOpenJDKGitRepo
```
karianna, line 993: Why did we move this up?

sbin/prepareWorkspace.sh
```diff
@@ -40,6 +40,48 @@ FREETYPE_FONT_SHARED_OBJECT_FILENAME="libfreetype.so*"
...
+   echo "--freetype-dir is set to system that is weird, but accepted. Try to use --skip-freetype"
+  elif [ -n "${BUILD_CONFIG[FREETYPE_DIRECTORY]}" ] ; then
+    echo "--freetype-dir is not accepted for JDK with bundled freetype."
+    exit 1
+  fi
+}
+
+checkNoBundledFreetypeJdkConfig() {
+  if [ "${BUILD_CONFIG[FREETYPE]}" = "false" ] && [ -n "${BUILD_CONFIG[FREETYPE_DIRECTORY]}" ] ; then
+    echo "--freetype-dir is declared together with --skip-freetype, that is invalid, and would build against system freetype"
```
karianna: I'm not sure what "and would build against system freetype" means

sbin/prepareWorkspace.sh
```diff
@@ -40,6 +40,48 @@ FREETYPE_FONT_SHARED_OBJECT_FILENAME="libfreetype.so*"
...
+
+checkBundledFreetypeJdkConfig() {
+  if [ "${BUILD_CONFIG[FREETYPE_DIRECTORY]}" = "bundled" ] ; then
+    if [ "${BUILD_CONFIG[FREETYPE]}" = "false" ] ; then
+      echo "--freetype-dir bundled is in contradiction with -skip-freetype"
+      exit 1
+    fi
+    echo "--freetype-dir is set to bundled that is weird, but accepted. It should be default."
+  elif [ "${BUILD_CONFIG[FREETYPE_DIRECTORY]}" = "system" ] ; then
+   echo "--freetype-dir is set to system that is weird, but accepted. Try to use --skip-freetype"
```
karianna:
```suggestion
   echo "--freetype-dir is set to system which is unusual, but accepted. Try to use --skip-freetype"
```

karianna: When we are saying "try to use --skip-freetype". I think we're really saying, please use --skip-freetype instead to get the same functionality?

sbin/build.sh
sxa: Similar to how our old "skip-freetype" option that we use doesn't mean "build without freetype" 😳😉

sxa: Similar to how our old "skip-freetype" option that we use doesn't mean "build without freetype" 😳😉

judovana: > This reads oddly, if we detect the freetype lib in sources (e.g., `${BUILD_CONFIG[WORKSPACE_DIR]}/libs/freetype`), would we not use that? I think the function may need renaming here.

Why? The  `isFreeTypeInSources` is doing what it should - testing whether the freetype is in sources, and checking configs appropriately based on it. See the https://github.com/adoptium/temurin-build/pull/4287#issuecomment-3582533466

To remove  function and replace eg by one-time check is tricky, because you need the jdk sources (or get to original mayhem with tag parsing or similar). To extend the name to `checkFreetypeInSourcesAndValdate` do not seem right, as the validation is consequence of check.

sbin/prepareWorkspace.sh
judovana: Thats the crucial part of being able to recognize, whether jdk have freetype in sources or not. To be able to find whether freetype is in sources, you need sources.  And many other behaviours - like downloadig or building  of dependences - depends on it. Again see the  https://github.com/adoptium/temurin-build/pull/4287#issuecomment-3582533466 and  https://github.com/adoptium/temurin-build/pull/4287#issuecomment-3584716770.

Before it, it was "minimalistic change", now it is not. But at the end several interesting corner cases were found and fixed.

If yuo have any suspicions this swap can break something, please tell! But then I'm afraid the change must go to imperfect tag checking or even drop it at all, as it was in uttermost beginning  . TY!

sbin/prepareWorkspace.sh
judovana: `${BUILD_CONFIG[FREETYPE_DIRECTORY]}` == `--freetype-dir` - unless set to bundled or system  (hopefully handled elsewhere in the `if`s)  is configuring the build to build against the given dir which contains freetype.
`${BUILD_CONFIG[FREETYPE]}` ==  `--skip-freetype` - is forcng the build to build agaisnt system freetype.

So having `--skip-freetype=true`  and  `--freetype-dir=path or bundled` is invalid combination, because one is forcing system, the other bundled or exact. Please see the https://github.com/adoptium/temurin-build/pull/4287#issuecomment-3582533466 and https://github.com/adoptium/temurin-build/pull/4287#issuecomment-3584716770.  It would answer most of the question

 This check is for version without freetype in sources, so bundled is not option here., and  thus `--freetype-dir` can be only dir. Thus `--skip-freetype` is invalidating it, and without worning, the build will happen against SYSTEM jdk (as `--skip-freetype` is forcing), instead to `dir` as you intended with `--freetype-dir` . I think the error message is thus correct.

sbin/prepareWorkspace.sh
judovana: correct.

judovana: Small adjust ion: https://github.com/adoptium/temurin-build/pull/4287/commits/5e63e97c6f8c40fbf5c5c5943c83eec07942907d

sbin/build.sh
judovana: @karianna  Please, note also related PRs: https://github.com/adoptium/temurin-build/issues/4306#issuecomment-3584889700

karianna: Its confusing because it reads that the test is for detecting that Freetype is available from the sources but then we select the setDefaultFreeType as opposed to the setFreeTypeFromSrcs.

judovana: I see. The naming of the worker methods is indeed legacy. The  `setDefaultFreeType`" was default, becasue it was the main case in original logic. The `setFreeTypeFromSrcs` was the only one setting any sources - because  it was in jdk8 block. Where in reality it was setting external sources. Renamed:

```
    setDefaultFreeType->setBundledFreeType
    setFreeTypeFromSrcs->setFreeTypeFromExternalSrcs
```

Wdyt?

andrew-m-leonard: sounds reasonable to me

andrew-m-leonard:
SBOM validation just needs fixing up now:
- Mac - https://ci.adoptium.net/job/sbom_validator_job/339/console
- Windows - https://ci.adoptium.net/job/sbom_validator_job/341/console
- AIX - https://ci.adoptium.net/job/sbom_validator_job/342/console

You should just need to fixup the values in here https://github.com/adoptium/temurin-build/blob/master/tooling/validateSBOMcontent.sh

judovana:
> SBOM validation fixup

Hi @andrew-m-leonard  Sorry if ti caused any troubles, I made the sbom validation changes in different PR:  https://github.com/adoptium/temurin-build/pull/4315  (there is more related changes, see: https://github.com/adoptium/temurin-build/pull/4287#issuecomment-3589700911  and  https://github.com/adoptium/temurin-build/issues/4306#issuecomment-3584889700)

As stated above, I can merge them all here, or do it separately.  But seems like theirs nature is more close to splitting them up. especially as the [ideal sbom change](https://github.com/adoptium/temurin-build/pull/4315) depends on [version of freetype in jdk8](https://github.com/openjdk/jdk8u-dev/pull/728) and on [adoptium build configs](https://github.com/adoptium/temurin-build/pull/4320)

I may - maybe - to do some minimalistic chnage to sbom here, just for  jdk8u482-b02 purposes, but it may lead jsut to more future work later. Also the sbom "if's" are so complex that I may simply miss the correct execution branch

andrew-m-leonard (approved):

sbin/build.sh
gnu-andrew: For reference (I see you changed to checking for the directory), the version is available in `common/autoconf/version-numbers` e.g. see https://github.com/openjdk/jdk8u/commit/c0a36d8fbe59b6fcd878cf14fed2c28c1c30c6eb  You don't need to rely on Git metadata for this. Checking for what you actually want rather than a particular version is always preferable though, if possible.

judovana: That would require again to clone the repo, before all those checks. That was something I originally really wanted to avoid.

PR #4289, opened 2025-10-28
Add build recipe generation to build.sh and integrate new workflows CLI into sbom.sh

Lukisorisch:
# Changes

## build.sh

New code that generates the build recipe for the workflows section of the SBOM.

Add function addTemurinBuildRecipeToSBOM, which is responsible for generating the workflow/recipe section in the SBOM. In detail, it:

1. Reads `makejdk-any-platform.args` from `workspace/config/` to get the original makejdk-any-platform.sh flags.

2. Reads `metadata/buildSource.txt` from the target directory to recover the GitHub repository URL and the exact commit hash used by the build.

3. Uses the reproducible build timestamp from `BUILD_REPRODUCIBLE_DATE` (or `BUILD_TIMESTAMP` if the first does not exist) for the `--build-reproducible-date '<timestamp>'` flag. (Only add if not present already)

4. Reads `metadata/BUILD_ARGS` and if it finds a `--use-adoptium-devkit` flag, extracts the devkit tag and adds `-C --use-adoptium-devkit <tag>` to the makejdk-any-platform command.

See below for an example and general recipe output.

## sbom.sh

Add wrappers around the TemurinGenSBOM CLI so shell code can author formulation.workflows:

workflows
```
addSBOMWorkflow <javaHome> <classpath> <jsonFile> <formulaRef> <workflowRef> <workflowUid> <workflowName> <taskTypes>
```

workflows.steps
```
addSBOMWorkflowStep <javaHome> <classpath> <jsonFile> <formulaRef> <workflowRef> <stepName> <description>
```

workflows.steps.commands
```
addSBOMWorkflowStepCmd <javaHome> <classpath> <jsonFile> <formulaRef> <workflowRef> <stepName> <executed>
```

Each wrapper calls the respective methods in TemurinGenSBOM.java with the right flags.

## Verification / OS support:

- [x] **Linux:** Verified (Produced a 100% reproducible build)
- [x] **Windows:** Verified (Produced a 100% reproducible build)
- [ ] **macOS:** Untested

## Example:

```
{
      "bom-ref" : "formula_temurin_build_script_21.0.10-beta+5-202512031805",
      "workflows" : [
        {
          "bom-ref" : "workflow_temurin_build_script_21.0.10-beta+5-202512031805",
          "uid" : "workflow_temurin_build_script_21.0.10-beta+5-202512031805",
          "name" : "Temurin Build Script",
          "taskTypes" : [
            "clone",
            "build"
          ],
          "steps" : [
            {
              "name" : "clone repo",
              "description" : "clone repository",
              "commands" : [
                {
                  "executed" : "git clone https://github.com/adoptium/temurin-build.git"
                }
              ]
            },
            {
              "name" : "cd into repository",
              "description" : "cd into temurin-build and checkout commit",
              "commands" : [
                {
                  "executed" : "cd temurin-build"
                },
                {
                  "executed" : "git checkout f4bcbeb80b2dc0a624721f09a91c925e6df2bc90"
                }
              ]
            },
            {
              "name" : "makejdk",
              "description" : "execute makejdk-any-platform.sh",
              "commands" : [
                {
                  "executed" : "bash ./makejdk-any-platform.sh --build-reproducible-date '2025-12-03 18:05:29' --jdk-boot-dir download --build-variant temurin -C --disable-ccache -C --disable-warnings-as-errors -T jdk-hotspot.tar.gz --create-sbom jdk21u"
                }
              ]
            }
          ]
        }
      ]
    }
```

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -1183,116 +1183,6 @@ generateSBoM() {
   addCycloneDXVersions
 
   # Generate the Workflow part containing the Build Recipe
-  addTemurinBuildRecipeToSBOM() {
-
-    if [[ -z "${fullVer}" || -z "${sbomJson}" ]]; then
-      echo "WARNING: 'fullVer' or 'sbomJson' file missing, cannot generate build recipe."
-      return 0
-    fi
-
-    local formulaName="formula_temurin_build_script_${fullVer}"
-    local workflowRef="workflow_temurin_build_script_${fullVer}"
-    local workflowUid="${workflowRef}"
-    local workflowName="Temurin Build Script"
-    local taskTypes="clone,build"
-
-    # Read makejdk-any-platform args
-    local makejdk_args_file="${BUILD_CONFIG[WORKSPACE_DIR]}/config/makejdk-any-platform.args"
-    if [[ ! -s "${makejdk_args_file}" ]]; then
-      echo "WARNING: makejdk-any-platform args file '${makejdk_args_file}' missing or empty, skipping build recipe generation." 1>&2
-      return 0
-    fi
-
-    local makejdk_args
-    makejdk_args="$(< "${makejdk_args_file}")"
-
-    # If there is one, replace the relative boot-jdk path with "download", since
-    # the bootjdk is probably in another directory or not even existent when running
-    # a reproducible build with this recipe.
-    # i.e. from --jdk-boot-dir <path> to --jdk-boot-dir download
-    makejdk_args="$(printf '%s\n' "${makejdk_args}" | sed -E 's/--jdk-boot-dir[[:space:]]+[^[:space:]]+/--jdk-boot-dir download/g')"
-
-    # Replace all <\"> and <"> with <'>.
-    # We need this to happen because double quotation marks need escaping
-    # and end up as \" in the SBoM, which bash confuses as a new line when
-    # running the recipe by copy pasting it in.
-    # i.e. from --build-reproducible-date "<date>" to --build-reproducible-date '<date>'
-    # and from --configure-args \" <args> \" to --configure-args '<args>'
-    makejdk_args="$(printf '%s\n' "${makejdk_args}" | sed -E 's/\\?"/'\''/g')"
-
-    # Git-Metadata i.e. buildSource.txt (Repo + Commit)
-    local build_src_file="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/buildSource.txt"
-    if [[ ! -s "${build_src_file}" ]]; then
-      echo "WARNING: buildSource metadata file '${build_src_file}' missing or empty, skipping build recipe generation." 1>&2
-      return 0
-    fi
-
-    local build_src_url
-    build_src_url="$(< "${build_src_file}")"
-
-    # Parse URL and account for different formats
-    local sha="${build_src_url##*/}"
-    local base_url="${build_src_url%%/commit/*}"
-    local repo_name="${base_url##*/}"
-    local removed_repo="${base_url%/*}"
-    local org_name="${removed_repo##*[/:]}"
-
-    local clone_url="https://github.com/${org_name}/${repo_name}.git"
-
-    # Build Timestamp
-    local buildStamp
-    buildStamp="${BUILD_CONFIG[BUILD_REPRODUCIBLE_DATE]:-${BUILD_CONFIG[BUILD_TIMESTAMP]:-}}"
-
-    # Normalise buildStamp: remove any surrounding " or ' to add single quotation marks later
-    buildStamp="${buildStamp%\"}"
-    buildStamp="${buildStamp#\"}"
-    buildStamp="${buildStamp%\'}"
-    buildStamp="${buildStamp#\'}"
-
-    # Get DevKit-Tag
-    local metadata_build_args_file="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/BUILD_ARGS"
-    local devkit_tag=""
-    if [[ -s "${metadata_build_args_file}" ]]; then
-      local build_args
-      build_args="$(< "${metadata_build_args_file}")"
-      # Search for --use-adoptium-devkit in build args
-      if [[ "${build_args}" =~ --use-adoptium-devkit[[:space:]]+([^[:space:]]+) ]]; then
-        devkit_tag="${BASH_REMATCH[1]}"
-      fi
-    fi
-    
-    # Build makejdk-any-platform command:
-    # Base: bash ./makejdk-any-platform.sh
-    # Add --build-reproducible-date <buildStamp> only if not in args already
-    # Add -C --use-adoptium-devkit <devkit_tag> only if not in args already
-    # Then concatenate the rest
-    local makejdk_cmd="bash ./makejdk-any-platform.sh"
-
-    if [[ -n "${buildStamp}" && ${makejdk_args} != *"--build-reproducible-date"* ]]; then
-      makejdk_cmd+=" --build-reproducible-date '${buildStamp}'"
-    fi
-
-    if [[ -n "${devkit_tag}" && ${makejdk_args} != *"--use-adoptium-devkit"* ]]; then
-      makejdk_cmd+=" -C --use-adoptium-devkit ${devkit_tag}"
-    fi
-
-    makejdk_cmd+=" ${makejdk_args}"
-
-    # Workflow
-    addSBOMWorkflow "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "${workflowUid}" "${workflowName}" "${taskTypes}"
-
-    # Steps
-    addSBOMWorkflowStep "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "clone repo" "clone repository"
-    addSBOMWorkflowStep "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "cd into repository" "cd into temurin-build and checkout commit"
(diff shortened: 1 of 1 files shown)


sbin/build.sh
```diff
@@ -1067,6 +1067,25 @@ generateSBoM() {
   # Add CycloneDX versions
   addCycloneDXVersions
 
+  local formulaName="formula_temurin_build_script_1.0_jdk21u"
```
andrew-m-leonard: We will need to construct these names based on the given builds VERSION etc...
I suggest something like:
```
local formulaName="formula_temurin_build_script_${fullVer}"
```
fullVer from https://github.com/adoptium/temurin-build/blob/f4bcbeb80b2dc0a624721f09a91c925e6df2bc90/sbin/build.sh#L1010C9-L1010C16

sbin/build.sh
```diff
@@ -1067,6 +1067,25 @@ generateSBoM() {
...
+  # Create workflow under the formula (formula/workflow are created if missing)
+  addSBOMWorkflow "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "${workflowUid}" "${workflowName}" "${taskTypes}"
+
+  # Steps
+  addSBOMWorkflowStep "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "clone repo" "clone repository"
+  addSBOMWorkflowStep "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "cd into repository" "cd into temurin-build"
+  addSBOMWorkflowStep "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "makejdk" "execute makejdk-anyplatform.sh"
+
+  # Commands
+  addSBOMWorkflowStepCmd "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "clone repo" "git clone git@github.com:adoptium/temurin-build"
```
andrew-m-leonard: The clone needs to be for the exact temurin-build "commit" that is used for this build, so that in running this recipe it will exactly reproduce this build.
Use ${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[TARGET_DIR]}/metadata/buildSource.txt from https://github.com/adoptium/temurin-build/blob/f4bcbeb80b2dc0a624721f09a91c925e6df2bc90/sbin/build.sh#L1128

sbin/build.sh
```diff
@@ -1067,6 +1067,25 @@ generateSBoM() {
...
+  addSBOMWorkflowStepCmd "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "cd into repository" "cd temurin-build"
+  addSBOMWorkflowStepCmd "${javaHome}" "${classpath}" "${sbomJson}" "${formulaName}" "${workflowRef}" "makejdk" "bash ./makejdk-any-platform.sh jdk21u --with-version-string=21.0.2+13-202312052047 --with-vendor-version-string=202312052047"
```
andrew-m-leonard: Need this builds full exact makejdk-any-platform.sh params...
https://github.com/adoptium/temurin-build/blob/f4bcbeb80b2dc0a624721f09a91c925e6df2bc90/sbin/build.sh#L1145

andrew-m-leonard:
I would suggest in trying to test this, is you do a local build on your laptop, save the built JDK1 tar.gz, then try and rebuild it using just the SBOM formulation recipe in a completely fresh woerkspace, then see if it is identical to JDK1 tar.gz ?

andrew-m-leonard:
Test build https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-linux-x64-temurin/402/

andrew-m-leonard:
@Lukisorisch You need to merge latest "master" into this to update...

sbin/build.sh
```diff
@@ -1172,6 +1172,119 @@ generateSBoM() {
...
+
+    if [[ -z "${fullVer}" || -z "${sbomJson}" ]]; then
+      echo "WARNING: 'fullVer' or 'sbomJson' file missing, cannot generate build recipe."
+      return 0
+    fi
+
+    local formulaName="formula_temurin_build_script_${fullVer}"
+    local workflowRef="workflow_temurin_build_script_${fullVer}"
+    local workflowUid="${workflowRef}"
+    local workflowName="temurin build script ${fullVer}"
```
andrew-m-leonard: I think the workflowName should be constant for the type of formula this is, so that someone who has a script looking for the "Temurin build formula" can find it in any given SBOM. The Ref and Uid will be unique...
See if you can find any examples online...?

Lukisorisch: Makes sense. I will change that. Thanks!

sbin/build.sh
```diff
@@ -1182,6 +1182,119 @@ generateSBoM() {
   # Add CycloneDX versions
   addCycloneDXVersions
 
+  # Generate the Workflow part containing the Build Recipe
+  addTemurinBuildRecipeToSBOM() {
+
+    if [[ -z "${fullVer}" || -z "${sbomJson}" ]]; then
+      echo "WARNING: 'fullVer' or 'sbomJson' file missing, cannot generate build recipe."
```
adamfarley: This seems to indicate that the file itself is missing, rather than us having an empty string.

Also, if the other warnings redirect stdout to stderr. Should we do so here as well?

```suggestion
      echo "WARNING: 'fullVer' or 'sbomJson' variable/s are empty. Cannot generate build recipe." 1>&2
```

sbin/build.sh
```diff
@@ -1182,6 +1182,119 @@ generateSBoM() {
   # Add CycloneDX versions
   addCycloneDXVersions
 
+  # Generate the Workflow part containing the Build Recipe
+  addTemurinBuildRecipeToSBOM() {
```
andrew-m-leonard: Can you move the function definition outside of the function this is in, as it makes this function hard to follow

sbin/build.sh
```diff
@@ -1182,6 +1182,119 @@ generateSBoM() {
...
+    fi
+
+    local makejdk_args
+    makejdk_args="$(< "${makejdk_args_file}")"
+
+    # If there is one, replace the relative boot-jdk path with "download", since
+    # the bootjdk is probably in another directory or not even existent when running
+    # a reproducible build with this recipe.
+    # i.e. from --jdk-boot-dir <path> to --jdk-boot-dir download
+    makejdk_args="$(printf '%s\n' "${makejdk_args}" | sed -E 's/--jdk-boot-dir[[:space:]]+[^[:space:]]+/--jdk-boot-dir download/g')"
```
adamfarley: This regex is good. I think it could be improved by adding the ability to handle paths with speech marks and spaces. For example:


```suggestion
    makejdk_args="$(printf '%s\n' "${makejdk_args}" | sed -E -e 's/--jdk-boot-dir[[:space:]]+"[^"]+"/--jdk-boot-dir download/g' -e 's/--jdk-boot-dir[[:space:]]+[^[:space:]]+/--jdk-boot-dir download/g')"
```

sbin/build.sh
Lukisorisch: Good catch! I updated the wording and added the redirection. Thank you!

sbin/build.sh
Lukisorisch: I moved the function definition to the other helper functions.

sbin/build.sh
Lukisorisch: Good point, I didn't account for that. I updated the regex as you suggested, thanks!

sbin/build.sh
```diff
@@ -1182,6 +1182,9 @@ generateSBoM() {
   # Add CycloneDX versions
   addCycloneDXVersions
 
+  # Generate the Workflow part containing the Build Recipe
+  addTemurinBuildRecipeToSBOM
```
adamfarley, line 1186: I think this needs to be moved below the function.

Lukisorisch: Are you sure the function call shouldn't be in `generateSBOM`? I moved the definition below the function but kept the call where it was.

adamfarley: I was concerned that calling a function before it was defined would cause problems in Bash.

It turns out that, since both functions are not called until further down the file, this is not a problem.

sbin/build.sh
```diff
@@ -1430,6 +1433,117 @@ addCycloneDXVersions() {
 
 # Below add versions to sbom | Facilitate reproducible builds
 
+# Generate the Workflow part containing the Build Recipe
+addTemurinBuildRecipeToSBOM() {
+
+  if [[ -z "${fullVer}" || -z "${sbomJson}" ]]; then
```
andrew-m-leonard, line 1439: These variable are "local" from the parent function, so really they should be passed in like the other addSBOM functions do..? (even though it seems to work, it could be problematic..)

I think at least the following need to be $ args:
"${javaHome}" "${classpath}" "${sbomJson}" "${fullVer}"

Lukisorisch: I completely agree! However, when I wrote the function I took inspiration from all the other helper functions called in generateSBoM and noticed that they (addGLIBCforLinux, addGCC, addALSAVersion etc) all are currently implemented with this exact scoping as well for ${javaHome}, ${classpath}, and ${sbomJson}. So without function arguments...

How do you want to proceed? I'm happy to open another PR and refactor all helper functions to be implemented with function parameters if you want

Lukisorisch: here for example

https://github.com/Lukisorisch/temurin-build/blob/e5209a425297abd76a37820cd3fc868a99e93855/sbin/build.sh#L1580

being called without parameters:

https://github.com/Lukisorisch/temurin-build/blob/e5209a425297abd76a37820cd3fc868a99e93855/sbin/build.sh#L1174

andrew-m-leonard: good spot, yeah you're right

andrew-m-leonard: ok I checked, and it seems in bash sub-functions can reference parent scoped local variables. 
So I am happy with it as-is.

andrew-m-leonard:
Windows CI failures look unrelated to this PR, more likely github runner Windows image update, raised https://github.com/adoptium/temurin-build/issues/4404

PR #4299, opened 2025-11-07
Extract SBOM validation so it can be added into the build pipeline

adamfarley:
This task includes:
- Extracting the SBOM validation code from release_download_test.sh into a separate script so it can be run on its own. - Done
- Create a framework to validate and set up the validation parameters/prerequisites  before execution. - Done
- Adapt the validation scripting to allow non-GA builds - Done
- Check SHA/SCM validity only where relevant - Done
- Create a job so SBOMs can be tested without running a build. - [Done](https://ci.adoptium.net/job/sbom_validator_job)

Solo test job run: https://ci.adoptium.net/job/sbom_validator_job/13/

Half-resolves https://github.com/adoptium/ci-jenkins-pipelines/issues/1252

### DIFF
--- a/tooling/validateSBOM.sh
+++ b/tooling/validateSBOM.sh
@@ -85,10 +85,10 @@ arg_parser() {
                                  -e "^jdk8u[0-9][0-9]*-dryrun-ga-aarch32-[0-9][0-9]*\$"
     TAG_CHECK="$?"
   else
-    echo "$SOURCE_TAG" | grep -q "^jdk-[0-9][0-9\.\+]*_adopt\$" \
-                              -e "^jdk-[0-9][0-9\.\+]*\$" \
-                              -e "^jdk-[0-9][0-9\.\+]*-dryrun-ga\$" \
-                              -e "^jdk-[0-9][0-9\.\+]*-ga\$"
+    echo "$SOURCE_TAG" | grep -q -e "^jdk-[0-9][0-9\.\+]*_adopt\$" \
+                                 -e "^jdk-[0-9][0-9\.\+]*\$" \
+                                 -e "^jdk-[0-9][0-9\.\+]*-dryrun-ga\$" \
+                                 -e "^jdk-[0-9][0-9\.\+]*-ga\$"
     TAG_CHECK="$?"
   fi
   


karianna (approved):

sxa:
Can you provide some clarity on this statement please? There shouldn't be any reason why it's not consistent between builds and if there is an issue there then we need to resolve it, not just remove the check. The builds occur within a fixed docker container.
```
    [ "${MAJORVERSION}" -lt 20 ] && EXPECTED_FREETYPE=0 # Skip check as freetype version is umpredictable.
```

adamfarley:
> Can you provide some clarity on this statement please? There shouldn't be any reason why it's not consistent between builds and if there is an issue there then we need to resolve it, not just remove the check. The builds occur within a fixed docker container.
> 
> ```
>     [ "${MAJORVERSION}" -lt 20 ] && EXPECTED_FREETYPE=0 # Skip check as freetype version is umpredictable.
> ```

I've spoken with @steelhead31 and it seems the FreeType version that ends up in the sbom is the same as the version in the devkit (not the version that's on the system's build image).

- Devkit: 2.6.5
- Build image: 2.10.1

See here for more details: https://github.com/adoptium/infrastructure/issues/4131#issuecomment-3517101260

So I'm removing the freetype check exemption for JDK17u on RiscV, and setting the expected version to the devkit one (2.6.5).

adamfarley:
@sxa - I think I've addressed the problem you raised.
Before re-reviewing, would you mind running download_and_sbom_validation using this PR's repo and branch, to make sure I haven't broken anything there?

tooling/validateSBOM.sh
```diff
@@ -0,0 +1,246 @@
...
+#                      - E.g. For JDK17.0.1+35 the major version is 17.
+# - SOURCE_TAG:        The tag of the source code build to make the aforementioned JDK.
+# - SBOM_LOCATION:     The location of the SBOM to be verified.
+#                      - This can be a file location or a web URL, but must be an absolute path.
+#
+########################################################################################################################
+arg_parser() {
+  if [ $# -ne 3 ]; then
+    echo "ERROR: validateSBOM.sh did not receive 3 arguments."
+    echo "Arguments should be: JDK_MAJOR_VERSION SOURCE_TAG SBOM_LOCATION"
```
sxa: Maybe include an example in this output?

tooling/validateSBOM.sh
```diff
@@ -0,0 +1,246 @@
...
+    echo "$SOURCE_TAG" | grep -q "^jdk8u[0-9][0-9]*-dryrun-ga\$"
+    TAG_CHECK="${TAG_CHECK}$?"
+    echo "$SOURCE_TAG" | grep -q "^jdk8u[0-9][0-9]*-aarch32-[0-9][0-9]*\$"
+    TAG_CHECK="${TAG_CHECK}$?"
+    echo "$SOURCE_TAG" | grep -q "^jdk8u[0-9][0-9]*-ga-aarch32-[0-9][0-9]*\$"
+    TAG_CHECK="${TAG_CHECK}$?"
+    echo "$SOURCE_TAG" | grep -q "^jdk8u[0-9][0-9]*-dryrun-ga-aarch32-[0-9][0-9]*\$"
+    TAG_CHECK="${TAG_CHECK}$?"
+  else
+    echo "$SOURCE_TAG" | grep -q "^jdk-[0-9][0-9\.\+]*_adopt\$"
```
sxa: This is an odd structure for checking this - any reason not to use `grep -e` / `egrep` to check for all valid patterns at once? (Or even a `case` statement if you want to have the patterns split out)

tooling/validateSBOM.sh
```diff
@@ -0,0 +1,246 @@
...
+    exit 1
+  fi
+
+  # Now we check that the third argument is a valid link.
+  if echo "$SBOM_LOCATION" | grep -q "^https.*"; then
+    if ! curl -L "$SBOM_LOCATION" -O; then
+      echo "ERROR: SBOM_LOCATION could not be downloaded." 
+      exit 1
+    fi
+    SBOM_NAME=$(echo "${SBOM_LOCATION}" | sed 's:.*/::')
```
sxa: This is slightly more efficient as it doesn't spawn other processes and I think it does the same as your code, but I'm ok either way.
```suggestion
    SBOM_NAME=$(SBOM_LOCATION##*/}
```

sxa:
> Before re-reviewing, would you mind running download_and_sbom_validation using this PR's repo and branch, to make sure I haven't broken anything there?

Yep testing now but it looks like it'll be ok ...

Note to self - I'm not sure we have this documented but for these scripts to work you need the following packages installed:
- `git curl libicu gpg unzip binutils`

sxa (changes requested):
The permissions on `validateSBOM.sh` are not set correctly ... The other scripts are executable. This results in the following error:
```
./release_download_test.sh: line 413: /root/temurin-build/tooling/validateSBOM.sh: Permission denied
```

tooling/validateSBOM.sh
adamfarley: Sure, done.

tooling/validateSBOM.sh
adamfarley: Sure, done.

tooling/validateSBOM.sh
adamfarley: I had to tweak this a little, but I see what you meant. Change made.

adamfarley:
> The permissions on `validateSBOM.sh` are not set correctly ... The other scripts are executable. This results in the following error:
> 
> ```
> ./release_download_test.sh: line 413: /root/temurin-build/tooling/validateSBOM.sh: Permission denied
> ```

@sxa - All sorted. :)

tooling/validateSBOM.sh
```diff
@@ -0,0 +1,238 @@
...
+    echo "$SOURCE_TAG" | grep -q -e "^jdk8u[0-9][0-9]*-b[0-9][0-9]*_adopt\$" \
+                                 -e "^jdk8u[0-9][0-9]*-b[0-9][0-9]*\$" \
+                                 -e "^jdk8u[0-9][0-9]*-ga\$" \
+                                 -e "^jdk8u[0-9][0-9]*-dryrun-ga\$" \
+                                 -e "^jdk8u[0-9][0-9]*-aarch32-[0-9][0-9]*\$" \
+                                 -e "^jdk8u[0-9][0-9]*-ga-aarch32-[0-9][0-9]*\$" \
+                                 -e "^jdk8u[0-9][0-9]*-dryrun-ga-aarch32-[0-9][0-9]*\$"
+    TAG_CHECK="$?"
+  else
+    echo "$SOURCE_TAG" | grep -q "^jdk-[0-9][0-9\.\+]*_adopt\$" \
```
sxa: Something's going wrong here as I'm getting an error:
```
grep: ^jdk-[0-9][0-9\.\+]*_adopt$: No such file or directory
```
The first parameter will definitely need a `-e` on it as a minimum (as you have on the previous block) but I think something else needs to be fixed here. (Don't accept this suggestion in the github UI as it'll mess up the formatting ;-) )
```suggestion
    echo "$SOURCE_TAG" | grep -q -e "^jdk-[0-9][0-9\.\+]*_adopt\$" \
```
The above error occurred when I ran: `./validateSBOM.sh 17 jdk-17.0.17+10 https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.17%2B10/OpenJDK17U-sbom_riscv64_linux_hotspot_17.0.17_10.json`

sxa (changes requested):

tooling/validateSBOM.sh
adamfarley: Fixed, thank you.

adamfarley:
Thanks @sxa and @karianna 😎

Issue #4302, opened 2025-11-11
JDK version should be parsed out of version-numbers.conf when SCM_REF is not specified


adamfarley:
**Summary**

If a build is run with SCM_REF="", we should identify the version from version-numbers.conf and not the latest tag.

**Why?**

Because we currently use the latest tag, and that causes problems when OpenJDK thinks it's version 2 but we think it's version 1.

See [here](https://adoptium.slack.com/archives/C9C272Z1T/p1762867591625079) for more details.

And here for when this conflict causes problems: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk25u/job/jdk25u-windows-x64-temurin/10

**How?**

Right now I'm looking at putting a change [here](https://github.com/adoptium/temurin-build/blob/f4bcbeb80b2dc0a624721f09a91c925e6df2bc90/sbin/build.sh#L318):
```
    version=$(getOpenJDKTag)
    # TODO remove pending #1016
    version=${version%_adopt}
    version=${version#aarch64-shenandoah-}
  fi
```

So just before the "fi", we call a method that:
- Takes the latest tag as an argument
- Finds version-numbers.conf
- Parses the contents (with build number 0)
- Returns whichever tag is later

| Latest tag | Parsed conf | Method returns |
| ----------- | ------------ | ------------------ |
| 21.0.1+8 | 21.0.2 | 21.0.2+0 |
| 21.0.1+8 | 21.0.1 | 21.0.1+8 |
| 21.0.1+8 | Error | 21.0.1+8 |

Error = 
- Could not find version-numbers.conf (or equivalent)
- Could not parse version-numbers.conf

PR #4304, opened 2025-11-13
Version numbers conf parsing script

adamfarley:
This ensures that we're using the OpenJDK version string as defined in the version_numbers.conf file rather than the git tag (if the two are different).

We do this because conflicts have caused problems in the past, where we tell (e.g.) an installer creation job that various dependencies are in the folder for version 1, even though the folders are named after version 2 in version_numbers.conf.

Resolves https://github.com/adoptium/temurin-build/issues/4302

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -268,48 +268,25 @@ versionNumbersFileParser() {
   [ "$jdkVersion" -ge 17 ] && numbersFile="${buildSrc}/make/conf/version-numbers.conf"
   [ ! -r "${numbersFile}" ] && echo "ERROR: build.sh: ${funcName}: JDK version file not found: ${numbersFile}" >&2 && exit 1
 
-  # File parsing logic
-  error="false"
-  fileVersionArray=()
+  fileVersionString=""
+  error=""
   if [ "$jdkVersion" -eq 8 ]; then
-    fileVersionArray+=("$(grep "JDK_MINOR_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')") || error="true"
-    fileVersionArray+=("$(grep "JDK_UPDATE_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')") || error="true"
+    # jdk8 uses this format: jdk8u482-b01
+    fileVersionString="jdk8u$(grep "JDK_UPDATE_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
+   [[ ! "${fileVersionString}" =~ ^jdk8u[0-9]+$  || $error ]] && echo "ERROR: build.sh: ${funcName}: version file could not be parsed." >&2 && exit 1
   else
     # File parsing logic for jdk11+.
-    fileVersionArray+=("$(grep "DEFAULT_VERSION_FEATURE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')") || error="true"
-    fileVersionArray+=("$(grep "DEFAULT_VERSION_INTERIM" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')") || error="true"
-    fileVersionArray+=("$(grep "DEFAULT_VERSION_UPDATE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')") || error="true"
-    fileVersionArray+=("$(grep "DEFAULT_VERSION_PATCH" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')") || error="true"
-  fi
+    patchNo="$(grep "DEFAULT_VERSION_PATCH" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
+    updateNo="$(grep "DEFAULT_VERSION_UPDATE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
+    interimNo="$(grep "DEFAULT_VERSION_INTERIM" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
+    featureNo="$(grep "DEFAULT_VERSION_FEATURE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
 
-  # Check for file parsing errors.
-  if [ "${error}" == "true" ]; then
-    echo "ERROR: build.sh: ${funcName}: ${numbersFile##*/} was found but could not be parsed." >&2
-    exit 1
-  fi
+    [[ ! "${patchNo}" =~ ^0$ ]] && fileVersionString=".${patchNo}"
+    [[ ! "${updateNo}.${fileVersionString}" =~ ^[0\.]+$ ]] && fileVersionString=".${updateNo}${fileVersionString}"
+    [[ ! "${interimNo}.${fileVersionString}" =~ ^[0\.]+$ ]] && fileVersionString=".${interimNo}${fileVersionString}"
+    fileVersionString="jdk-${featureNo}${fileVersionString}"
 
-  # Turn version numbers array into a valid version string.
-  if [ "$jdkVersion" -gt 8 ]; then
-    # JDK11 and above uses this format: jdk-21.0.10.1+2
-    fileVersionString="jdk-${fileVersionArray[0]}"
-    if [ "${fileVersionArray[3]}" -gt 0 ]; then
-      fileVersionString="${fileVersionString}.${fileVersionArray[1]}.${fileVersionArray[2]}.${fileVersionArray[3]}"
-    elif [ "${fileVersionArray[2]}" -gt 0 ]; then
-      fileVersionString="${fileVersionString}.${fileVersionArray[1]}.${fileVersionArray[2]}"
-    elif [ "${fileVersionArray[1]}" -gt 0 ]; then
-      fileVersionString="${fileVersionString}.${fileVersionArray[1]}"
-    fi
-  else
-    # jdk8 uses this format: jdk8u482-b01
-    fileVersionString="jdk8u${fileVersionArray[1]}"
-  fi
-
-  # Check the version string is valid.
-  if [[ ! "${fileVersionString}" =~ ^jdk8u[0-9]+$ ]]; then
-    if [[ ! "${fileVersionString}" =~ ^jdk\-[0-9]+[0-9\.]*$ ]]; then
-      echo "ERROR: build.sh: ${funcName}: The version string parsed from the version numbers file does not use a valid format." >&2
-      exit 1
-    fi
+    [[ ! "${fileVersionString}" =~ ^jdk\-[0-9]+[0-9\.]*$ || $error ]] && echo "ERROR: build.sh: ${funcName}: version file could not be parsed." >&2 && exit 1
   fi
 
   # Returning the formatted jdk version string.
@@ -333,10 +310,9 @@ compareToOpenJDKFileVersion() {
     exit 1
   fi
 
-  # This verifies that arg1 was in an expected format for a jdk version string.
-  # This provides some assurance that we can parse that string in this function.
-  # JDK11 and above uses this format: jdk-21.0.10+2
-  # jdk8 uses this format: jdk8u482-b01
+  # Check if arg 1 looks like a jdk version string.
+  # Example: JDK11+: jdk-21.0.10+2
+  # Example: JDK8  : jdk8u482-b01
   if [[ ! "$1" =~ ^jdk\-[0-9]+[0-9\.]*(\+[0-9]+)?$ ]]; then
     if [[ ! "$1" =~ ^jdk8u[0-9]+(\-b[0-9][0-9]+)?$ ]]; then
       echo "ERROR: build.sh: ${funcName}: The JDK version passed to this function did not match the expected format." >&2
@@ -357,19 +333,16 @@ compareToOpenJDKFileVersion() {
     # The file version matches the function argument.
     echo "$1"
     return
-  else
-    # The file version does not match the function argument.
-    # Returning the file version.
-    if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -eq 8 ]; then
-      fileVersionString+="-b00"
-    else
-      fileVersionString+="+0"
-    fi
-
-    echo "WARNING: build.sh: ${funcName}: JDK version in source does not match the supplied version (likely the latest git tag)." >&2
-    echo "WARNING: The JDK version in source will be used instead." >&2
-    echo "${fileVersionString}"
   fi
+
+  # The file version does not match the function argument.
+  # Returning the file version.
+  [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -eq 8 ] && fileVersionString+="-b00"
+  [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -gt 8 ] && fileVersionString+="+0"
+
+  echo "WARNING: build.sh: ${funcName}: JDK version in source does not match the supplied version (likely the latest git tag)." >&2
+  echo "WARNING: The JDK version in source will be used instead." >&2
+  echo "${fileVersionString}"
 }
 
 getOpenJdkVersion() {


adamfarley:
Test run on jdk25u, where we see the issue we're trying to solve.

https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk25u/job/jdk25u-windows-x64-temurin/13/

Update: Specifying a default build number of 0 causes problems (detailed below), so I'm changing it to "1" and rerunning.

https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk25u/job/jdk25u-windows-x64-temurin/14/

When we specify a build number of 0, the openjdk build scripts interpret that as there being no build number at all.

This means that the vendor string and folder names use build number "0", while everything else uses an empty string.

So we end up having the same problem. The installer jobs are passed a build number of \"null\", while the folder names are using "+0". Installer job creation fails.

So I changed it to "1" to avoid that logic path.

Update 2: Test passed completely. Ready for review.

karianna:
General comment, how does the git tag get out of sync with that conf file?  Should one or the other not be corrected?

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
   fi
 }
 
+# This attempts to compare the jdk version from version-numbers.conf with arg 1.
+# We will then return whichever version is bigger/later.
+# e.g. 17.0.1+32 > 17.0.0+64
+# If any errors or unusual circumstances are detected, we simply return arg1 to avoid destabilising the build.
+compareToOpenJDKFileVersion(){
+  # First, sanity checking on the arg.
+  if [ $# -eq 0 ]; then
```
karianna, line 304: What about 2 args?

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+    echo "compareToOpenJDKFileVersion_was_called_with_no_args"
+    exit 1
+  fi
+  if [[ ! "$1" =~ ^jdk\-[0-9]+[0-9\.]*\+[0-9]+$ ]]; then
```
karianna: Probably worth adding a comment with some examples of what you're picking up here.

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+    if [[ ! "$1" =~ ^8u[0-9]+\-b[0-9][0-9]$ ]]; then
+      echo "$1"
+      exit 1
+    fi
+  fi
+
+  build_src="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
```
karianna: why _ in this variable name but camel case elsewhere?

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+  fileVersionArray=("0" "0" "0" "0" "1")
+  error="false"
+  numbersFile=""
+
+  # Parse the version-numbers.conf or the equivalent.
+  if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -gt 8 ]; then
+    numbersFile="${build_src}/make/conf/version-numbers.conf"
+    if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -eq 11 ]; then
+      numbersFile="${build_src}/make/autoconf/version-numbers"
```
karianna: Was this only for 11 or 11+?

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+    fi
+
+    # Check the file exists.
+    if [ ! -r "${numbersFile}" ]; then
+      echo "$1"
```
karianna, line 319: Moar context in the error message?

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+      exit 1
+    fi
+
+    # File parsing logic for jdk11+
+    fileVersionArray[0]=$(grep "DEFAULT_VERSION_FEATURE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
+    fileVersionArray[1]=$(grep "DEFAULT_VERSION_INTERIM" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
+    fileVersionArray[2]=$(grep "DEFAULT_VERSION_UPDATE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
+    fileVersionArray[3]=$(grep "DEFAULT_VERSION_PATCH" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
+    fileVersionArray[4]="1"
```
karianna: This is the temurin patch number?

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+  else
+    # File location for jdk8.
+    numbersFile="${build_src}/common/autoconf/version-numbers"
+    # Check the file exists.
+    if [ ! -r "${numbersFile}" ]; then
```
karianna: could extract dup logic

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+      echo "$1"
+      exit 1
+    fi
+
+    # File parsing logic for jdk8.
```
karianna: So positions [1] and [2] always have 0?

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+    fileVersionArray[0]=$(grep "JDK_MINOR_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
+    fileVersionArray[3]=$(grep "JDK_UPDATE_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
+    fileVersionArray[4]="01"
+  fi
+
+  # Compare and set fileVersionArray to the biggest one.
+  # If file version is bigger, then set build number to 0 (or 00 for jdk8).
+  if [ "${error}" == "true" ]; then
+    echo "$1"
```
karianna, line 328: moar context

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+
+  # If there were no errors, then compare the two versions and set fileVersionArray to the biggest/latest one.
+  argFullVersionArray=("0" "0" "0" "0" "1")
+
+  # Now we turn arg 1 into an array of numbers.
+  spaceSeperatedArg=$(echo "$1" | sed -e 's/jdk//' -e 's/-//' -e 's/[^0-9]/ /g')
+  # If we somehow produced anything other than a space-seperated list of integers,
+  # quit and return the function arg.
+  if [[ ! "${spaceSeperatedArg}" =~ ^[0-9\ ]+$ ]]; then
+    echo "$1"
```
karianna: More context

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+
+  # Then we turn the list of numbers into consistent version number arrays.
+  # So 17.0.1+23 becomes 17.0.1.0+23, to allow for equal comparisons.
+  if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -gt 8 ]; then
+    argFullVersionArray[0]="${argArray[0]}"
+    [ "${#argArray[@]}" -gt 1 ] && argFullVersionArray[4]="${argArray[1]}"
+    [ "${#argArray[@]}" -gt 2 ] && argFullVersionArray[1]="${argArray[1]}" && argFullVersionArray[4]="${argArray[2]}"
+    [ "${#argArray[@]}" -gt 3 ] && argFullVersionArray[2]="${argArray[2]}" && argFullVersionArray[4]="${argArray[3]}"
+    [ "${#argArray[@]}" -gt 4 ] && argFullVersionArray[3]="${argArray[3]}" && argFullVersionArray[4]="${argArray[4]}"
+  else
```
karianna: This is the java 8 case right?

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+    echo "$1"
+    exit 1
+  fi
+
+  # And then we compare the other version numbers:
+  for i in {1..4}; do
+    if [ "${fileVersionArray[$i]}" -gt "${argFullVersionArray[$i]}" ]; then 
+      break
+    elif [ "${argFullVersionArray[$i]}" -gt "${fileVersionArray[$i]}" ]; then
+      echo "$1"
```
karianna, line 319: Are these success or failurre cases?

sbin/build.sh
```diff
@@ -253,6 +253,169 @@ patchFreetypeWindows() {
...
+    versionString="jdk8u${fileVersionArray[3]}-b${fileVersionArray[4]}"
+
+    # Check to ensure that we've produced a valid version string.
+    if [[ ! "${versionString}" =~ ^8u[0-9]+\-b[0-9][0-9]+$ ]]; then
+      echo "$1"
+      exit 1
+    fi
+  fi
+
+  echo "${versionString}"
```
karianna: probably add more context here

sbin/build.sh
adamfarley: My thought at the time was that this is the only scenario where we cannot just return arg1 if something goes wrong.

I can return arg 1 if 2 args are passed, but we have to be careful in how we can express errors given that all stdout is captured in the subshell we run the function in.

adamfarley:
> General comment, how does the git tag get out of sync with that conf file? Should one or the other not be corrected?

My understanding is that while jdk25u is in STS mode with Oracle (until it gets transitioned to the LTS owner in a few months), some of the development happens internally, and as such the code continues to grow and progress through versions without the appropriate tags being added.

In short, we're not 100% sure, we just know it happens.

This change allows us to bypass the issue and ensure that we can do the following in the run up to the January release:
- We can produce installers for JDK25u on Windows
- We produce Windows JDKs with the correct folder names
- Our Temurin vendor version strings show the same version as the rest of the java -version output

sbin/build.sh
adamfarley: Ok

sbin/build.sh
adamfarley: An oversight. Fixed. Good spot. 👍

sbin/build.sh
adamfarley: "${build_src}/make/autoconf/version-numbers" is only for 11. I'll add a comment to clarify.

sbin/build.sh
adamfarley: I've upgraded all the error messages.

sbin/build.sh
adamfarley: No, this is the default build number. We can't set it to "0", because then configure treats it as us building without a build number.

sbin/build.sh
adamfarley: done

sbin/build.sh
adamfarley: Yes. To my knowledge, JDK8 doesn't use interim or update numbers.

sbin/build.sh
adamfarley: Yes. I've added a comment to clarify.

sbin/build.sh
adamfarley: Success. It means we have identified which version is the right one for us to use.

I've added a comment.

sbin/build.sh
adamfarley: Done.

adamfarley:
> Long function is also long, if possible split it up?

@karianna - I split the method into two; extracting the file parsing logic. Both functions are now smaller than their contemporaries in the same script.

Here's another test run: https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk25u/job/jdk25u-windows-x64-temurin/15/

Update: Passed.

sbin/build.sh
```diff
@@ -315,6 +519,21 @@ getOpenJdkVersion() {
     # TODO remove pending #1016
     version=${version%_adopt}
     version=${version#aarch64-shenandoah-}
+
+    # Now we check if the version in the code is later than the version we have so far.
+    # This prevents an issue where the git repo tags are not updated to match the hard-coded version string.
+    # If SCM_REF is specified, we don't do this check as SCM_REF gets priority.
+    tag_ref="SCM_REF"
```
andrew-m-leonard: did you try this here?
```
if [ -n "${BUILD_CONFIG[SCM_REF]}" ]; then
```

adamfarley: I'm sure I did but I don't see the test run.

Tried it, and the issue happened again.

```
    if [ -n "${BUILD_CONFIG[SCM_REF]}" ]; then
      version=$(compareToOpenJDKFileVersion "$version")
    fi
```

```
/Users/runner/work/temurin-build/temurin-build/sbin/build.sh: line 535: SCM_REF: unbound variable
```
https://github.com/adoptium/temurin-build/actions/runs/19569946267/job/56040820712?pr=4304

karianna:
> > General comment, how does the git tag get out of sync with that conf file? Should one or the other not be corrected?
> 
> My understanding is that while jdk25u is in STS mode with Oracle (until it gets transitioned to the LTS owner in a few months), some of the development happens internally, and as such the code continues to grow and progress through versions without the appropriate tags being added.
> 
> In short, we're not 100% sure, we just know it happens.
> 
> This change allows us to bypass the issue and ensure that we can do the following in the run up to the January release:
> 
> * We can produce installers for JDK25u on Windows
> * We produce Windows JDKs with the correct folder names
> * Our Temurin vendor version strings show the same version as the rest of the java -version output

OK, can we add this reasoning as a comment block in that case?  Stops folks from ripping this out in the future.

sbin/build.sh
```diff
@@ -315,6 +519,13 @@ getOpenJdkVersion() {
     # TODO remove pending #1016
     version=${version%_adopt}
     version=${version#aarch64-shenandoah-}
+
+    # Now we check if the version in the code is later than the version we have so far.
+    # This prevents an issue where the git repo tags are not updated to match the hard-coded version string.
+    # If SCM_REF is specified, we don't do this check as SCM_REF gets priority.
+    if [ -n "${BUILD_CONFIG[TAG]}" ]; then
```
andrew-m-leonard: it would be better to do this check within the code doing the "tag determination logic" (getOpenJDKTag), as there are various User build Use cases, where BUILD_CONFIG[TAG] or [BRANCH] are specified which could be SHAs or branches, or determined via a different mechanism for 2nd pass Windows&Mac builds.

getOpenJDKTag() handles the various sub-use cases where we don't want to do this, and then goes on to determine the "tag" by searching the git repository, which is where we want to do this.

I suggest in that code, we :
- Determine tag from git repo as existing...
- Determine using versionNumbersFileParser(), if the determined tag from git repo does not start with the value from versionNumbersFileParser(), if not then change to versionNumbersFileParser()+("+0" or "-b00")

sbin/build.sh
```diff
@@ -253,6 +253,210 @@ patchFreetypeWindows() {
...
+    # File parsing logic for jdk8.
+    fileVersionString=$(grep "JDK_MINOR_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
+    fileVersionString+=" 0 0 "
+    fileVersionString+=$(grep "JDK_UPDATE_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
+  else
+    # File parsing logic for jdk11+.
+    fileVersionString=$(grep "DEFAULT_VERSION_FEATURE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+' | awk '{print $1" "}') || error="true"
+    fileVersionString+=$(grep "DEFAULT_VERSION_INTERIM" "${numbersFile}" | head -1 | grep -Eo '[0-9]+' | awk '{print $1" "}') || error="true"
+    fileVersionString+=$(grep "DEFAULT_VERSION_UPDATE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+' | awk '{print $1" "}') || error="true"
+    fileVersionString+=$(grep "DEFAULT_VERSION_PATCH" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
```
andrew-m-leonard: This logic needs to match what OpenJDK does to form the VERSION_NUMBER, see https://github.com/openjdk/jdk21u/blob/46687d38472dafe8cacda2de61fb34a2b73ee194/make/autoconf/jdk-version.m4#L453
ie.it needs to strip any trailing ".0" from the $VERSION_FEATURE.$VERSION_INTERIM.$VERSION_UPDATE.$VERSION_PATCH

sbin/build.sh
```diff
@@ -253,6 +253,210 @@ patchFreetypeWindows() {
...
+    exit 1
+  fi
+
+  error="false"
+  fileVersionString=""
+  if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -eq 8 ]; then
+    # File parsing logic for jdk8.
+    fileVersionString=$(grep "JDK_MINOR_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
+    fileVersionString+=" 0 0 "
+    fileVersionString+=$(grep "JDK_UPDATE_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+') || error="true"
```
andrew-m-leonard: This needs to match for jdk8, just the jdk8u<JDK_UPDATE_VERSION> , as that is what get taken from the build number to form the JDK folder name, and is checked by the Installers here https://github.com/adoptium/installer/blob/c5f8f53df892f0e43b9533601ca9e937b62adc74/wix/Build.OpenJDK_generic.cmd#L145C1-L145C160

sbin/build.sh
adamfarley: The formatting logic does this in compareToOpenJDKFileVersion. 

I've moved that logic into the versionNumbersFileParser function to enable the string comparison.

sbin/build.sh
adamfarley: Ok.

sbin/build.sh
adamfarley: Ok.

karianna:
@adamfarley there's an erroneous '}'

adamfarley:
> @adamfarley there's an erroneous '}'

Ah yes. I removed an elif by accident. All sorted.

Linter still has concerns though. Fixing them now.

sbin/build.sh
```diff
@@ -253,6 +253,134 @@ patchFreetypeWindows() {
...
+    return
+  else
+    # The file version does not match the function argument.
+    # Returning the file version.
+    if [ "${BUILD_CONFIG[OPENJDK_FEATURE_NUMBER]}" -eq 8 ]; then
+      fileVersionString+="-b00"
+    else
+      fileVersionString+="+0"
+    fi
+    echo "${fileVersionString}"
```
andrew-m-leonard: We should issue a "WARNING" here that discovered tag does not match OpenJDK src version etc...

adamfarley: Done.

adamfarley:
And here's a jenkins helper tweak to accommodate build number 0 (so VersionInfo doesn't return a null).

https://github.com/adoptium/jenkins-helper/pull/63

Testing here:

https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk25u/job/jdk25u-windows-x64-temurin/18/

It passed the relevant stages, but failed SBOM validation.

I've written a fix and tested it here: https://ci.adoptium.net/job/sbom_validator_job/272 - passed.

tooling/validateSBOMcontent.sh
```diff
@@ -22,6 +22,8 @@ SBOMFILE="$1"
 MAJORVERSION="$2"
 EXPECTED_SCM_REF="$3"
 
+[ "${EXPECTED_SCM_REF}" = "scm_not_specified" ] && EXPECTED_SCM_REF="null"
```
andrew-m-leonard: Not keen on having special "strings", best just allow blank value if not specified

tooling/validateSBOMcontent.sh
```diff
@@ -134,15 +134,25 @@ echo -n "Checking for JDK source SHA validity: "
 GITSHA=$(jq '.components[].properties[] | select(.name|test("OpenJDK Source Commit")) | .value' "$1" | tr -d \" | uniq)
 GITREPO=$(echo "$GITSHA" | cut -d/ -f1-5)
 GITSHA=$( echo "$GITSHA" | cut -d/ -f7)
-if ! git ls-remote "${GITREPO}" | grep "${GITSHA}"; then
-  echo "ERROR: git sha of source repo not found"
-  echo "GITREPO: ${GITREPO}"
-  echo "GITSHA: ${GITSHA}"
-  RC=1
+if [ -z "${EXPECTED_SCM_REF}" ]; then
+  commitData=$(curl --silent "https://api.github.com/repos/adoptium/jdk25u/commits/${GITSHA}")
```
andrew-m-leonard: This is hard coded to jdk25u, and should probably just do a http header and http response check?
eg:
```
if ! curl --fail -I "$GIT_COMMIT_URL"; then
```

tooling/validateSBOMcontent.sh
adamfarley: Ok, testing here: https://ci.adoptium.net/job/build-scripts/job/openjdk25-pipeline/95/

tooling/validateSBOMcontent.sh
adamfarley: Fixed.

adamfarley:
Here is the PR for passing an empty string into the sbom validator when scm is not set.

https://github.com/adoptium/ci-jenkins-pipelines/pull/1330

sbin/build.sh
```diff
@@ -253,6 +253,137 @@ patchFreetypeWindows() {
...
 
+# Returns the version numbers in version-numbers.conf as a space-seperated list of integers.
+# e.g. 17 0 1 5
+# or 8 0 0 432
+# Build number will not be included, as that is not stored in this file.
+versionNumbersFileParser() {
+  funcName="versionNumbersFileParser"
+  # Find version-numbers.conf or the equivalent and confirm we can read it.
+  buildSrc="${BUILD_CONFIG[WORKSPACE_DIR]}/${BUILD_CONFIG[WORKING_DIR]}/${BUILD_CONFIG[OPENJDK_SOURCE_DIR]}"
+  # File location for jdk8.
```
sxa: This seems like a lot of lines to read the numbers. Can we not do something simpler here like:
```
if [ -r "${buildSrc}/make/autoconf/version-numbers"]; then
    numbersFile="${buildSrc}/make/autoconf/version-numbers"
elif [ -r "${buildSrc}/make/conf/version-numbers.conf" ]; then
   numbersFile="${buildSrc}/make/conf/version-numbers.conf"
[...]
else
    echo "ERROR: build.sh: ${funcName}: The file expected to contain the JDK version numbers could not be found and/or read." >&2
    echo "ERROR: build.sh: ${funcName}: The expected location was: ${numbersFile}" >&2
    exit 1
fi
```
Or even:
```
for CONFFILE in make/autoconf/version-numbers make/conf/version-numbers.conf; do
  if [ -r "${buildSrc}/$CONFFILE"]; then
    numbersFile="${buildSrc}/$CONFFILE"
  fi
done
[ -z "$numbersFile" ] && echo Throw some error && exit 1
```

sbin/build.sh
```diff
@@ -253,6 +253,137 @@ patchFreetypeWindows() {
...
+  # Check the version numbers file exists.
+  if [ ! -r "${numbersFile}" ]; then
+    echo "ERROR: build.sh: ${funcName}: The file expected to contain the JDK version numbers could not be found and/or read." >&2
+    echo "ERROR: build.sh: ${funcName}: The expected location was: ${numbersFile}" >&2
+    exit 1
+  fi
+
+  error="false"
+
+  fileVersionArray=()
```
sxa: Could this be done with a single more efficient `awk` command to do a single pass on the file e.g.
`awk -F= '/^JDK_MAJOR_VERSION=/{MAJOR=$2} /^JDK_MINOR_VERSION=/{MINOR=$2} END{print MAJOR " " MINOR}' common/autoconf/version-numbers`
which would give `1 8` for jdk8u for example? The `END` part of the awk can spit it out in whatever format you want  avoiding the need for the "arrange the file version array into a standard format for that feature number" version later on.

sbin/build.sh
adamfarley: Ok, done.

sbin/build.sh
adamfarley: Yes and no. I think readability decreases if we add "don't include any number where that number and below are 0" to the awk command.

So I refactored the logic in the shell script to omit the array entirely, and changed some other parts to help make the whole thing shorter.

adamfarley:
Merged. Thanks for the reviews and your help refining this code.
@sxa @andrew-m-leonard @karianna 😎

adamfarley:
On request, here's an additional test run against a repo without any tags:

https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk25u/job/jdk25u-windows-x64-temurin/35

Update: It passed correctly. We received two warnings:
- The first said there were no tags, so we'd use the default tag of 25.0.0
- The second said that 25.0.0 didn't match the source, so we override the default with 25.0.3

PR #4320, opened 2025-11-27
All build-configs should use default (bundeld) free type

judovana:
Since jdk8u482.b02 all maintained JDKs have in tree freetype. All jdks should use it unless there is necessary reason to do so.

See https://github.com/adoptium/temurin-build/pull/4287#issuecomment-3582533466


judovana:
The GHAs looks better then I expected (maybe rings a  red bell a bit). Will double check the logs. The failure on alpine is epxcted, as it can not build "byu default downloaded freetype", and thus must wait for https://github.com/adoptium/temurin-build/pull/4287 to pass.

Issue #2545, opened 2021-03-23
Reverse PR #2539 when Scaleway investigation is complete.


karianna:
See https://github.com/AdoptOpenJDK/openjdk-build/pull/2539#issue-597265703

PR #4323, opened 2025-11-28
Remove arm32 debug checks that were added for Scaleway/arm32

sxa:
This reverts commit a9a9e37b02d40a33f4c16ac0e81ce6b7967daef6.

This was added for debugging problems on Scaleway arm32 systems which we no longer use. This should have been reverted ages ago

Fixes https://github.com/adoptium/temurin-build/issues/2545

Noting that this was also causing problems if you run on a machine (e.g. build container) when nothing was running as the jenkins user as it fell over on `ps -fu jenkins` which was returning nothing)

### DIFF
--- a/tooling/validateSBOM.sh
+++ b/tooling/validateSBOM.sh
@@ -112,8 +112,6 @@ arg_parser() {
     fi
     SBOM_NAME=${SBOM_LOCATION##*/}
     SBOM_LOCATION="${WORKSPACE_DIR}/${SBOM_NAME}"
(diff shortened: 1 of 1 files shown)


sxa:
Noting that there are are other things a few lines up which were specfic to the Scaleway systems and should be removed.
Also mentioning that it's quite common for arm32 builds to end up running with `Cores to use: 1` which should probably be looked into as a separate issue.

karianna (approved):

Issue #4316, opened 2025-11-25
Solaris pipelines should not produce sha256 file for SBOMs


sxa:
We do not publish a SHA56 checksum for the SBOM files. At the moment the solaris `simple` jobs do produce one and so if you attempt to use the release tool with **/OpenJDK* then you need to explicitly exclude the `*sbom*sha256.txt` file.

The best solution here would be to ensure that it is not generated as part of https://github.com/adoptium/temurin-build/blob/a44863d150ed2303b6833a02d313b7defb8ab15f/sbin/solaris/build-simple.sh#L158 so that there is no risk of it appearing and being published.

SehrishHussain:
@sxa I’d like to pick this up.

I want to confirm the expected direction:
Are we specifically looking to skip generating checksum files for SBOM artifacts in the Solaris `build-simple.sh` script, so the output stays consistent with the other platforms?

sxa:
Correct @SehrishHussain! Should be an relatively easy fix to the script. Thanks.

sxa:
> Correct [@SehrishHussain](https://github.com/SehrishHussain)! Should be an relatively easy fix to the script. Thanks.

Now slightly regretting saying that ;-)

SehrishHussain:
@sxa the culprit is my novice scripting skills and the daily 22 hour power breakouts on my part! But I will get there!

sxa:
Haha no I put the delay down to the fact we had some other problems that got in the way of testing your fix that had to be resolved. You're doing well - this is the sort of thing that can happen in open-source :-)

PR #4254, opened 2025-09-12
Correcting and/or exempting linter shellcheck code warnings

adamfarley:
Also includes a change to the PR build pipeline workflow to exempt the solaris files as we don't test solaris in those builds.

**Status**

Linter check - passed.

Test run: Looks like this branch needs to be hard-coded in [the job](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-x64-temurin-simpletest). Created a copy of the job and giving it a shot:

https://ci.adoptium.net/job/adam_farley_solaris_test_job_copy/

PR #4326, opened 2025-11-30
Skip generating checksum files for SBOM artifacts in Solaris build

SehrishHussain:
This PR modifies the Solaris build-simple.sh script to skip generating .sha256.txt checksum files for SBOM JSON artifacts. Previously, the script generated checksums for all artifacts, including SBOMs, which are not intended to be published. Closes https://github.com/adoptium/temurin-build/issues/4316

**Changes:**

- Added a conditional check in the for FILE in OpenJDK* loop to skip checksum generation for files matching *sbom.json.
- Metadata for SBOM files is still created, with the sha256 field left empty.
- Checksum generation for other artifacts remains unchanged.

**Local Testing:**
Created a test workspace with dummy artifacts (OpenJDK-fake.tar.gz and OpenJDK-sbom.json).

**Verified that:**

- Non-SBOM files generate .sha256.txt as expected.
- SBOM files skip checksum creation and metadata includes an empty sha256 field.
- Confirmed script logic works as intended without errors.
- This change ensures Solaris builds behave consistently with other platforms and prevents unnecessary SBOM checksums from being generated.

**Note: Reason for opening a new PR**
Due to Windows removing the executable flag during local edits, the previous commit did not preserve the executable bit, which caused CI/linter failures. This PR restores the executable permission and includes the SBOM checksum skip changes together.


SehrishHussain:
@karianna The linter is showing warnings for lines in `build-simple.sh` that I haven’t modified. I was building and testing this repo on a Windows OS, which may be causing some env related differences. Do you have any suggestions on how to handle these linter warnings?

judovana:
Still you can replace modified ` "$FILE"` by ` "${FILE}" ;)

SehrishHussain:
@judovana Thank you for your patience. I've implemented the recommended `${FILE}` syntax for safer shell usage.

The persistent lint errors are from pre existing lines in the sections I modified. To avoid changing unrelated logic, I'd appreciate any suggestions you might have.

judovana:
Probably best what you can do, is to open new PR, where - without any logic change, you would fix all the linter issues. Once it is merged, you can rebase this PR, and it will automagically pass.   I know it is ungrateful,  but best to not mix logical changes and refactoring, and still make all bots happy. Sorry for that :( Thanx a lot for PR! All the best, J.

I guess the issues paeared, as there were no changes in this file, and in meantime linter got updated. So new issues popped up. AS there is quite a few of them, would be worthy to fix them for the future.

karianna (approved):

sbin/solaris/build-simple.sh
```diff
@@ -155,12 +155,20 @@ scp -prP "${SSH_PORT}" $SSH_OPTS "${SSH_TARGET}:temurin-build/build-farm/workspa
...
+        sha256=$(cut -d' ' -f1 "${FILE}.sha256.txt")
+    fi
+
+    # Metadata filename: SBOM files get <name>-metadata.json (consistent with naming)
+    if [[ "$FILE" == *sbom.json ]]; then
+        metadata_file="${FILE%.*}-metadata.json"
+    else
+        metadata_file="${FILE}.json"
     fi
     createMetadataFile "$metadata_file" "${TARGET_ARCH}" "$SCM_REF" metadata/buildSource.txt metadata/version.txt "$sha256"
```
sxa, line 176: This is going to make `$sha256` in the metadata file an empty string I think, which may cause problems for the API which uses the metadata file. We may have to still generate the sha256 checksum, but not leave the `.sha256.txt` file around. I'll tag in @andrew-m-leonard to look over this since he's a bit more familiar with this than I am :-)

SehrishHussain: @sxa Thank you for pointing this out! I’m still learning my way around this part of the codebase, so I really appreciate the guidance. That makes sense I wasn’t fully aware of the impact on the metadata flow and the API. I’ll wait for Andrew’s input and adjust the change accordingly.

sxa:
> Probably best what you can do, is to open new PR, where - without any logic change, you would fix all the linter issues. 

+1 from me to that, although I'd also be ok with merging this without it first being rebased on a second PR if this hasn't made the linter any worse :-)

I thought we'd fixed the linter errors in here somehow. Maybe the linter just got more picky ...

EDIT: Yeah https://github.com/adoptium/temurin-build/pull/4254/files should have excluded this file. FYI @adamfarley

adamfarley:
> EDIT: Yeah https://github.com/adoptium/temurin-build/pull/4254/files should have excluded this file. FYI @adamfarley

It did exclude this file. That PR only included sbin/solaris/dotests.sh (plus a yml file and a comment elsewhere).

sbin/solaris/build-simple.sh
andrew-m-leonard: Yes, we should ensure the metadata file has the correct sha256 value in it, but do not generate a sbom.sha256.txt file.
Just remove the sbom.sha256.txt after we have the sha256 value I suspect.

sbin/solaris/build-simple.sh
```diff
@@ -155,12 +155,20 @@ scp -prP "${SSH_PORT}" $SSH_OPTS "${SSH_TARGET}:temurin-build/build-farm/workspa
 cd workspace/target || exit 1
 for FILE in OpenJDK*; do
     echo Creating metadata for ${FILE}
-    sha256sum $FILE > $FILE.sha256.txt
-    sha256=$(cat $FILE.sha256.txt | cut -d' ' -f1)
-    if [[ "$FILE" =~ .*sbom.*\.json ]]; then
-        metadata_file=${FILE%.*}-metadata.json
+    # Skip checksum generation for SBOM files
+    if [[ "${FILE}" == *sbom.json ]]; then
```
sxa: This isn't the correct pattern as the filenames don't end in `sbom.json`. It needs to match a file like this:
```
OpenJDK8U-sbom_sparcv9_solaris_hotspot_8u482b01-ea.json
```
(and the equivalent `x64` one instead of `sparcv9`)

sbin/solaris/build-simple.sh
SehrishHussain: @andrew-m-leonard and @sxa Updated the script to generate `SHA256` for all files while removing leftover `.sha256.txt` for SBOMs and setting appropriate metadata filenames. please review.

PR #4330, opened 2025-12-03
Fix ShellCheck warnings in sbin/solaris/build-simple.sh (lint only)

SehrishHussain:
This PR contains only lint and formatting fixes (no logic changes) to address ShellCheck warnings that started failing CI after the linter was updated. (With reference to failed lint checks in PR #4326 )

Changes:
- Replaced `echo|sed` leading-zero removal with parameter expansion for `build_trim`
- Constructed `pre`/`opt` with printf to avoid literal backslash/quote issues
- Grouped metadata writes into a single redirect block to avoid repeated `>>` (SC2129)
- Used double quoted expansions for JSON fields to preserve quotes/backslashes
- Replaced `sed` substitutions for `PUBLISH_NAME` with parameter expansion
- Escaped inner quotes in exports inside the `ssh` remote command to satisfy SC2027
- Ensured `sbin/solaris/build-simple.sh` has the executable bit set

Rationale:
- Tried these edits to be minimal and safe, meant to remove ShellCheck warnings
- No functional changes to the build logic; output format is unchanged

After this PR is merged, I will rebase my SBOM checksum patch branch so CI will pass cleanly.

### DIFF
--- a/sbin/solaris/build-simple.sh
+++ b/sbin/solaris/build-simple.sh
@@ -50,59 +50,62 @@ createMetadataFile() {
       ver_opt=""
       semver_adopt_build_num=""
       semver_opt=""
-      build_trim=$(echo "$build" | sed 's/^0*//')
+      build_trim="${build##0}"
     elif [[ "$ver" =~ $beta_pattern ]]; then
       update=${BASH_REMATCH[1]}
-      pre="\"${BASH_REMATCH[2]}\""
-      opt="\"${BASH_REMATCH[3]}\""
+      pre=$(printf '"%s"' "${BASH_REMATCH[2]}")
+      opt=$(printf '"%s"' "${BASH_REMATCH[3]}")
       build=${BASH_REMATCH[4]}
       adopt_build_num="0"
       ver_pre="-${BASH_REMATCH[2]}"
       ver_opt="-${BASH_REMATCH[3]}"
       semver_adopt_build_num=".$adopt_build_num"
       semver_opt=".${BASH_REMATCH[3]}"
-      build_trim=$(echo $build | sed 's/^0*//')
+      build_trim="${build##0}"
     else
       echo "ERROR: Unable to determine metadata parameters"
       exit 1
     fi
 
-    echo '{' > $metadata_file
-    echo '"vendor": "Eclipse Adoptium",' >> $metadata_file
-    echo '"os": "solaris",' >> $metadata_file
-    echo '"arch": "'$arch'",' >> $metadata_file
-    echo '"variant": "temurin",' >> $metadata_file
-    echo '"version": {' >> $metadata_file
-    echo '    "minor": 0,' >> $metadata_file
-    echo '    "patch": null,' >> $metadata_file
-    echo '    "msi_product_version": "8.0.'$update'.'$build_trim'",' >> $metadata_file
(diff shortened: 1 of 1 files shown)


SehrishHussain:
Hi @andrew-m-leonard  @judovana  @sxa  I opened this lint only PR to fix the ShellCheck issues that started appearing after the linter update. I’m still learning parts of this script, so I tried to keep the changes strictly non functional and would really appreciate a careful review to ensure nothing breaks.

**Summary of changes**

- Replaced `echo | sed` patterns with parameter expansion for `build_trim`.
- Rewrote `pre` and `opt` construction using safer quoting (`printf`) instead of literal backslash escaped strings.
- Grouped multiple `echo >>` calls into a single `{ … } >> file` block to satisfy SC2129.
- Updated `PUBLISH_NAME` substitutions to use parameter expansion instead of `sed`.
- Fixed inner quoting inside remote `ssh "..."` exports so the remote shell receives valid values.
- Restored executable bit on` sbin/solaris/build-simple.sh.`

All previous ShellCheck errors are resolved locally.

I avoided touching parsing/logic code (e.g., version parsing) to reduce risk. If you see any place where quoting/expansion changes could alter behavior on Solaris or legacy shells, please flag it I’ll revert or update immediately.
If you prefer a different style (e.g., keep `sed` for portability), tell me and I’ll adjust the PR.
Thanks in advance for your careful review and patience while I get up to speed I want to make sure this is safe for all supported platforms.

judovana:
Nice work. TY! Were you able to run the file and try outputs?

SehrishHussain:
@judovana thank you, yes I ran local sanity checks on the script to make sure the lint related changes didn’t introduce any regressions.

**I tested locally**

- Exported all required environment variables `(WORKSPACE, JAVA_TO_BUILD, VARIANT, ARCHITECTURE, TARGET_OS, VERSION, SCM_REF, boot JDK paths`, etc.)
- Verified that the script starts and runs both `--help` and `--generate-metadata `without syntax errors.
- Confirmed that the workspace/target directory is created and writable. It currently contains Linux build artifacts (`sbom.json, tar.gz`), which shows workspace paths resolve correctly.

**Solaris specific failures** (expected)

- Since I'm running on Linux wsl and did not build any Solaris binaries, the script correctly reported:
- Missing Solaris artifacts (`sha256sum: 'OpenJDK*': No such file or directory`)
- Missing metadata files (metadata/version.txt, metadata/buildSource.txt)
- Expected` scp/port` parsing errors related to Solaris remote build logic (`Bad port 'JDK7_BOOT_DIR=...'`)

These failures are normal for a non Solaris environment and indicate that the code paths behave as expected given the missing Solaris build outputs. I believe these lint fixes did not alter the logic or introduced breaking changes. :/ All ears for your suggestions.

karianna (approved):

judovana:
Cool. I second to Martijn. Maybe hint,  I consider rebasing on master instead of merging of master, as slightly better approach. But I guess thats based on personal taste only. ty!

SehrishHussain:
@judovana I rebase my branch on top of the latest master instead of merging. Let me know if you suggest any further adjustments.

SehrishHussain:
@sxa Please take a look at these lint changes, if there are no further concerns, I'll move forward with the merge.

judovana:
ok. Good  job. TY!

PR #4331, opened 2025-12-03
Removing Win32 and Solaris x64 and SparcV9 from Autotriage

adamfarley:
Also updating the regexes to exclude false failures and to correctly recognise a new JCK remote test trigger failure.

Test: https://github.com/adamfarley/temurin-build/actions/runs/19901201844/job/57045242574

Test run passed.


adamfarley:
Thanks @karianna and @andrew-m-leonard 😎

Issue #4335, opened 2025-12-11
SBOM validator fails with "Invalid SHA" error


adamfarley:
**What are you trying to do?**

Validate an SBOM using a curl command to validate the SHA for the build source's latest commit.

**Expected behaviour:**

This curl command should produce a return code of "0".

```
curl --silent --fail -I "https://github.com/adoptium/jdk17u/commit/3c17462cb874247d89da1514e9dae32046eef6"
```

**Observed behaviour:**

But instead it returns an error, possibly due to GitHub filtering out https requests from a curl agent.

**Any other comments:**

Will look into using the github api instead.

PR #4338, opened 2025-12-12
Adding github api check to SBOM sha validation

adamfarley:
When checking that the sha provided in the SBOM is valid, we sometimes encounter an intermittent problem where github returns an unusual HTML response code even when the commit exists. The api seemed unaffected.

Here is a backup check that uses the api if the usual url causes problems again.

Resolves https://github.com/adoptium/temurin-build/issues/4335


adamfarley:
Tested extensively locally, where all code paths were proven to work.

Any linter fails will be addressed Monday.

adamfarley:
Thanks @steelhead31 and @Haroon-Khel 😎

PR #4344, opened 2025-12-16
Rename remote tests appropriately

### LINKED ISSUE #4343: Relabel remote tests in the weekly triage issue

View the triage output with correctly labelled elements.

**Expected behaviour:**

The remote tests should be clearly labelled as remote tests, and we shouldn't be guessing at what sort of test they are, as that can cause inaccurate conclusions.

**Observed behaviour:**

We assume the remote tests are a specific type of test, and announce that they are that test.

**Any other comments:**

While we're editing this code, we should also block the display of the error text for the failing tests. This makes the issue shorter and easier to read.

adamfarley:
Renaming the remote test category to prevent confusion in the public weekly triage issues.

Resolves #4343

### DIFF
--- a/tooling/build_autotriage/build_autotriage.sh
+++ b/tooling/build_autotriage/build_autotriage.sh
@@ -470,9 +470,15 @@ generateOutputFile() {
             preventable="no"
           fi
           jobTriageOutput+="Preventable: ${preventable}\n"
-          jobTriageOutput+="\`\`\`\n"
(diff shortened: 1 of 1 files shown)


adamfarley:
Testing here: https://github.com/adamfarley/temurin-build/actions/runs/20275671762

Update: Passed. Here is the issue generated: https://github.com/adamfarley/temurin-build/issues/108

PR #4347, opened 2025-12-22
switch version parsing to use standard commands

sxa:
`grep -Eo` is not standard grep syntax and is incompatible with Solaris producing errors when this is run. The incompatible code was introduced in https://github.com/adoptium/temurin-build/pull/4304.

Change to using `awk` as it avoids the fork of additional processes to parse the files and is standard across UNIX-based systems.

@judovana has suggested using a simpler `grep` construct but personally I'm losing interest in testing things on this platform at the moment but happy if someone wishes to propose and test an alternative ;-)

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -276,7 +276,8 @@ versionNumbersFileParser() {
     [[ ! "${fileVersionString}" =~ ^jdk8u[0-9]+$ || $error ]] && echo "ERROR: build.sh: ${funcName}: version file could not be parsed." >&2 && exit 1
   else
     # File parsing logic for jdk11+.
-    patchNo="$(  awk -F= '/^DEFAULT_VERSION_PATCH/  {print$2}' "${numbersFile}")" || error="true"
+    # We use awk as a POSIX std (to be consistent with platforms like Solaris)
+    patchNo="$(  awk -F= '/^DEFAULT_VERSION_PATCH/{print$2}' "${numbersFile}")" || error="true"
     updateNo="$( awk -F= '/^DEFAULT_VERSION_UPDATE/ {print$2}' "${numbersFile}")" || error="true"
     interimNo="$(awk -F= '/^DEFAULT_VERSION_INTERIM/{print$2}' "${numbersFile}")" || error="true"
     featureNo="$(awk -F= '/^DEFAULT_VERSION_FEATURE/{print$2}' "${numbersFile}")" || error="true"


sbin/build.sh
```diff
@@ -272,14 +272,14 @@ versionNumbersFileParser() {
...
-   [[ ! "${fileVersionString}" =~ ^jdk8u[0-9]+$  || $error ]] && echo "ERROR: build.sh: ${funcName}: version file could not be parsed." >&2 && exit 1
+    fileVersionString="jdk8u$(awk -F= '/^JDK_UPDATE_VERSION/{print$2}' "${numbersFile}")" || error="true"
+    [[ ! "${fileVersionString}" =~ ^jdk8u[0-9]+$  || $error ]] && echo "ERROR: build.sh: ${funcName}: version file could not be parsed." >&2 && exit 1
   else
     # File parsing logic for jdk11+.
-    patchNo="$(grep "DEFAULT_VERSION_PATCH" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
-    updateNo="$(grep "DEFAULT_VERSION_UPDATE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
-    interimNo="$(grep "DEFAULT_VERSION_INTERIM" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
-    featureNo="$(grep "DEFAULT_VERSION_FEATURE" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
+    patchNo="$(  awk -F= '/^DEFAULT_VERSION_PATCH/  {print$2}' "${numbersFile}")" || error="true"
```
karianna:
```suggestion
    # We use awk as a POSIX std (to be consistent with platforms like Solaris)
    patchNo="$(  awk -F= '/^DEFAULT_VERSION_PATCH/{print$2}' "${numbersFile}")" || error="true"
```

sbin/build.sh
```diff
@@ -272,14 +272,14 @@ versionNumbersFileParser() {
...
+    updateNo="$( awk -F= '/^DEFAULT_VERSION_UPDATE/ {print$2}' "${numbersFile}")" || error="true"
```
karianna, line 281:
```suggestion
    updateNo="$( awk -F= '/^DEFAULT_VERSION_UPDATE/{print$2}' "${numbersFile}")" || error="true"
```

sbin/build.sh
```diff
@@ -272,14 +272,14 @@ versionNumbersFileParser() {
   error=""
   if [ "$jdkVersion" -eq 8 ]; then
     # jdk8 uses this format: jdk8u482-b01
-    fileVersionString="jdk8u$(grep "JDK_UPDATE_VERSION" "${numbersFile}" | head -1 | grep -Eo '[0-9]+')" || error="true"
-   [[ ! "${fileVersionString}" =~ ^jdk8u[0-9]+$  || $error ]] && echo "ERROR: build.sh: ${funcName}: version file could not be parsed." >&2 && exit 1
+    fileVersionString="jdk8u$(awk -F= '/^JDK_UPDATE_VERSION/{print$2}' "${numbersFile}")" || error="true"
+    [[ ! "${fileVersionString}" =~ ^jdk8u[0-9]+$  || $error ]] && echo "ERROR: build.sh: ${funcName}: version file could not be parsed." >&2 && exit 1
```
karianna:
```suggestion
    [[ ! "${fileVersionString}" =~ ^jdk8u[0-9]+$ || $error ]] && echo "ERROR: build.sh: ${funcName}: version file could not be parsed." >&2 && exit 1
```

sbin/build.sh
karianna, line 281:
```suggestion
    updateNo="$( awk -F= '/^DEFAULT_VERSION_UPDATE/{print$2}' "${numbersFile}")" || error="true"
```

sxa: Duplicate of previous comment?

sxa: The additional spaces in this and the previous line are intentional to make the block more "human readable" by lining up the four parsing lines as follows (and fortunately the linter hasn't moaned about it ;-) ):

Image

sbin/build.sh
sxa: I'd prefer "POSIX standard" rather than the implication that Solaris is "unusual", although this comment in the new version doesn't tell you what the alternative would have been so I'm not sure the comment adds much here anyway. I can add this in if you really want it though. Let me know your thoughts.

sbin/build.sh
judovana: I sometimes seem bad when I do that. I mean ok for development, not ok for commit. But next time, I align it for work again, and undo for commit... next time I will feel better when keeping the align spaces in commit:)

sbin/build.sh
karianna: Updated comment

karianna (approved):

adamfarley (approved):
LGTM

Issue #4351, opened 2026-01-02
jdk-26+29 or later builds are not creating debugimages


andrew-m-leonard:
https://bugs.openjdk.org/browse/JDK-8351842 has changed the upstream behavior for where debug symbol files are located, they are no longer bundled within the JDK, but only in the "symbols" image. Unless --with-external-symbols-in-bundles=full is specified.

Investigate whether either:
1. We should build jdk-26+ using --with-external-symbols-in-bundles=full
2. Update the debug image file [extraction](https://github.com/adoptium/temurin-build/blob/740d4af9cc1d6ca0412e442c43e4fe93ac52df30/sbin/build.sh#L1863) to pull from the symbols image

judovana:
In the fedora specfiles I had copied the debug symbols backs (which is practically `--with-external-symbols-in-bundles=full` and it immediately bitten me back. So tt update the debug image extraction to pull from the symbols in theirs new location seems like better thing to do    However I'm reconstructing the debuginfo rpms later, and I was  unable to find better solution to match it with debugsources then to copy it back. So if future debuginfo rpms are in radar, the `--with-external-symbols-in-bundles=full` may be the only way.

PR #4356, opened 2026-01-07
Jdk26 debugimages

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/4351

Updated build.sh script to locate jdk-26+ debug symbols within the "symbols" image and collect them from there for debugimage archiving

Test builds:
- jdk21 : https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk21u/job/jdk21u-linux-x64-temurin/399/artifact/workspace/target/ - GOOD
- jdk26 : https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk26/job/jdk26-linux-x64-temurin/11/artifact/workspace/target/ - GOOD

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ For reference, OpenJDK version numbers look like 1.8.0_162-b12 (for Java 8) or
 (162) or the 3rd position in the semVer version string (4).
 This is typically used in conjunction with -b.
 
---use-adoptium-devkit <Adoptium DevKit release tag>
+--use-adoptium-devkit <Adoptium_DevKit_release_tag>
 Download and use the given DevKit from https://github.com/adoptium/devkit-binaries/releases.
 The DevKit is downloaded and unpacked to WORKSPACE_DIR/WORKING_DIR/devkit
 and will add the configure arg --with-devkit=WORKSPACE_DIR/WORKING_DIR/devkit.
--- a/makejdk-any-platform.1
+++ b/makejdk-any-platform.1
@@ -198,6 +198,11 @@ For reference, OpenJDK version numbers look like 1.8.0_162-b12 (for Java 8) or
 (162) or the 3rd position in the semVer version string (4).
 This is typically used in conjunction with \fB<-b>\fR.
 .TP
+.BR \-\-use-adoptium-devkit " " \fI<Adoptium_DevKit_release_tag>\fR
+Download and use the given DevKit from https://github.com/adoptium/devkit-binaries/releases.
+The DevKit is downloaded and unpacked to WORKSPACE_DIR/WORKING_DIR/devkit
+and will add the configure arg --with-devkit=WORKSPACE_DIR/WORKING_DIR/devkit.
+.TP
 .BR \-\-use-jep319-certs
 Use certs defined in JEP319 in Java 8/9. Deprecated, has no effect.
 .TP
--- a/sbin/solaris/build-simple.sh
+++ b/sbin/solaris/build-simple.sh
@@ -157,24 +157,24 @@ scp -prP "${SSH_PORT}" $SSH_OPTS "${SSH_TARGET}:temurin-build/build-farm/workspa
 
 cd workspace/target || exit 1
 for FILE in OpenJDK*; do
-    echo Creating metadata for ${FILE}
-    # Skip checksum generation for SBOM files
-    if [[ "${FILE}" == *sbom.json ]]; then
-        echo "Skipping checksum generation for SBOM: $FILE"
+    echo "Creating metadata for ${FILE}"
+
+    if [[ "${FILE}" == *-sbom_* ]]; then
+        echo "SBOM detected – removing upstream checksum if present"
+        rm -f "${FILE}.sha256.txt"
         sha256=""
+        metadata_file="${FILE%.*}-metadata.json"
     else
         sha256sum "${FILE}" > "${FILE}.sha256.txt"
         sha256=$(cut -d' ' -f1 "${FILE}.sha256.txt")
-    fi
-
-    # Metadata filename: SBOM files get <name>-metadata.json (consistent with naming)
-    if [[ "${FILE}" == *sbom.json ]]; then
-        metadata_file="${FILE%.*}-metadata.json"
-    else
         metadata_file="${FILE}.json"
     fi
-    createMetadataFile "$metadata_file" "${TARGET_ARCH}" "$SCM_REF" metadata/buildSource.txt metadata/version.txt "$sha256"
+
+    createMetadataFile "$metadata_file" "${TARGET_ARCH}" "$SCM_REF" \
+        metadata/buildSource.txt metadata/version.txt "$sha256"
 done
+
+
 # Simple test job uses filenames.txt to determine the correct filenames to pull down
 ls -1 > filenames.txt
 cd ../../..


karianna:
@andrew-m-leonard Looks like a bunch of build failures - related?

jerboaa:
> @andrew-m-leonard Looks like a bunch of build failures - related?

The alpine one looks related:
```
find: symbols: No such file or directory
Error: Process completed with exit code 1.
```
See: https://github.com/adoptium/temurin-build/actions/runs/20788717421/job/59733430746?pr=4356#step:6:3242

adamfarley:
The x64 Linux failure looks like the alpine failure:
```
 Creating directory structure for static libs 'lib/static/linux-amd64/glibc'
/home/jenkins/workspace/build/src/build/linux-x86_64-server-release/images
moving static-libs-graal to jdk-27+3-static-libs
find: 'symbols': No such file or directory
```

jerboaa:
Hmm, the [test build](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk26/job/jdk26-linux-x64-temurin/11/artifact/workspace/target/) has `--with-native-debug-symbols=external` and the CI Linux build has `--with-native-debug-symbols=none`. I don't know why. But for the `symbols` image to get produced (on JDK 26+), one needs `--with-native-debug-symbols=external`, AFAIK.

jerboaa:
> Hmm, the [test build](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk26/job/jdk26-linux-x64-temurin/11/artifact/workspace/target/) has `--with-native-debug-symbols=external` and the CI Linux build has `--with-native-debug-symbols=none`. I don't know why.

I have a theory (explicitly sets it to `none`):
https://github.com/adoptium/temurin-build/actions/runs/20788717421/job/59733430534?pr=4356#step:6:10
... and the default build has it set to external.

On a related note: we need to ensure that the `symbols-image` target is being built (it's included in `product-images` but not sure which we use).

andrew-m-leonard:
temurin-build builds with no symbols by default which is what the CI action build does

The symbols search previously is checking the "jdk" folder which always exists..., now checking the symbols image which may not exist if no symbols requested... so "find" needed a | true

jerboaa:
> temurin-build builds with no symbols by default which is what the CI action build does

Why? The tarball shouldn't include the debug symbol files, but the debugimage always should (also for release). This is what `--with-native-debug-symbols=external` would provide. Right now, the debugimage would be useless for our release builds. IMO it's fine to separate this out as a separate PR, but it should be fixed. At least for JDK 26+ where the fix is easy: Configure with `--with-native-debug-symbols=external` and package the `symbols` image.

andrew-m-leonard:
> > temurin-build builds with no symbols by default which is what the CI action build does
> 
> Why? The tarball shouldn't include the debug symbol files, but the debugimage always should (also for release). This is what `--with-native-debug-symbols=external` would provide. Right now, the debugimage would be useless for our release builds. IMO it's fine to separate this out as a separate PR, but it should be fixed. At least for JDK 26+ where the fix is easy: Configure with `--with-native-debug-symbols=external` and package the `symbols` image.

So Temurin is always built with --with-native-debug-symbols=external
as we always use build arg [--create-debug-image](https://github.com/adoptium/temurin-build/blob/9ca299c1548b68e0513feeb87b0a8bebf672fc5c/sbin/common/config_init.sh#L262)
https://github.com/adoptium/temurin-build/blob/9ca299c1548b68e0513feeb87b0a8bebf672fc5c/build-farm/make-adopt-build-farm.sh#L237

jerboaa:
Thanks. I'm confused, though. There seem to be contradictory statements.

## CI builds *without* debug symbols

It has `--with-native-debug-symbols=none` explicitly set.
https://github.com/adoptium/temurin-build/blob/9ca299c1548b68e0513feeb87b0a8bebf672fc5c/.github/workflows/build.yml#L117

## Build scripts build *with* debug symbols

It turns on `--with-native-debug-symbols=external` iff `--create-debug-image` is given. The latter is the default.
https://github.com/adoptium/temurin-build/blob/9ca299c1548b68e0513feeb87b0a8bebf672fc5c/build-farm/make-adopt-build-farm.sh#L236-L237

So CI is testing something other than we do by default? Seems strange.

sbin/build.sh
```diff
@@ -1861,26 +1861,43 @@ cleanAndMoveArchiveFiles() {
...
       ;;
     darwin)
       # on MacOSX, we want to take the files within the .dSYM folders
-      debugSymbols=$(find "${jdkTargetPath}" -type d -name "*.dSYM" | xargs -I {} find "{}" -type f)
+      debugSymbols=$(find "${symbolsLocation}" -type d -name "*.dSYM" | xargs -I {} find "{}" -type f || true)
       ;;
     *)
       # on other platforms, we want to take .debuginfo files
-      debugSymbols=$(find "${jdkTargetPath}" -type f -name "*.debuginfo")
+      debugSymbols=$(find "${symbolsLocation}" -type f -name "*.debuginfo" || true)
```
jerboaa, line 1885: Perhaps there should be a comment why the `|| true` is there. Something like:

```
# When --with-native-debug-symbols=none is explicitly set then there won't be a `symbols` image on JDK 26+.
# Therefore, we need a '|| true' at the end.
```

judovana: I quite like   `cmd || echo  "what happened and why"` instead of just `true` with comment...

andrew-m-leonard:
> Thanks. I'm confused, though. There seem to be contradictory statements.
> ## CI builds _without_ debug symbols
> 
> It has `--with-native-debug-symbols=none` explicitly set.
> 
> https://github.com/adoptium/temurin-build/blob/9ca299c1548b68e0513feeb87b0a8bebf672fc5c/.github/workflows/build.yml#L117
> 
> ## Build scripts build _with_ debug symbols
> 
> It turns on `--with-native-debug-symbols=external` iff `--create-debug-image` is given. The latter is the default.
> 
> https://github.com/adoptium/temurin-build/blob/9ca299c1548b68e0513feeb87b0a8bebf672fc5c/build-farm/make-adopt-build-farm.sh#L236-L237
> 
> So CI is testing something other than we do by default? Seems strange.

Yes, CI testing does not build symbols, the reason being because we don't have enough disk space on github runners. We use the "free" github runners which only have 40Gb disks. We've already had to stop some CI builds (jdk11u Windows) because there's not enough disk space to build them. I could try enabling debug, but literally we're within 1Gb of free space in places.

sbin/build.sh
andrew-m-leonard: I've added a comment

andrew-m-leonard:
> Thanks. I'm confused, though. There seem to be contradictory statements.
> ## CI builds _without_ debug symbols
> 
> It has `--with-native-debug-symbols=none` explicitly set.
> 
> https://github.com/adoptium/temurin-build/blob/9ca299c1548b68e0513feeb87b0a8bebf672fc5c/.github/workflows/build.yml#L117
> 
> ## Build scripts build _with_ debug symbols
> 
> It turns on `--with-native-debug-symbols=external` iff `--create-debug-image` is given. The latter is the default.
> 
> https://github.com/adoptium/temurin-build/blob/9ca299c1548b68e0513feeb87b0a8bebf672fc5c/build-farm/make-adopt-build-farm.sh#L236-L237
> 
> So CI is testing something other than we do by default? Seems strange.

Just to clarify, with temurin-build scripts there are different ways to build, either using the "build-farm", or using "makejdk-any-platform" direct, The former is what is used by Jenkins type CI, the later is what is used by github CI and typically "users"

sbin/build.sh
jerboaa: @andrew-m-leonard Not seeing the comment. Am I missing something?

jerboaa: Oh, I guess it's `Find debug symbols if they were built (ie.--with-native-debug-symbols=external)`. That should be fine.

Issue #4354, opened 2026-01-06
JDK8u: Head: MacOS builds fail due to Clang error


adamfarley:
Thank you for reporting a broken build! Please fill in the following details:
- **Jenkins link to the failing log:**
  - https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-mac-x64-temurin/586
- **Which platforms (OS+arch) are affected?:**
  - MacOS
- **If you look at the historic logs, does the problem appear to be specific to one machine?:**
  - This appears to be caused by an upstream change: https://bugs.openjdk.org/browse/JDK-8317970
- **Link to any upstream project bug report:**
  - Issue raised upstream: https://bugs.openjdk.org/browse/JDK-8374589

```
clang: error: invalid version number in '-mmacosx-version-min=11.00.00'
```

adamfarley:
Theory 1: This issue is caused by the new mmacosx-version-min value having the wrong format.
  - Proven incorrect, as changing the value to the previous format (but the same value) still results in failure:
    - [Example 1:](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-mac-x64-temurin/591/console) clang: error: invalid version number in '-mmacosx-version-min=11.0'
    - [Example 2:](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-mac-x64-temurin/592) clang: error: invalid version number in '-mmacosx-version-min=11' 
- Theory 2: This issue is caused by the mechanism to validate the OS, which is rooted in the Xcode SDK. For JDK8, the default SDK is MacOSX10.15.sdk, which may have no concept of macos11.

Assuming theory 2 is the right one, here are some options for fixes:
1. To override mmacosx-version-min via --with-extra-cxxflags=' -mmacosx-version-min=11.0.0' - 
    - Doesn't work. [Attempts to do this](https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-mac-x64-temurin/601/) result in the same "invalid version number" error, as if the 11 min had not been overridden.
    - Adding a cflags arg gets us further, but ldflags are still a problem, and those have the min added after the user's args.
    - I don't see a clean way to override the built-in min without altering the source, but there may be unclean ways (see option 2).
2. Find some other way to override all instances of mmacosx-version-min via arguments or environment variables - IN PROGRESS
[...]

PR #4360, opened 2026-01-08
Switching xcode version for jdk8u builds on mac

adamfarley:
Along with an optional SDKNAME override if the jdk config script fails to use the new Xcode SDK.

The latter will be removed if the test run uses the correct SDK (once the infrastructure re-work is complete).

Resolves https://github.com/adoptium/temurin-build/issues/4354

### DIFF
--- a/build-farm/platform-specific-configurations/mac.sh
+++ b/build-farm/platform-specific-configurations/mac.sh
@@ -74,12 +74,7 @@ else
   fi
 fi
 
-
-#if [ "${JAVA_TO_BUILD}" == "${JDK8_VERSION}" ]
-#then
-#  export SDKNAME="${XCODE_SWITCH_PATH}/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk"
-#  # The configure option '--with-macosx-codesign-identity' is supported in JDK8 OpenJ9 and JDK11 and JDK14+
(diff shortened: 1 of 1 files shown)


adamfarley:
If we need to tell the OpenJDK toolchain scripting (used during the configure stage of an OpenJDK build) to use a different version of the xcode sdk, we can replace [this line and the one below it](https://github.com/adoptium/temurin-build/blob/9ca299c1548b68e0513feeb87b0a8bebf672fc5c/build-farm/platform-specific-configurations/mac.sh#L79) with this code:

```
if [ "${JAVA_TO_BUILD}" == "${JDK8_VERSION}" ]
then
  export SDKNAME="${XCODE_SWITCH_PATH}/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk"
  # The configure option '--with-macosx-codesign-identity' is supported in JDK8 OpenJ9 and JDK11 and JDK14+
elif [[ ( "$JAVA_FEATURE_VERSION" -eq 11 ) || ( "$JAVA_FEATURE_VERSION" -ge 14 ) ]]
```

Specifically, we may need to do this if the xcode SDK printed to the console (during the configure stage) is under version 11, as that may result in [this issue](https://github.com/adoptium/temurin-build/issues/4354) continuing to occur even if we're using the newer xcode.

adamfarley:
Thanks @andrew-m-leonard and @karianna

Issue #4361, opened 2026-01-09
Not all Reproducible Build compare tests are running successfully


andrew-m-leonard:
Ref: https://adoptium.slack.com/archives/C09NW3L2J/p1767872296748789
```
jdk21u Reproducibility:  x64Linux(100 %), aarch64Linux(?%), ppc64leLinux(100 %), x64Windows(?%), x64Mac(99.86 %), aarch64Mac(99.73 %)
```

andrew-m-leonard:
All Linux and Windows tests should now be 100%

PR #4366, opened 2026-01-12
Fix up reproducible build script tests

andrew-m-leonard:
Fixes Linux and Windows for issue: https://github.com/adoptium/temurin-build/issues/4361

**Linux:**
- Use rhel ubi8 public image, so we have standard support for aarch64 available
- Add support for s390x_linux using rhel ubi8 image
- Remove old configure_args formatting, now obsolete
- Filter --with-jobs= configure arg as needs re-evaluating on re-build host by standard openjdk make query
- Correct userDevkitLocation param
- Slim down to be able to fit within a 30Gb docker volume when using DevKit:
  - Shallow clone temurin-build git commit
  - Cut un-needed yum dependent gcc downloads when using DevKit
  - Yum install gcc tooling and prereqs only when not using DevKit
- Update obtaining of required BootJDK to be more flexible, and automatically download a suitable JDK, as some platforms (eg.s390x) might reference SBOM bootjdks that are not available from the API.

**Windows:**
 - Correction to use the Program Files(x86) to find vs2022
 - Correct userDevkitLocation param

Test Grinders:
- x64Linux - https://ci.adoptium.net/job/Grinder/15874/ - SUCCESS
- pLinux - https://ci.adoptium.net/job/Grinder/15876/ - SUCCESS
- aarch64Linux - https://ci.adoptium.net/job/Grinder/15875/ - SUCCESS
- s390xLinux - https://ci.adoptium.net/job/Grinder/15873/ - SUCCESS

### DIFF
--- a/tooling/reproducible/linux_repro_build_compare.sh
+++ b/tooling/reproducible/linux_repro_build_compare.sh
@@ -127,7 +127,7 @@ setTemurinBuildArgs() {
   buildArgs="$(echo "$buildArgs" | sed -e "s|--jdk-boot-dir [^ ]*|--jdk-boot-dir /usr/lib/jvm/jdk-${bootJdk}|")"
 
   if [[ "${using_DEVKIT}" == "true" ]] && [[ -n "${userDevkitLocation}" ]]; then
-    buildArgs="${buildArgs} --user-devkit-location ${userDevkitLocation}"
+    buildArgs="--user-devkit-location ${userDevkitLocation} ${buildArgs}"
   fi
 
   echo "${buildArgs}"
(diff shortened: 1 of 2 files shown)


tooling/reproducible/linux_repro_build_compare.sh
```diff
@@ -199,7 +199,8 @@ else
 fi
 
 echo "Rebuild args for makejdk_any_platform.sh are: $TEMURIN_BUILD_ARGS"
-echo " cd temurin-build && ./makejdk-any-platform.sh $TEMURIN_BUILD_ARGS > build.log 2>&1" | sh
+#echo " cd temurin-build && ./makejdk-any-platform.sh $TEMURIN_BUILD_ARGS > build.log 2>&1" | sh
+echo " cd temurin-build && ./makejdk-any-platform.sh $TEMURIN_BUILD_ARGS" | sh
```
judovana: why so?

andrew-m-leonard: just debugging so I can see what fails in stdout!

judovana: I see:(( Thank you! Crossing fingers!

adamfarley (approved):
LGTM

Issue #4377, opened 2026-02-03
Update windows reproducible build script to support attestation verification


andrew-m-leonard:
The [windows_repro_build_compare.sh](https://github.com/adoptium/temurin-build/blob/master/tooling/reproducible/windows_repro_build_compare.sh) script currently performs a standard Adoptium temurin-build scripts re-build and compare, installing any required dependencies into Adoptium CI node locations.

To support user/3rd party based Windows Attestation verification builds, whereby any required tooling/dependencies have already been securely downloaded and verified, the windows_repro_build_compare.sh needs to support a "Attestation Verification" mode, whereby it does not installed prereqs, and performs a standard OpenJDK "configure and make" using purely upstream OpenJDK github source. This independently verifies a given Temurin build, without use of temurin-build scripts, and not using an Adoptium based mirror of OpenJDK source. This then provides a more secure verification to the integrity of a Temurin release build.

PR #4378, opened 2026-02-03
Update windows_repro_build_compare.sh to support Attestation verify build

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/4377

Update windows_repro_build_compare.sh to provide an attestation verification mode of securely re-building a given Temurin JDK.
- Re-worked the arguments parsing and added --attestation-verify option
- Update BootJDK download to perform GPG verify
- In "Attestation Mode" : 
  - Any missing prereqs must have been already installed
  - A USER_DEVKIT_LOCATION is mandatory, as for secure verification the user/3rd party must have independently created/installed the necessary VS2022 and ReDist DLL devkit
  - Upstream openjdk github source is used as the basis of the re-build
  - Standard OpenJDK "configure & make" build process is used, using the configure_args from the SBOM
  - Target EA builds (Release builds get away with it!) to be attested requires to have been built with SBOM fix (https://github.com/adoptium/temurin-build/issues/4379) for 100% compare

Grinder running updated Rebuild_Same_JDK_Reproducibility_Test_win against an EA build: https://ci.adoptium.net/view/Test_grinder/job/Grinder/16103/ - SUCCESS - 100%

### DIFF
--- a/tooling/reproducible/windows_repro_build_compare.sh
+++ b/tooling/reproducible/windows_repro_build_compare.sh
@@ -21,6 +21,9 @@
 
 set -e
 
+# Adoptium's public GPG key used for GPG signatures
+ADOPTIUM_PUBLIC_GPG_KEY="0x3B04D753C9050D9A5D343F39843C48A565F8F04B"
+
 # Read Parameters
 SBOM_URL=""
 TARBALL_URL=""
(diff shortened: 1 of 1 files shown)


test/system/reproducibleCompare/playlist.xml
```diff
@@ -69,7 +69,7 @@
 	<test>
 		<testCaseName>Rebuild_Same_JDK_Reproducibility_Test_win</testCaseName>
 		<command>unset SPEC;\
-		$(TEST_ROOT)$(D)system$(D)reproducibleCompare$(D)windows_repro_build_compare.sh $(SBOM_FILE) $(JDK_FILE) $(REPORTDIR); \
+		$(TEST_ROOT)$(D)system$(D)reproducibleCompare$(D)windows_repro_build_compare.sh --sbom-url $(SBOM_FILE) --jdk-url $(JDK_FILE) --report-dir $(REPORTDIR) --user-devkit-location https://github.com/adoptium/devkit-binaries/releases/download/vs2022_redist_14.40.33807_10.0.26100.1742/vs2022_redist_14.40.33807_10.0.26100.1742.zip --attestation-verify; \
```
karianna: Is that devkit verison worth extracting to a var at the top of the file?

tooling/reproducible/windows_repro_build_compare.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2001,SC2086
```
karianna, line 2: Worth saying why we added these?

andrew-m-leonard: Sure i'll add a comment, it's because there are numerous sed lines that if you quote as suggested don't work on windows

test/system/reproducibleCompare/playlist.xml
andrew-m-leonard: WINDOWS_ADOPTIUM_DEVKIT variable added

tooling/reproducible/windows_repro_build_compare.sh
andrew-m-leonard: added

Issue #4381, opened 2026-02-04
jdk25u Reproducible build tests failing due to jdk-25.0.3+0 version


andrew-m-leonard:
Temurin builds of HEAD Oracle managed versions for which there are no tags yet, makes up the jdk-25.0.3+0, this is necessary for the Temurin build script expectations and to satisfy the Windows Installer WIX requirements to have a "+0" build level.
Unfortunately the actual hotspot build of a non-"build" HEAD for such is actually "jdk-25.0.3", thus the JVM setting "java.runtime.version" is this without a build : 
```
java.runtime.version = 25.0.3-beta-ea
```
The logic added here : https://github.com/adamfarley/temurin-build/blob/ce3a48460a8e9740033553a9db0db6852aa60b8c/sbin/build.sh#L341
generates a "+0" for the Temurin version which gets put into the JDK home folder name to satisfy the Installers requirements.

Unfortunately this means the JVM component version which is correct in the SBOM does not match the generated JDK home folder name which has the additional "+0".

andrew-m-leonard:
Closing as won't fix, as jdk25u is now non-Oracle managed for jdk-25.0.3, and in fact the first jdk-25.0.3+1 tag has just dropped... :-)

PR #4382, opened 2026-02-04
Make sure reproducible build uses correct tag version default +0 for example jdk25u

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/4381

Allow linux reproducible build script to handle jdk25u(or oracle managed version) HEAD builds, which don't have any tags, which means we default to a "+0" build, as JVM only has jdk-25.0.3. This logic is only done for the Linux reproducible script, as Windows & Mac do not use the version in the same way.

Check for "+" in JVM version, and if no build present, add "+0" default.

jdk-25.0.3 Reproducible test re-run : 
- linux : https://ci.adoptium.net/job/Grinder/16081/
-


andrew-m-leonard:
jdk-25.0.3 is now non-Oracle managed.. won't fix.

Issue #4387, opened 2026-02-06
validate release download and SBOM validation checks fail due to missing SBOM path and _adopt tag


andrew-m-leonard:
download_and_sbom_validation is failing due to : 
- validateSBOM.sh assumes passed SBOM is full path to file location, release_download_test.sh just passed the file name
- validateSBOM.sh expected SCM_REF needs to be the _adopt tag, not just the release tag name
- jdk8u aarch32 cannot check for expected SCM_REF as the actual tag is jdk8u aarch32 port format, not the standard jdk8u tag name

PR #4388, opened 2026-02-06
Release Download and SBOM test fixes

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/4387

- release_download_test for Temurin needs to add  "_adopt" to the github release tag to get the correct SCM_REF
- release_download_test needs to pass full path to SBOM, as validateSBOMcontent.sh changes to its own workspace
- Add input param verbose output so we can see params in console
- Correct jdk8u aarch32 validateSBOM.sh tag check which was missing the build part, and _adopt variant
- Only validate jdk8u linux aarch32 expected SCM_REF if the specified SOURCE_TAG is a correct aarch32_YYYYMMDD format tag, which will not be the case when called from release_download_test for a Temurin github release tag

### DIFF
--- a/tooling/validateSBOM.sh
+++ b/tooling/validateSBOM.sh
@@ -100,7 +100,7 @@ arg_parser() {
   fi
 
   # If SOURCE_TAG is not the expected full jdk8u aarch32 port date format, then skip expected source tag check
-  if [ -n "$SOURCE_TAG" ] && [ "$JDK_MAJOR_VERSION" -eq "8" ] && echo $SBOM_LOCATION | grep _arm_linux_; then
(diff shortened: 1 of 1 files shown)


tooling/validateSBOM.sh
```diff
@@ -94,6 +94,14 @@ arg_parser() {
     SOURCE_TAG=""
   fi
 
+  # If SOURCE_TAG is not the expected full jdk8u aarch32 port date format, then skip expected source tag check
```
karianna, line 102: Nitpick, I wonder if this sort of thing can be extracted into a function that basically says if (jdk8Aarch32) then X

karianna (approved):

tooling/validateSBOM.sh
andrew-m-leonard: i've corrected the previous logic, so it's a little more than that now, I think i'll keep it as it is, as logic only used here

PR #4397, opened 2026-02-24
Adding jdk27 to autotriage

adamfarley:
JDK26 is no longer head. Adjusting autotriage to match.


adamfarley:
Thanks @Haroon-Khel @tellison and @steelhead31 😎

Issue #4404, opened 2026-03-06
temurin-build Windows CI builds failing due to github runner image changes??


andrew-m-leonard:
ref: https://github.com/adoptium/temurin-build/actions/runs/22757111264/job/66004348046?pr=4394
```
mk-ca-bundle.pl generates 144 certificates
FATAL: Startup failure (dev note: apps_startup()) for openssl
100000000A000000:error:80000002:system library:process_include:No such file or directory:crypto/conf/conf_def.c:821:calling stat(/etc/crypto-policies/back-ends/opensslcnf.config)
100000000A000000:error:07000075:configuration file routines:ssl_module_init:ssl command section empty:crypto/conf/conf_ssl.c:94:name=system_default, value=crypto_policy
100000000A000000:error:0700006D:configuration file routines:module_run:module initialization error:crypto/conf/conf_mod.c:274:module=ssl_conf, value=ssl_module retcode=-1      
```
Possibly we need to specifically install openssl?

gawk failures too:
```
Creating support/native/java.desktop/libjsound/static/jsound.lib from 17 file(s)
Creating support/native/jdk.jpackage/jpackageapplauncherw/static/jpackageapplauncherw.lib from 23 file(s)
Creating support/native/jdk.hotspot.agent/libsaproc/static/saproc.lib from 3 file(s)
gawk: cmd. line:1: error: ? * + or {interval} not preceded by valid subpattern: /??_7.*@@6B@/
make[3]: *** [lib/JvmMapfile.gmk:133: /cygdrive/d/a/temurin-build/temurin-build/workspace/build/src/build/windows-x86_64-server-release/hotspot/variant-server/libjvm/symbols-objects] Error 1
make[3]: *** Deleting file '/cygdrive/d/a/temurin-build/temurin-build/workspace/build/src/build/windows-x86_64-server-release/hotspot/variant-server/libjvm/symbols-objects'
make[2]: *** [make/Main.gmk:252: hotspot-server-libs] Error 2
```

andrew-m-leonard:
Upstream bug for the gawk failures which is being backported to jdk17u and jdk21u https://bugs.openjdk.org/browse/JDK-8314555

andrew-m-leonard:
works now

PR #4406, opened 2026-03-09
Force jdk-25+ Linux builds to use LC_ALL=C to ensure Adoptium Temurin builds on Centos7 are reproducible across OS distributions

andrew-m-leonard:
Fixes https://github.com/adoptium/infrastructure/issues/4289 for current Centos7 build environment.

Force linux (except for Alpine) builds to use "C" (as aarch64 is already..), thus ensuring language neutral LC_ALL and thus consistent sort order across OS distributions. This then allows reproducible builds across Centos/RHEL7,8,9, Ubtuntu, ....
- Alpine was already choosing C.UTF-8 due to OpenJDK make not finding locale in BusyBox environment.
- Riscv64 left as-is, as riscv64 does have the correct language neutral C.UTF-8 already
- Added SBOM value "Build LC_ALL" to indicate LC_ALL value built with..

Test build: https://ci.adoptium.net/job/build-scripts/job/openjdk25-pipeline/150/ - SUCCESSFUL, all AQA-tests consistent with existing Temurin builds..
Test build only with Alpine & Riscv64 exclusion and SBOM value "Build LC_ALL" added: https://ci.adoptium.net/job/build-scripts/job/openjdk25-pipeline/154/ - GOOD (CycloneDX CLI had a http download 504 glitch re-run fine)

### DIFF
--- a/sbin/build.sh
+++ b/sbin/build.sh
@@ -941,7 +941,7 @@ executeTemplatedFile() {
   # For current jdk-25+ Temurin builds we need to force LC_ALL locale of "C" to avoid OpenJDK make choosing en_US.UTF-8 due to lack of C.UTF-8 on Centos7/RHEL7
   # en_US.UTF-8 is not language neutral causing sort order differences on differing linux OS distributions.
   # See ref: https://github.com/adoptium/infrastructure/issues/4289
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
Windows CI failure due to issue: https://github.com/adoptium/temurin-build/issues/4404

PR #4429, opened 2026-03-20
Add reproducible-verification (formerly attestation-verify) SBOM formulation recipe

Lukisorisch:
## Resolves #4420: Add reproducible-verification (formerly attestation-verify) SBOM formulation recipe

This PR adds the reproducible-verification recipe to the SBOM for documentation purposes, and to aid 3rd party vendors.

The current implementation generates two seperate formulas (one for the build script and one for the attestation verify), each containing one workflow. The alternative would be to group both workflows under a single formula.

## Resulting recipe

```
    {
      "bom-ref" : "formula_temurin_reproducible_verification_21.0.11-beta+9-202511161915",
      "workflows" : [
        {
          "bom-ref" : "workflow_temurin_reproducible_verification_21.0.11-beta+9-202511161915",
          "uid" : "workflow_temurin_reproducible_verification_21.0.11-beta+9-202511161915",
          "name" : "Temurin Reproducible Verification",
          "taskTypes" : [
            "clone",
            "test"
          ],
          "steps" : [
            {
              "name" : "clone repo",
              "description" : "clone repository",
              "commands" : [
                {
                  "executed" : "git clone https://github.com/adoptium/temurin-build.git"
                }
              ]
            },
            {
              "name" : "execute verification",
              "description" : "run reproducible build compare script",
              "commands" : [
                {
                  "executed" : "temurin-build/tooling/reproducible/linux_repro_build_compare.sh --sbom-url 'SBOM_FILE_OR_URL' --jdk-url 'JDK_FILE_OR_URL' --user-devkit-location 'ADOPTIUM_DEVKIT_FILE_OR_URL' --reproducible-verification --build-workspace 'FULL_PATH_TO_AN_EXISTING_BUILD_FOLDER'"
                }
              ]
            }
          ]
        }
      ]
    }
```

PR #4442, opened 2026-04-20
Convert SBOM Generation from Java to Python

### LINKED ISSUE #4421: To provide wider usage of the Temurin SBOM & CDXA generation clients, provide a python client package
The Temurin SBOM & CDXA utilities are currently only available as Java classes: https://github.com/adoptium/temurin-build/tree/master/cyclonedx-lib/src/temurin/sbom

Investigate the feasibility of converting them to be python3 classes? 

I suspect we could use:
- https://github.com/CycloneDX/cyclonedx-python-lib
  - https://pypi.org/project/cyclonedx-python-lib/

Lukisorisch:
## Description

This PR transitions our SBOM generation from the Java tool to the official [cyclonedx-python-lib](https://github.com/CycloneDX/cyclonedx-python-lib).

The changes have only been tested on Linux so far.

Resolves #4421.
**Note:** Please merge #4429 before merging this PR.

### 1.

Currently, the official cyclonedx-python-lib does not natively support the CycloneDX 1.6 formulation objects/array. To still migrate to python while still having a valid JSON without patching the library's internals, I have implemented a temporary "post-processing" script (temporary_sbom_post_processing.py). It injects the formulation workflows into the finished JSON at the very end of the pipeline i.e. at the end of generateSBoM().

All locations related to this temporary workaround are tagged with "TODO (CycloneDX 1.6)" for easy searching. Once the upstream library adds native support, the migration will look like this:

```text
MIGRATION (once native formulation support arrives in cyclonedx-python-lib):
1. Delete cyclonedx-lib/temporary_sbom_post_processing.py
2. Update everything marked with TODO (CycloneDX 1.6) back to the correct state
3. Implement native formulation logic (--addWorkflow, --addWorkflowStep, --addWorkflowStepCmd) in cyclonedx-lib temurin_gen_sbom.py, replacing the _skip_formulation() stubs with real CycloneDX lib calls.
4. The Workflow/Step/Cmd calls in addTemurinBuildRecipeToSBOM and addReproducibleVerificationRecipeToSBOM will then work natively.
```

You can find this text in the source code of build.sh as well.

### 2.

For some reason, the cyclonedx-python-lib has a adifferent key ordering in the json. For now, I fixed this using a hardcoded constant that defines the order of keys. This can be removed later, if I get confirmation that the order is indeed irrelevant (as it should be for jsons anyway)

### 3.

To avoid externally-managed-environment errors, the build script creates a temporary Python virtual environment now (sbom_venv) for the SBOM generation. sbom_venv has been added to .gitignore.

## Showcase

Here are two SBOMs (not from the same JDK or build script), one generated with the current Java implementation, the other generated with the new Python implementation. 

[jdk-hotspot-sbom-python.json](https://github.com/user-attachments/files/26876458/jdk-hotspot-sbom-python.json)
[jdk-hotspot-sbom-java.json](https://github.com/user-attachments/files/26876459/jdk-hotspot-sbom-java.json)


andrew-m-leonard:
I am not sure injecting the missing support via a script is going to be a great solution in the long term, as it’s likely to be vulnerable to schema issues, and adds an extra step that the TemurinGenSBOM.py can’t use directly. The managing python venv is also a typical issue with Python, which we haven't had to resolve with temurin build.sh since we don't currently use Python.

I have been also doing some work with the CycloneDX Attestation tool, which is also not available in python.

I think although this may work, it's not architecturally the right direction, given the python libraries limited schema support.

Lukisorisch:
> I am not sure injecting the missing support via a script is going to be a great solution in the long term, as it’s likely to be vulnerable to schema issues, and adds an extra step that the TemurinGenSBOM.py can’t use directly. The managing python venv is also a typical issue with Python, which we haven't had to resolve with temurin build.sh since we don't currently use Python.
> 
> I have been also doing some work with the CycloneDX Attestation tool, which is also not available in python.
> 
> I think although this may work, it's not architecturally the right direction, given the python libraries limited schema support.

I agree, the "injection script" would only be a temporary solution until the official CycloneDX 1.6 support for python is released though.
The venv issue on the other hand is a bummer, yes.

Feel free to archive this PR.

nathanandrewturner (approved):

PR #4467, opened 2026-05-18
Remove Solaris support and Windows x86-32 build support

gdams:
Solaris is no longer shipped, so this PR removes all Solaris/SunOS/SPARC code, platform configs, documentation, and test references.

Windows x86-32 (32-bit) is also no longer built — only x64 and aarch64 are supported. This removes all x86-32 build configurations, platform config blocks, test architecture enums, Azure DevOps stages, reproducible build handling, devkit documentation, and SBOM validation for x86-32.

### Files deleted
- `SOLARIS.md`
- `build-farm/platform-specific-configurations/solaris.sh`
- `sbin/solaris/build-simple.sh`
- `sbin/solaris/dotests.sh`
- `sbin/solaris/simplepipe.groovy`

### Files modified (23)
Build scripts, platform configs, CI workflows, documentation, tests, and tooling updated to remove Solaris and x86-32 references.

28 files changed, 23 insertions(+), 745 deletions(-)


sxa (commented):
I'm not going to put a hard block on this but I would personally prefer that we leave some of this in so that individuals who want to build their own "Temurin" on the platforms we no longer build for are still able to do it with our helper scripts.

PR #4481, opened 2026-05-28
Enable windows_repro_build_compare.sh to handle Windows aarch64 cross-compiled target

### LINKED ISSUE #4480: windows_repro_build_compare.sh dow not handle cross-compiled Windows Aarch64
Windows aarch64 is cross-compiled on x64 Windows, thus the repro_compare.sh scripts try to use the built JDKs as "work JDKs" to run things like jmod/jimage as part of the compare expansion process. These fail with invalid executable, as they are aarch64 binaries !

windows_repro_build_compare.sh and the repro_compare.sh scripts need to take an optional --work-jdk argument to specify a suitable machine JDK to use for compare expansion.

andrew-m-leonard:
Fixes https://github.com/adoptium/temurin-build/issues/4480

Assisted-by: IBM Bob

- repro_compare.sh has been extended with an optional 6th argument which is a suitable WORK_JDK. the repro_common.sh scripts have been augmented to use this.
- windows_repro_build_compare.sh has also been augmented to check the buildArgs to see if this is a Windows aarch64 cross-compile target, and use that as the MS VisualStudio targetArch

### DIFF
--- a/tooling/reproducible/repro_common.sh
+++ b/tooling/reproducible/repro_common.sh
@@ -19,13 +19,13 @@ TEMURIN_TOOLS_BINREPL="temurin.tools.BinRepl"
 function expandJDK() {
   local JDK_DIR="$1"
   local OS="$2"
-  local work_JDK="$3"
+  local WORK_JDK="$3"
 
   local JDK_HOME_DIR="$JDK_DIR"
 
   if [[ "$OS" =~ Darwin* ]]; then
     JDK_HOME_DIR="$JDK_DIR/Contents/Home"
-    work_JDK="$3/Contents/Home"
+    WORK_JDK="$3/Contents/Home"
   fi
 
   echo "$(date +%T) : Expanding various components to enable comparisons ${JDK_DIR} (original files will be removed):"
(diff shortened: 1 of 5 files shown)


tooling/reproducible/repro_common.sh
```diff
@@ -19,19 +19,16 @@ TEMURIN_TOOLS_BINREPL="temurin.tools.BinRepl"
...
+  local work_JDK="$3"
 
   local JDK_HOME_DIR="$JDK_DIR"
-  local JDK_COPY="${1}_CP"
-  local JDK_COPY_BIN="${JDK_COPY}/bin"
 
   if [[ "$OS" =~ Darwin* ]]; then
     JDK_HOME_DIR="$JDK_DIR/Contents/Home"
-    JDK_COPY_BIN="$JDK_COPY/Contents/Home/bin"
+    work_JDK="$3/Contents/Home"
```
adamfarley: I think work_JDK should be WORK_JDK to match the naming convention used for other variables.

adamfarley (approved):
Lgtm. 1 Optional style suggestion.

tooling/reproducible/repro_common.sh
andrew-m-leonard: fixed

adamfarley: Thanks

karianna (approved):

